PR #38, opened 2015-11-19
win32: Show drive letter instead of device in dropdown

jviotti:
Physical drive names are long and cause the dropdowns to overflow.

Fixes: https://github.com/resin-io/herostratus/issues/23


jviotti:
Tested in OS X.

jviotti:
Tested on Windows:

capture

PR #46, opened 2015-11-23
Eject Windows drives after successful burning

jviotti:
Use the `removedrive` for this task:

https://github.com/jviotti/removedrive

Fixes: https://github.com/resin-io/herostratus/issues/40


jviotti:
Tested in OS X for regressions.

jviotti:
Works on Windows, however "Done" is showed before the drive is actually ejected.

PR #47, opened 2015-11-24
Add finished state before done

jviotti:
Currently we say "done" as soon as the progress is equal to 100, however
we don't wait for the drive to be ejected/unmounted.

This commit introduces a new state, called "Finishing..." which is shown
when the progress is 100, but the drive has not been unmounted.


jviotti:
Tested in OS X.

jviotti:
Tested in Windows.

PR #70, opened 2016-01-02
Autoselect the drive if only one drive is present

### LINKED ISSUE #55: Autoselect drive if there is only a single one

alexandrosm:
An attempt to resolve #55. Looking forward to your feedback, I'm sure there's ways to improve this.

### DIFF
--- a/build/browser/app.js
+++ b/build/browser/app.js
@@ -60,6 +60,12 @@ app.controller('AppController', function($q, DriveScannerService, SelectionState
     this.selection.clear();
     this.writer.setProgress(0);
     this.scanner.start(2000);
+
+    this.scanner.scan().then(function (res) {
+      if(res.length == 1){
+        self.selectDrive(res[0]);
+      }
+    });
   };
 
   this.restart();
@@ -226,6 +232,8 @@ driveScanner.service('DriveScannerService', function($q, DriveScannerRefreshServ
     if (!angular.equals(self.drives, drives)) {
       self.drives = drives;
     }
+
+    return "bla"
   };
 
   /**
@@ -273,7 +281,7 @@ driveScanner.service('DriveScannerService', function($q, DriveScannerRefreshServ
 
 });
 
-}).call(this,"/lib/browser/modules")
+}).call(this,"/lib\\browser\\modules")
 },{"angular":9,"lodash":10,"path":undefined}],3:[function(require,module,exports){
 (function (__dirname){
 /* The MIT License
@@ -440,7 +448,7 @@ imageWriter.service('ImageWriterService', function($q, $timeout) {
 
 });
 
-}).call(this,"/lib/browser/modules")
+}).call(this,"/lib\\browser\\modules")
 },{"angular":9,"path":undefined}],4:[function(require,module,exports){
 /* The MIT License
  *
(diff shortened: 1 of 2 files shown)


build/browser/app.js
```diff
@@ -273,7 +279,7 @@ driveScanner.service('DriveScannerService', function($q, DriveScannerRefreshServ
 
 });
 
-}).call(this,"/lib/browser/modules")
+}).call(this,"/lib\\browser\\modules")
```
jviotti: Uhm, I don't like how Browserify is compiling with backslashes on Windows, but not a problem I guess.

build/browser/app.js
```diff
@@ -9148,7 +9154,7 @@ angular.module("template/typeahead/typeahead-popup.html", []).run(["$templateCac
 !angular.$$csp() && angular.element(document).find('head').prepend('<style type="text/css">.ng-animate.item:not(.left):not(.right){-webkit-transition:0s ease-in-out left;transition:0s ease-in-out left}</style>');
 },{}],8:[function(require,module,exports){
 /**
- * @license AngularJS v1.4.8
+ * @license AngularJS v1.5.0-beta.2
```
jviotti: Can you omit the angular version bump change from this commit, to keep it clean? I'll send a new PR later on upgrading the dependencies.

lib/browser/app.js
```diff
@@ -59,6 +59,12 @@ app.controller('AppController', function($q, DriveScannerService, SelectionState
     this.selection.clear();
     this.writer.setProgress(0);
     this.scanner.start(2000);
+
+    this.scanner.scan().then(function (res) {
```
jviotti: Really low priority, but we don't put a space between the `function` keyword and the open parenthesis in JavaScript.

lib/browser/app.js
```diff
@@ -59,6 +59,12 @@ app.controller('AppController', function($q, DriveScannerService, SelectionState
...
+      if(res.length === 1){
```
jviotti: Another very low priority one, but there is a space missing between the closing parenthesis and the open curly brace.

jviotti:
Can we squash the Hound fix into the main commit to keep the history clean?

jviotti:
I've commented on some style related minor issues. I'll review in much more detail the actual logic tomorrow.

lib/browser/app.js
jviotti: I think this solution is fine except for non-critical issue: The drives are scanned twice at the same time when the app starts or get's restarted due to `scanner.start()` and `scanner.scan()`.

I have the following idea:

- Let's make `this.scanner.start()` accept a callback function that is called whenever the drives are scanned successfully, returning the array of detected drives. An alternative, which seems more "node-friendly" is to make `this.scanner.start()` return an `EventEmitter` instance that emits an event.

Then here, in the application controller, we can listen to this event and inject the appropriate logic.

jviotti: I don't like calling this local variable `res`. Maybe something like `drives` or `detectedDrives` to better express it's meaning?

jviotti: I just noticed that the current approach also means that after the first scan, a single drive will not be selected by default anymore. The approach mentioned above should resolve this.

alexandrosm:
@jviotti let me know if you want me to address the comments or if you'll proceed with the EventEmmitter approach, in which case ignore this PR altogether. I think it's the right solution to go for anyway.

jviotti:
@alexandrosm The `EventEmitter` approach can be easily made in this PR. I can update the PR for you if you like, but thought you might want to get some fun doing it yourself.

jviotti:
@alexandrosm I've improved this PR and created a new one here: https://github.com/resin-io/resin-etcher/pull/91

- I've done some surgery to your original commits to squash them together and omit the angular version bump stuff.

- I've implement and tested the EventEmitter approach.

- I've improved the feature by only auto-selecting the drive after an image has been selected (otherwise the drive is selected even when the second step is still disabled).

- I've made sure it works after the first scan.

- I've made sure it doesn't keep selecting the same single drive over and over again.

- I've added a log line to determine when a drive was auto-selected.

Issue #74, opened 2016-01-05
Add a "burn another" button on completion


jviotti:
The completion screen will then have two buttons:

1. Completely start over by reseting every selection.

2. Burning the same image you selected before to another drive (basically keeps the image selection and resets the rest).

taahirisaacs:
This method should work @jviotti 

![screen shot 2016-01-18 at 10 14 57 am](https://cloud.githubusercontent.com/assets/15251376/12386215/64077c92-bdcc-11e5-9dfd-d5a58c5672cc.png)

jviotti:
@fedealconada Can you implement the above design? (only the separation between the buttons and explanation. The new footer design is coming in another PR).

jviotti:
Re-opening as a reminder to finish the design side of this feature.

PR #80, opened 2016-01-13
Solving issue 74

fedealconada:
- Creation of button “Burn image again”.
- Modification of restart() function
* now can receive the image as parameter
* if image is passed, means that the button “Burn image again” was
clicked

### DIFF
--- a/build/browser/app.js
+++ b/build/browser/app.js
@@ -48,9 +48,9 @@ app.controller('AppController', function($q, DriveScannerService, SelectionState
   this.writer = ImageWriterService;
   this.scanner = DriveScannerService;
 
-  this.restart = function() {
+  this.restart = function(options) {
     console.debug('Restarting');
-    this.selection.clear();
+    this.selection.clear(options);
     this.writer.setProgress(0);
     this.scanner.start(2000);
   };
@@ -481,7 +481,7 @@ pathModule.filter('basename', function() {
 /**
  * @module ResinEtcher.selection-state
  */
-
+var _ = require('lodash');
 var angular = require('angular');
 var selectionState = angular.module('ResinEtcher.selection-state', []);
 
@@ -588,20 +588,30 @@ selectionState.service('SelectionStateService', function() {
   };
 
   /**
-   * @summary Clear all selections
+   * @summary Clear selections
    * @function
    * @public
    *
+   * @param {Object} options - options
+   * @param {Boolean} [options.preserveImage] - preserve image
+   *
    * @example
    * SelectionStateService.clear();
+   *
+   * @example
+   * SelectionStateService.clear({ preserveImage: true });
    */
-  this.clear = function() {
-    selection = {};
+  this.clear = function(options) {
+    if (options && options.preserveImage) {
+      selection = _.pick(selection, 'image');
+    } else {
+      selection = {};
+    }
   };
 
 });
 
-},{"angular":9}],6:[function(require,module,exports){
+},{"angular":9,"lodash":10}],6:[function(require,module,exports){
 require('./ui-bootstrap-tpls');
 module.exports = 'ui.bootstrap';
 
--- a/lib/browser/modules/selection-state.js
+++ b/lib/browser/modules/selection-state.js
@@ -17,7 +17,7 @@
 /**
  * @module ResinEtcher.selection-state
  */
-
+var _ = require('lodash');
 var angular = require('angular');
 var selectionState = angular.module('ResinEtcher.selection-state', []);
 
@@ -124,15 +124,25 @@ selectionState.service('SelectionStateService', function() {
   };
 
   /**
-   * @summary Clear all selections
+   * @summary Clear selections
    * @function
    * @public
    *
+   * @param {Object} options - options
+   * @param {Boolean} [options.preserveImage] - preserve image
+   *
    * @example
    * SelectionStateService.clear();
+   *
+   * @example
+   * SelectionStateService.clear({ preserveImage: true });
    */
-  this.clear = function() {
-    selection = {};
+  this.clear = function(options) {
+    if (options && options.preserveImage) {
+      selection = _.pick(selection, 'image');
+    } else {
+      selection = {};
+    }
   };
 
 });
--- a/lib/index.html
+++ b/lib/index.html
@@ -97,9 +97,13 @@
               <p class="soft">Safely ejected and ready for use</p>
 
               <div class="space-vertical-large">
-                <hero-button ng-click="app.restart()">
+                <hero-button ng-click="app.restart({ preserveImage: true })">
                   <span class="glyphicon glyphicon-repeat" aria-hidden="true"></span>
-                  Lets do it again
+                  Burn image again
+                </hero-button>
+                <hero-button ng-click="app.restart()">
+                  <span class="glyphicon glyphicon-cd" aria-hidden="true"></span>
+                  Burn another image
                 </hero-button>
               </div>
             </div>
--- a/lib/index.html
+++ b/lib/index.html
@@ -15,10 +15,8 @@
     <script src="../build/browser/app.js"></script>
   </head>
   <body ng-app="ResinEtcher" ng-controller="AppController as app" style="display: none">
-
     <div class="content row middle-xs space-horizontal-large">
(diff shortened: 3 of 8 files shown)


lib/index.html
```diff
@@ -97,9 +97,13 @@
               <p class="soft">Safely ejected and ready for use</p>
 
               <div class="space-vertical-large">
-                <hero-button ng-click="app.restart()">
+                <hero-button ng-click="app.restart(app.selection.getImage())">
```
jviotti: This looks nice, however the meaning of passing an image to a restart function is not very straightforward at first sight. A more expressive approach would be allowing to pass an opt-in option to `app.restart()` asking to preserve the image, so we would do something like:

```html
<hero-button ng-click="app.restart({ preserveImage: true })">
```

Which reads a lot better and natural.

The `app.restart()` options could be directly passed to `app.selection.clear()`, where you can implement and unit test.

jviotti:
Thank you very much for the PR! A couple of comments:

- It's considered a good practice to squash commits that include very minor changes, review fixes, or commit updates. You can find more information here: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html.

- I noticed you put the number of the issue this PR fixes in the title. You can make GitHub understand, and automatically close the references issue once the PR is merged by adding it like this: `Fixes: #74` (see https://help.github.com/articles/closing-issues-via-commit-messages/). It's also nice to add a descriptive message, like "Add Burn image again button to completion screen. Fixes #74".

jviotti:
Would you also mind attaching a screenshot of the updated completion screen here, so we more easily evaluate design-wise?

fedealconada:
@jviotti Here is the screenshot of how it looks like now...
screen shot 2016-01-14 at 19 52 22

lib/browser/modules/selection-state.js
```diff
@@ -124,15 +124,26 @@ selectionState.service('SelectionStateService', function() {
   };
 
   /**
-   * @summary Clear all selections
+   * @summary Clear selections based on parameter passed
```
jviotti: I think just "Clear selections" is fine here.

lib/browser/modules/selection-state.js
```diff
@@ -124,15 +124,26 @@ selectionState.service('SelectionStateService', function() {
...
    * SelectionStateService.clear();
+   *
+   * @example
+   * SelectionStateService.clear(options);
    */
-  this.clear = function() {
-    selection = {};
+  this.clear = function(options) {
+    if (typeof options === 'object'){
+      if (options.preserveImage){
```
jviotti: You can shorten:

```js
this.clear = function(options) {
    if (typeof options === 'object'){
        if (options.preserveImage){
            ...
```

To:

```js
if (options && options.preserveImage) {
    ....
}
```

lib/browser/modules/selection-state.js
```diff
@@ -124,15 +124,26 @@ selectionState.service('SelectionStateService', function() {
...
+        delete selection.drive;
```
jviotti: Instead of saying "if preserve image, delete the drive", which sounds weird at first, but makes sense since `drive` is the only other possible selection, you can do something like this:

```js
if (options && options.preserveImage) {
    selection = _.pick(selection, 'image');
} else {
    selection = {};
}
```

This has the added benefit that if we decide to store something else than a drive or an image as a selection, we won't need to remember to modify this function to delete the new property.

You can check `_.pick` documentation here: https://lodash.com/docs#pick.

lib/browser/modules/selection-state.js
```diff
@@ -124,15 +124,26 @@ selectionState.service('SelectionStateService', function() {
...
+   *
    * @example
    * SelectionStateService.clear();
+   *
+   * @example
+   * SelectionStateService.clear(options);
    */
-  this.clear = function() {
-    selection = {};
+  this.clear = function(options) {
```
jviotti: The new behaviour of this function should be unit tested in `tests/browser/modules/selection-state.spec.js`: https://github.com/resin-io/resin-etcher/blob/master/tests/browser/modules/selection-state.spec.js#L154.

lib/browser/modules/selection-state.js
```diff
@@ -124,15 +125,24 @@ selectionState.service('SelectionStateService', function() {
...
+    if (options && options.preserveImage) {
+      selection = _.pick(selection, 'image');
+    }else{
```
jviotti: Can you add spacing between the curly braces?

```js
} else {
```

lib/index.html
```diff
@@ -95,12 +94,27 @@
...
               <p class="soft">Safely ejected and ready for use</p>
-
-              <div class="space-vertical-large">
-                <hero-button ng-click="app.restart()">
-                  <span class="glyphicon glyphicon-repeat" aria-hidden="true"></span>
-                  Lets do it again
-                </hero-button>
+              <div class="col-xs center-xs space-vertical-large">
+                <div class="col-xs-6 soft vertical-right-separator">
+                      <div class="row center-xs">
```
jviotti: There are some indentation issues in this part. Can you fix them?

tests/browser/modules/selection-state.spec.js
```diff
@@ -44,6 +44,36 @@ describe('Browser: SelectionState', function() {
 
     });
 
+    describe('given a burn-image-again state', function() {
```
jviotti: Notice that `SelectionStateService` doesn't know that "burn-image-again" means. In fact, it's completely isolated from the fact that the information it holds is used to burn something, therefore the title should be more generic, like "given the preserveImage option in clear()".

jviotti: Actually, notice the `.clear()` describe block in line 154. At that point, the nested `beforeEach` blocks setup a scenario where an image and a drive are selected, which is just what we need. I'd add a `describe('given the preserveImage option')` inside `describe('.clear()')` and move all the new test cases there.

lib/index.html
```diff
@@ -18,7 +18,6 @@
 
     <div class="content row middle-xs space-horizontal-large">
       <div class="col-xs">
-
```
jviotti: Is this accidental? Can you un-stage from the commit?

lib/browser/modules/selection-state.js
```diff
@@ -124,15 +125,24 @@ selectionState.service('SelectionStateService', function() {
   };
 
   /**
-   * @summary Clear all selections
+   * @summary Clear selections
    * @function
    * @public
    *
+   * @param {Object} options - options
```
jviotti: You can document the new option like this:

```js
@param {Object} options - options
@param {Boolean} [options.preserveImage] - preserve image
```

lib/browser/modules/selection-state.js
```diff
@@ -124,15 +125,24 @@ selectionState.service('SelectionStateService', function() {
...
+   *
    * @example
    * SelectionStateService.clear();
+   *
+   * @example
+   * SelectionStateService.clear(options);
```
jviotti: What about being more explicit in this example?

```js
@example
SelectionStateService.clear({ preserveImage: true });
```

build/browser/app.js
```diff
@@ -484,6 +484,7 @@ pathModule.filter('basename', function() {
 
 var angular = require('angular');
 var selectionState = angular.module('ResinEtcher.selection-state', []);
+var _ = require('lodash');
```
jviotti: Can you move this require statement above `var angular = require('angular');`?

jviotti:
@fedealconada The completion screen design still needs some work, but I'd really like to get the main functionality merged. Let's focus on https://github.com/fedealconada/resin-etcher/commit/b2a37c400b619deda64d4debf582dd7015437928 and we'll create another PR for the design.

jviotti:
@fedealconada I merged the commit successfully: https://github.com/resin-io/resin-etcher/commit/6331ee4aa1062133d62159af176e521c65f97c90! 

Sadly GitHub doesn't seems to detect this and mark the PR as merged automatically, so I'm closing it myself!

Thanks a lot for the contribution!

I'd be great if you can open a new PR to address the styling (don't forget to sync your fork first: https://help.github.com/articles/syncing-a-fork/)!

PR #104, opened 2016-01-19
Integrate Etcher with TrackJS

jviotti:


jviotti:
Confirmed it works fine if there is no internet connection.

PR #136, opened 2016-01-27
Change design. Fixes #74

### LINKED ISSUE #74: Add a "burn another" button on completion

fedealconada:


lib/index.html
```diff
@@ -99,16 +99,27 @@
...
-                <hero-button ng-click="app.restart({ preserveImage: true })">
-                  <span class="glyphicon glyphicon-repeat" aria-hidden="true"></span>
-                  Burn image again
-                </hero-button>
-                <hero-button ng-click="app.restart()">
-                  <span class="glyphicon glyphicon-cd" aria-hidden="true"></span>
-                  Burn another image
-                </hero-button>
+              <div class="col-xs center-xs space-vertical-large">
+                <div class="col-xs-6 soft vertical-right-separator">
```
jviotti: Having `col` as direct children of another `col` doesn't look right. `col` elements should only contain `box`. See http://flexboxgrid.com.

lib/index.html
```diff
@@ -99,16 +99,27 @@
...
+                  <div class="row center-xs">
```
jviotti: These should be `box` instead of `row`. `row` elements are meant to contain `col` elements.

lib/index.html
```diff
@@ -99,16 +99,27 @@
...
-              <div class="space-vertical-large">
-                <hero-button ng-click="app.restart({ preserveImage: true })">
-                  <span class="glyphicon glyphicon-repeat" aria-hidden="true"></span>
-                  Burn image again
-                </hero-button>
-                <hero-button ng-click="app.restart()">
-                  <span class="glyphicon glyphicon-cd" aria-hidden="true"></span>
-                  Burn another image
-                </hero-button>
+              <div class="col-xs center-xs space-vertical-large">
```
jviotti: Modifiers such as `center-xs` are only meant to be added to `row` elements. This pattern happens in various places below.

jviotti:
Closing due to inactivity. The relevant changes were made in https://github.com/resin-io/etcher/pull/143

PR #207, opened 2016-03-11
Enable OS X in Travis CI

jviotti:


jviotti:
Rejecting this PR since we'll be looking forward to do the signing in CircleCI instead.

Issue #280, opened 2016-04-11
Audio/visual cue on burn completion


jviotti:
Provide audio feedback when image writing is complete, or maybe a single dock bounce or system notification.

PR #283, opened 2016-04-11
Implement OS notifications on burn completion

jviotti:
Its helpful to have an auditive/visual cue when a burn operation
completed.

Instead of adding a setting entry to enable/disable notifications, you
can use the standard way to control notifications from your operating
system. For example, in OS X, you might go to "System Preferences" ->
"Notifications" and disable notifications for "Etcher".

Fixes: https://github.com/resin-io/etcher/issues/280


jviotti:
Pending notification text review.

PR #299, opened 2016-04-12
Use SUDO_USER environment variable for default file selection path.

abresas:
Etcher is usually run with sudo as it needs to permission to write
to removable drives. This results to electron opening root user's
file directories by default.

If etcher is run with UID 0 and SUDO_USER is set, with this solution
etcher will open /home/SUDO_USER on linux and /Users/SUDO_USER on
darwin.

This should not have any effect otherwise.

### DIFF
--- a/lib/src/dialog.js
+++ b/lib/src/dialog.js
@@ -22,6 +22,37 @@ const zipImage = require('resin-zip-image');
 const packageJSON = require('../../package.json');
 
 /**
+ * @summary Get default directory for file selection dialog.
+ * @function
+ * @private
+ *
+ * @description
+ * This function is meant to be used with electron dialog.showOpenDialog method.
+ *
+ * If etcher is run with sudo, it will return the sudo user's home directory,
+ * as long as the SUDO_USER environment variable is set.
+ *
+ * In every other case the function will return undefined so that electron
+ * decides what the default path should be.
+ *
+ * Works on linux and darwin platforms only.
+ */
+const getDefaultSelectionPath = function() {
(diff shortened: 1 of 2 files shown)


abresas:
@jviotti updated to use `user-homedir` module

lib/src/dialog.js
```diff
@@ -41,22 +65,26 @@ const packageJSON = require('../../package.json');
...
-      ]
-    }, function(files) {
-      return resolve(files || []);
-    });
+  exports.getDefaultSelectionPath()
+  .then(function(defaultPath) {
+    return new Bluebird(function(resolve) {
+      electron.dialog.showOpenDialog({
+        properties: [ 'openFile' ],
+	defaultPath: defaultPath,
```
jviotti: There seems to be an indentation issue here.

lib/src/dialog.js
```diff
@@ -19,9 +19,33 @@
...
+ *
+ * If etcher is run with sudo, it will return the sudo user's home directory,
+ * as long as the SUDO_USER environment variable is set.
+ *
+ * In every other case the function will return undefined so that electron
+ * decides what the default path should be.
+ *
+ * Works on linux and darwin platforms only.
+ */
+exports.getDefaultSelectionPath = Bluebird.method(function() {
```
jviotti: I believe this function can be removed now that we have all required functionality in `user-homedir`.

jviotti:
I'll be closing this PR since we're moving towards only elevating the process that performs the writing, which will solve the root issue.

Issue #332, opened 2016-04-15
Add more command line output (verbose/debug output and -h/--help)


mccollam:
In trying to debug a hang (which appears to be already filed as issue #307) I was unable to see any information about what Etcher was doing internally at any step.  I expected a "-v" or "--verbose" switch (or similar) to enable this but that did not work.

Additionally, running Etcher with "-h" or "--help" to find the list of supported command line switches produced no output, which is unusual for a tool of this kind.

jviotti:
Hi @mccollam, what operating system are you running?

The problem with the output is that to perform elevation on startup, the app first runs without privileges, and then spawns a process to run itself again with privileges once you provide your password to the sudo dialog.

Sadly the output streams are not inherited by the children process (maybe there is a way to do it on `sudo-prompt`, but I wonder how that would play in Windows, or OS X because of `open`), so you get no output.

I'm planning to add a command line option to omit the elevation dialog, so you can take care of it yourself, and thus prevent the app from forking a new process.

There is a fix coming for the issue you linked though, if you're interested, I can ping you when it lands to `master` so you can check yourself.

mccollam:
This is on Ubuntu 15.10 x86_64.

I don't see the behavior you are describing -- if I launch as a normal user I just get a dialog box telling me to rerun as root.

![image](https://cloud.githubusercontent.com/assets/3144447/14575555/86dc794e-0330-11e6-8094-a9bc880a79ba.png)

I was planning to try to submit a simple patch to provide -h at least but if it's already in the works I'll hold off. :)

jviotti:
Hi @mccollam ,

Yeah, GNU/Linux doesn't yet have the elevation dialog (see https://github.com/resin-io/etcher/issues/277) sadly, but in GNU/Linux you should be able to see logging. Since Electron runs in two processes: the main and renderer processes, things logged in one will not appear in the other. I believe this is the issue you have, since the `stream.push() after EOF` only happens in the renderer process. Notice you can open DevTools by pressing Ctrl+Alt+I.

> I was planning to try to submit a simple patch to provide -h at least but if it's already in the works I'll hold off. :)

Sounds great! By "working on it", I mean its planned, but work didn't start yet. I'd be awesome to see a patch :)

jviotti:
[...]

PR #333, opened 2016-04-16
Add support for command line arguments (including passing in an image file)

mccollam:
This allows parsing of arguments on the command line as well as passing in an image filename when invoking Etcher.  This streamlines the process when launching Etcher from the command line with an image in the current directory.

(I'm sure there's a better way to pass the filename in but I'm not sure what it is :/ )

I tested this on Ubuntu 15.04.  It closes issue #332.

### DIFF
--- a/lib/etcher.js
+++ b/lib/etcher.js
@@ -92,7 +92,9 @@ electron.app.on('ready', function() {
       }, 100);
 
       if (args._.length === 1) {
-        mainWindow.webContents.executeJavaScript('angular.element(document.getElementById("appcontainer")).scope().app.selectImage("' + args._[0] + '");');
+        mainWindow.webContents.executeJavaScript(
+          'angular.element(document.getElementById("appcontainer")).scope().app.selectImage("' + args._[0] + '");'
+        );
       }
     });
 
--- a/lib/etcher.js
+++ b/lib/etcher.js
@@ -43,7 +43,7 @@ electron.app.on('ready', function() {
 
   if (args._.length > 1) {
     console.log('Only one image can be specified.');
-    console.log('(See ' + process.argv[0] + ' -h for more information.)');
+    console.log('(Run "' + process.argv[0] + '" -h for more information.)');
     return process.exit(0);
   }
 
--- a/lib/etcher.js
+++ b/lib/etcher.js
@@ -19,11 +19,46 @@
 const electron = require('electron');
 const path = require('path');
 const elevate = require('./elevate');
+const fs = require('fs');
+const args = require('minimist')(process.argv.slice(2));
 let mainWindow = null;
 
 electron.app.on('window-all-closed', electron.app.quit);
 
 electron.app.on('ready', function() {
+
+  if (args.h || args.help) {
+    console.log('Etcher - A better way to flash OS images to SD cards');
+    console.log('http://etcher.io\n');
+    console.log('Usage:');
+    console.log('   ' + process.argv[0] + ' [options] [image]\n');
+    console.log('Options:');
+    console.log('   -h | --help     This information');
+    console.log('\n');
+    console.log('Supported image formats:');
+    console.log('   .img .iso .zip');
+    console.log('\n');
+    return process.exit(0);
+  }
+
+  if (args._.length > 1) {
+    console.log('Only one image can be specified.');
+    console.log('(See ' + process.argv[0] + ' -h for more information.)');
+    return process.exit(0);
+  }
+
+  if (args._.length === 1) {
+
+    // Check for image file existence and pass into the writer
+    var stat;
+    try {
+      stat = fs.statSync(args._[0]);
+    }
+    catch (e) {
+      console.log('Cannot find image ' + args._[0]);
+      process.exit(0);
+    }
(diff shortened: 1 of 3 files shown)


jviotti:
I'd love this to use something like [yargs](https://github.com/yargs/yargs) instead of `minimist` (which is used by `yargs` internally anyway).

Most of the things you implemented here can be shortened a lot and described in a more declarative way by using `yargs` "usage" feature (which auto generates available options for example), "aliases", etc.

lib/etcher.js
```diff
@@ -19,12 +19,47 @@
...
 const elevate = require('./elevate');
+const fs = require('fs');
+const args = require('minimist')(process.argv.slice(2));
 let mainWindow = null;
 
 electron.app.on('window-all-closed', electron.app.quit);
 
 electron.app.on('ready', function() {
 
+  if (args.h || args.help) {
```
jviotti: `yargs` aliases could take care of this automatically.

lib/etcher.js
```diff
@@ -19,12 +19,47 @@
...
+    console.log('Etcher - A better way to flash OS images to SD cards');
```
jviotti: `yards` "usage" feature could help a lot here. Also, by declaring the accepted options, the "Options" section can be automatically generated, which would help having the options and their documentation in sync.

lib/etcher.js
```diff
@@ -19,12 +19,47 @@
...
+    console.log('http://etcher.io\n');
+    console.log('Usage:');
+    console.log('   ' + process.argv[0] + ' [options] [image]\n');
+    console.log('Options:');
+    console.log('   -h | --help     This information');
+    console.log('\n');
+    console.log('Supported image formats:');
```
jviotti: You can use `yargs` "epilog" for this.

lib/etcher.js
```diff
@@ -19,12 +19,47 @@
...
+    console.log('   .img .iso .zip');
+    console.log('\n');
+    return process.exit(0);
+  }
+
+  if (args._.length > 1) {
```
jviotti: You could use `yargs` "check" feature to implement this.

lib/etcher.js
```diff
@@ -19,12 +19,47 @@
...
+    console.log('Only one image can be specified.');
+    console.log('(Run "' + process.argv[0] + '" -h for more information.)');
+    return process.exit(0);
+  }
+
+  if (args._.length === 1) {
+
+    // Check for image file existence and pass into the writer
```
jviotti: This could also be part of a "check" block.

lib/etcher.js
```diff
@@ -19,12 +19,47 @@
...
+    var stat;
+    try {
+      stat = fs.statSync(args._[0]);
```
jviotti: `fs.access()` is the recommended way to perform these kind of checks now that `fs.exists()` has been deprecated (https://github.com/nodejs/node/issues/1592). Alternatively, I'd be happy to include a new dependency that provides an "existence" check functionality (potentially around `fs.access()`) to more clearly communicate the intention.

For example, having a `fileExists(file)` call reads much nicer than a `stat` or `fs.access()`, for which you have to understand that the real reason you're calling that is to check that the file exists.

lib/etcher.js
```diff
@@ -56,6 +91,11 @@ electron.app.on('ready', function() {
         mainWindow.show();
       }, 100);
 
+      if (args._.length === 1) {
+        mainWindow.webContents.executeJavaScript(
+          'angular.element(document.getElementById("appcontainer")).scope().app.selectImage("' + args._[0] + '");'
```
jviotti: This looks OK, however I'd like to find another way to do this rather than relying on evaluating a string as JavaScript, and knowing too much about the controller's API from outside the Angular's world. I'm thinking about something like this:

- Have a new file under `src/`, which runs on the main process, to store the CLI options state.
- From the renderer process, remotely require this file (with `electron.remote`) when the Angular application starts to get the values. For modularisation reasons, creating a new Angular service to encapsulate the details of fetching the appropriate remote file would be great.
- Call the appropriate functions (like `selectImage()`) from the Angular's world, if certain options exist.

mccollam:
Awesome, thanks for the feedback!  I've moved things over to yargs (which is indeed a lot cleaner and fewer lines of code).  I'll start poking at cleaning up the option passing next.  Thank you for the pointers!

jviotti:
Awesome @mccollam , looking forward to that!

imrehg:
@mccollam out of curiosity, is this still a work in progress?

mccollam:
No -- I haven't poked at this in a while as I believe it is already solved in mainline Etcher at this point.  I could pick it back up if that's not the case -- I have learned a lot more about how Electron works since I started this!

jviotti:
This hasn't been implemented in Etcher yet, but I'd be glad to accept it if you pick it back up!

alexandrosm:
@mccollam shall we put this PR out of its misery?

mccollam:
@alexandrosm yes please; I thought this was already closed.  With Etcher having a CLI version this does not seem as necessary.

PR #390, opened 2016-05-02
Fix child writer process not working on packaged app

jviotti:
Setting `ELECTRON_RUN_AS_NODE=1` causes the node process to ignore the
default application location ('resources/app.asar') when no argument has
been passed.

This works fine when spawning the child process on develoment mode since
we're pointing to `lib/start.js` manually anyway, but fails when being
packaged since the electron binary tries to run the actual image as a
JavaScript program, failing with weird scary error messages.

As a workaround, we make use of an utility package called
`electron-is-running-in-asar`, which as the name implies, attempts to
detect whether the application is being ran from inside an asar package
or not. If the first case holds, we push the path to the `app.asar`
ourselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ before_install:
 
 install:
   - npm install
+  - npm install -g bower
+  - bower install
 
 before_script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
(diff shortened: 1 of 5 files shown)


jviotti:
Testing status:

- [ ] OS X in development mode.
- [ ] OS X in packaged mode.
- [ ] GNU/Linux in development mode.
- [ ] GNU/Linux in packaged mode.
- [ ] Windows in development mode.
- [ ] Windows in packaged mode.

PR #395, opened 2016-05-09
Implement cross-platform build scripts

jviotti:
With the addition of native addons to Etcher, its no longer possible to
package the application for all operating system in a single host.

Given we need to build on Windows, we're creating cross-platform bash
and batch scripts and deprecating the original Makefile.


jviotti:
We're lacking archiving back to `.asar` in Windows, but omitting the `.dlls` in the way.

PR #427, opened 2016-05-23
Upgrade electron to v1.1.1

jviotti:


jviotti:
There are some upgrade deprecations we must handle before this can get merged.

PR #454, opened 2016-06-07
Annotation for EncloseJS to ignore missing dependency

igorklopov:


jviotti:
Hi @igorklopov ,

Thanks for the annotation. I believe this is EncloseJS specific right? If so, is this documented anywhere (I couldn't find it in the EncloseJS website)? I'd like to add a comment explaining where this comes.

igorklopov:
Yes. It is EncloseJS-specific. Unfortunately docs are outdated. I will write full documentation soon. Please comment that it is just from EncloseJS :)

PR #542, opened 2016-06-28
docs: commit guidelines

jviotti:
This document will serve as a more expanded guide on the commit
conventions we are going to be introducing in this project mainly for
the purpose of `CHANGELOG.md` generation.

See: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit

### DIFF
--- a/docs/COMMIT_GUIDELINES.md
+++ b/docs/COMMIT_GUIDELINES.md
@@ -1,0 +1,248 @@
+Commit Guidelines
+=================
+
+We enforce certain rules on commits with the following goals in mind:
+
+- Be able to reliably auto-generate the `CHANGELOG.md` *without* any human
+intervention.
+- Be able to automatically and correctly increment the semver version number
+based on what was done since the last release.
+- Be able to get a quick overview of what happened to the project by glancing
+over the commit history.
+- Be able to automatically reference relevant changes from a dependency
+upgrade.
+
+The guidelines are inspired by the [AngularJS git commit guidelines][angular-commit-guidelines].
+
+Commit structure
+----------------
+
+Each commit message consists of a header, a body and a footer. The header has a
+special format that includes a type, a scope and a subject.
+
+```
+<type>(<scope>): <subject>
+<BLANK LINE>
+<body>
+<BLANK LINE>
+<footer>
+```
+
+The subject should not contain more than 70 characters, including the type and
+scope, and the body should be wrapped at 72 characters.
+
+Type
+----
+
+Must be one of the following:
+
+- `feat`: A new feature.
+- `fix`: A bug fix.
+- `minifix`: A minimal fix that doesn't warrant an entry in the CHANGELOG.
+- `docs`: Documentation only changes.
+- `style`: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc).
+- `refactor`: A code change that neither fixes a bug nor adds a feature.
+- `perf`: A code change that improves performance.
+- `test`: Adding missing tests.
+- `chore`: Changes to the build process or auxiliary tools and libraries.
+- `upgrade`: A version upgrade of a project dependency.
+
+Scope
+-----
+
+The scope is required for types that make sense, such as `feat`, `fix`,
+`test`, etc. Certain commit types, such as `chore` might not have a clearly
+defined scope, in which case its better to omit it.
+
+When it applies, the scope must be either `GUI` or `CLI`.
+
+A commit that takes part in both the GUI and CLI scopes, and makes more logical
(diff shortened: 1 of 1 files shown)


petrosagg:
Does this mean that if I don't include a `Changelog-Entry:` tag my change won't be included in the changelog?

docs/COMMIT_GUIDELINES.md
```diff
@@ -0,0 +1,245 @@
...
+-----
+
+The scope is required for types that make sense, such as `feat`, `fix`,
+`test`, etc. Certain commit types, such as `chore` might not have a clearly
+defined scope, in which case its better to omit it.
+
+When it applies, the scope must be either `GUI` or `CLI`.
+
+A commit that takes part in both the GUI and CLI scopes of the code should
+probably be split in commits that apply to each scope separately.
```
petrosagg: I think the commits should be split by logical change, so a change needs both GUI and CLI changes that depend on each other then the scope can be set to nothing

jviotti:
@petrosagg The commit title doesn't come into play for the generation of the CHANGELOG. The purpose is that the commit title can be set to something that makes more sense for a developer, while the `Changelog-Entry` can rephrase the commit title, if necessary, to make it user-oriented.

The addition of a separate tag to introduce the CHANGELOG entry text was motivated by the following reasons:

- Suppose I've just released `v1.0.0`. I then keep submitting PRs, and I introduced a subtle regression on one of them, that for some reason slipped undetected until several PRs later, but I still didn't release a new version. At that point, I'll submit a `fix:` commit that solve such regression, but that commit doesn't make sense to be included in the CHANGELOG for the new version, since it fixes a problem that happened in between.

- I believe the commit subject should describe the changes introduced in the commit in a nutshell, which might not be the best CHANGELOG entry in some cases (it works in most cases though, but still this is an edge case I encountered already a couple of times). For example, I had an issue with Windows where `EPERM` would be thrown when flashing to GPT drives. The fix was to use `diskpart`'s `clean` command instead of manually handle the deletion of the partition table, since GPT is duplicated at the end of the drive, etc. A nice commit subject, from the point of view of the developer, would be something like `Use diskpart clean command to erase partition table on Windows`, since that gives a very nice summary of what the commit does, without any possible confusion. Putting that on the CHANGELOG might not make a lot of sense for the user. Instead, something like `Fix EPERM errors when flashing to GPT drives in Windows` is way better. Of course, I could put that in the commit subject, but the previous one if much more descriptive to me. I encountered some other cases where the distinction was even deeper, but I can't remember right now.

- There can be very low-level fixes to the code that don't warrant an entry in the CHANGELOG, but are still fixes. If we blindly put everything that contains a `fix:` or `feat:` type, we'll have to delete them manually.

- The Angular's convention is to only include `fix:` and `feat:` commits in the CHANGELOG, but of course, there can be certain commits from other types that make a lot of sense in the CHANGELOG. Consider an important performance improvement, or very useful documentation addition that should have a place in the CHANGELOG.

Of course all these issues can be solved if we generate the CHANGELOG but then require the maintainers to inspect it before publishing (which is what the AngularJS team does), however taking care of these edge cases allows us to fully-automate the process.

docs/COMMIT_GUIDELINES.md
jviotti: Yeah maybe. At least in Etcher, the CLI and GUI code are very decoupled, and I very rarely encountered a case where such changes would make more sense together, but I can definitely see that happening in some cases. I'll fix the paragraph.

jviotti: Fixed!

alexandrosm:
@jviotti I believe your concerns are covered with the following two changes to the protocol:

1. introduce a type of ignorable fix label (fix-small, fix-ignore, minifix, etc) that doesn't make it to the changelog

2. the explicit changelog entry is used when present, but otherwise the commit message is.

It's critical that the scheme is easy/convenient to use, otherwise the team won't use it, and then it's for naught.

jviotti:
@alexandrosm 

> 1. introduce a type of ignorable fix label (fix-small, fix-ignore, minifix, etc) that doesn't make it to the changelog

I like this idea. In terms of wording, I personally like `minifix`, but happy to discuss if there are objections.

> 2. the explicit changelog entry is used when present, but otherwise the commit message is.

Makes perfect sense. So basically the rules for inclusion in the CHANGELOG would the following:

- The commit includes an explicit `Changelog-Entry`.
- The commit type is either `fix` or `feat`, and if it doesn't include a `Changelog-Entry`, the commit title is instead used.

alexandrosm:
SGTM. @petrosagg ?

jviotti:
Ping @petrosagg

PR #572, opened 2016-07-13
Use comments instead of plaintext in issue template

WasabiFan:
Doing it like this means that one does not have to delete the note before posting; it simply won't appear.


jviotti:
Ah, very nice! I didn't know we could use HTML comments in Markdown, thanks a lot!

Issue #571, opened 2016-07-13
"Unknown error" when writing .img.xz file on Windows


WasabiFan:
- **Etcher version:** 1.0.0-beta.10
- **Operating system and architecture:** Windows 10 x64, Insider Preview build 14385
- **Do you see any meaningful error information on DevTools?** Do I see anything? Yes. Meaningful? No. Just an error object with the same information as shown below.

When I attempt to flash an image (`.img.xz`) to an SD card, the operation often opens the UAC dialog for verification and then sits there with the message "Starting..." until I quit the application. When it gets past that, it progresses to around the 80% mark and then slows, before throwing an error (pictured below). The image that was written is corrupted, so Etcher did not successfully complete the imaging operation by the time the error was thrown.

I can't find any useful info in the dev console and there are no other logs as far as I can tell, so past this screenshot I don't have anything to go off of.

![image](https://cloud.githubusercontent.com/assets/3310349/16792661/43f834a8-487f-11e6-9ca3-74aeb1ba4228.png)

jviotti:
Hi @WasabiFan ,

Thanks for reporting this issue and the valuable information. Can you share a link to that exact same image you're trying to flash to see if I can reproduce? Also, you mention a `.img.xz`, but in the screenshot, I see an uncompressed `.img` instead.

Does this happen only with this image? If so, do you have another USB Drive or SD Card to try burning this exact same image to it, and see if the error still happens?

Regarding DevTools, is there a stack trace when expanding the Error object?

WasabiFan:
> Can you share a link to that exact same image you're trying to flash to see if I can reproduce?

Yep, it's [here](https://bintray.com/ev3dev/nightly/ev3dev-jessie#files). It is the one called `ev3-nightly-2016-07-13-bootroot.img.xz`. The purpose of burning these images is to test this new nightly build system, so it is possible that the image is invalid; keep that in mind. Obviously, regardless of its validity, the burner should never throw these errors if it can be helped.

> Also, you mention a .img.xz, but in the screenshot, I see an uncompressed .img instead.

Ah, unfortunate window placement :wink: It _is_ a `.img.xz`.

> Does this happen only with this image? If so, do you have another USB Drive or SD Card to try burning this exact same image to it, and see if the error still happens?

[...]

PR #579, opened 2016-07-19
fix(GUI): check drive size agains uncompressed image size

jviotti:
We were currently checking if a drive was large enough for an image by checking
the image file size, completely ignoring compression. This results on a small
chance of the uncompressed image not actually fitting in the drive, and
throwing more weird errors later on.

In order to mitigate this, we use the new `.getEstimatedFinalSize()` function
from `etcher-image-stream`, which returns the uncompressed size in the case of
compressed images.

Fixes: https://github.com/resin-io/etcher/issues/571
Change-Type: patch
Changelog-Entry: Check if drive is large enough using the final uncompressed size of the image.

### DIFF
--- a/lib/gui/os/dialog/services/dialog.js
+++ b/lib/gui/os/dialog/services/dialog.js
@@ -17,7 +17,7 @@
 'use strict';
 
 const _ = require('lodash');
-const fs = require('fs');
+const imageStream = require('etcher-image-stream');
 const electron = require('electron');
 
 module.exports = function($q, SupportedFormatsModel) {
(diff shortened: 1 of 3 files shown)


jviotti:
This PR fails because of the following issues:

- https://github.com/addaleax/lzma-native/issues/26
- https://github.com/addaleax/lzma-native/issues/25

jviotti:
Closing this since I've updated the shrink-wrap file in another PR, and got tons of conflicts which I don't feel like solving here. Replaying the extremely simple changes introduced from this PR in another branch based on the newest `master` is much easier.

PR #707, opened 2016-09-20
docs: etcher extended archives specification

jviotti:
This document aims to describe in detail the structure of an Etcher
Extended Archive.

### DIFF
--- a/docs/EXTENDED-ARCHIVES.md
+++ b/docs/EXTENDED-ARCHIVES.md
@@ -121,23 +121,28 @@ Here's an example of a real-world manifest for Raspbian Jessie
       "background": "#535760",
       "text": "#FFFFFF",
       "primary": "#5793db"
-    }
-  },
-  "image": {
-    "name": "Raspbian Jessie",
-    "version": "May 2016",
-    "url": "https://www.raspberrypi.org/downloads/raspbian/",
-    "supportUrl": "https://retropie.org.uk/forum/",
-    "logo": "raspbian.svg",
-    "checksumType": "sha1",
-    "path": "raspbian-jessie.img",
-    "bmap": "raspbian-jessie.img.bmap",
-    "instructions": "raspbian-jessie.markdown",
-    "releaseNotes": "raspbian-jessie-changelog.txt",
-    "configurationSchema": "raspbian-jessie-schema.json",
-    "recommendedDriveSize": 4294967296,
-    "updateUrl": "https://downloads.raspberrypi.org/raspbian_latest",
-    "etag": "c0170-53152af2-533d18ef29fc0"
+    },
+    "images": [
+      {
+        "name": "Raspbian Jessie",
+        "version": "May 2016",
+        "url": "https://www.raspberrypi.org/downloads/raspbian/",
+        "supportUrl": "https://retropie.org.uk/forum/",
+        "logo": "raspbian.svg",
+        "checksum": {
+          "type": "sha1",
+          "path": "raspbian-jessie-checksum.txt"
+        },
+        "path": "raspbian-jessie.img",
+        "bmap": "raspbian-jessie.img.bmap",
+        "instructions": "raspbian-jessie.markdown",
+        "releaseNotes": "raspbian-jessie-changelog.txt",
+        "configurationSchema": "raspbian-jessie-schema.json",
+        "recommendedDriveSize": 4294967296,
+        "updateUrl": "https://downloads.raspberrypi.org/raspbian_latest",
+        "etag": "c0170-53152af2-533d18ef29fc0"
+      }
+    ]
   }
 }
 ```
@@ -171,67 +176,71 @@ You may declare the following colors, in hexadecimal format:
 
 The client decides how to use them and where, if at all.
 
-#### `image.name (String)`
+#### `images[].name (String)`
 
 The human-friendly name of the image.
 
-#### `image.version (String)`
+#### `images[].version (String)`
 
 The version of the image.
 
-#### `image.url (String)`
+#### `images[].url (String)`
 
 The main url of the image.
 
-#### `image.supportUrl (String)`
+#### `images[].supportUrl (String)`
 
 The url where users can get general support for this image.
 
 This could be a link to a forum, IRC room, troubleshooting page, support form,
 etc.
 
-#### `image.logo (String)`
+#### `images[].logo (String)`
 
 The path to a logo that represents the image, relative to the `.meta`
 directory.
 
-#### `image.releaseDate (String)`
+#### `images[].releaseDate (String)`
 
 The release date timestamp. The date should conform to [ISO 8601][iso8601]
 standard.
 
-#### `image.checksumType (String)`
+#### `images[].checksum.type (String)`
 
 The checksum type. The current possible values are: `sha1`, `sha256`, `crc32`,
 and `md5`.
 
-See the `.meta/checksum` file.
+#### `images[].checksum.path (String)`
 
-#### `image.path (String)`
+The path to a file containing the checksum, relative to the root of the
+archive. This file **can't be inside `.meta/`** and should be included **at the
+tail of the archive**.
+
+#### `images[].path (String)`
 
 The path to the image, relative to the root of the archive.
 
-#### `image.bmap (String)`
+#### `images[].bmap (String)`
 
 The path a [`bmap`][bmap] file for the image, relative to the `.meta`
 directory.
 
-#### `image.instructions (String)`
+#### `images[].instructions (String)`
 
 The path a markdown post-flash instructions file for the image, relative to the
 `.meta` directory.
 
-#### `image.releaseNotes (String)`
+#### `images[].releaseNotes (String)`
 
 The path to a plain text file describing the image version's release notes,
 relative to the `.meta` directory.
 
-#### `image.configurationSchema (String)`
+#### `images[].configurationSchema (String)`
 
 The path to a [Reconfix][reconfix] image configuration schema, relative to the
 `.meta` directory.
 
-#### `image.recommendedDriveSize (Number)`
+#### `images[].recommendedDriveSize (Number)`
 
 The minimum recommended drive size to flash this image, in bytes.
 
@@ -239,7 +248,7 @@ The use case for this option is that while a drive might be large enough to
 contain the image, it might not be large enough to deliver a good experience
 when actually using the application or operating system contained in the image.
 
-#### `image.updateUrl (String)`
+#### `images[].updateUrl (String)`
 
 The url where the client can request the latest version of the image.
 
@@ -254,14 +263,14 @@ image.
 If the HTTP response from `updateUrl` doesn't contain an ETag this mechanism is
 ignored.
 
-#### `image.etag (String)`
+#### `images[].etag (String)`
 
 The [HTTP Etag][etag] of the current image.
 
 Clients may use this to check if the image resolved by `updateUrl` is different
 to the one we have locally.
 
-#### `image.expiryDate (String)`
+#### `images[].expiryDate (String)`
 
 The timestamp to determine the expiration date of the image. The date should
 conform to [ISO 8601][iso8601] standard.
@@ -284,15 +293,17 @@ For example:
     ..
     "extensions": {
       ..
-    }
+    },
     ..
-  },
-  "image": {
-    ..
-    "extensions": {
-      ..
-    }
-    ..
+    "images": [
+      {
+        ..
+        "extensions": {
+          ..
+        }
+        ..
+      }
+    ]
   }
 }
 ```
@@ -306,12 +317,13 @@ compressed bytes to the drive.
 
 This file should be references by `.meta/manifest.json`'s `path` property.
 
-### `.meta/checksum`
+### `.meta/<checksum file name>`
 
 For example: `64c7ed611929ea5178fbb69b5a5f29cc9cc7c157`
 
 This file contains the checksum matching the type specified in the
-`checksumType` field of the image section of `.meta/manifest.json`.
+`images[].checksum.type` field of an image section of `.meta/manifest.json`,
+and it should be referenced by the `images[].checksum.path` property.
 
 The reason the actual checksum is at the end of the archive is that if any tool
 modifies the image inside the extended archive, it will need to update the
--- a/docs/EXTENDED-ARCHIVES.md
+++ b/docs/EXTENDED-ARCHIVES.md
@@ -75,13 +75,10 @@ Archive layout
 The contents of an extended archive are the followings:
 
 - **REQUIRED** `.meta/manifest.json`
-- `.meta/publisher(2x)?.[svg|png]`
-- `.meta/image(2x)?.[svg|png]`
 - `.meta/schema.json`
-- `.meta/image.bmap`
 - `.meta/instructions.markdown`
 - `.meta/release-notes.txt`
-- **REQUIRED** `image.<extension>`
+- **REQUIRED** `<image name>.<extension>`
 - `.meta/checksum`
 
 ### Order
@@ -119,6 +116,7 @@ Here's an example of a real-world publisher manifest for RetroPie:
 {
   "name": "RetroPie",
   "url": "https://retropie.org.uk",
+  "logo": "retropie.svg",
   "supportUrl": "https://retropie.org.uk/forum/",
   "colorScheme": {
     "background": "#535760",
@@ -137,6 +135,11 @@ The display human-friendly name of the publisher.
 ##### `url (String)`
 
 The main url of the publisher, usually the landing page.
+
+##### `logo (String)`
+
+The path to a logo that represents the publisher, relative to the `.meta`
+directory.
 
 ##### `supportUrl (String)`
 
@@ -168,7 +171,10 @@ Here's an example of a real-world image manifest for Raspbian Jessie:
   "name": "Raspbian Jessie",
   "version": "May 2016",
   "url": "https://www.raspberrypi.org/downloads/raspbian/",
+  "logo": "raspbian.svg",
   "checksumType": "sha1",
+  "path": "raspbian-jessie.img",
+  "bmap": "raspbian-jessie.img.bmap",
   "recommendedDriveSize": 4294967296,
   "updateUrl": "https://downloads.raspberrypi.org/raspbian_latest",
   "etag": "c0170-53152af2-533d18ef29fc0"
@@ -189,6 +195,11 @@ The version of the image.
 
 The main url of the image.
 
+##### `logo (String)`
+
+The path to a logo that represents the image, relative to the `.meta`
+directory.
+
 ##### `releaseDate (String)`
 
 The release date timestamp. The date should conform to [ISO 8601][iso8601]
@@ -200,6 +211,15 @@ The checksum type. The current possible values are: `sha1`, `sha256`, `crc32`,
 and `md5`.
 
 See the `.meta/checksum` file.
+
+##### `path (String)`
+
+The path to the image, relative to the root of the archive.
+
+##### `bmap (String)`
+
+The path a [`bmap`][bmap] file for the image, relative to the `.meta`
+directory.
 
 ##### `recommendedDriveSize (Number)`
 
@@ -267,21 +287,9 @@ For example:
 }
 ```
 
-### `.meta/publisher(2x)?.[svg|png]`
-
-A graphic that represents the image publisher.
-
-### `.meta/image(2x)?.[svg|png]`
-
-A graphic that represents the image.
-
 ### `.meta/schema.json`
 
 A [Reconfix][reconfix] image configuration schema.
-
-### `.meta/image.bmap`
-
-A [bmap][bmap] file for the extended archive image.
 
 ### `.meta/instructions.markdown`
 
@@ -291,15 +299,14 @@ A markdown file including post-flash instructions.
 
 A plain text file describing the image version's release notes.
 
-### `image.<extension>`
+### `<image name>.<extension>`
 
-The image file **in uncompressed form**. Clients should not attempt to
+An image file **in uncompressed form**. Clients should not attempt to
 decompress the image file since the extended archive already includes
 compression. If the image is compressed, then the client should write
 compressed bytes to the drive.
 
-An archive containing more than one image file is not a well-formed archive. If
-this is the case, use only the first image.
+This file should be references by `.meta/manifest.json`'s `path` property.
 
 ### `.meta/checksum`
 
--- a/docs/EXTENDED-ARCHIVES.md
+++ b/docs/EXTENDED-ARCHIVES.md
@@ -3,22 +3,16 @@ Etcher Extended Archives
 
 Etcher extended archives are archive files including an image and a special
 metadata directory including information that describes the image, device type,
-and vendor in detail, enabling Etcher to make use of such information to
+and publisher in detail, enabling Etcher to make use of such information to
 deliver advanced features and a more pleasant experience to the user.
 
 Entities
 --------
 
-### Vendor
-
-A vendor represents a manufacturer of a device type and potential publisher.
-
-*Some examples of image vendors are: Beagleboard, Raspberry Pi Foundation,
-Intel, and ODroid.*
-
 ### Publisher
 
-A publisher represents an entity that maintains and publishes official images.
+A publisher represents an entity that maintains and publishes operating system
+images.
 
 *Some examples of image publishers are: Tizen, RetroPie, Resin.io, and CoreOS.*
 
@@ -50,7 +44,8 @@ The file format should have the following characteristics:
 We base this format on the [`tar`][tar] archive format, which has the following
 benefits:
 
-- It's a nearly universal archive format (easy to find robust implementations)
+- It's a common archive format, so it'd be easy to find robust modules to work
+  with in any platform ecosystem
 
 - It allows us to easily impose a file order, which greatly helps with
   streamability. This is not trivial with `zip`, for instance
@@ -80,9 +75,9 @@ Archive layout
 The contents of an extended archive are the followings:
 
 - **REQUIRED** `.meta/manifest.json`
-- `.meta/publisher.svg`
-- `.meta/device.svg`
-- `.meta/image.svg`
+- `.meta/publisher(2x)?.[svg|png]`
+- `.meta/device(2x)?.[svg|png]`
+- `.meta/image(2x)?.[svg|png]`
 - `.meta/schema.json`
 - `.meta/image.bmap`
 - `.meta/instructions.markdown`
@@ -101,9 +96,10 @@ described above ensures the maximum UX perceived efficiency.
 
 ### Graphics
 
-We chose [`svg`][svg] over other image formats since clients can display it on
-any kind of screen resolution without requiring multi-layer files, or multiple
-files workarounds.
+Any graphic included in the archive can be either an [`svg`][svg] file, or a
+[`png`][png] file with the `<filename>.png` and `<filename>@2x.png` convention.
+
+If both are present, [`svg`][svg] should be favoured by the client.
 
 Archive files
 -------------
@@ -164,13 +160,15 @@ You may declare the following colors, in hexadecimal format:
 - `text`: The text color
 - `primary`: The primary color
 
+The client decides how to use them and where, if at all.
+
 #### Device
 
 Here's an example of a real-world device manifest for Raspberry Pi 3:
 
 ```json
 {
-  "displayName": "Raspberry Pi 3",
+  "name": "Raspberry Pi 3",
   "url": "https://www.raspberrypi.org/products/raspberry-pi-3-model-b/",
   "arch": "armv7hf",
   "slug": "raspberrypi3",
@@ -183,7 +181,7 @@ Here's an example of a real-world device manifest for Raspberry Pi 3:
 
 Describe a device with any of the following properties:
 
-##### `displayName (String)`
+##### `name (String)`
 
 The human-friendly name of the device.
 
@@ -191,7 +189,7 @@ The human-friendly name of the device.
 
 The url to the landing page of the device.
 
-##### `arch (String[])`
+##### `arch (String)`
 
 The device architecture.
 
@@ -242,10 +240,9 @@ Here's an example of a real-world image manifest for Raspbian Jessie:
 
 ```json
 {
-  "displayName": "Raspbian Jessie",
+  "name": "Raspbian Jessie",
   "version": "May 2016",
   "url": "https://www.raspberrypi.org/downloads/raspbian/",
-  "releaseNotesUrl": "http://downloads.raspberrypi.org/raspbian/release_notes.txt",
   "checksumType": "sha1",
   "recommendedDriveSize": 4294967296,
   "updateUrl": "https://downloads.raspberrypi.org/raspbian_latest",
@@ -260,7 +257,7 @@ Here's an example of a real-world image manifest for Raspbian Jessie:
 
 Describe an image with any of the following properties:
 
-##### `displayName (String)`
+##### `name (String)`
 
 The human-friendly name of the image.
 
@@ -272,11 +269,6 @@ The version of the image.
 
 The main url of the image.
 
-##### `releaseNotesUrl (String)`
-
-The url to the image's release notes or CHANGELOG. This URL will take
-precedence over `.meta/release-notes.txt` if defined.
-
 ##### `releaseDate (String)`
 
 The release date timestamp. The date should conform to [ISO 8601][iso8601]
@@ -284,7 +276,8 @@ standard.
 
 ##### `checksumType (String)`
 
-The checksum type. The current possible values are: `sha1`, `crc32`, and `md5`.
+The checksum type. The current possible values are: `sha1`, `sha256`, `crc32`,
+and `md5`.
 
 See the `.meta/checksum` file.
 
@@ -294,7 +287,7 @@ The minimum recommended drive size to flash this image, in bytes.
 
 The use case for this option is that while a drive might be large enough to
 contain the image, it might not be large enough to deliver a good experience
-when actual using the application or operating system contained in the image.
+when actually using the application or operating system contained in the image.
 
 ##### `supportedDevices (String[])`
 
@@ -327,7 +320,7 @@ to the one we have locally.
 The timestamp to determine the expiration date of the image. The date should
 conform to [ISO 8601][iso8601] standard.
 
-Clients may use this information to refuse flashing the image.
+Clients may use this information to warn against flashing the image.
 
 #### Extensions
 
@@ -365,13 +358,17 @@ For example:
 }
 ```
 
-### `.meta/publisher.svg`
+### `.meta/publisher(2x)?.[svg|png]`
 
-An [`svg`][svg] logo of the publisher.
+A graphic that represents the image publisher.
 
-### `.meta/device.svg`
+### `.meta/device(2x)?.[svg|png]`
 
-An [`svg`][svg] logo of the device.
+A graphic that represents the target device.
+
+### `.meta/image(2x)?.[svg|png]`
+
+A graphic that represents the image.
 
 ### `.meta/schema.json`
 
@@ -387,10 +384,7 @@ A markdown file including post-flash instructions.
 
 ### `.meta/release-notes.txt`
 
-A plain text file describing the image version's release notes. The client
-should read this file only if the `releaseNotesUrl` on the `image` section of
-the `.meta/manifest.json` is not defined, returns an unsuccessful HTTP code,
-or the client is offline.
+A plain text file describing the image version's release notes.
 
 ### `image.<extension>`
 
@@ -422,6 +416,7 @@ image has been processed.
 [nodejs]: https://nodejs.org
 [lzma-native]: https://github.com/addaleax/lzma-native
 [svg]: https://developer.mozilla.org/en-US/docs/Web/SVG
+[png]: http://www.libpng.org/pub/png/
 [reconfix]: https://github.com/resin-io/reconfix
 [bmap]: https://source.tizen.org/documentation/reference/bmaptool/introduction
 [iso8601]: https://en.wikipedia.org/wiki/ISO_8601
--- a/docs/EXTENDED-ARCHIVES.md
+++ b/docs/EXTENDED-ARCHIVES.md
@@ -76,7 +76,6 @@ The contents of an extended archive are the followings:
 
 - **REQUIRED** `.meta/manifest.json`
 - `.meta/publisher(2x)?.[svg|png]`
-- `.meta/device(2x)?.[svg|png]`
 - `.meta/image(2x)?.[svg|png]`
 - `.meta/schema.json`
 - `.meta/image.bmap`
@@ -107,11 +106,9 @@ Archive files
 ### `.meta/manifest.json`
 
 This is the only required file inside `.meta`, which declares information about
-the image, the target device type, and the publisher. This file may contain the
-following top level properties:
+the image. This file may contain the following top level properties:
 
 - `publisher`: Information about the image publisher
-- `device`: Information about the image device type
 - `image`: Information about the image itself
 
 #### Publisher
@@ -162,78 +159,6 @@ You may declare the following colors, in hexadecimal format:
 
 The client decides how to use them and where, if at all.
 
-#### Device
-
-Here's an example of a real-world device manifest for Raspberry Pi 3:
-
-```json
-{
-  "name": "Raspberry Pi 3",
-  "url": "https://www.raspberrypi.org/products/raspberry-pi-3-model-b/",
-  "arch": "armv7hf",
-  "slug": "raspberrypi3",
-  "slugAliases": [
-    "raspberry-pi3",
-    "rpi3"
-  ]
-}
-```
-
-Describe a device with any of the following properties:
-
-##### `name (String)`
-
-The human-friendly name of the device.
-
-##### `url (String)`
-
-The url to the landing page of the device.
-
-##### `arch (String)`
-
-The device architecture.
-
-The current possible values are: `aarch64`, `amd64`, `armv5e`, `armv6`,
-`armv7hf`, and `i386`.
-
-##### `slug (String)`
-
-The device slug. Officially recognised slugs include:
-
-- `artik10`
-- `artik5`
-- `artik710`
-- `beaglebone-black`
-- `beaglebone-green-wifi`
-- `beaglebone-green`
-- `colibri-imx6dl`
-- `cybertan-ze250`
-- `edge`
-- `hummingboard`
-- `intel-edison`
-- `intel-nuc`
-- `nitrogen6x`
-- `odroid-c1`
-- `odroid-xu4`
-- `parallella`
-- `qemux86-64`
-- `qemux86`
-- `raspberry-pi`
-- `raspberry-pi2`
-- `raspberrypi3`
-- `ts4900`
-- `ts7700`
-- `via-vab820-quad`
-- `zynq-xz702`
-
-##### `slugAliases (String[])`
-
-The device slug aliases, if any.
-
-##### `compatibleWith (String[])`
-
-An array of the device slugs this device is compatible with.
-
 #### Image
 
 Here's an example of a real-world image manifest for Raspbian Jessie:
@@ -246,12 +171,7 @@ Here's an example of a real-world image manifest for Raspbian Jessie:
   "checksumType": "sha1",
   "recommendedDriveSize": 4294967296,
   "updateUrl": "https://downloads.raspberrypi.org/raspbian_latest",
-  "etag": "c0170-53152af2-533d18ef29fc0",
-  "supportedDevices": [
-    "raspberrypi",
-    "raspberrypi2",
-    "raspberrypi3"
-  ]
+  "etag": "c0170-53152af2-533d18ef29fc0"
 }
 ```
 
@@ -288,10 +208,6 @@ The minimum recommended drive size to flash this image, in bytes.
 The use case for this option is that while a drive might be large enough to
 contain the image, it might not be large enough to deliver a good experience
 when actually using the application or operating system contained in the image.
-
-##### `supportedDevices (String[])`
-
-An array of the device slugs this image supports.
 
 ##### `updateUrl (String)`
 
@@ -341,13 +257,6 @@ For example:
     }
     ..
   },
-  "device": {
-    ..
-    "extensions": {
-      ..
-    }
-    ..
-  },
   "image": {
     ..
     "extensions": {
@@ -361,10 +270,6 @@ For example:
 ### `.meta/publisher(2x)?.[svg|png]`
 
 A graphic that represents the image publisher.
-
-### `.meta/device(2x)?.[svg|png]`
-
-A graphic that represents the target device.
 
 ### `.meta/image(2x)?.[svg|png]`
 
--- a/docs/EXTENDED-ARCHIVES.md
+++ b/docs/EXTENDED-ARCHIVES.md
@@ -1,0 +1,428 @@
+Etcher Extended Archives
+========================
+
+Etcher extended archives are archive files including an image and a special
+metadata directory including information that describes the image, device type,
+and vendor in detail, enabling Etcher to make use of such information to
+deliver advanced features and a more pleasant experience to the user.
+
+Entities
+--------
+
+### Vendor
+
+A vendor represents a manufacturer of a device type and potential publisher.
+
+*Some examples of image vendors are: Beagleboard, Raspberry Pi Foundation,
+Intel, and ODroid.*
+
+### Publisher
+
+A publisher represents an entity that maintains and publishes official images.
+
+*Some examples of image publishers are: Tizen, RetroPie, Resin.io, and CoreOS.*
+
+### Device
+
+A device represents a specific device type manufactured by a vendor.
+
+*Some examples of devices are: Beaglebone Black, Beaglebone Green, Intel
+Edison, and Raspberry Pi 3.*
+
+### Client
+
+A client represents the program making use of the information about the other
+entities.
+
+*At the moment of this writing, the only existent client is the Etcher application.*
+
+File format
+-----------
+
+The file format should have the following characteristics:
+
+- Streamability: The extended archive allows the client to consume all its
+  information in a streaming fashion
+
+- Compression: The extended archive should be compressed in order to save
+  network resources
+
+We base this format on the [`tar`][tar] archive format, which has the following
+benefits:
+
+- It's a nearly universal archive format (easy to find robust implementations)
+
+- It allows us to easily impose a file order, which greatly helps with
+  streamability. This is not trivial with `zip`, for instance
+
+- It precedes each file with a header construct, making it streaming friendly
+
+The [`tar`][tar] archive is then compressed with the [`xz`][xz] data
+compression format, which has the following benefits:
+
+- A native, robust, performant, cross-platform and streaming-based
+  decompression library for [node.js][nodejs], called
+  [`lzma-native`][lzma-native], which we battle tested at Etcher
+
+- A way to calculate the precise uncompressed size of the file no matter the
+  size (which is a [known limitation of `gzip` after 4
+  GB](https://github.com/resin-io/etcher/issues/638)), and an operation
+  [`lzma-native`][lzma-native] supports out of the box. This is not possible
+  with `bzip2`
+
+- Decent compression ratio, and compression/decompression time
+
+The resulting file should contain the `.etch` extension.
+
+Archive layout
+--------------
+
+The contents of an extended archive are the followings:
+
+- **REQUIRED** `.meta/manifest.json`
+- `.meta/publisher.svg`
+- `.meta/device.svg`
+- `.meta/image.svg`
+- `.meta/schema.json`
+- `.meta/image.bmap`
+- `.meta/instructions.markdown`
+- `.meta/release-notes.txt`
+- **REQUIRED** `image.<extension>`
+- `.meta/checksum`
+
+### Order
+
+Notice that the order is important. The `.meta/` directory must come before the
+image data, since we need that information before being able to make use of the
+image.
+
+The order of the files inside `.meta/` is not mandatory, however the order
+described above ensures the maximum UX perceived efficiency.
+
+### Graphics
+
+We chose [`svg`][svg] over other image formats since clients can display it on
+any kind of screen resolution without requiring multi-layer files, or multiple
+files workarounds.
+
+Archive files
+-------------
+
+### `.meta/manifest.json`
+
+This is the only required file inside `.meta`, which declares information about
+the image, the target device type, and the publisher. This file may contain the
+following top level properties:
+
+- `publisher`: Information about the image publisher
+- `device`: Information about the image device type
+- `image`: Information about the image itself
+
+#### Publisher
+
+Here's an example of a real-world publisher manifest for RetroPie:
+
+```json
+{
+  "name": "RetroPie",
+  "url": "https://retropie.org.uk",
+  "supportUrl": "https://retropie.org.uk/forum/",
+  "colorScheme": {
+    "background": "#535760",
+    "text": "#FFFFFF",
+    "primary": "#5793db"
+  }
+}
+```
+
+Describe a publisher with any of the following properties:
+
+##### `name (String)`
+
+The display human-friendly name of the publisher.
+
+##### `url (String)`
+
+The main url of the publisher, usually the landing page.
+
+##### `supportUrl (String)`
+
+The publisher url where users can get general support.
+
+This could be a link to a forum, IRC room, troubleshooting page, support form,
+etc.
+
+##### `colorScheme (Object)`
+
+The publisher specific color scheme, if any.
+
+The client may allow a certain degree of branding based on this color scheme.
+
+You may declare the following colors, in hexadecimal format:
+
+- `background`: The background color
+- `text`: The text color
+- `primary`: The primary color
+
+#### Device
+
+Here's an example of a real-world device manifest for Raspberry Pi 3:
+
+```json
+{
+  "displayName": "Raspberry Pi 3",
+  "url": "https://www.raspberrypi.org/products/raspberry-pi-3-model-b/",
+  "arch": "armv7hf",
+  "slug": "raspberrypi3",
+  "slugAliases": [
+    "raspberry-pi3",
+    "rpi3"
+  ]
+}
+```
+
+Describe a device with any of the following properties:
+
+##### `displayName (String)`
+
+The human-friendly name of the device.
+
+##### `url (String)`
+
+The url to the landing page of the device.
+
+##### `arch (String[])`
+
+The device architecture.
+
+The current possible values are: `aarch64`, `amd64`, `armv5e`, `armv6`,
+`armv7hf`, and `i386`.
+
+##### `slug (String)`
+
+The device slug. Officially recognised slugs include:
+
+- `artik10`
+- `artik5`
+- `artik710`
+- `beaglebone-black`
+- `beaglebone-green-wifi`
+- `beaglebone-green`
+- `colibri-imx6dl`
+- `cybertan-ze250`
+- `edge`
+- `hummingboard`
+- `intel-edison`
+- `intel-nuc`
+- `nitrogen6x`
+- `odroid-c1`
+- `odroid-xu4`
+- `parallella`
+- `qemux86-64`
+- `qemux86`
+- `raspberry-pi`
+- `raspberry-pi2`
+- `raspberrypi3`
+- `ts4900`
+- `ts7700`
+- `via-vab820-quad`
+- `zynq-xz702`
+
+##### `slugAliases (String[])`
+
+The device slug aliases, if any.
+
+##### `compatibleWith (String[])`
+
+An array of the device slugs this device is compatible with.
+
+#### Image
+
+Here's an example of a real-world image manifest for Raspbian Jessie:
+
+```json
+{
+  "displayName": "Raspbian Jessie",
+  "version": "May 2016",
+  "url": "https://www.raspberrypi.org/downloads/raspbian/",
+  "releaseNotesUrl": "http://downloads.raspberrypi.org/raspbian/release_notes.txt",
+  "checksumType": "sha1",
+  "recommendedDriveSize": 4294967296,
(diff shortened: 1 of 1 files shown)


docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+should conform to [ISO 8601][iso8601] standard.
+
+Clients may use this information to refuse flashing the image.
+
+Client
+------
+
+> A client represents the program making use of the information about the other
+  entities. This is not something publishers declare themselves.
+
```
lurch: Huh? I don't see what the client manifest has to do with anything... Does this info belong in a separate page?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+The current possible values are: `arm`, `arm64`, `ia32`, `mips`, `mipsel`,
+`ppc`, `ppc64`, `s390`, `s390x`, `x32`, `x64`, and `x86`.
+
+Extensions
+----------
+
+Each of the entities described above accept an object property called
+`extensions`, which you can use to provide your own custom properties to the
+manifests, in a way that are specific to your application. It is every
```
lurch: every -> very

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+encouraged that you put those custom properties inside the `extensions`
+namespace to avoid potential clashes with other properties that may be
+introduced in the future.
```
lurch: Could you give an example of this?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+
+_Some examples of images are: Raspbian Jessie, ResinOS, CoreOS, and Ostro._
+
+Here's an example of a real-world image manifest for Raspbian Jessie:
+
+```json
+{
+  "displayName": "Raspbian Jessie",
+  "version": "May 2016",
+  "url": "https://www.raspbian.org",
```
lurch: This should actually be "https://www.raspberrypi.org/downloads/raspbian/" because the RPF-version of Raspbian (which most people will be using) is different to the 'pure' upstream Raspbian from "https://www.raspbian.org"

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+  "url": "http://beagleboard.org",
+  "supportUrl": "http://beagleboard.org/support/",
+  "logoPath": "vendor.svg",
+  "colorScheme": {
+    "backgroundColor": "#535760",
+    "foregroundColor": "#FFFFFF",
+    "primaryColor": "#5793db"
+  }
+}
+```
```
lurch: The example JSON snippets here and elsewhere should probably be shown how they would appear inside the final `manifest.json` (because people love copy'n'pasting) e.g.
```json
{
  "vendor": {
    "name": "Beagleboard",
    "url": "http://beagleboard.org",
    "supportUrl": "http://beagleboard.org/support/",
    "logoPath": "vendor.svg",
    "colorScheme": {
      "backgroundColor": "#535760",
      "foregroundColor": "#FFFFFF",
      "primaryColor": "#5793db"
    }
  }
}
 ```

File: docs/EXTENDED-ARCHIVES.md [hunk 6]
```diff
@@ -0,0 +1,344 @@
...
+
+Each of the top level properties are described in detail below.
+
+Vendor
+------
+
+> A vendor represents a manufacturer of a device type and potential publisher
+  of official images.
```
alexandrosm: it's the publisher. In fact, we may want to rename "vendor" to "publisher" to avoid confusion.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+### `displayName (String)`
+
+The human-friendly name of the device.
+
+### `arch (String[])`
+
+The device architecture.
+
+The current possible values are: `aarch64`, `amd64`, `armv5e`, `armv7hf`,
+`i386`, and `rpi`.
```
alexandrosm: ah.. "rpi" is a legacy string inside the resin.io stack that basically stands for "armv6". But I agree, that shouldn't be in here.

docs/EXTENDED-ARCHIVES.md
alexandrosm: Fair. Though obviously this will have whatever the publisher decides to put in it, in the general case :P

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+
+The version of the image.
+
+### `url (String)`
+
+The main url of the image.
+
+### `releaseNotesUrl (String)`
+
+The url to the image's release notes or CHANGELOG.
```
alexandrosm: Nice!

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+An array of the device slugs this image supports.
+
+### `updateUrl (String)`
+
+The url where the client can request the latest version of the image.
+
+The way this works is that you provide a generic url that redirects to the
+latest image resource. The client will follow all redirects until it finds the
+final image, and will check if the resource matches the currently selected
+image.
```
alexandrosm: I think the way we had thought of it is that etags not matching means the user gets a "there's a newer version of the image available. do you want to download it?"

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+
+### `etag (String)`
+
+The [HTTP Etag][etag] of the current image.
+
+This can be used to check if the image resolved by `updateUrl` is different to
+the one we have locally.
```
alexandrosm: then the mechanism wouldn't work :(

docs/EXTENDED-ARCHIVES.md
alexandrosm: the idea here (and also in the "client" stuff mentioned above) is that this data should be made available to the instructions page to pick from. Not sure if this was discussed in this doc, but part of it is having the ability to embed post-write instructions. Those may often be related to the OS of the user, or need added context (like the stuff that goes in the "extensions" segment. Also, if others want to process the .meta file for other reasons, having this "freeform space" will prevent the publisher from adding top-level properties that may lead to namespace collisions with future features of the extended archives.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+
+Declaring a vendor-specific color scheme allows a certain degree of branding to
+be applied to the application on the fly as soon as a vendored image is
+selected.
+
+You may declare the following colors, in hexadecimal format:
+
+- `backgroundColor`: The application background color
+- `foregroundColor`: The application foreground color
+- `primaryColor`: The application primary color
```
alexandrosm: Or perhaps we should find a name that's cleaner :). This has slightly bothered me too, but now that someone else raises it too, maybe it'd make sense to get this cleared up.

docs/EXTENDED-ARCHIVES.md
lurch: "I think the way we had thought of it..."
Yeah, that's what I assumed, but it might be helpful to "spell it out" in this doc ;)

I guess the problem with etags is you can only tell if they're the same, or different.
If the locally-downloaded image is of a beta-version, but the version hosted at the `updateUrl` is an older stable version, I guess you'd be told the online version is newer, even if it's not?

docs/EXTENDED-ARCHIVES.md
lurch: Given that this doc is suggesting adding a `manifest.json` inside the ZIP image file, perhaps it would make sense to also have an (optional) `updateManifestUrl` field, which would point to an online `manifest.json` file, which would allow richer version comparisons?
For comparison: https://downloads.raspberrypi.org/os_list_v3.json and https://downloads.raspberrypi.org/raspbian/os.json (and also https://downloads.raspberrypi.org/raspbian/partitions.json )

docs/EXTENDED-ARCHIVES.md
lurch: I guess I was asking more for a syntax-example (again, because people love copying'n'pasting), rather than a what-would-it-be-used-for example ;-)
The doc talks about both "properties" and "namespace" so I guess the intended syntax isn't very clear.

docs/EXTENDED-ARCHIVES.md
lurch: ...and in the "general case", the RPF currently don't provide bmap files ;-)

docs/EXTENDED-ARCHIVES.md
lurch: Cool. I guess there's also no reason not to add other possible arch-es in here too?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+
+Device
+------
```
alexandrosm: @jviotti -- thought we had support for that kind of relationship in the format? Any idea how it would be expressed?

docs/EXTENDED-ARCHIVES.md
alexandrosm: Depends what you mean. ARM has been pretty profilgate with architectures and I would like to keep this list fairly high-level. If you mean adding MIPS etc, that's great. But there's some micro-variants of ARM architectures that can really muddy things up and need some more thought.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+### `photoPath (String)`:
+
+The path to a photo of the device in JPEG format.
+
+### `compatibleWith (String[])`
+
+An array of the device slugs this device is compatible with.
+
+Declaring compatibility allows images that target a specific device to be also
+used in other device types that are known to work with it.
```
alexandrosm: We certainly have standardised slugs in resin.io, and it would be nice if we can kickstart a kind of device type ontology, but we have to be careful about adding some pragmatic guardrails to that sort of thing because otherwise we end up with the Semantic Web, writing in RDF and OWL-DL.

docs/EXTENDED-ARCHIVES.md
alexandrosm: Maybe we can tempt them?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+
+A image may be described with any of the following properties:
+
+### `displayName (String)`
+
+The human-friendly name of the image.
+
+### `version (String)`
+
+The version of the image.
```
alexandrosm: no harm in that.

docs/EXTENDED-ARCHIVES.md
alexandrosm: Yeah, there needs to be some more thinking through here. To be completely honest, the way we designed this was by looking at how the RPF does it now, so that it would be "automatically supported" by them, while being a fairly standard pattern that others can implement also. Happy to hear alternatives, and though not necessary, "RPF compatibility" is a definite nice-to-have

docs/EXTENDED-ARCHIVES.md
lurch: > If you mean adding MIPS etc, that's great.

Yeah that's what I was thinking, after seeing the huge list of arch-es in the 'Client' section below.
BTW it seems a bit odd that you refer to `amd64` and `i386` here, but then `x64` and `x86` in the 'Client arch' section?

docs/EXTENDED-ARCHIVES.md
lurch: Yeah, that's why I was suggesting that this document should define a set of 'standard' slug names, so that each publisher doesn't end up using their own slightly-different variations.

docs/EXTENDED-ARCHIVES.md
lurch: If the bmap files are easy to create (i.e. can be plumbed into https://github.com/RPi-Distro/pi-gen ), and they don't increase the download size for 'regular' `dd`-users too much, then they might be tempted.

docs/EXTENDED-ARCHIVES.md
lurch: NOOBS works by comparing `release_date`s ( see https://github.com/raspberrypi/noobs/blob/master/recovery/mainwindow.cpp#L1157 ), and only offers to download an OS from the internet if it has a newer `release_date` than the locally-available OS.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+### `version (String)`
+
+The client version.
+
+### `os (String)`
+
+The operating system running the client.
+
+The current possible values are: `aix`, `darwin`, `freebsd`, `linux`,
+`openbsd`, `sunos`, and `win32`.
```
lurch: Perhaps there should also be a Client `os_version` field?
(and again, this document should specify the allowed values, to stop different publishers using slightly different values to mean the same thing)

lurch:
IMHO this document should also explicitly specify which fields are mandatory, which fields are strongly recommended, and which fields are entirely optional.

docs/EXTENDED-ARCHIVES.md
alexandrosm: how arches are named is a massive chaos. One thing is for sure, there will be one canonical name, but a bunch of alternatives for each. no way around that. Not much to comment on the specifics here, leaving it to @jviotti

docs/EXTENDED-ARCHIVES.md
alexandrosm: makes sense. We could certainly seed it with the standard stuff we use on resin.io, and in fact we could probably follow that up with a bit more work that generates slugs for a ton of other boards, maybe as an open source repo that takes PRs from everyone, as a coordination point so we don't step on each others' toes. But first and foremost is to define the borders of a "device type".

docs/EXTENDED-ARCHIVES.md
alexandrosm: I do think it matches those requirements. bmaps are tiny, and dd-users can use bmap-tools from the commandline to speed up their writes also. Bmaps is not an etcher exclusive. Sounds extremely interesting.

docs/EXTENDED-ARCHIVES.md
alexandrosm: I am slowly realising that you've done this before :)

docs/EXTENDED-ARCHIVES.md
alexandrosm: These links are incredibly useful and we should probably borrow heavily from them instead of starting from scratch. this here (https://www.freedesktop.org/software/systemd/man/os-release.html) is the other relevant reference I've dug up.

alexandrosm:
@lurch so far as I know, every item is optional. @jviotti can correct me if I'm wrong, but the general spirit is to help publishers help us help them. There's not much compulsion in the whole setup, the point is -- if you give Etcher this info, it can improve your users' experience in this specific way.

docs/EXTENDED-ARCHIVES.md
lurch: There's a Python script that gets run on the RPF server, that automatically builds os_list_v3.json from the os.json and partitions.json files for the individual OSes, before doing the push-to-CDN. By 'duplicating' the info into a single os_list.json file, it means NOOBS only has to do one http request in order to display the list of network-installable OSes.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+screen resolutions.
+
+### `checksumType (String)`
+
+The checksum type. The current possible values are: `sha1`, `crc32`, and `md5`.
+
+### `checksum (String)`
+
+The actual checksum according to the type declared in `checksumType`.
+
```
1j01: Does it make sense to be able to include multiple checksums for forwards compatibility? Perhaps as an object like:
```
"checksums": {
  "md5": "f0e99e4adc9d8fe04a944d2257f32028"
}
```

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,344 @@
...
+### `bytesToZeroOutFromTheBeginning (Number)`
+
+The number of bytes to zero out from the beginning before starting the flash
+process. This option currently only applies when using bmaps.
+
+### `recommendedDriveSize (Number)`
+
+The recommended drive size to flash this image, in bytes.
```
1j01: Minimum, I'm guessing?

docs/EXTENDED-ARCHIVES.md
lurch: I guess we _could_, but I doubt many image-publishers would bother creating multiple checksums of their image files.
And if there _are_ multiple checksums, what logic should Etcher use to select which one to use? The fastest one? The 'most secure' one?

docs/EXTENDED-ARCHIVES.md
lurch: Nope, 'recommended' is correct. The minimum size is the actual size of the disk-image itself, and I believe the recommended size is so that the image-publisher can indicate "you _could_ flash this image to an XXXXX bytes SDcard (where XXXXX is the raw-disk-image size), but it might not work very well, and you'll likely have a better user-experience if your SDcard is at least YYYYY bytes".

alexandrosm:
well, it's "recommended minimum", in the sense that, in general, having
more won't hurt, having less might.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

alexandrosm:
for the hash question, i guess Etcher can use its own internal
prioritisation?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

docs/EXTENDED-ARCHIVES.md
1j01: That's what I was thinking. It's like a minimum recommended size. The alternative being an absolute 'ideal size', which would be weird. I think "minimum recommended drive size" would be clearer, but it could also be good to actually include the message that would be shown in the documentation, or a description of it like "The user will be informed that they may have a better experience if they use a drive of at least `recommendedDriveSize` bytes."

docs/EXTENDED-ARCHIVES.md
1j01: Yeah, I guess it would be the 'fastest one that's still effective checking against for corruption', which might vary per machine given for instance [hardware acceleration of SHA](https://en.wikipedia.org/wiki/Intel_SHA_extensions).
But also, another image writing tool could want to make use of the metadata and verify the image but doesn't support all the functions that etcher does. Theoretically. Just trying to think of the future. This would probably be one of the main pieces of metadata another program might want to leverage.

lurch: *If* this feature proves to be useful / desired, I guess we could write a tool which takes an extended-archive as input, and automatically updates the checksums section to include other variants.

jviotti:
@lurch How is the `.etch` format going? What do you think about closing this PR, and taking the feedback that its here to create a new specification document?

alexandrosm:
@jviotti, maybe we should ammend this PR, so we keep the conversation
started here?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

On Mon, Dec 26, 2016 at 10:19 PM, Juan Cruz Viotti <notifications@github.com

jviotti:
That works for me. @lurch feel free to take over this PR!

docs/EXTENDED-ARCHIVES.md
jviotti: This is addressed by the `supportedDevices` array in the image manifest.

docs/EXTENDED-ARCHIVES.md
jviotti: > We could certainly seed it with the standard stuff we use on resin.io

Done!

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,413 @@
...
+- `ts4900`
+- `ts7700`
+- `via-vab820-quad`
+- `zynq-xz702`
+
+##### `slugAliases (String[])`
+
+The device slug aliases, if any.
+
+##### `compatibleWith (String[])`
```
jviotti: @alexandrosm Can you remind me what was the use case for this? Declaring device compatibility, as well as declaring what device types are compatible with the image included in the extended archive is useless if we don't allow an extended archive to include multiple images (e.g: it is assumed that the image shipped with the extended archive works with the declared device type).

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,413 @@
...
+
+The client may allow a certain degree of branding based on this color scheme.
+
+You may declare the following colors, in hexadecimal format:
+
+- `background`: The background color
+- `text`: The text color
+- `primary`: The primary color
+
+#### Device
```
jviotti: Should an image declare all the device types it work with, instead of just one? How does knowing about compatible device types helps us? Will we somehow attempt to detect the device type, or will we ask the user what device types they want to flash to?

jviotti: I'm starting to think there should be a collection of device manifests somewhere centralized (e.g: in an Etcher server) for the sole reason of asking the user: What is your device type? And then allowing them to download and flash extended archives that we host that match that criteria.

In that case, we can leave out the device manifests from the extended archive, and only allow images to declare a list of device types slugs they support.

@alexandrosm Am I missing any use case we discussed at some point?

lurch: I guess having the device manifests on a central server prevents the use-case where you download Etcher and an extended archive on an online computer, copy it to e.g. a USB stick, and then later use it on an offline computer?

For the initial version of this, maybe the device-type stuff should be informational only? i.e. when selecting an extended archive, Etcher will tell you what device(s) it's compatible with, so you'll know before-flashing if you've accidentally selected the wrong extended archive, instead of waiting until after flashing and then discovering your device doesn't boot?

docs/EXTENDED-ARCHIVES.md
lurch: Wasn't there some discussion once about having "fat extended archives" that contained multiple images? Or maybe I'm just getting confused with something else.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+### Vendor
+
+A vendor represents a manufacturer of a device type and potential publisher.
+
+*Some examples of image vendors are: Beagleboard, Raspberry Pi Foundation,
+Intel, and ODroid.*
+
+### Publisher
+
+A publisher represents an entity that maintains and publishes official images.
```
lurch: Do we need the word 'official' in here? Extended Archives should be creatable by anyone.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+The file format should have the following characteristics:
+
+- Streamability: The extended archive allows the client to consume all its
+  information in a streaming fashion
+
+- Compression: The extended archive should be compressed in order to save
+  network resources
+
+We base this format on the [`tar`][tar] archive format, which has the following
+benefits:
```
lurch: Oh? This is a major change. I thought the idea was to use `.zip` format, so that extended-archive files where still usable by people on Windows (which has built-in `.zip` support), but who don't (for whatever reason) want to use Etcher?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+entities.
+
+*At the moment of this writing, the only existent client is the Etcher application.*
+
+File format
+-----------
+
+The file format should have the following characteristics:
+
+- Streamability: The extended archive allows the client to consume all its
```
lurch: IMHO the (possibly multi-gigabyte) disk-image itself should be streamable, but I don't think the metadata itself needs to be streamable (since it'll be small enough to buffer in RAM).
I still wonder if we ought to do something like the RPF does with it's NOOBS images regarding the separation of metadata (including image logos etc.) and (compressed) disk-images. And then for 'downloadable' (offline) images we could have a zip including both the metadata and the image-file.
Given that `.zip` is the lowest-common-denominator compression format, but it's not streamable, I think it _might_ make sense to have separate formats for the offline-download images (where we only want a single file, compressed in .zip), and the online-streamable images (where we want streamability, but there's no problems in having multiple separate files, as long as they're all reachable from a "starting URL"). Does that make sense?

Hmmm, maybe something like:
On the server: `metadata.json`, `metadata_XX.zip`, `diskimage_XX.img.xz` (with `metadata.json` being a file at a well-known/stable URL, which would be what Etcher 'consumes', which then contains links to `metadata_XX.zip` and `diskimage_XX.img.xz`)
In the downloaded `diskimage_XX.zip`: `diskimage_XX.img`, `.meta/metadata.json`, `.meta/<other metadata files pointed to by metadata.json>`

This makes it easy for publishers to start enhancing their current `diskimage_XX.zip` files to support the `.meta/` metadata directory (i.e. only a small change to their current workflow), and then in the future if they want to they can also start supporting online-streamable Etcher images. And it should be easy for us to supply them with a script which converts an offline-zipfile into the separate online-streamable files (and also the inverse, if desired).

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+- Compression: The extended archive should be compressed in order to save
+  network resources
+
+We base this format on the [`tar`][tar] archive format, which has the following
+benefits:
+
+- It's a nearly universal archive format (easy to find robust implementations)
+
+- It allows us to easily impose a file order, which greatly helps with
+  streamability. This is not trivial with `zip`, for instance
```
lurch: Obviously requiring files to be in a specific order would require a custom tar-creation program, which _slightly_ negates the "nearly universal" from the previous point (at least from the creation side).

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+  decompression library for [node.js][nodejs], called
+  [`lzma-native`][lzma-native], which we battle tested at Etcher
+
+- A way to calculate the precise uncompressed size of the file no matter the
+  size (which is a [known limitation of `gzip` after 4
+  GB](https://github.com/resin-io/etcher/issues/638)), and an operation
+  [`lzma-native`][lzma-native] supports out of the box. This is not possible
+  with `bzip2`
+
+- Decent compression ratio, and compression/decompression time
```
lurch: IMHO the compression time isn't relevant, as it's only a one-time thing (done by the publisher) and doesn't impact the end-user Etcher users (which IMHO is a more important consideration).

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+The resulting file should contain the `.etch` extension.
+
+Archive layout
+--------------
+
+The contents of an extended archive are the followings:
+
+- **REQUIRED** `.meta/manifest.json`
+- `.meta/publisher.svg`
```
lurch: IMHO it's only `manifest.json` that needs a fixed name, and all the other files (which should still be in the `.meta` directory) can be named whatever the publisher wants, for flexibility, as long as the `manifest.json` refers to them.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+- `.meta/device.svg`
+- `.meta/image.svg`
+- `.meta/schema.json`
+- `.meta/image.bmap`
+- `.meta/instructions.markdown`
+- `.meta/release-notes.txt`
+- **REQUIRED** `image.<extension>`
+- `.meta/checksum`
```
lurch: This should probably just be a field in `manifest.json`, rather than being a separate file?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+Notice that the order is important. The `.meta/` directory must come before the
+image data, since we need that information before being able to make use of the
+image.
+
+The order of the files inside `.meta/` is not mandatory, however the order
+described above ensures the maximum UX perceived efficiency.
+
+### Graphics
+
+We chose [`svg`][svg] over other image formats since clients can display it on
```
lurch: Would it hurt if we supported both `.svg` and `.png`, with SVG being the recommended option? Easy to imagine that many publishers might have only bitmap logos, but not yet vector logos.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+This is the only required file inside `.meta`, which declares information about
+the image, the target device type, and the publisher. This file may contain the
+following top level properties:
+
+- `publisher`: Information about the image publisher
+- `device`: Information about the image device type
+- `image`: Information about the image itself
+
+#### Publisher
```
lurch: For future extensibility, it might not hurt to have an optional publisher `logo` too?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+The publisher specific color scheme, if any.
+
+The client may allow a certain degree of branding based on this color scheme.
+
+You may declare the following colors, in hexadecimal format:
+
+- `background`: The background color
+- `text`: The text color
+- `primary`: The primary color
```
lurch: I guess we ought to provide an example of where each of these colours are used in Etcher, and show how Etcher looks when different colours are chosen.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+#### Device
+
+Here's an example of a real-world device manifest for Raspberry Pi 3:
+
+```json
+{
+  "displayName": "Raspberry Pi 3",
```
lurch: Should this be just `name`, for similarity with the publisher section?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+##### `displayName (String)`
+
+The human-friendly name of the device.
+
+##### `url (String)`
+
+The url to the landing page of the device.
+
+##### `arch (String[])`
```
lurch: What does the `String[]` here mean? Surely a device will only have one `arch`?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+An array of the device slugs this device is compatible with.
+
+#### Image
+
+Here's an example of a real-world image manifest for Raspbian Jessie:
+
+```json
+{
+  "displayName": "Raspbian Jessie",
```
lurch: Again, maybe just 'name' is appropriate.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+precedence over `.meta/release-notes.txt` if defined.
+
+##### `releaseDate (String)`
+
+The release date timestamp. The date should conform to [ISO 8601][iso8601]
+standard.
+
+##### `checksumType (String)`
+
+The checksum type. The current possible values are: `sha1`, `crc32`, and `md5`.
```
lurch: Doesn't hurt to add `sha256` too?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+See the `.meta/checksum` file.
+
+##### `recommendedDriveSize (Number)`
+
+The minimum recommended drive size to flash this image, in bytes.
+
+The use case for this option is that while a drive might be large enough to
+contain the image, it might not be large enough to deliver a good experience
+when actual using the application or operating system contained in the image.
```
lurch: actual -> actually

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+##### `supportedDevices (String[])`
+
+An array of the device slugs this image supports.
+
+##### `updateUrl (String)`
```
lurch: If we go with my earlier suggestion of splitting out the image and metadata, this could instead be the URL of the online manifest file, and then it could just compare the releaseDate in this (local) manifest file, with the releaseDate in the online manifest file, and we wouldn't need to bother with the Etag stuff (which may not be supported by all webservers anyway?)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+The release date timestamp. The date should conform to [ISO 8601][iso8601]
+standard.
+
+##### `checksumType (String)`
+
+The checksum type. The current possible values are: `sha1`, `crc32`, and `md5`.
+
+See the `.meta/checksum` file.
+
+##### `recommendedDriveSize (Number)`
```
lurch: We should also include the actualImageSize, so we know (before downloading and decompressing the image) if the currently-selected drive is big enough.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+The minimum recommended drive size to flash this image, in bytes.
+
+The use case for this option is that while a drive might be large enough to
+contain the image, it might not be large enough to deliver a good experience
+when actual using the application or operating system contained in the image.
+
+##### `supportedDevices (String[])`
+
+An array of the device slugs this image supports.
```
lurch: `slugs`, or `deviceTypes`?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+Clients may use this to check if the image resolved by `updateUrl` is different
+to the one we have locally.
+
+##### `expiryDate (String)`
+
+The timestamp to determine the expiration date of the image. The date should
+conform to [ISO 8601][iso8601] standard.
+
+Clients may use this information to refuse flashing the image.
```
lurch: If we're using an 'open' (and non-encrypted) compression method such as `.zip` or `.tar` then obviously a determined user will still be able to extract the image and write it manually, so maybe this should just be a warning, rather than an outright refusal?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+-------------
+
+### `.meta/manifest.json`
+
+This is the only required file inside `.meta`, which declares information about
+the image, the target device type, and the publisher. This file may contain the
+following top level properties:
+
+- `publisher`: Information about the image publisher
+- `device`: Information about the image device type
```
lurch: This should clarify that `device` can be an array, e.g. a Raspbian image which supports Pi1, Pi2 and Pi3.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+Etcher extended archives are archive files including an image and a special
+metadata directory including information that describes the image, device type,
+and vendor in detail, enabling Etcher to make use of such information to
+deliver advanced features and a more pleasant experience to the user.
+
+Entities
+--------
+
+### Vendor
```
lurch: Seems like 'vendor' isn't really used elsewhere in this document?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+### Device
+
+A device represents a specific device type manufactured by a vendor.
+
+*Some examples of devices are: Beaglebone Black, Beaglebone Green, Intel
+Edison, and Raspberry Pi 3.*
+
+### Client
+
+A client represents the program making use of the information about the other
```
lurch: Again, seems that 'client' isn't really an entity now?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+##### `displayName (String)`
+
+The human-friendly name of the image.
+
+##### `version (String)`
+
+The version of the image.
+
+##### `url (String)`
```
lurch: Perhaps there should also be an optional supportUrl field for image, in case a publisher has many images, and dedicated support-urls for each specific image?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+    ..
+  },
+  "image": {
+    ..
+    "extensions": {
+      ..
+    }
+    ..
+  }
+}
```
lurch: Maybe also a generic `extensions` at the top-level, outside of publisher, device and image?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+```
+
+### `.meta/publisher.svg`
+
+An [`svg`][svg] logo of the publisher.
```
lurch: The path to this should be pointed to from the publisher section (with `publisher.svg` just being a recommended default)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+### `.meta/device.svg`
+
+An [`svg`][svg] logo of the device.
```
lurch: The path to this should be pointed to from the device section (with `device.svg` just being a recommended default)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+
+A [bmap][bmap] file for the extended archive image.
+
+### `.meta/instructions.markdown`
+
+A markdown file including post-flash instructions.
+
+### `.meta/release-notes.txt`
+
+A plain text file describing the image version's release notes. The client
```
lurch: As mentioned in a previous comment, I think this should be explicitly pointed to in the image section, and maybe it makes more sense to _only_ have releaseNotes in the metadata, and get rid of releaseNotesUrl entirely? (since the former will always be the same version as the included image, but the URL might be non-versioned, and so display 'incorrect' notes (too new) for an old image)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
...
+### `.meta/checksum`
+
+For example: `64c7ed611929ea5178fbb69b5a5f29cc9cc7c157`
+
+This file contains the checksum matching the type specified in the
+`checksumType` field of the image section of `.meta/manifest.json`.
+
+The reason the actual checksum is at the end of the archive is that if any tool
+modifies the image inside the extended archive, it will need to update the
+checksum.
```
lurch: Any tool that modifies the image, could just as easily update the checksum field in manifest.json (rather than rewriting a separate file)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -0,0 +1,428 @@
+Etcher Extended Archives
+========================
+
+Etcher extended archives are archive files including an image and a special
+metadata directory including information that describes the image, device type,
+and vendor in detail, enabling Etcher to make use of such information to
```
lurch: vendor -> publisher

jviotti: Fixed!

docs/EXTENDED-ARCHIVES.md
jviotti: You're right. Deleted!

docs/EXTENDED-ARCHIVES.md
jviotti: Removed!

docs/EXTENDED-ARCHIVES.md
jviotti: We use the word "client" in various parts of the document, so I though describing precisely what we mean by that can clarify things. What do you think? We can move it to a section called "glossary" or something like that if the word "entity" is the confusing part.

docs/EXTENDED-ARCHIVES.md
jviotti: The problem with this is that suddenly publishers need to upload and maintain a tree of versioned metadata, which extra burden compared to simply including it on the file. Pinging @alexandrosm for this.

docs/EXTENDED-ARCHIVES.md
jviotti: This is still under discussion, but looks like we're dropping `.zip` for our own custom format, even if that has a usability drawback. The discussions will continue on the architecture call on Monday (after the company call), so you should definitely join us!

docs/EXTENDED-ARCHIVES.md
jviotti: Yeah, by universal I meant that we can find good battle-tested modules to work with them, not necessarily usability. I'll make that clearer!

docs/EXTENDED-ARCHIVES.md
jviotti: It does affect users who don't have fast internet connections, and it also affects publishers, since they have to host larger images. For big publishers, a hundreds of megabytes less can mean quite some dollars saved.

docs/EXTENDED-ARCHIVES.md
jviotti: We might simplify this whole thing for now (e.g: remove the device type declarations, etc), so it might not be relevant anymore, but relying on naming conventions is much simpler than having to reference them IMHO.

docs/EXTENDED-ARCHIVES.md
jviotti: Read the detailed description of this file at the bottom of the document. We plan to have an Etcher Server, which will take an extended archive and configuration values, and will modify the image itself after applying Reconfix configurations. Since the image will be modified, the checksum will need to be updated. Having the checksum at the tail of the archive makes streaming much simpler.

docs/EXTENDED-ARCHIVES.md
jviotti: OK, let's support both `.svg` and the common `@2x` PNG naming convention.

docs/EXTENDED-ARCHIVES.md
jviotti: Sounds reasonable. We will remove this stuff from the first implementation, but we will re-discuss this in the future.

docs/EXTENDED-ARCHIVES.md
jviotti: We do have one. Check `.meta/publisher.svg`

docs/EXTENDED-ARCHIVES.md
jviotti: This is not implemented yet. In any case, the extended archive spec should be generic (not Etcher specific), so it'd be up the client how it makes use of these, or if it uses it at all.

docs/EXTENDED-ARCHIVES.md
jviotti: Good catch!

docs/EXTENDED-ARCHIVES.md
jviotti: We did discuss it at some point, but looks like we're only going to support a single one for now (we'll open this for discussion again in the near future).

docs/EXTENDED-ARCHIVES.md
jviotti: Agree! Changed!

docs/EXTENDED-ARCHIVES.md
jviotti: Done!

docs/EXTENDED-ARCHIVES.md
jviotti: Since we're going to impose ordering of the files in the archive, the actual file information will be accessible from the tar header right before the image (after the metadata), so I don't think also having it here has an advantage for us. Of course it'd be faster, but then is also a point where data can get out of sync, and we should try to prevent that from happening where possible.

docs/EXTENDED-ARCHIVES.md
jviotti: Good catch!

docs/EXTENDED-ARCHIVES.md
jviotti: Fair enough, updated.

docs/EXTENDED-ARCHIVES.md
jviotti: Yeah, I'll update the spec to use references.

docs/EXTENDED-ARCHIVES.md
jviotti: Sounds good and simpler. Let's go for it.

docs/EXTENDED-ARCHIVES.md
jviotti: The problem is streaming (and particular modifying the contents on the fly while streaming, which is a requisite for the Image Maker). If the checksum is at the tail, we can start piping the archive to the user while doing the modifications, and write the updated checksum at the tail.

docs/EXTENDED-ARCHIVES.md
lurch: Note that in my earlier comment I was referring to "compression time", not "compression ratio" ;-)

docs/EXTENDED-ARCHIVES.md
lurch: Will do. Remind me in flowdock nearer the time ;-)

docs/EXTENDED-ARCHIVES.md
lurch: In my experience, allowing flexibility by manually specifying filenames might seem "more difficult" at first, but it gives much more flexibility later on.

Also, when files are being hosted online (possibly by a CDN), and a typo has been made, it's often much easier to just edit the (small) manifest and upload a new one, than it is to upload a new version of the renamed (big) file. (A CDN I was using previously only supported upload and delete, not rename)

docs/EXTENDED-ARCHIVES.md
lurch: Oh, first I've heard of an "Etcher Server". I thought the idea was that it'd be Etcher itself that did the configuration (using reconfix), after the image had been written to disk and verified (with the original checksum) ?

AFAIK it won't be possible for Reconfix to work in a streaming fashion (since you can't know which parts of the filesystem will be modified), so you'll still need to finish the whole reconfix process before starting to stream the image. So there's no reason the checksum can't be included in the data (i.e. manifest.json) before the image itself?

docs/EXTENDED-ARCHIVES.md
lurch: But for this sort of thing it's always good to have a "reference/example image" so people can know what sort of thing to expect, even if Etcher itself doesn't support it yet. (Otherwise I suspect publishers won't bother using this feature)

docs/EXTENDED-ARCHIVES.md
WasabiFan: Keep in mind that there are two parts to what @jviotti said: the publisher has to host it, _and_ the user is dependent on that publisher. So an extended archive that you download may stop functioning uncontrollably if the publisher changes something on their end, and there isn't an easy way to take a pre-downloaded "metadata-only" image and link the image file locally.

docs/EXTENDED-ARCHIVES.md
lurch: And of course it might be worth still going with this approach, even if the manifest is embedded inside the extended archive itself.

docs/EXTENDED-ARCHIVES.md
lurch: I was just thinking - another advantage of including the actualImageSize in the manifest, is it gets around the problem of not being able to determine the image size when compressed with .gz or .bzip2 (if the disk image is external to the extended archive)

And given that the extended archive will need to be created with custom tooling in order to get the correct file-order, there's no reason why that same tool couldn't update the actualImageSize in the manifest, so it'd never be out of date.

docs/EXTENDED-ARCHIVES.md
lurch: Yeah, "glossary" or "definitions" would be a good place for it. When I see "entities" I'm expecting that to be talking about actual things in the archive / manifest.

docs/EXTENDED-ARCHIVES.md
lurch: Hmm, another thought re: the custom tooling - it might be nice if Etcher had a test-mode (for publishers) where it could be pointed to a directory containing all the files that _will_ get included into the extended archive, so that files can be edited and tweaked as necessary on-disk, before the extended-archive gets created from them (which will be a slow process, given that we're including and compressing disk images)

docs/EXTENDED-ARCHIVES.md
lurch: Hmm, I guess ImageMaker would also need to read checksumType from the manifest.json, so it knows what kind of checksum to append to the image (unless ImageMaker also rewrites the checksumType in the manifest too)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -80,9 +75,9 @@ Archive layout
 The contents of an extended archive are the followings:
 
 - **REQUIRED** `.meta/manifest.json`
-- `.meta/publisher.svg`
-- `.meta/device.svg`
-- `.meta/image.svg`
+- `.meta/publisher(2x)?.[svg|png]`
```
lurch: Probably better to split this out to separate files, as it currently makes it look like `.meta/publisher2x.svg` would be a valid option (and not everyone reading this will grok regexes).

docs/EXTENDED-ARCHIVES.md
```diff
@@ -101,9 +96,10 @@ described above ensures the maximum UX perceived efficiency.
 
 ### Graphics
 
-We chose [`svg`][svg] over other image formats since clients can display it on
-any kind of screen resolution without requiring multi-layer files, or multiple
-files workarounds.
+Any graphic included in the archive can be either an [`svg`][svg] file, or a
+[`png`][png] file with the `<filename>.png` and `<filename>@2x.png` convention.
```
lurch: I think we should allow:
* no filenames
* `filename.svg` only
* `filename.png` only
* `filename.png` and `filename@2x.png`
* `filename.png` and `filename@2x.png` and `filename.svg` (include both PNG and SVG in case some clients don't support displaying SVGs?)

The more flexibility we allow, the easier it'll be for publishers to start using the extra features.

I guess that also suggests that maybe graphics should be referenced in the manifest by the extension-less part of the filename, so the client is then free to choose the 'best' extension from those available?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -162,78 +159,6 @@ You may declare the following colors, in hexadecimal format:
 
 The client decides how to use them and where, if at all.
 
-#### Device
```
lurch: Is it a bit premature to be removing this? Isn't knowing what device(s) an image is suitable for actually quite valuable?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -201,6 +212,15 @@ and `md5`.
...
 See the `.meta/checksum` file.
 
+##### `path (String)`
+
+The path to the image, relative to the root of the archive.
+
+##### `bmap (String)`
+
+The path a [`bmap`][bmap] file for the image, relative to the `.meta`
+directory.
```
lurch: It feels like this conceptually belongs alongside the image, rather than in the `.meta` subdirectory? Just a gut-feeling.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -201,6 +212,15 @@ and `md5`.
 
 See the `.meta/checksum` file.
 
+##### `path (String)`
+
+The path to the image, relative to the root of the archive.
```
lurch: Possibly worth highlighting that this has a different relative path to the other files referenced here (e.g. by making it bold?)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -291,15 +299,14 @@ A markdown file including post-flash instructions.
...
 
-The image file **in uncompressed form**. Clients should not attempt to
+An image file **in uncompressed form**. Clients should not attempt to
 decompress the image file since the extended archive already includes
 compression. If the image is compressed, then the client should write
 compressed bytes to the drive.
 
-An archive containing more than one image file is not a well-formed archive. If
-this is the case, use only the first image.
+This file should be references by `.meta/manifest.json`'s `path` property.
```
lurch: references -> referenced

docs/EXTENDED-ARCHIVES.md
```diff
@@ -267,22 +287,10 @@ For example:
...
-
 ### `.meta/schema.json`
 
 A [Reconfix][reconfix] image configuration schema.
 
-### `.meta/image.bmap`
-
-A [bmap][bmap] file for the extended archive image.
-
 ### `.meta/instructions.markdown`
```
lurch: For consistency, the schema and instructions should probably also be referenced by the 'image' section in the manifest.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -201,6 +212,15 @@ and `md5`.
 
 See the `.meta/checksum` file.
 
+##### `path (String)`
+
+The path to the image, relative to the root of the archive.
+
+##### `bmap (String)`
+
+The path a [`bmap`][bmap] file for the image, relative to the `.meta`
```
lurch: "The path to a "

docs/EXTENDED-ARCHIVES.md
```diff
@@ -267,22 +287,10 @@ For example:
...
 
 A [Reconfix][reconfix] image configuration schema.
 
-### `.meta/image.bmap`
-
-A [bmap][bmap] file for the extended archive image.
-
 ### `.meta/instructions.markdown`
 
 A markdown file including post-flash instructions.
```
lurch: "containing" is probably better than "including" here, as "including" implies it might include other things too ;)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -291,15 +299,14 @@ A markdown file including post-flash instructions.
 
 A plain text file describing the image version's release notes.
```
lurch: Any good reason why the Instructions are markdown, but the ReleaseNotes are plain text? Maybe we should allow both markdown or plain text for both?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -75,28 +75,32 @@ Archive layout
...
 
 ### Graphics
 
 Any graphic included in the archive can be either an [`svg`][svg] file, or a
-[`png`][png] file with the `<filename>.png` and `<filename>@2x.png` convention.
+[`png`][png] file.
 
 If both are present, [`svg`][svg] should be favoured by the client.
 
+Ifa [`png`][png] file path is declared and the client detects that the host is
```
lurch: "If a"

docs/EXTENDED-ARCHIVES.md
```diff
@@ -75,28 +75,32 @@ Archive layout
...
+running a high definition screen, it may search for a `<filename>@2x.png` file
+in the archive.
+
+### Paths
+
+The `.meta/manifest.json` file may refer to other files in the package. If the
+property searches for the file relative to `.meta`, then it should not allow
+accessing the parent directory with `..`.
```
lurch: Probably easier to just rephrase this as filenames can't include any path segments, i.e. no `/`

docs/EXTENDED-ARCHIVES.md
```diff
@@ -175,6 +179,9 @@ Here's an example of a real-world image manifest for Raspbian Jessie:
   "checksumType": "sha1",
   "path": "raspbian-jessie.img",
   "bmap": "raspbian-jessie.img.bmap",
+  "instructions": "raspbian-jessie.markdown",
+  "releaseNotes": "raspbian-jessie-changelog.txt",
+  "configurationSchema": "raspbian-jessie-schema.json",
```
lurch: Would `reconfixSchema` be a better name? configurationSchema sounds a bit too generic IMHO.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -221,6 +228,21 @@ The path to the image, relative to the root of the archive.
 The path a [`bmap`][bmap] file for the image, relative to the `.meta`
 directory.
 
+##### `instructions (String)`
+
+The path a markdown post-flash instructions file for the image, relative to the
```
lurch: "path to a"

docs/EXTENDED-ARCHIVES.md
lurch: And another good reason for referencing filenames in the manifest, is that we'll instantly know whether we need to stream in more of the extended archive to get more metadata files or not.

docs/EXTENDED-ARCHIVES.md
lurch: And similarly the tooling could also automatically generate the necessary checksum value when creating the extended archive.

docs/EXTENDED-ARCHIVES.md
```diff
@@ -107,44 +107,57 @@ Archive files
...
 {
-  "name": "RetroPie",
-  "url": "https://retropie.org.uk",
-  "logo": "retropie.svg",
-  "colorScheme": {
-    "background": "#535760",
-    "text": "#FFFFFF",
-    "primary": "#5793db"
+  "publisher": {
+    "name": "Raspberry Pi",
```
lurch: "Raspberry Pi Foundation" ;-)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -107,44 +107,57 @@ Archive files
...
+      "background": "#535760",
+      "text": "#FFFFFF",
+      "primary": "#5793db"
+    }
+  },
+  "image": {
+    "name": "Raspbian Jessie",
+    "version": "May 2016",
+    "url": "https://www.raspberrypi.org/downloads/raspbian/",
+    "supportUrl": "https://retropie.org.uk/forum/",
```
lurch: Ooops ;)

docs/EXTENDED-ARCHIVES.md
```diff
@@ -121,23 +121,28 @@ Here's an example of a real-world manifest for Raspbian Jessie
...
+        "path": "raspbian-jessie.img",
+        "bmap": "raspbian-jessie.img.bmap",
+        "instructions": "raspbian-jessie.markdown",
+        "releaseNotes": "raspbian-jessie-changelog.txt",
+        "configurationSchema": "raspbian-jessie-schema.json",
+        "recommendedDriveSize": 4294967296,
+        "updateUrl": "https://downloads.raspberrypi.org/raspbian_latest",
+        "etag": "c0170-53152af2-533d18ef29fc0"
+      }
+    ]
```
lurch: I guess including Raspbian Lite in here too would make a good example

docs/EXTENDED-ARCHIVES.md
```diff
@@ -306,12 +317,13 @@ compressed bytes to the drive.
 
 This file should be references by `.meta/manifest.json`'s `path` property.
 
-### `.meta/checksum`
+### `.meta/<checksum file name>`
```
lurch: A few lines earlier, you said 'can't be inside `.meta/`' ?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -171,75 +176,79 @@ You may declare the following colors, in hexadecimal format:
...
 The checksum type. The current possible values are: `sha1`, `sha256`, `crc32`,
 and `md5`.
 
-See the `.meta/checksum` file.
+#### `images[].checksum.path (String)`
 
-#### `image.path (String)`
+The path to a file containing the checksum, relative to the root of the
+archive. This file **can't be inside `.meta/`** and should be included **at the
+tail of the archive**.
```
lurch: If we're including multiple images in an extended archive, I guess a more accurate restriction would be that the checksum has to come after it's associated image file, and not necessarily at the end of the archive.
I.e. you could have
```
.meta/manifest.json
<metadata files>
image1.img
image2.img
checksum1
checksum2
```
or
```
.meta/manifest.json
<metadata files>
image1.img
checksum1
image2.img
checksum2
```

For extendability purposes (e.g. there was a suggestion elsewhere about supporting multiple checksum types in the future), I'm beginning to wonder if the checksum should actually be a JSON file, rather than just flat-text?

docs/EXTENDED-ARCHIVES.md
```diff
@@ -112,15 +112,17 @@ the image.
...
+  {
+    "publisher": {
+      "name": "Raspberry Pi",
+      "url": "https://www.raspberrypi.org",
+      "logo": "raspberrypi.svg",
+      "colorScheme": {
+        "background": "#535760",
+        "text": "#FFFFFF",
+        "primary": "#5793db"
+      }
```
lurch: I guess an alternative way of doing this, would be to have an array of publishers and an array of images, and then each image could reference a publisher by `name`.

docs/EXTENDED-ARCHIVES.md
lurch: I was thinking that any file that you explicitly downloaded would always be the big `.zip` file (with all the metadata included) and the online-streamable-files would involve the user copy'n'pasting a `https://www.publisher.com/images/manifest.json` URL into Etcher, and Etcher would then download the files it needed on demand.

docs/EXTENDED-ARCHIVES.md
lurch: BTW what's the use-case for having multiple publishers in a single extended archive?

lurch:
And if we are going for our own extended image file-format, then the updateUrl and Etag stuff probably doesn't make sense?
IMHO the updateUrl (if present) should then point at the online version of this extended archive.

Oooh, and there's another interesting thought - including the `manifest.json` inside the extended archive and then only downloading enough of the file (on demand) to parse all the metadata, will only work on webservers that support HTTP Range Requests (which is hopefully most of them?). I wonder how / if that would work with the ImageMaker stuff?

jviotti:
I'm closing this PR since we're taking a new approach towards image metadata. We can always recover this from the list of closed PRs if needed in the future.

1j01:
What's the new approach?

jviotti:
@1j01 Instead of having the metadata included in an image archive, it will be available in a central repository. That simplifies a lot of streaming problems we've been having and allows us to provide a "catalog" of images we know about in Etcher itself.

ensonic:
How are you going to associate the image with the manifest?

WasabiFan:
I think the idea is that you wouldn't download the image yourself, you would browse for an image and download it (along with the associated metadata) from within the app. I'd hope there was also an app-specific link protocol mechanism to link directly to an image from the archive.

lurch:
There's some discussion about our future plans in https://resin.io/blog/the-future-of-etcher/

Issue #632, opened 2016-08-08
Creating Debian Linux installation package 


imrehg:
Create install images as Debian/Ubuntu `deb` files, and possible other types like `rpm` as well.

The idea would be to avoid the "this is how you make it executable, this is how you run it from the file manager", etc issues, but getting as simple as possible to install Etcher on common distros. Reported by @limator

dlech:
I tried packaging this for Debian/Ubuntu about a month ago since I have a bit of experience packaging software, but I quickly gave up. Debian policy is to build *everything* from source, so the packaging infrastructure is geared towards this.

The easiest thing to do is to have http://lanuchpad.net build the packages and host the ppa. However, the launchpad build servers do no allow Internet access during the build. So, we cannot fetch node and bower packages during the build process. This means in order to use launchpad, we also would have to package all of the node and bower packages as debian packages. This sounds like too much work for me.

I suppose we could do like the Arch Linux package and just throw the AppImage into a debian package. There seems like a reason why I decided this was not a good idea and didn't try it, but I can't remember what is was right now.

imrehg:
Hi @dlech thanks for the comments! Yeah, our Launchpad setup has stalled a bit, and good to hear your comments, makes sense.

### Debian/Ubuntu

As for packaging, as far as I remember, Ubuntu/Debian packages have the sources in there, so you think the build path could be adjusted by splitting the package downloads and the build to two clearly defined steps. Then could prepare a zip with all sources, which launchpad then builds? As I have no experience with their build system yet, would love to hear if this makes any sense or there are any gaps in the process. And whether @jviotti thinks that the build can be split like that :)

### Arch

For ArchLinux, you made me think that indeed building it from source would make a lot of sense, will check it out, now that I have more experience building it locally (and some node bugs that were affecting the build were fixed up).

dlech:
> so you think the build path could be adjusted by splitting the package downloads and the build to two clearly defined steps. Then could prepare a zip with all sources, which launchpad then builds?

This could work. I don't know anything about how node and bower work though, so it would take me quite a long time to figure out how to do this. If someone can figure out how to get all of the source files zipped up, I will be glad to do the rest of the debian packaging.

glensc:
i guess you can attach the source tarballs to github release page as well:
https://github.com/resin-io/etcher/releases

i'm also interested building from source the rpm package. but considering this is electron app, the node dependencies need to be at least included in the source package.

jviotti:
[...]

PR #733, opened 2016-09-30
Debian Packaging

dlech:
Obviously, you do not want to merge this as-is because of the changes to `travis.yml`. This PR is just so we can discuss the changes for now.

Most everything lives in the `debian/` folder, so no problems there. The question is how to fit it in with your release.

- Will you be hosting the package on your own bintray account?
- Do you want the debian package to be built as part of the travis build?
- If this is merged, then I would expect the package maintainer to change to resin.io and that the debian changelog is updated by an etcher maintainer as part of the release.

Or, if you would like me to take more responsibility, I can continue to maintain the debian packaging the way I have it setup now. 

Related: #632


debian/patches/disable-updates.patch
```diff
@@ -0,0 +1,11 @@
+--- etcher-1.0.0~beta.15.orig/lib/gui/app.js
++++ etcher-1.0.0~beta.15/lib/gui/app.js
+@@ -65,7 +65,7 @@ const app = angular.module('Etcher', [
+ app.run((AnalyticsService, UpdateNotifierService, SelectionStateModel) => {
+   AnalyticsService.logEvent('Application start');
+ 
+-  if (UpdateNotifierService.shouldCheckForUpdates() && !process.env.ETCHER_DISABLE_UPDATES) {
++  if (false && UpdateNotifierService.shouldCheckForUpdates() && !process.env.ETCHER_DISABLE_UPDATES) {
```
jviotti: We allow disabling update notifications using the `ETCHER_DISABLE_UPDATES` environment variable (see https://github.com/resin-io/etcher/blob/master/docs/USER-DOCUMENTATION.md#disable-update-notifications). Is there a way we can re-use that instead of patching directly? I'm thinking that you could provide a script that calls the Etcher executable setting that env var first. Happy to provide such option in another form that helps you.

debian/patches/do-not-use-upx.patch
```diff
@@ -0,0 +1,34 @@
...
+   echo "Dependency missing: python" 1>&2
+   exit 1
+@@ -102,11 +97,6 @@ function package_x86 {
+ 
+   mv $output_package/Etcher $output_package/etcher
+   chmod a+x $output_package/*.so*
+-
+-  # UPX fails for some reason with some other so libraries
+-  # other than libnode.so in the x86 build
+-  upx -9 $output_package/etcher $output_package/libnode.so
```
jviotti: Any particular reason for disabling UPX?

jviotti:
This looks really good, amazing work! The way it works at the moment is that you're just hosting the `.deb` packages in bintray, which should be manually downloaded and installed with `dpkg`, right? I guess the ideal approach would be to host a PPA at Launchpad, or would that lock us to just Ubuntu?

I propose the following, but let me know your thoughts:

Lets find a way to merge this to the main Etcher repository and build and publish to a Resin.io/Etcher Launchpad account on Travis, but still have you listed as the official maintainer of the package.

Also, do you think this package would be acceptable for submission to official Debian/Ubuntu repositories? If not, what would be needed from Etcher/Electron side?

debian/patches/disable-updates.patch
dlech: Creating a symlink was easier, but yes, we can certainly do this.

debian/patches/do-not-use-upx.patch
dlech: The debian package tools automatically determine which shared library dependencies are needed by inspecting the binaries. If they are compressed with upx, then the debian tools cannot read them.

After I disabled this, it seems to me that the application starts a bit faster as well. This could just be my imagination though.

dlech:
> The way it works at the moment is that you're just hosting the .deb packages in bintray, which should be manually downloaded and installed with dpkg, right?

No, bintray generates debian archive metadata, so you can use it as a PPA. See https://github.com/dlech/etcher/wiki

> guess the ideal approach would be to host a PPA at Launchpad, or would that lock us to just Ubuntu?

Since the package is build for one distro, it would be the same as I have it now - users would just install the `precise` version no matter what distro they actually use. However, in order to use Launchpad, you have to upload a source package only. They do not allow the upload of binary packages. The source package is then built by Launchpad build servers. These build servers do now allow an Internet connection during build! So in order to make this work with Launchpad, the "source" package would have to include the electron package already completely built.

So really, I don't see any difference between hosting on bintray or launchpad.

> Lets find a way to merge this to the main Etcher repository (so it gets official support and resources) and build and publish to a Resin.io/Etcher Launchpad account on Travis, but still have you listed as the official maintainer of the package.

OK

> Also, do you think this package would be acceptable for submission to official Debian/Ubuntu repositories?

I highly doubt it. Debian has a strict policy of building everything from source. No binaries are allowed in source packages. There is also a policy of not statically linking executables. Maybe there is some chance of it being included in the nonfree section of debian. I'm not sure what exceptions to the policy are tolerated there.

> If not, what would be needed from Etcher/Electron side?

[Here is the debian bug to get the atom text editor in debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=747824). Basically, every node package dependency of the whole build stack needs to be packaged. Only a small percentage of these are already packaged and some may not be the right version. Then every bower dependency will need to be packaged as well. Then you can finally start thinking about an official debian package.

dlech:
I have a few of the issues worked out. Some questions:

1. Package names: Since you plan on having both a graphical and command line version, what should the package names be? I am using `etcher` for graphical version since that is probably what most people will install. So `etcher-cli` can be the command line version.
2. Executable names: along the same lines, we can't have the same executable name for both the graphical and command line versions. I think using `etcher` for the command line version makes the most sense since it will by typed the most. I am using `etcher-ui` for the graphical version. However, these are at odds with the package names.
3. Launchpad or Bintray: If you want to do Launchpad, I will have to figure out a way to create a "source" package to upload. If you go with bintray, then you can actually use anything you like for the distribution name. For example, you could call the distribution "etcher". I think this might reduce confusion so that people don't think the package only works on precise. Also, I don't think you can change the component on Launchpad. So if you want a beta channel, it would require two ppas, whereas on bintray, you can have a "main" component and a "beta" component.
4. Icon name: one thing you may have not noticed is that I changed the name of an icon file in the `assets/` folder. This probably breaks the appimage build. If we can change the name without breaking stuff, great. If not I will have to find another way to make it work.

jviotti:
> Package names: Since you plan on having both a graphical and command line version, what should the package names be? I am using etcher for graphical version since that is probably what most people will install. So etcher-cli can be the command line version.
> Executable names: along the same lines, we can't have the same executable name for both the graphical and command line versions. I think using etcher for the command line version makes the most sense since it will by typed the most. I am using etcher-ui for the graphical version. However, these are at odds with the package names.

Hm, are you aware of any Linux application which faces the same issue (they have both GUI and CLI executables)? How do they solve it?

> Launchpad or Bintray: If you want to do Launchpad, I will have to figure out a way to create a "source" package to upload. If you go with bintray, then you can actually use anything you like for the distribution name. For example, you could call the distribution "etcher". I think this might reduce confusion so that people don't think the package only works on precise. Also, I don't think you can change the component on Launchpad. So if you want a beta channel, it would require two ppas, whereas on bintray, you can have a "main" component and a "beta" component.

Got it, lets go with Bintray in that case.

> Icon name: one thing you may have not noticed is that I changed the name of an icon file in the assets/ folder. This probably breaks the appimage build. If we can change the name without breaking stuff, great. If not I will have to find another way to make it work.

Happy to make such change! Can you send a separate PR for this, and rebase this once its merged afterwards?

***

I noticed there are some files that are duplicated inside `debian/`, like the LICENSE. Could we copy `LICENSE` during the building phase instead of committing a copy? I guess `debian/` could be an skeleton, but the complete directory is put together by a script on a temporary location. 

Also, I see some debian packaging specific commands in `.travis.yml`, like `quilt`, `debuild`, etc. Can we move everything specific to debian package generator as a new `debian` command in `./scripts/build/linux.sh`? I guess you can also check if the command equals `debian` and conditionally omit the `upx` portions to avoid having to maintain patches. What do you think?

I don't like the `create-bintray-json.sh` script very much, and the fact that is not very clear how to publish to Bintray outside of Travis CI (lets say for whatever reason I need to do it myself).

Can we create a new script in (maybe at `scripts/publish-bintray.sh`, and rename the current `scripts/publish.sh` to `scripts/publish-aws.sh`, or even better, extend the existing script to accept `bintray` or `aws` commands) and put all the logic to reliably publish to Bintray there, and make Travis CI just call that script? Would this eradicate the need for a `bintray.json`?

dlech:
> Hm, are you aware of any Linux application which faces the same issue (they have both GUI and CLI executables)? How do they solve it?

All of the ones I can think of off hand are a little different since they involve a daemon so there is usually something like `foo-applet` for a tray icon and `fooctl` for a command line application. The other ones I can think of have a completely different name and are described as "graphical front end for foo".

I guess I will have to do some more digging to see if I can come up with anything.

dlech:
> I noticed there are some files that are duplicated inside debian/, like the LICENSE. Could we copy LICENSE during the building phase instead of committing a copy? I guess debian/ could be an skeleton, but the complete directory is put together by a script on a temporary location.

It is rather unconventional, but I suppose we could do something like this. Most of the time the debian packaging exists separate from upstream, so I am used to this duplication.

The license is copied because the debian/copyright file is in a special machine-readable format, so it is not an exact copy. The .desktop file is copied as well, but again, it is not an exact copy - it is configured exactly for the debian package.

dlech:
> Can we move everything specific to debian package generator as a new debian command in ./scripts/build/linux.sh?

Sure.

jviotti:
@alexandrosm Any thoughts here?

The `etcher-ui` name is not that bad, but UI stands for "user interface", and of course both the CLI and the GUI have user interfaces. `etcher-gui` might be better. Alternatively, we can follow what other Linux applications do by noting the GUI technology as part of the executable name (think `kdesudo` for `sudo`), and do something like `etcher-electron`, which also opens the possibility of creating a non-electron GUI in the future without having to be concerned about naming again.

dlech:
> I guess you can also check if the command equals debian and conditionally omit the upx portions to avoid having to maintain patches. What do you think?

The current patches are the same as #737, so if you merge that, they will be dropped from debian/patches.

dlech:
> I don't like the create-bintray-json.sh script very much, and the fact that is not very clear how to publish to Bintray outside of Travis CI (lets say for whatever reason I need to do it myself).

Why not? :smiley: 

It is easy to construct a `curl` command to upload, so you don't need to rely on the travis magic. See https://bintray.com/docs/api/#_debian_upload.

You can also manually upload files to bintray using the website, so no problems there.

dlech:
> Can we create a new script in (maybe at scripts/publish-bintray.sh, and rename the current scripts/publish.sh to scripts/publish-aws.sh, or even better, extend the existing script to accept bintray or aws commands) and put all the logic to reliably publish to Bintray there, and make Travis CI just call that script? Would this eradicate the need for a bintray.json?

Yes, we can do this. And yes, we can do it without bintray.json.

jviotti:
@dlech I see. I vote to create a publishing script and calling it from Travis CI instead of relying on Travis CI releasing magic, since we'll be extended the Travis CI builds to actually produce and publish AppImages and DMGs, so having a single script entry point will ease that transition when the time comes.

dlech:
I also found [debian-electron-installer](https://github.com/unindented/electron-installer-debian). I don't particularly like it as a matter of personal preference, but if you want to automatically generate the info in the `debian/` folder, this seems like the way to go.

jviotti:
@dlech Can we do something like this in the build script?

```sh
mkdir -p etcher-release
cp -rf debian etcher-release/debian
cp LICENSE etcher-release/debian/LICENSE
# Build .deb based on etcher-release/debian
```

dlech:
You could, but there are easier ways to get the same result.

jviotti:
I'm happy with both. Is not a big deal in any case, just trying to avoid duplication if its easy to do so.

dlech:
Closing in favor of #747

Issue #716, opened 2016-09-23
Suggestion for displaying image metadata (logo and description)


dlech:
Currently, when you load an image with metadata, it will look something like this:

![etcher_009](https://cloud.githubusercontent.com/assets/963645/18795511/90969e00-818a-11e6-8998-79d66d1271f1.png)

I would like to suggest that instead of replacing the filename with the "name" from the metadata, that "SELECT IMAGE" is replaced instead to show that the image has been selected. It would look something like this.

![etcher_009_mod](https://cloud.githubusercontent.com/assets/963645/18795654/2eae6d66-818b-11e6-8e71-da65f7b12cd8.png)

BertLindeman:
Maybe not really at it's place here, but could there be a little better contrast between the gray background and the gray text? Or maybe a slightly larger font size? I have to take a photo to be able to read the text. As an example in the image above the text "show full file name".

jviotti:
Hi @dlech ,

Thanks a lot for the suggestion. I definitely agree with it, and will make the changes very soon.

@BertLindeman 

Thanks for raising this up. What do you think @taahirisaacs ?

lurch:
I agree that small grey text on a grey background is not the easiest thing to read (I imagine it's even worse on a screen with reduced colour depth!)

Also: would it be possible to display the full-image-path (e.g. as a tooltip) when mouse-hovering over the "SHOW FULL FILE NAME" ?

WasabiFan:
> Also: would it be possible to display the full-image-path (e.g. as a tooltip) when mouse-hovering over the "SHOW FULL FILE NAME" ?

See https://github.com/resin-io/etcher/issues/634#issuecomment-240274982

jviotti:
I guess we could, but is much harder to protect from long file names causing the tooltip from overflowing the application screen. For simplicity purposes, I think the modal is the way to go, but maybe we can show something else as the tooltip, like the image basename, or truncate the absolute file path, although that might not be very useful anyway.

lurch:
> but is much harder to protect from long file names causing the tooltip from overflowing the application screen

Ahh, I hadn't considered that. In which case it probably makes sense to keep the "SHOW FULL FILE NAME" button as a modal, and instead show the full basename of the image when mouse-hovering over the truncated image name ("nightly-ev3d...09-04.img.xz" in the screenshot above).
[...]

PR #736, opened 2016-09-30
Change where the "Image Name" gets displayed, and display basename(imagepath) as a tooltip on the truncated imagename

lurch:
Fixes a couple of the problems identified in #716

### DIFF
--- a/lib/gui/pages/main/templates/main.tpl.html
+++ b/lib/gui/pages/main/templates/main.tpl.html
@@ -2,7 +2,7 @@
   <div class="col-xs" ng-controller="ImageSelectionController as image">
     <div class="box text-center" os-dropzone="image.selectImage($file)">
       <svg-icon class="center-block" path="{{ main.selection.getImageLogo() || '../../../assets/image.svg' }}"></svg-icon>
-      <span class="icon-caption">SELECT IMAGE</span>
+      <span class="icon-caption">{{ main.selection.getImageName() || "SELECT IMAGE" }}</span>
       <span class="badge space-top-medium">1</span>
 
       <div class="space-vertical-large">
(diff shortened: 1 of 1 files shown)


jviotti:
Looks good, but can we split these two changes as two separate commits? We rely on certain commit conventions to be able to auto-generate CHANGELOG, which becomes an issue if two changes are mixed in a single patch.

lurch:
No problem, will do.

lurch:
#745 

I can't create a PR for the other part of this PR yet though, because the two changes would conflict. (so after #745 has been merged into master, I can create a new PR)

WasabiFan:
You could also just add two separate commits to the same PR. As long as you follow their commit conventions, it should work out fine.

lurch:
I dunno what resin.io's commit conventions are yet, but I _personally_ prefer one commit per PR.

WasabiFan:
You can look at any commit in their commit history to see the items they used. Compared to waiting for the first PR before making the second round of changes, adding an additional commit seems a lot simpler! In the end you wouldn't even need to generate a merge commit if they asked GitHub to rebase on merge.

lurch:
> You can look at any commit in their commit history to see the items they used.

Good point :+1: 

@jviotti would #745 be a 'feat(GUI)' or a 'style(GUI)' ? (or something else?)

alexandrosm:
but this raises the point that we should document what we're doing with our
commit convention. It's really cool stuff using the versionist project, and
I'm sure juanchi will want to spend more time writing it up next week, when
he's freed up from the first batch of the reconfiguration work.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

WasabiFan:
> but this raises the point that we should document what we're doing with our commit convention.

Isn't that this? https://github.com/resin-io/etcher/blob/master/docs/COMMIT-GUIDELINES.md

alexandrosm:
@lurch -- there you go :)

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lurch:
Oops, I managed to miss that! \*embarrased\*

@alexandrosm Maybe you could make use of https://help.github.com/articles/setting-guidelines-for-repository-contributors/ ?

jviotti:
The tooltip is definitely a feature IMO, while the other one is more of a style. We should definitely write a CONTRIBUTING document that serves as an entry point to the other documentation files (like a toc, or something like that).

alexandrosm:
@lurch *I* managed to miss that also. I should have had more faith in
@jviotti's relentless drive to document everything :D

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

PR #737, opened 2016-09-30
Move upx compression to appimage command in linux build script

dlech:
Debian packaging tools cannot analyze the binaries if they are compressed
with upx. If we move the compression to the appimage command, then the
output of the package command can be used for building debian packages.


jviotti:
Looks great! I left a suggestion to workaround the UPX patches at https://github.com/resin-io/etcher/pull/733, which you can still do if you like.

PR #745, opened 2016-10-03
Display basename(imagepath) as a tooltip on the truncated imagename

lurch:


lurch:
Thanks for merging! I'll try to add a "better" commit message on my next PR ;-)

Issue #756, opened 2016-10-12
Error: No polkit authentication agent found.


TheSandman86:
- **Etcher version: 1.0.0-beta.15**
- **Operating system and architecture: Xubuntu 16.04.1 64 bits**
- **Do you see any meaningful error information on DevTools?**

I'm getting this error while triying to flash an image on a SD Card,
This is what i see on the DevTools console:
Wed Oct 12 2016 11:56:07 GMT-0300 (ART) Flash error
tracker.js:10 Wed Oct 12 2016 11:56:29 GMT-0300 (ART) Flash ({"image":"/home/mauricio/Downloads/ubuntu-16.04.1-desktop-amd64.iso","device":"/dev/mmcblk0"})
tracker.js:10 WRITER: Tailing /tmp/etcher-2553487Od3NFKK75z.log
tracker.js:10 WRITER: Tail starting:
tracker.js:10 WRITER: filename: /tmp/etcher-2553487Od3NFKK75z.log
tracker.js:10 WRITER: options: {}
tracker.js:10 WRITER: Attempting to elevate
tracker.js:10 Error: { Error: No polkit authentication agent found.
    at /tmp/.mount_7QwaJI/usr/bin/resources/app.asar/node_modules/sudo-prompt/index.js:132:23
    at ChildProcess.exithandler (child_process.js:221:5)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:850:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:215:5)
  cause: 
   Error: No polkit authentication agent found.
       at /tmp/.mount_7QwaJI/usr/bin/resources/app.asar/node_modules/sudo-prompt/index.js:132:23
       at ChildProcess.exithandler (child_process.js:221:5)
       at emitTwo (events.js:106:13)
       at ChildProcess.emit (events.js:191:7)
       at maybeClose (internal/child_process.js:850:16)
       at Process.ChildProcess._handle.onexit (internal/child_process.js:215:5),
  isOperational: true }

    at Socket.child.stderr.on (/tmp/.mount_7QwaJI/usr/bin/resources/app.asar/lib/src/child-writer/index.js:83:29)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:172:18)
    at Socket.Readable.push (_stream_readable.js:130:10)
[...]

PR #758, opened 2016-10-12
chore(debian): add depedency on polkit package

dlech:
A policy kit authentication agent is needed for privilege escalation.

See: https://github.com/resin-io/etcher/issues/756#issuecomment-253298322

### DIFF
--- a/scripts/build/debian/config.json
+++ b/scripts/build/debian/config.json
@@ -39,7 +39,8 @@
     "libxrandr2",
     "libxrender1",
     "libxss1",
-    "libxtst6"
+    "libxtst6",
+    "polkit-1-auth-agent | policykit-1-gnome | polkit-kde-1"
   ],
   "recommends": [
     "libnotify4"


scripts/build/debian/config.json
```diff
@@ -39,7 +39,8 @@
     "libxrandr2",
     "libxrender1",
     "libxss1",
-    "libxtst6"
+    "libxtst6",
+    "polkit-1-auth-agent | policykit-1-gnome"
```
dlech: Yes. However, it was only created in 2016, so older OSes won't have that package. There are several packages that `Provides: policykit-1-gnome`, so that package name provides the same sort of virtual package function, so hopefully this will cover most cases. Although now that I am thinking about it, I should check to see if there is a KDE version as well.

dlech:
updated with KDE polkit package

jviotti:
Looks awesome! Thanks a lot!

PR #764, opened 2016-10-19
Include Etcher version in installer/archive names

### LINKED ISSUE #763: Name installer executables (and portable versions too) so that they include the Etcher version number
Currently, my Downloads folder looks like this:

image

You'll notice that, aside from the debug builds I downloaded for diagnosing an error, I have a bunch of Etcher installer executables which are auto-incrementing with a suffixed number as I download new ones. For me at least, I have had to (and likely will have to in the future) go back to old versions of Etcher for various reasons, and attempting to discern which version is which is pretty much impossible with it like that. _Yes, I could name them myself or delete old versions, but I'm a lazy programmer who ain't got time for that._

I think it would be helpful if Etcher installers were named such that they included the version number of the app that they contained. Not only would this prevent name collisions (and by extension auto-incremented names) but it would also make it much easier to figure out which version you downloaded a week ago. It would be even nicer if, for the case where one is running special builds, the installer name included the user-supplied name of the person who built the thing (which would be set as part of the build process).

WasabiFan:
This works towards #763 on Windows, in theory. I haven't tested it (because installing the dependencies would be a pain) but it seems to be the right thing to do. This does not make any changes for other OSes.

### DIFF
--- a/scripts/build/windows.bat
+++ b/scripts/build/windows.bat
@@ -132,8 +132,6 @@ if "%arch%"=="x64" (
   set electron_arch=x64
 )
 
-set package_name=Etcher-win32-%arch%
-
 for /f %%i in (' "node -e ""console.log(require('./package.json').devDependencies['electron-prebuilt'])""" ') do (
   set electron_version=%%i
 )
@@ -157,6 +155,8 @@ for /f %%i in (' "node -e ""console.log(require('./package.json').version)""" ')
 for /f %%i in (' "node -v" ') do (
   set node_version=%%i
 )
+
+set package_name=Etcher-%etcher_version%-win32-%arch%
 
 :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 :: Configure NPM to build native addons for Electron correctly
@@ -214,8 +214,10 @@ call %electron_packager% . %application_name%^
 set package_output=%output_build_directory%\%package_name%
 
 if not "%arch%"=="%electron_arch%" (
-  move %output_build_directory%\Etcher-win32-%electron_arch% %package_output%
-)
+    move %output_build_directory%\Etcher-win32-%electron_arch% %output_build_directory%\Etcher-win32-%arch%
+) 
+
+move %output_build_directory%\Etcher-win32-%arch% %package_output%
 
 :: Omit *.dll and *.node files from the asar package, otherwise
 :: `process.dlopen` and `module.require` can't load them correctly.
@@ -242,7 +244,7 @@ cd %package_output%^
 :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 
 set installer_tmp_output=%output_build_directory%\win32-%arch%-tmp-installer
-set installer_output=%output_directory%\Etcher-win32-%arch%.exe
+set installer_output=%output_directory%\%package_name%.exe
 
 call %electron_builder% %package_output%^
  --platform=win^
--- a/scripts/build/darwin.sh
+++ b/scripts/build/darwin.sh
@@ -114,33 +114,33 @@ function installer_zip {
   mkdir -p $output_directory
   sign $source_directory/$APPLICATION_NAME.app
   pushd $source_directory
-  zip -r -9 Etcher-darwin-x64.zip $APPLICATION_NAME.app
+  zip -r -9 Etcher-$APPLICATION_VERSION-darwin-x64.zip $APPLICATION_NAME.app
   popd
-  mv $source_directory/Etcher-darwin-x64.zip $output_directory
+  mv $source_directory/Etcher-$APPLICATION_VERSION-darwin-x64.zip $output_directory
 }
 
 function installer_dmg {
   local source_directory=$1
   local output_directory=$2
-  local temporal_dmg=$source_directory.dmg
+  local temporary_dmg=$source_directory.dmg
   local volume_directory=/Volumes/$APPLICATION_NAME
   local volume_app=$volume_directory/$APPLICATION_NAME.app
 
   # Make sure any previous DMG was unmounted
   hdiutil detach $volume_directory || true
 
-  # Create temporal read-write DMG image
-  rm -f $temporal_dmg
+  # Create temporary read-write DMG image
+  rm -f $temporary_dmg
   hdiutil create \
     -srcfolder $source_directory \
     -volname "$APPLICATION_NAME" \
     -fs HFS+ \
     -fsargs "-c c=64,a=16,e=16" \
     -format UDRW \
-    -size 600M $temporal_dmg
+    -size 600M $temporary_dmg
 
-  # Mount temporal DMG image, so we can modify it
-  hdiutil attach $temporal_dmg -readwrite -noverify
+  # Mount temporary DMG image, so we can modify it
+  hdiutil attach $temporary_dmg -readwrite -noverify
 
   # Wait for a bit to ensure the image is mounted
   sleep 2
@@ -204,20 +204,20 @@ function installer_dmg {
 
   sign $volume_app
 
-  # Unmount temporal DMG image.
+  # Unmount temporary DMG image.
   hdiutil detach $volume_directory
 
-  # Convert temporal DMG image into a production-ready
+  # Convert temporary DMG image into a production-ready
   # compressed and read-only DMG image.
   mkdir -p $output_directory
   rm -f $output_directory/Etcher-darwin-x64.dmg
-  hdiutil convert $temporal_dmg \
+  hdiutil convert $temporary_dmg \
     -format UDZO \
     -imagekey zlib-level=9 \
-    -o $output_directory/Etcher-darwin-x64.dmg
+    -o $output_directory/Etcher-$APPLICATION_VERSION-darwin-x64.dmg
 
-  # Cleanup temporal DMG image.
-  rm $temporal_dmg
+  # Cleanup temporary DMG image.
+  rm $temporary_dmg
 
 }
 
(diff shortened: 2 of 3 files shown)


lurch:
IMHO if Etcher makes this change, it should do so consistently for all platforms.

WasabiFan:
Yes, I agree. However this PR only does it for Windows. I may propose equivalent changes for the other platforms in this PR later, but this was the simplest initial implementation to start the discussion.

jviotti:
Looks great! Lets keep expanding this PR to include support for other platforms. Happy to comment out patches here for other platforms so they can be integrated in this same PR.

WasabiFan:
I'm not really sure how this applies to the other platforms, given that on Linux there's no installer (just the new Debian package from @dlech) and I really have no idea what's on OS X. Would you be able to point out what I should change or just make it yourself (you can push to my PR branch or bring my commit into your own)?

dlech:
There was a discussion somewhere about packaging the Linux AppImage in a zip to take care of setting the executable permission. When this is done, the zip should have the version name, but not necessarily the AppImage. So, probably nothing to do for Linux yet.

For macOS, the the .app is packaged in a .dmg, so you could make the same sort of change for the .dmg name [here](https://github.com/resin-io/etcher/blob/feb120bf8c9de1c17d8cbad647eb3b674fb2ce44/scripts/build/darwin.sh#L125)

WasabiFan:
Is "temporal" on the line that @dlech mentioned (and the other uses of the word) supposed to be "temporary"?

WasabiFan:
I updated this PR to include the equivalent change for the macOS build script. Keep in mind that none of this has been tested, so someone will need to do that before this is merged.

I made the change in a different location from that which @dlech suggested because this seemed like the `temporal_dmg` string wasn't actually the name of the final product.

We'll need to make sure that version numbers don't include any characters that may break this; presumably since it is retrieved from the `package.json` it will just be periods and numbers.

jviotti:
@WasabiFan 

> Is "temporal" on the line that @dlech mentioned (and the other uses of the word) supposed to be "temporary"?

Yeah, it should be `temporary`, my bad! 

> I updated this PR to include the equivalent change for the macOS build script. Keep in mind that none of this has been tested, so someone will need to do that before this is merged.

I'll give it a go for you now.

> I made the change in a different location from that which @dlech suggested because this seemed like the temporal_dmg string wasn't actually the name of the final product.

That's right. The temporary DMG gets converted into the actual final DMG at https://github.com/resin-io/etcher/blob/feb120bf8c9de1c17d8cbad647eb3b674fb2ce44/scripts/build/darwin.sh#L214

> We'll need to make sure that version numbers don't include any characters that may break this; presumably since it is retrieved from the package.json it will just be periods and numbers.

Right. Remember the hyphen as well (from `-beta.X`).

jviotti:
I can confirm this works great in OS X :+1:

jviotti:
@dlech 

> There was a discussion somewhere about packaging the Linux AppImage in a zip to take care of setting the executable permission. When this is done, the zip should have the version name, but not necessarily the AppImage. So, probably nothing to do for Linux yet.

I guess most users will extract from the zip, delete the zip, and just handle the AppImage, which can end up in the same problem: many AppImages, and not clear which version is which.

Maybe we should add the versions to both? In such case, lets add the version to the AppImage in this PR as well.

dlech:
I consider the AppImage like a .app on macOS. When the user extracts it, they will copy it over an existing version, so you want the file name to be the same. I don't think most people would want two different versions installed at one time. Having two AppImages at one time would also complicate things with the .desktop file. Do you have 2 menu icons since there are 2 versions?

jviotti:
I've tested the Windows changes, and the resulting file in `etcher-release/installers` still seems to lack the version. I fixed it with the following change:

```diff
diff --git a/scripts/build/windows.bat b/scripts/build/windows.bat
index f0d809c..cf8c5ea 100644
--- a/scripts/build/windows.bat
+++ b/scripts/build/windows.bat
@@ -231,7 +231,7 @@ upx -9 %package_output%\*.dll
 :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::

 set installer_tmp_output=%output_build_directory%\win32-%arch%-tmp-installer
-set installer_output=%output_directory%\Etcher-win32-%arch%.exe
+set installer_output=%output_directory%\%package_name%.exe

 call %electron_builder% %package_output%^
  --platform=win^
```

Yet another repetition instead of using `%package_name%`, I should really put some time to refactor these scripts :)

jviotti:
@dlech 

> I consider the AppImage like a .app on macOS. When the user extracts it, they will copy it over an existing version, so you want the file name to be the same. I don't think most people would want two different versions installed at one time. Having two AppImages at one time would also complicate things with the .desktop file. Do you have 2 menu icons since there are 2 versions?

Its a great point, however I've seen testers having multiple AppImage builds/versions and use those to report back information in issues. I'm with you that AppImages are like OS X `*.app`s and that having the version there would look a bit awkward indeed.

Lets just have it on the zip for consistency with the OS X package (where the `zip`/`dmg` contains the version but not the actual `*.app` inside).

WasabiFan:
I just made the fix for Windows and renamed the "temporal" variable; is there anything else you'd like me to do? I haven't been bothering with commit messages because I assume you're just going to squash this anyway.

jviotti:
@WasabiFan I just sent a PR to distribute the AppImage inside a ZIP for GNU/Linux (see https://github.com/resin-io/etcher/pull/772). I propose to wait for that PR to get merged and rebase this one, so we can add the version number for the ZIP as well and have all OSes covered by this single PR.

jviotti:
As a side note, this PR now has conflicts on `darwin.sh`, caused by https://github.com/resin-io/etcher/pull/771. I basically just prefixed all variable definitions with `local`, so should be easy to resolve.

jviotti:
The PR has been merged!

WasabiFan:
Sounds good; I'll rebase onto master and make the equivalent changes for Linux in a few hours.

lurch:
...and then once *this* PR has been merged too, I'll continue with my script refactoring :-)
https://github.com/resin-io/etcher/pull/767#issuecomment-255804649

jviotti:
@lurch WRT script refactoring, are you in the Gitter channel? I have some ideas that might be worth discussing whenever you can (ping me).

WasabiFan:
I just finished these changes and squashed my commits. Can I get someone to test this on the three platforms to make sure I didn't screw anything up?

jviotti:
Hi @WasabiFan ,

Thanks a lot for the amazing work! I'll be doing some testing and will update the list below:

- [ ] Windows
- [ ] macOS
- [ ] Linux

jviotti:
@WasabiFan I spotted a minor issue on the Windows scripts. Turns out `electron-packager` will always output a package with its own naming convention, ignoring what we put in `%package_output%` (namely `Etcher-win32-%arch%`), which results in some ENOENT errors later on, that I didn't catch last time since in batch scripts, things continue even if the return value of the command is not 0 (and therefore get lost).

The following patch fixes everything for me:

```diff
diff --git a/scripts/build/windows.bat b/scripts/build/windows.bat
index ff53912..487d56b 100644
--- a/scripts/build/windows.bat
+++ b/scripts/build/windows.bat
@@ -212,6 +212,7 @@ call %electron_packager% . %application_name%^
  --out=%output_build_directory%

 set package_output=%output_build_directory%\%package_name%
+move %output_build_directory%\Etcher-win32-%arch% %package_output%

 if not "%arch%"=="%electron_arch%" (
   move %output_build_directory%\Etcher-win32-%electron_arch% %package_output%
```

jviotti:
@WasabiFan Everything works flawlessly! Just the Windows fix I pointed out above and we should be ready to merge.

WasabiFan:
The line in the `if` statement following the part you referenced is the same as the one you added; can it be removed? I can't tell exactly what it does.

jviotti:
@WasabiFan I'm quite sure it can be removed, but let me give it a go so I don't make you lose more time with yet another change in case it doesn't!

WasabiFan:
OK, sounds good; it really isn't much of an issue for me to spend a few extra minutes on it. Also note that you can push directly to my PR branch to amend my changes.

jviotti:
@WasabiFan I can confirm we can indeed remove that conditional! 

> Also note that you can push directly to my PR branch to amend my changes.

Ah, cool, didn't know that! Maybe for next time!

WasabiFan:
I'll make that change then!

> Ah, cool, I didn't know that! Maybe for next time!

I'm pretty sure it is a new feature.

image

WasabiFan:
Updated!

scripts/build/windows.bat
```diff
@@ -212,10 +212,7 @@ call %electron_packager% . %application_name%^
  --out=%output_build_directory%
 
 set package_output=%output_build_directory%\%package_name%
-
-if not "%arch%"=="%electron_arch%" (
-  move %output_build_directory%\Etcher-win32-%electron_arch% %package_output%
-)
```
lurch: I think this if-test should be restored, with the line inside the if changed to:
```
  move %output_build_directory%\Etcher-win32-%electron_arch% %output_build_directory%\Etcher-win32-%arch%
```

because if %arch% is "x86" then %electron_arch% will be "ia32" https://github.com/WasabiFan/etcher/blob/47d0594019cf5d775c3f9ee118e099a853be4baf/scripts/build/windows.bat#L127

WasabiFan: That seems to make sense, however @jviotti decided that it wasn't needed; was that just a case of not testing for 32-bit?

jviotti: Totally right. The line should be changed to what @lurch suggested. I did try a 32-bit build, however I guess I had stuff on my `etcher-release` before running the command, which made the issue slip undetected. Sorry for that!

WasabiFan: ✨ No problem, PR updated.

jviotti:
Merged! Thanks a lot for the amazing work @WasabiFan !

WasabiFan:
Great! Let's hope I didn't break anything in the process :laughing:

PR #767, opened 2016-10-24
chore: fix output_package unbound variable in GNU/Linux

jviotti:
The `$output_package` variable is defined inside the `package_x$ARCH`
functions, which means that if we run the following commands:

```sh
./scripts/build/linux.sh package x64
./scripts/build/linux.sh appimage x64
```

The `$output_package` variable will not be defined when the script runs
the `appimage` command.

To workaround this issue, we declare a new `$package_directory` variable
inside the `appimage` command clause.

### DIFF
--- a/scripts/build/linux.sh
+++ b/scripts/build/linux.sh
@@ -197,15 +197,19 @@ if [ "$COMMAND" == "debian" ] || [ "$COMMAND" == "all" ]; then
 fi
 
 if [ "$COMMAND" == "appimage" ] || [ "$COMMAND" == "all" ]; then
+  package_directory=etcher-release/Etcher-linux-$ARCH
+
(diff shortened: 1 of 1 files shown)


lurch:
`etcher-release` occurs in this file 8 separate times, `Etcher-linux-` appears in the file 9 separate times, and the `package_x86` and `package_x64` functions are almost identical so could be easily combined into one.

Would you like me to add additional tidy-ups to the script, or is it simply not worth it?

jviotti:
@lurch I definitely agree. This particular file has a lot of repetition and is not very organized anymore. I have a todo item in my personal list to get through it at some point, but if you can do it instead, that'd be awesome, lets get it some good refactoring!

Issue #738, opened 2016-10-01
 Error: Command failed


Olivier3869:
- **Etcher version:** 1.0.0-beta.15
- **Operating system and architecture:** Linux Mint 18 
- **Do you see any meaningful error information on DevTools?**

```
{ Error: Command failed: /usr/bin/sudo -n -E -- sh -c 'mkdir -p /tmp/.mount_0MCj2p-elevated && mount -o loop -o ro /home/olivier/Documents/Etcher-linux-x64.AppImage /tmp/.mount_0MCj2p-elevated && env ELECTRON_RUN_AS_NODE=1 ETCHER_INTERNAL_LOG_FILE=/tmp/etcher-20612vdxJVz3egp9o.log /tmp/.mount_0MCj2p-elevated/usr/bin/etcher /tmp/.mount_0MCj2p-elevated/usr/bin/resources/app.asar/lib/src/child-writer/writer-proxy.js /tmp/.mount_0MCj2p-elevated/usr/bin/resources/app.asar /media/olivier/Softs/Images\ ISO\ d\'OS/Antivirus/kav_rescue_10.iso --robot --drive /dev/sde --unmount --check --log /tmp/etcher-20612vdxJVz3egp9o.log ; sleep 1 ; umount /tmp/.mount_0MCj2p-elevated'
/bin/sh: 1: Syntax error: Unterminated quoted string

    at ChildProcess.exithandler (child_process.js:214:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:850:16)
    at Socket.<anonymous> (internal/child_process.js:323:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at Pipe._handle.close [as _onclose] (net.js:492:12)
  cause: 
```

lurch:
Looks like this is caused by the fact you have a `'` in your filepath - I guess https://github.com/resin-io/etcher/blob/master/lib/src/child-writer/writer-proxy.js#L143 needs to fixed to properly escape quote characters.

jviotti:
Good catch @lurch !

ilansavir:
Is there a solution for this?

it happens randomly

jviotti:
@ilansavir, I haven't had time to look into this yet, but if @lurch 's theory is correct, removing the quote or any other special character from the image name should be enough.

lurch:
[...]

PR #768, opened 2016-10-24
fix(GUI): correctly escape single quotes when running inside AppImage

jviotti:
If the application is running from inside an AppImage, the command that
is elevated gets passed as an argument to `sh -c` surrounded by single
quotes.

This means that the usual way of escaping single quotes (with a
backslash), takes no effect:

> Enclosing characters in single quotes preserves the literal value of
> each character within the quotes. A single quote may not occur between
> single quotes, even when preceded by a backslash.
>
> From http://stackoverflow.com/a/1250098/1641422

In this case, the solution is really hacky and obscure: replace single
quotes with `'"'"'`, which ends up being parsed as `'` by the shell.

The following changes have been introduced to accomodate to the above
workaround:

- Rename `utils.escapeWhiteSpacesFromArguments` to
  `utils.escapeArguments`.

- Accept an `willBeSurroundedInSingleQuotes` boolean option in
  `utils.escapeArguments`.

- Make `writer-proxy.js` pass the `willBeSurroundedInSingleQuotes` when
  elevating inside an AppImage.

See: http://stackoverflow.com/a/1250279/1641422
Fixes: https://github.com/resin-io/etcher/issues/738
Change-Type: patch
Changelog-Entry: Fix "Command Failed" error when attempting to flash an image containing a single quote anywhere in the path in GNU/Linux.

### DIFF
--- a/lib/src/child-writer/utils.js
+++ b/lib/src/child-writer/utils.js
@@ -111,19 +111,32 @@ exports.getCLIWriterArguments = (options) => {
 };
 
 /**
- * @summary Escape white spaces from arguments
+ * @summary Escape arguments
  * @function
  * @public
  *
  * @param {String[]} argv - argv
+ * @param {Object} [options={}] - options
+ * @param {Boolean} [options.willBeSurroundedInSingleQuotes=false] - will be surrounded in single quotes
  * @returns {String[]} escaped arguments
  *
  * @example
- * const escapedArguments = utils.escapeWhiteSpacesFromArguments(process.argv);
+ * const escapedArguments = utils.escapeArguments(process.argv, {
+ *   willBeSurroundedInSingleQuotes: true
+ * });
  */
-exports.escapeWhiteSpacesFromArguments = (argv) => {
+exports.escapeArguments = (argv, options = {}) => {
   return _.map(argv, (argument) => {
-    return argument.replace(/\s/g, '\\ ');
+    const escapedArgument = argument.replace(/\s/g, '\\ ');
+
+    if (options.willBeSurroundedInSingleQuotes) {
+
+      // See http://stackoverflow.com/a/1250279/1641422
+      return escapedArgument.replace(/'/g, '\'"\'"\'');
+
+    }
+
+    return escapedArgument;
   });
 };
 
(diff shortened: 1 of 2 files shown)


lurch:
Could the commit-message (and PR-comment) be updated to refer to `utils.escapeArguments` (which is what the function is actually named), rather than `utils.escapedArguments` ?

`willBeSurroundedInSingleQuotes` seems like a bit of a verbose option name? I wonder if `singleQuoteStyle` would be neater? (although it's not quite as self-explanatory)

If `willBeSurroundedInSingleQuotes` is false, IMHO that implies that the escaped-string will be surrounded in double-quotes instead. So if `willBeSurroundedInSingleQuotes` is false, should the function  escape `"` chars in the filename?

I guess this is the sort of thing that ideally ought to have unit-tests... ;)

jviotti:
> Could the commit-message (and PR-comment) be updated to refer to utils.escapeArguments (which is what the function is actually named), rather than utils.escapedArguments ?

Oh, great catch! Will update now...

> `willBeSurroundedInSingleQuotes` seems like a bit of a verbose option name? I wonder if `singleQuoteStyle` would be neater? (although it's not quite as self-explanatory)

I can't think of any name that makes the fact we're referring specifically to surrounding the command crystal clear. I think the verbosity its outweighed by the clarity of the name.

> If `willBeSurroundedInSingleQuotes` is false, IMHO that implies that the escaped-string will be surrounded in double-quotes instead. So if `willBeSurroundedInSingleQuotes` is false, should the function escape " chars in the filename?

If such option is false, it means that the arguments in question won't be surrounded at all. In the case of allowing arguments to be surrounded by double quotes, a new option should probably be introduced.

> I guess this is the sort of thing that ideally ought to have unit-tests... ;)

The whole `lib/src/child-writer` module is a big, dirty, temporal hack that I really want to get rid of. The ideal solution would be to implement cross platform elevation as a NodeJS C++ Addon that communicates with the native APIs directly, therefore deprecating a lot of the hacks we're throwing it here. I'd rather push to start work on the "better" solution.

jviotti:
This escaping code could also be killed if we find a way to properly execute binaries as `root` inside a mounted AppImage.

lurch:
See also #773 
Rather than trying to 'escape' all the special-characters in the string, maybe it'd be better to just double-quote the image filename?

jviotti:
I think this PR can be closed in favor of https://github.com/resin-io/etcher/pull/775, but I have to test first.

jviotti:
Closing this in favor of https://github.com/resin-io/etcher/pull/775. Such PR is a better solution to this problem.

Issue #760, opened 2016-10-15
Windows portable version


tooomm:
- **Operating system and architecture:** Win 10 64bit

I would highly appreciate to see a portable version of this sweet tool! ❤️

jviotti:
Hi @tooomm,

Thanks for the suggestion! We do build a portable version at one point during the packaging, but only re-distribute the final installer. Uploading the portable version as well sounds like a good idea!

tooomm:
You already somehow build it? That sounds great.
Makes just distributing it an easier task. :)

jviotti:
The installer gets created around an standalone version of the app. Of course, the standalone version consists of many files (dlls and other resources) but can be run from anywhere. 

Is this what you're looking for, or do you mean a single standalone executable instead?

lurch:
I'm assuming he meant something along the lines of http://portableapps.com/ i.e. that can just be run by extracting a zipfile to a local folder?

tooomm:
Actually 
>many files (dlls and other resources) but can be run from anywhere

is pretty much what i'm looking for. 
Is there are a version hosted somehwere that i can test?

I don't use suits like portableapps.com, having just a single executable file is easier to handle for sure!

jviotti:
@tooomm

Sounds good! We don't have builds like that uploaded anywhere, but you can build one from source if you like (see https://github.com/resin-io/etcher/blob/master/docs/RUNNING-LOCALLY.md. Once you have a locally running build, you can run `.\scripts\build\windows.bat package <arch>` to get the portable build).

tooomm:
I happily wait for a official distribution then. ;)

DJuego:
Me too. 👍  Please! :-) 
Thanks!
DJuego

P.S.  I have discovered this tool from ev3dev project.

lurch:
Just in case the people who've commented on this issue haven't seen this yet:
https://talk.resin.io/t/etcher-1-0-0-beta16-release/379

PR #769, opened 2016-10-24
chore: create standalone Windows builds

jviotti:
This PR zips the package that is built by running:

```sh
.\scripts\build\windows.bat package <arch>
```

And puts it in `etcher-release/installers`.

Fixes: https://github.com/resin-io/etcher/issues/760
Change-Type: minor
Changelog-Entry: Publish standalone Windows builds.

### DIFF
--- a/scripts/build/windows.bat
+++ b/scripts/build/windows.bat
@@ -102,10 +102,17 @@ if %ERRORLEVEL% neq 0 (
   exit /b 1
 )
 
-:: Check that asar is installed.
+:: Check that python is installed.
 where python >nul 2>nul
 if %ERRORLEVEL% neq 0 (
   echo Dependency missing: python 1>&2
+  exit /b 1
+)
+
+:: Check that 7z is installed.
+where 7z >nul 2>nul
+if %ERRORLEVEL% neq 0 (
+  echo Dependency missing: 7z 1>&2
   exit /b 1
 )
 
@@ -225,6 +232,10 @@ signtool sign^
 signtool verify /pa /v %package_output%\Etcher.exe
 
 upx -9 %package_output%\*.dll
+
+cd %package_output%^
+ && 7z a -tzip ..\installers\%package_name%.zip *^
+ && cd ..\..
 
 :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 :: Generate installer
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,6 +22,7 @@ install:
   - npm install -g npm bower rimraf asar
   - choco install nsis -version 2.51
   - choco install upx
+  - choco install 7zip.commandline
   - set PATH=C:\Program Files (x86)\Windows Kits\8.1\bin\x86;%PATH%
   - set PATH=C:\Program Files (x86)\NSIS;%PATH%
   - set PATH=C:\Ruby22\bin;%PATH%
--- a/docs/RUNNING-LOCALLY.md
+++ b/docs/RUNNING-LOCALLY.md
@@ -21,6 +21,7 @@ Prerequisites
 - [Asar](https://github.com/electron/asar)
 - [NSIS v2.51](http://nsis.sourceforge.net/Main_Page) (v3.x won't work)
 - [Visual Studio Community 2013](https://www.visualstudio.com/en-us/news/vs2013-community-vs.aspx)
+- [7z](http://www.7-zip.org) (command line version)
 
 ### OS X
 


lurch:
There's a couple of places where `Etcher-win32-%arch%` is used, which I guess could be replaced by `%package_name%` ?
(Might impact #764 as well?)

jviotti:
Great catch! Fixed!

jviotti:
Hm, we should probably make sure 7z is installed in Appveyor to avoid the missing dependency error.

scripts/build/windows.bat
```diff
@@ -226,6 +233,10 @@ signtool verify /pa /v %package_output%\Etcher.exe
 
 upx -9 %package_output%\*.dll
 
+cd %package_output%^
+ && 7z a -tzip ..\installers\%package_name%.zip *^
```
WasabiFan: Why use 7zip? It seems like it would be both easier in usage and environment setup to just use something built-in, e.g. http://stackoverflow.com/questions/17546016/how-can-you-zip-or-unzip-from-the-command-prompt-using-only-windows-built-in-ca

jviotti: I see, I didn't know Windows provided any built-in zip support we could use programmatically, so I just headed straight to 7z. I'll give the PowerShell and VBS suggestions a try and will see how that goes.

lurch: I'm not sure how 'clean' those 3rd-party scripts are? Looks like some might work in some environments, and some in others.
Seems like just using 7zip might be more reliable?

jviotti: We can always include `7z.exe` on this repo and get the best out of both worlds. Thoughts?

jviotti: As a side note, the VBS script doesn't seem to work at all.

lurch: I'm not sure I'd count 7-zip as "sketchy" ? :-S

@WasabiFan The Windows built-in zip has problems with files bigger than 4GB (which happens quite often with disk images), so if you want to manually extract those you're _forced_ to use something like 7-zip.

@jviotti From a technical point of view, you've listed 7-zip as a runtime dependency, and I guess strictly speaking it's only a build / packaging dependency.

WasabiFan: > I'm not sure I'd count 7-zip as "sketchy" ? :-S

I think a large part of my aversion stems from the fact that they used to host their downloads on SourceForge and redirected you there to download. Even today, they have links pointing there, and their web page is ugly and unkempt to the point where I still wouldn't download it unless forced to.

The more important idea here is that minimizing dependencies is important. Adding a dependency means adding one more thing that a contributor must download and means opening the process up to more uncontrollable failure modes. For example, what happens when an unknowing contributor installs 7-zip (as prompted by the failing build message) but fails to check the "Add to path" option? They'll continue getting that error even though they have installed the app already. This kind of an issue can easily dissuade prospective contributors.

> As a side note, the VBS script doesn't seem to work at all, and the PowerShell one only works on Windows 8 and newer.

Hmmm, that's unfortunate. I'd contend that people shouldn't be on pre-Windows 8 versions anyway, but in general that isn't a maintainable position 😞 In that case, including the binary seems like the best bet.

Honestly, it might make sense for you to take the binaries for as many of Etcher's build deps as possible and push them to a server so that they could then be downloaded during the build, greatly simplifying the work on a contributor. That's a larger-scale plan though.

lurch: > their web page is ugly and unkempt to the point where I still wouldn't download it unless forced to.

I guess some people prefer working on improving their software, to working on prettifying their website? ;-)

> I'd contend that people shouldn't be on pre-Windows 8 versions anyway

My Windows laptop is still chugging away quite happily on Win7.

> Honestly, it might make sense for you to take the binaries for as many of Etcher's build deps as possible and push them to a server so that they could then be downloaded during the build, greatly simplifying the work on a contributor.

IIRC that's kinda the idea behind [Vagrant](https://www.vagrantup.com/) - but as you say, a larger-scale project.

jviotti: > @jviotti From a technical point of view, you've listed 7-zip as a runtime dependency, and I guess strictly speaking it's only a packaging dependency.

Yeah, the build scripts take care of both initializing the project for development purposes, and dealing with packaging, which is inconvenient. I propose to handle it in the build script refactoring, and get it fixed there for once and for all.

***

I propose to merge this PR to get the ball rolling. With the refactoring @lurch is working on, the script will not complain about packaging-specific dependencies, which means the required dependencies needed to get the project up and running for development purposes will be very few.

In the case of packaging, we're looking forward to move all that to CI servers (code-signing and security policies from the company are stopping us a bit here, but we'll get that sorted), which means very few people will actually do it on their machines anyway.

alexandrosm:
speaking of vagrant, maybe there's a middle ground whereby we package all
the build deps in a docker container? newer Docker (claims to) run linux
containers on all platforms, could that be viable?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

dlech:
Re docker: yes, in fact I have done this already to build the linux version of etcher because I didn't want to install all of the build deps on my computer. If you want to open a new issue, I will share my Dockerfile.

jviotti:
Hi @dlech ,

@lurch is currently working on this exact same problem, and I believe he's considering using Docker to cross-compile for different Linux architectures. Maybe worth sharing with him in Gitter or something like that?

lurch:
I'm afraid I'm working for another client today, so will be too busy to pop into Gitter.
But jut FYI I tried using https://hub.docker.com/r/i686/ubuntu/ to build the x86 version of the Etcher installers on a x64 host machine. After fixing a few problems along the way (and installing the missing dependencies as I went) it eventually failed on trying to mount the FUSE volume, and I run out of time in trying to debug that further.

But yeah, feel free to open a new issue for further discussions :-)

tooomm:
Cool, can I grab that somewhere? Will it be available on the webpage at one point?

jviotti:
Hi @tooomm ,

I've created a custom build for you: https://www.dropbox.com/s/556asankd5e2mnf/Etcher-win32-x64.zip?dl=0. The standalone `zip` will be available on the website for the next version!

tooomm:
Thank you @jviotti 
It works, just tested it... Looking forward to the next release and download from the website!

lurch:
Looks like this is now live on https://www.etcher.io/  :-)

PR #772, opened 2016-10-24
chore: distribute AppImage inside a zip

jviotti:
AppImages need to be marked as executables before they can be used.
Distributing them directly means that virtually all web browsers will
automatically remove the execution permissions, leading us to have to
explain users how to add it back with `chmod`, etc.

For simplicity purposes, we'll be distributing AppImages inside ZIPs,
which ensure the execution permissions are added back when the user
decompresses it.

### DIFF
--- a/scripts/build/linux.sh
+++ b/scripts/build/linux.sh
@@ -156,6 +156,12 @@ function installer {
   app_dir_create $source_directory $architecture $appdir_temporary_location
   rm -f $output_file
   ./scripts/build/AppImages/AppImageAssistant-$architecture $appdir_temporary_location $output_file
+
+  pushd $output_directory
+  zip Etcher-linux-$architecture.zip Etcher-linux-$architecture.AppImage
+  rm Etcher-linux-$architecture.AppImage
+  popd
+
   rm -rf $appdir_temporary_location
 }
 


scripts/build/linux.sh
```diff
@@ -156,6 +156,12 @@ function installer {
   app_dir_create $source_directory $architecture $appdir_temporary_location
   rm -f $output_file
   ./scripts/build/AppImages/AppImageAssistant-$architecture $appdir_temporary_location $output_file
+
+  pushd $output_directory
+  zip Etcher-linux-$architecture.zip Etcher-linux-$architecture.AppImage
```
dlech: will you add the version to .zip file name later?

jviotti: @dlech Yeah, it will be added in the following PR (https://github.com/resin-io/etcher/pull/764#issuecomment-255871914), which will be rebased with these changes.

probonopd:
Please do not put AppImages into any kind of archive. The reasons are given on https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages#common-mistake

jviotti:
@probonopd We put it in a ZIP due to the massive amount of (newbie) users getting confused, and asking for support (even though we had instructions in our website).

Breaking appimaged desktop integration sounds like a good argument, though. See https://github.com/resin-io/etcher/pull/1703. What do you think about adding a page in appimage.org with very clear instructions on how to run AppImages, targeted at newbies? We could link to it from various places (e.g. our own website) and educate users that way.

probonopd:
Definitely on our TODO list for the upcoming, revamped website. In the meantime, there is http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80 - feel free to copy it.

Issue #642, opened 2016-08-22
Windows 10 Unexpected end of json


Bottersnike:
- Etcher version 1.0.0-beta.13
- Windows 10 64bit
- DevTools printed the following:
```
tracker.js:10 Error: undefined:1
{

SyntaxError: Unexpected end of JSON input
    at Object.parse (native)
    at EventEmitter.tail.on (C:\Program Files (x86)\Etcher\resources\app.asar\lib\src\child-writer\writer-proxy.js:76:25)
    at emitOne (events.js:96:13)
    at EventEmitter.emit (events.js:188:7)
    at ReadStream.<anonymous> (C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\file-tail\index.js:52:19)
    at emitOne (events.js:96:13)
    at ReadStream.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:172:18)
    at ReadStream.Readable.push (_stream_readable.js:130:10)
    at onread (fs.js:1743:12)
    at FSReqWrap.wrapper [as oncomplete] (fs.js:675:17)

    at Socket.child.stderr.on (C:\Program Files (x86)\Etcher\resources\app.asar\lib\src\child-writer\index.js:85:29)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:172:18)
    at Socket.Readable.push (_stream_readable.js:130:10)
    at Pipe.onread (net.js:542:20)

```

It stopped at 45%. I'm retrying now and will edit this with the result of a retry.

jviotti:
Hey there,

Thanks a lot for reporting. This is an interesting issue. This is what I
believe its happening:

The Etcher GUI basically spawns the Etcher CLI as a child process that
emits JSON lines that are then read and interpreted by the GUI. Each
line is supposed to contains a fully valid JSON object, but that seems
to fail in your setup for some reason (the buffer gets full, or the data
is flushed before the whole line has been received for some reason).

Is this something you can actively reproduce?

If so, you can see that we print the path to a temporary file in
DevTools (which you can open by pressing `Ctrl+Alt+I`) right when you
start a flash, which is the location where the JSON lines are temporary
saved.

Can you start a flash, wait until it fails with this error, and without
closing Etcher (otherwise the temp file will be wiped out), send us what
the temporary file contains?

I'm wondering if your system is emitting a huge speed number, or
something else that could trigger this.

[...]

PR #774, opened 2016-10-25
fix(GUI): emit progress back to parent using node-ipc

jviotti:
This PR makes use of `node-ipc` to emit progress information from the
child CLI to the GUI process rather than the tailing approach we've
been doing until now.

This change was motivated by the following Electron fix which landed in
v1.4.4: https://github.com/electron/electron/pull/7578. Before such fix,
Electron would not output anything to stdout/stderr if the Electron
process was running with `ELECTRON_RUN_AS_NODE` under Windows, which
forced us to implement `--log` option in the Etcher CLI to output state
information to a file.

Since this issue is fixed, we can consume the Etcher CLI output from
within `child_process.spawn`, which opens more interesting possibilities
for sharing information between both processes.

This coindentally fixes a Windows issue where the tailing module would
receive malformed JSON, causing Etcher to crash at `JSON.parse`. The
reason of this problem was a bug in the tailing module we were using.

Fixes: https://github.com/resin-io/etcher/issues/642
Change-Type: patch
Changelog-Entry: Fix "Unexpected end of JSON" error in Windows.


lib/src/child-writer/writer-proxy.js
```diff
@@ -197,18 +165,23 @@ return isElevated().then((elevated) => {
...
+        const child = childProcess.spawn(EXECUTABLE, ETCHER_ARGUMENTS);
+        child.on('error', reject);
+        child.on('close', resolve);
+
+        const emitMessage = (data) => {
+          ipc.of[process.env.IPC_SERVER_ID].emit('message', data.toString());
+        };
+
+        child.stdout.on('data', emitMessage);
+        child.stderr.on('data', emitMessage);
```
lurch: I guess this means that the server can't tell whether the 'message' has come from stdout or stderr of the child? Would it be worth setting up two IPC channels, so the server can deal with errors different to normal output?

docs/CLI.md
```diff
@@ -38,7 +38,6 @@ Options
   --drive, -d    drive
   --check, -c    validate write
   --robot, -r    parse-able output without interactivity
-  --log, -l      output log file
```
lurch: Why does the ability for the parent and child to communicate via IPC mean that the ability to write a logfile should be removed? Surely it's the 'robot' option that should be removed, if anything?
Apologies if I'm misunderstanding things! ;)

lib/src/child-writer/writer-proxy.js
jviotti: @lurch Errors are marked with an `error` type property anyway, so there is no need for the distinction.

jviotti:
I just realized I have to update `lib/src/child-writer/README.md` to reflect the new changes.

lurch:
So, to check that I'm understanding what this is doing...

Before now, the etcher-gui used to launch the etcher-cli (with elevated privileges, as it's the etcher-cli that actually writes to disks), with the etcher-gui reading and parsing the 'robot' output of etcher-cli in order to update the disk-flashing status in the GUI.

After this PR, the etcher-gui generates a random-hex-string to use as the IPC name, creates a couple of IPC channels, sets a couple of environment variables, and launches the etcher-cli (again with elevated privileges). The etcher-cli then reads in the two env-vars, connects to the parent's IPC channels, and then uses IPC to report disk-flashing progress to the parent (etcher-gui) process.

Is that about right?

lib/src/child-writer/index.js
```diff
@@ -56,20 +58,46 @@ const EXIT_CODES = require('../exit-codes');
...
+    entryPoint: rendererUtils.getApplicationEntryPoint(),
+    image: image,
+    device: drive.device,
+    validateWriteOnSuccess: options.validateWriteOnSuccess,
+    unmountOnSuccess: options.unmountOnSuccess
+  });
+
+  // There might be multiple Etcher instances running at
+  // the same time, therefore we must ensure each IPC
+  // server/client has a randomly generated name.
```
WasabiFan: I already asked this somewhere else, but why not use the host PID or system time instead? It seems like this opens you up to potential collisions where you don't need to.

lib/src/child-writer/writer-proxy.js
lurch: How does that work, when the messages are just passed as strings?
https://github.com/resin-io/etcher/pull/774/files#diff-dcf350e93272ccc9aabbc8169ec80230R178

lib/src/child-writer/index.js
lurch: 20-random-bytes is 160-bit 'security' ;-)
Perhaps using a PID or time value would be too easy for other (malicious) processes to guess? \*shrug\*

lib/src/child-writer/index.js
```diff
@@ -56,20 +58,46 @@ const EXIT_CODES = require('../exit-codes');
...
+  });
+
+  ipc.server.on('message', (data) => {
+    const message = JSON.parse(data);
 
+    // The error object is decomposed by the CLI for serialisation
+    // purposes. We compose it back to an `Error` here in order
+    // to provide better encapsulation.
+    if (message.command === 'error') {
+      const error = new Error(message.data.message);
```
WasabiFan: Doesn't this mean that the call stack will be lost (the error object will populate the stack dump with the location of this code)? The error reporting here should be focused on pretty intently, because this is by far the most common issue I see people have with Etcher: Cryptic errors bubbling up from the CLI.

lib/src/child-writer/writer-proxy.js
WasabiFan: What happens when data emitted over the pipe is invalid?

lib/src/child-writer/index.js
jviotti: I haven't considered PIDs, which is way more elegant, but as @lurch pointed out, it'd allow other potentially malicious processes to inject information to the GUI.

Even if a malicious process sends stuff to the IPC server, there is no possible harm to be done to the user. The IPC server is just used to display progress information, so I guess we're safe to go with PIDs.

lib/src/child-writer/writer-proxy.js
jviotti: @lurch The string is a JSON stringified object.

lib/src/child-writer/index.js
jviotti: Very, very good point. I'll fix it.

lib/src/child-writer/index.js
WasabiFan: If there is a malicious program running on the machine, I'm sure it has better things to do than mess with Etcher :laughing: Maybe include the string "Etcher" to prevent other processes which also happen to use numbers from interfering.

lib/src/child-writer/writer-proxy.js
jviotti: If this handler received invalid data, its because the Etcher CLI is sending invalid data (e.g: non-valid JSON strings), which should be fixed in the CLI.

lib/src/child-writer/index.js
lurch: Yeah, PIDs are more guessable, but guaranteed-unique as @WasabiFan said in his earlier comment.

Hopefully a malicious program telling the etcher-gui that the disk progress is -15% or +250% won't cause the GUI to crash ;-)

jviotti:
@lurch The main difference is that before, we piped the output of the Etcher CLI to a file, and then use the writer proxy to tail such file as a way to consume the information. Now, the Etcher CLI directly communicates with the parent process (the GUI) using IPC, so no more temporary files are involved in the process.

> EDIT: Ah, I see from your earlier comment that there's actually a writer-proxy program which facilitates communication between the cli and gui portions. How come? ;) Some kind of "clean API barrier"?

Exactly, the GUI is not concerned at all about how the writing takes place and how the information is retrieved.

lurch:
> Now, the Etcher CLI directly communicates with the parent process (the GUI) using IPC, so no more temporary files are involved in the process.

Hmmm, I'm afraid I'm still confused. Where does the writer-proxy fit in? Or is it now "bypassed"?
Is the `--robot` option still needed?

lib/src/child-writer/writer-proxy.js
WasabiFan: Agreed, however I think it's a case worth handling. It would be better to show an error because invalid data was detected than throw an uncontrolled error because the parse failed, potentially compromising the write process.

lurch: Yup, the more robust the error-handling can become, the better.

jviotti:
@lurch The "child-writer" module that lives in `lib/src/child-writer` has the sole purpose of decoupling the GUI from the dirty work of forking the CLI with the right arguments, and retrieving the progress state, exit state, etc. The whole purpose of such piece of code is to provide encapsulation and a nice to use interface. By making use of it, the GUI can trigger a flash by simply doing:

```js
const child = childWriter.write('path/to/image', '/dev/disk2', {
  validateWriteOnSuccess: true,
  unmountOnSuccess: true
});

child.on('error', (error) => {
  console.error(error);
});

child.on('done', () => {
  console.log('The flashing completed!');
});

child.on('progress', (progress) => {
  console.log(progress);
});
```

## What the child writer module used to to before

The child writer module entry point would:

- Create a temporary file.
- Build a list of command line arguments to pass to the CLI.
- Execute the writer proxy as a child process using `.fork()`, which enabled NodeJS's built-in IPC technology, which can be used by calling `process.send()` from the child process (see https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options):

> The returned ChildProcess will have an additional communication channel built-in that allows messages to be passed back and forth between the parent and child. See child.send() for details.

Now that the writer proxy has the control, the following things happen:

- Since this process can talk to the child process entry point (which was called from the GUI) using NodeJS's build-in IPC technology, it starts tailing the temporary file created before, and pipes anything it gets to the GUI using `process.send()`.
- The writer proxy checks if its running as `root`/Administrator.
- If so, carries on its way. Otherwise, it prompts the user for elevation and *re-executes itself*.
- Now that the writer proxy is running with full permissions (either because it had them, or because it re-executed itself), the writer proxy executes the Etcher CLI as a child process, passing the `--log` option to force the CLI to output any information printed at runtime to the temporary file created by the child writer entry point, which is being tailed by the writer proxy. The `--robot` option complements `--log` by making the CLI output progress state in a way that can be easily parsed (e.g: stringified JSON lines).
- When the CLI starts executing, and starts printing progress state, it writes it to the temporary file. The writer proxy receives those messages since its tailing the temporary file, which are then passed to the child writer entry point through IPC communication.

## What the child writer module does now

The child writer module entry point would:

- Open an IPC server using [`node-ipc`](https://www.npmjs.com/package/node-ipc).
- Spawn the writer proxy as a child process. It prints its stdout and stderr for debugging purposes, but we are not really concerned with what the writer proxy print now.

Now that the writer proxy has the control, the following things happen:

- The writer proxy checks if its running as `root`/Administrator.
- Now that the writer proxy is running with full permissions (either because it had them, or because it re-executed itself), the writer proxy executes the Etcher CLI as a child process, this time without `--log`, since it will pipe its `stdout`/`stderr` to the IPC server created by the child writer entry point directly.

***

As you can see, things got really simplified after this PR. The writer proxy basically has the responsibilities of making sure we have enough privileges (and doing something about it if not), and spawning the CLI, piping its messages to the IPC server.

The `--robot` option is only concerned with making the CLI output easily parseable by a program, which is really helpful of course independently of these changes.

jviotti:
Please let me know if you have any questions, or if you think `lib/src/child-writer/README.md` could be improved to better explain this.

lib/src/child-writer/writer-proxy.js
jviotti: Sounds good! Keep in mind that at this point, the writer proxy simply blindly pipes anything it gets from the CLI to the IPC server created by the child writer, therefore, the error handling logic will take place in the IPC server message handler, which is the place where we actually try to "interpret" the data, and thus the place we can better know if its valid or not.

lib/src/child-writer/index.js
```diff
@@ -56,20 +57,50 @@ const EXIT_CODES = require('../exit-codes');
...
+  ipc.config.id = process.env.IPC_SERVER_ID;
+  ipc.config.silent = true;
+  ipc.serve();
+
+  ipc.server.on('error', (error) => {
+    emitter.emit('error', error);
+  });
+
+  ipc.server.on('message', (data) => {
+    const message = JSON.parse(data);
```
WasabiFan: Could this be surrounded in a `try...catch` which would also emit an error, in the same way that not having a `command` or `data` property would?

jviotti: Done! I've not re-used the error message from the error thrown by `JSON.parse` since its completely unfriendly and non-informational.

lurch:
Thanks for the detailed explanation @jviotti ! :+1: 

I guess I was getting confused because this PR actually does **two** things (changes from tailing a logfile, to just reading stdout; and changes the IPC mechanism slightly), whereas I thought it was only doing one thing.
If you don't  mind, could you explain a bit more about how/why the IPC mechanism changed?

lurch:
Regarding the 'if' at https://github.com/resin-io/etcher/pull/774/files#diff-dcf350e93272ccc9aabbc8169ec80230R47 - should it have an 'else' block after it?

jviotti:
Yeah, there are two changes that somewhat depend on each other.

> If you don't mind, could you explain a bit more about how/why the IPC mechanism changed?

So the NodeJS built-in IPC mechanism only allows communication between a process that is directly forked by its parent. Given that we need to communicate from a process to potentially its "grandfather" process (if elevation took place), the built-in IPC mechanism falls short, unless we make the process communicate with its parent, which then communicates with its parent, which is way more complex.

In order to simplify this, we make use of [`node-ipc`](https://github.com/RIAEvangelist/node-ipc), which is another (third-party) IPC implementation that allows communication between any process, independently on their relationship.

jviotti:
> Regarding the 'if' at https://github.com/resin-io/etcher/pull/774/files#diff-dcf350e93272ccc9aabbc8169ec80230R47 - should it have an 'else' block after it?

The body of the `if` conditional has a `return`, which makes the `else` unnecessary.

WasabiFan:
I think @lurch was pointing out that the if statement special-cases Windows, but there isn't an equivalent for the other platforms.

lurch:
...and it's such a big chunk-o-code that it's not immediately obvious that there's a `return` buried in there, whereas adding an `else` would make the code-flow more obvious, IMHO.

lurch:
> Given that we need to communicate from a process to potentially its "grandfather" process (if elevation took place), the built-in IPC mechanism falls short, unless we make the process communicate with its parent, which then communicates with its parent, which is way more complex.

Aha! That helps, thankyou :-)

jviotti:
I rather have an `if`-guard with a `return` statement rather than `else` and indenting the whole block below, which is even larger. The Windows section has the `return` statement very close to the `if` (2 lines below), so I guess its not a big deal.

lurch:
> The Windows section has the return statement very close to the if (2 lines below), so I guess its not a big deal.

Fair enough - I guess I was just confused by all the nested functions / try-catch stuff, etc.

Just to clarify - does this mean that the `if (process.env.APPIMAGE && process.env.APPDIR) {` block is the "linux part" and the `return sudoPrompt.execAsync(command, {` is the "darwin / OSX part" ?

jviotti:
So Windows makes use of `elevator` (which is the Windows block we're referring to). `sudo-prompt` applies to both GNU/Linux *and* OS X, so that part is the same for both. The part that checks for `APPIMAGE` and `APPDIR` is inside an [`_.attempt`](https://lodash.com/docs/4.16.4#attempt) function, which gets executed instantly. Inside the function, we return a specialized command if we detect we're running inside an AppImage, or a generic one that works for both GNU/Linux and OS X if not.

Issue #637, opened 2016-08-19
Etcher leaves behind process after burning and exiting the UI


imrehg:
- Etcher 1.0.0-beta13
- Arch Linux 64bit

Starting Etcher and burning a file, afterwards exiting leaves an etcher process behind. It does not happen if the program is started, not burned, and exited. I guess this means something is up with the elevation?

Here's my system after 3 start-burn-and-exit:

```
$ps uax |grep etcher
greg      1376  0.2  1.5 974088 127336 pts/9   Sl   11:56   0:01 /tmp/.mount_zADdyB/usr/bin/etcher /tmp/.mount_zADdyB/usr/bin/resources/app.asar/lib/src/child-writer/writer-proxy.js /tmp/.mount_zADdyB/usr/bin/resources/app.asar /home/greg/Downloads/FreeBSD-10.3-RELEASE-arm-armv6-RPI-B.img.xz --robot --drive /dev/sdf --unmount --check --log /tmp/etcher-563Ws0zjh0K4YSQ.log
greg      9482  0.0  1.5 965892 125976 pts/3   Sl   10:35   0:02 /tmp/.mount_8MehiU/usr/bin/etcher /tmp/.mount_8MehiU/usr/bin/resources/app.asar/lib/src/child-writer/writer-proxy.js /tmp/.mount_8MehiU/usr/bin/resources/app.asar /home/greg/Downloads/ada/2016-03-25-28c.img.xz --robot --drive /dev/sdf --unmount --check --log /tmp/etcher-8512x2nVk1SDGXhM.log
greg     21982  0.0  1.5 958720 126156 pts/7   Sl   11:26   0:01 /tmp/.mount_3avsg6/usr/bin/etcher /tmp/.mount_3avsg6/usr/bin/resources/app.asar/lib/src/child-writer/writer-proxy.js /tmp/.mount_3avsg6/usr/bin/resources/app.asar /home/greg/Downloads/resin-RapiroRPi3-1.8.0-1.13.0-7ad55d1c946b.img --robot --drive /dev/sdf --unmount --check --log /tmp/etcher-212983245IIEEg1b0.log
```

jviotti:
Hi @imrehg ,

Yeah, the problem is that is quite hard to ensure previously used custom mount points get unmounted. AppImages solve this issue by using FUSE, which ensures the AppImage is automatically unmounted when is no longer used, but FUSE prevented us from implementing elevation, so we're doing a basic mounting that is not guaranteed to go away.

[...]

Issue #773, opened 2016-10-25
Elaborate Error when writing to disk. What is wrong?


mendeldesign:
![schermafbeelding 2016-10-25 om 14 25 43](https://cloud.githubusercontent.com/assets/831082/19686078/f246373e-9abf-11e6-9903-116f7f954782.png)

![schermafbeelding 2016-10-25 om 14 26 07](https://cloud.githubusercontent.com/assets/831082/19685919/22907f0e-9abf-11e6-91d2-bb93b69c6dfd.png)

- **Etcher version: 1.0.0-beta.15**
- **Operating system and architecture: OSX 10.11.6**
- **Do you see any meaningful error information on DevTools?**

mendeldesign:
It seems to be saying that the image file was not ok, but that img file actually worked a month ago on a different SD card.

lurch:
Looks like it's going wrong because of the `&` in the file-path? See also #738

mendeldesign:
O wow you are right, dragged it to the desktop and it works. Thank you for the quick reply, I was already downloading new images from the raspberry website.

PR #775, opened 2016-10-26
feat(GUI): spawn CLI directly from the AppImage in GNU/Linux

jviotti:
If we pass relative paths as arguments to the AppImage, they get
resolved from `/tmp/.mount_XXXXXX/usr`. We can exploit this to run the
Etcher CLI directly from the AppImage, rather than having to mount it
ourselves:

```sh
ELECTRON_RUN_AS_NODE=1 Etcher-linux-x64.AppImage bin/resources/app.asar [OPTIONS]
```

By using this little trick, we get rid of both our custom mounting logic
and the need of surrounding the command to run in single quotes,
therefore avoiding a whole new kind of escaping issues.

Since running the CLI directly from the AppImage means that the
`desktopintegration` will be ran for the CLI, we pass the `SKIP`
environment variable to avoid having it prompt the user for
installation.

We also updated the `desktopintegration` script to the latest version,
which reduces the amount of logging plus other minor fixes.

Fixes: https://github.com/resin-io/etcher/issues/637
Change-Type: patch
Changelog-Entry: Fix Etcher leaving zombie processes behind in GNU/Linux.

### DIFF
--- a/scripts/build/AppImages/desktopintegration
+++ b/scripts/build/AppImages/desktopintegration
@@ -4,9 +4,23 @@
 # into the host system without special help from the host system.
 # If you want to use it, then place this in usr/bin/$APPNAME.wrapper
 # and set it as the Exec= line of the .desktop file in the AppImage.
-
+#
+# For example, to install the appropriate icons for Scribus,
+# put them into the AppDir at the following locations:
+#
+# ./usr/share/icons/default/128x128/apps/scribus.png
+# ./usr/share/icons/default/128x128/mimetypes/application-vnd.scribus.png
+#
+# Note that the filename application-vnd.scribus.png is derived from
+# and must be match MimeType=application/vnd.scribus; in scribus.desktop
+# (with "/" characters replaced by "-").
+#
+# Then, change Exec=scribus to Exec=scribus.wrapper and place the script
+# below in usr/bin/scribus.wrapper and make it executable.
+# When you run AppRun, then AppRun runs the wrapper script below
+# which in turn will run the main application.
+#
 # TODO:
-# Handle icons for mime types as well using xdg-icon-resource
 # Handle multiple versions of the same AppImage?
 # Handle removed AppImages? Currently we are just setting TryExec=
 # See http://specifications.freedesktop.org/thumbnail-spec/thumbnail-spec-latest.html#DELETE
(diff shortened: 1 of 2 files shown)


jviotti:
Hm, I'm suddenly getting a weird error after the `desktopintegration` upgrade:

```
bash: ./Etcher-linux-x64.AppImage: No such file or directory
```

lurch:
I haven't tested, but I'm not convinced that this'll fix #773 
(I still think just using double-quoted strings would be better than trying to 'escape' space characters)

jviotti:
I found the cause of the error: the `desktopintegration` script was committed without execution permissions.

> I haven't tested, but I'm not convinced that this'll fix #773 
(I still think just using double-quoted strings would be better than trying to 'escape' space characters)

Oops, wrong issue, sorry. I meant https://github.com/resin-io/etcher/issues/738, which I can confirm it has indeed being fixed by these changes. I'll definitely try the double quotes solution you suggested to see if can eradicate the other issues completely with it.

scripts/build/AppImages/desktopintegration
```diff
@@ -69,13 +111,13 @@ yesno()
   TITLE=$1
   TEXT=$2
   if [ -x /usr/bin/zenity ] ; then
-    LD_LIBRARY_PATH="" zenity --question --title="$TITLE" --text="$TEXT" || exit 0
+    LD_LIBRARY_PATH="" zenity --question --title="$TITLE" --text="$TEXT" 2>/dev/null || exit 0
   elif [ -x /usr/bin/kdialog ] ; then
     LD_LIBRARY_PATH="" kdialog --caption "Disk auswerfen?" --title "$TITLE" -yesno "$TEXT" || exit 0
```
jviotti: @lurch Hm, that is weird, I never noticed it. This script comes without any modifications from the AppImageKit project, so I suggest raising an issue at https://github.com/probonopd/AppImageKit, and I'll update accordingly afterwards.

jviotti: The original script lives in https://github.com/probonopd/AppImageKit/blob/master/desktopintegration

lurch:
> Oops, wrong issue, sorry. I meant #738

You never mentioned #773 , nothing to apologise for ;)
I was just trying to say that while this issue might fix the escaping issue in #738 I don't think it'll fix #773 as well.

jviotti:
Oh, I'm really confused. I thought I had referenced an issue, but maybe that is from another PR :) Yeah, lets try to double quote idea and see if that fixes the other one as well!

scripts/build/AppImages/desktopintegration
lurch: https://github.com/probonopd/AppImageKit/issues/265

PR #777, opened 2016-10-26
chore: Fix minor install script problems

lurch:
When running the install script as root (e.g. inside a Docker container), bower complained that it

"shouldn't" be run with sudo. When running the x86 install multiple times, it tried to move

Etcher-linux-ia32 inside the existing Etcher-linux-x86/ folder, rather than renaming it, which meant

later commands failed.

### DIFF
--- a/scripts/build/linux.sh
+++ b/scripts/build/linux.sh
@@ -88,7 +88,7 @@ function install {
 
   rm -rf node_modules bower_components
   npm install --build-from-source
-  bower install --production
+  bower install --production --allow-root
 }
 
 function package_x86 {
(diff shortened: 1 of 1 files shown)


lurch:
This is just a "stop-gap" fix until my other Makefile-based refactoring is finished :)

jviotti:
Looks great, thanks a lot!

lurch:
I tried using "commitizen", which is why the line-wrapping in the commit message has come out weird.

jviotti:
Yeah the commitizen plugin we're using is not "there" yet. I have a personal TODO item about forking it to spawn `$EDITOR` instead.

dlech:
> When running the install script as root (e.g. inside a Docker container),

Why do you not run as a normal user inside of a Docker container?

lurch:
\*shrug\*
I've never really used Docker before, so I was just using the `docker run -it ubuntu bash` that I'd seen in some tutorial somewhere.

jviotti:
I remember this particular issue bit @alexandrosm and I when we were trying to build Etcher as a Resin.io application, so I'm still happy with this fix.

PR #779, opened 2016-10-26
fix(GUI): surround paths in double quotes before elevating

jviotti:
This PR makes sure every command line argument that represents an
absolute path is surrounded by double quotes, to avoid any potential
escaping issue.

This simplifies a lot the various special character escaping routines we
had in place, since we now only have to make sure double quotes inside
the paths are escaped.

Fixes: https://github.com/resin-io/etcher/issues/773
Change-Type: patch
Changelog-Entry: Prevent escaping issues during elevation by surrounding paths in double quotes.

### DIFF
--- a/lib/src/child-writer/utils.js
+++ b/lib/src/child-writer/utils.js
@@ -16,9 +16,6 @@
 
 'use strict';
 
-const _ = require('lodash');
-const os = require('os');
-
 /**
  * @summary Get the explicit boolean form of an argument
  * @function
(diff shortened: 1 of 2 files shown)


lurch:
Cool! :tada: 
I haven't tested this yet, but it's great to see the removal of the clunky `escapeWhiteSpacesFromArguments` function :-)

Although I wonder why this is only done for absolute paths, rather than all paths?

jviotti:
@lurch My line of thinking was that we should only surround paths in double quotes, and since every path that we handle at this point is ensured to be in absolute form, checking for `path.isAbsolute()` would be enough. There is no easy way to determine if something is a relative path if we wanted to cover those as well, however I just realized that surrounding everything, including options, is harmless, for example:

```sh
"env" "ELECTRON_RUN_AS_NODE=1" "IPC_SERVER_ID=etcher-server-1320" "IPC_CLIENT_ID=etcher-client-1320" "SKIP=1" "/Users/jviotti/Projects/resin/etcher/node_modules/electron-prebuilt/dist/Electron.app/Contents/Frameworks/Electron Helper.app/Contents/MacOS/Electron Helper" "/Users/jviotti/Projects/resin/etcher/lib/src/child-writer/writer-proxy.js" "/Users/jviotti/Projects/resin/etcher/lib/start.js" "/Users/jviotti/Downloads/Images/core (os).iso" "--robot" "--drive" "/dev/disk2" "--unmount" "--check"
```

I'll update the PR accordingly.

PR #783, opened 2016-10-28
docs(GUI): explain why some drives are not bootable out of the box

jviotti:
The usual case is images requiring special treatment. We recommend the
users to use specialised tooling for those cases.

### DIFF
--- a/docs/USER-DOCUMENTATION.md
+++ b/docs/USER-DOCUMENTATION.md
@@ -3,6 +3,20 @@ Etcher User Documentation
 
 This document contains how-tos and FAQs oriented to Etcher users.
 
+Why is my drive not bootable?
+-----------------------------
+
+Etcher copies images to drives byte by byte, without doing any transformation
+to the final device, which means images that require special treatment to be
+made bootable, like Windows images, will not work out of the box. In these
+cases, the general advice is to use software specific to those kind of
+images, usually available from the image publishers themselves.
+
+Images known to require special treatment:
+
+- Microsoft Windows (use [Windows USB/DVD Download Tool][windows-usb-tool], or
+  [Rufus][rufus]).
+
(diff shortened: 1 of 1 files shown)


docs/USER-DOCUMENTATION.md
```diff
@@ -3,6 +3,15 @@ Etcher User Documentation
 
 This document contains how-tos and FAQs oriented to Etcher users.
 
+Why is my drive not bootable?
+-----------------------------
+
+Etcher copies images to drives byte by byte, without doing any transformation
+to the final device, which means images that require special treatment to be
+made bootable, like Windows images, will not work out of the box. In these
+cases, the general advice is to use tooling specific to those kind of images,
```
lurch: Perhaps "software" would be better here than "tooling" ?

lurch:
Is it worth also explaining that Etcher doesn't set up "persistent" partitions like Unetbootin can do? (or should that come under a separate PR?)

docs/USER-DOCUMENTATION.md
jviotti: Fixed!

alexandrosm:
you may want to include specific problematic cases, as well as the specific
tools? if we start listing them here, we will have them ready when we want
a list of what to include in our warnings feature.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

jviotti:
@alexandrosm Done! 

@lurch Good idea! I'll do it in a separate PR.

lurch:
It also looks like "Windows 10 IoT Core Dashboard" is needed for writing Windows 10 IoT to a SD card
https://developer.microsoft.com/en-us/windows/iot/Docs/GetStarted/rpi3/sdcard/stable/GetStartedStep1.htm

PR #786, opened 2016-10-29
docs(GUI): generic appimage name in desktopintegration SKIP instructions

jviotti:
We're now including the version as part of the filename, which means
`Etcher-linux-x64` is no longer 100% accurate. As a solution, we include
a filename with "placeholders" for the version and architecture, to
avoid harcoding a specific version.

See: https://github.com/resin-io/etcher/commit/18e9d1eb11104daab92448f07a3cdfeaa4735c15#commitcomment-19619181
See: https://github.com/resin-io/etcher/pull/764

### DIFF
--- a/docs/USER-DOCUMENTATION.md
+++ b/docs/USER-DOCUMENTATION.md
@@ -20,7 +20,7 @@ Alternatively, set the `SKIP` environment variable before executing the
 AppImage:
 
 ```sh
-SKIP=1 ./Etcher-linux-x64.AppImage
+SKIP=1 ./Etcher-<version>-linux-<arch>.AppImage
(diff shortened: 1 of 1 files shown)


lurch:
Is this correct?? AFAICT #764 only modifies the name of the zip file, not the name of the AppImage itself?

dlech:
The current file name in the 1.0.0-beta.16 download is `Etcher-linux-x64.AppImage`, so `<arch>` is correct, but `<version>` is not.

jviotti:
Totally right! I'll fix in a second. I'm so happy to have various extra pair of eyes to catch this kind of stuff 🎉

PR #791, opened 2016-10-31
add travis triggers for website

craig-mulligan:
Trigger rebuilds of the website [`resin-io/etcher-homepage`](https://github.com/resin-io/etcher-homepage) via travis API. 

The build should only trigger **once** when commits are pushed to master. 

Depends on some work on [`resin-io/etcher-homepage`](https://github.com/resin-io/etcher-homepage) so we will need to wait for those PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: cpp
 sudo: false
-
 env:
+  matrix:
   - NODE_VERSION="6.1.0" RUBY_VERSION="2.3.1"
+  global:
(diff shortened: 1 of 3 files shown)


lurch:
I wonder if you might want to look into using something like https://github.com/gjtorikian/html-proofer , just like https://github.com/raspberrypi/documentation does?

craig-mulligan:
Looks very cool @lurch, we could test the markdown before triggering the travis build. I wish travis had a better system for dependent projects :/ though. 

Anyway, this would be good for the resin.io docs too @shaunmulligan should we add there and test it out?

WasabiFan:
I use html-proofer on one of the sites I maintain and although I do like it, the "verify links" functionality can be a bit temperamental. It often fails with "error code 0" for SSL sites on Travis whereas those same sites work fine locally. Our solution has been to just add all HTTPS sites to the ignore list for html-proofer. This may not be a problem for this site though, because there are so few external links anyway that you aren't likely to run into issues.

lurch:
> the "verify links" functionality can be a bit temperamental. It often fails with "error code 0" for SSL sites on Travis

Yeah, https://travis-ci.org/raspberrypi/documentation often sees the same issue :-/

jviotti:
@craig-mulligan I have the impression we should focus work on bringing the website to the main etcher repository rather than devising a build system to mitigate the fact they live in separate repositories. Maybe we should dive into it together next week?

As far as I understand, this PR will trigger a deployment to GitHub Pages every time we merge something to `master` here, but that doesn't mean a change in the website was necessarily made, right? Also, the etcher website currently deploys itself when we push to `gh-pages`, so why is this mechanism needed?

If we want to change the hardcoded URLs to the latest version, we'll have to send a commit to `etcher-homepage`, which will cause the website to be deployed as soon as that commit is merged, so the currently deploying script on this PR will always re-deploy an unmodified version, right?

craig-mulligan:
This is the setup (or the interim setup if we decide to merge the repos). 

Any commit on `resin-io/etcher-homepage` will trigger a build a Travis CI build that pulls latest downloads (and some data from this repo) and push the static files to it's gh-pages branch. 

Any commit on `resin-io/etcher` will trigger the same process as above. This way we can keep the dependent `resin-io/etcher-homepage` in sync with any changes here. 

If we merge the repo's and keep all source in master (rather than having a separate branch for the website) it will simplify the building (as we won't need this trigger script), so I'm up for that. 

Fine to close this for now and we can re-open when we chat about it next week.

craig-mulligan:
@jviotti here's [an example](https://github.com/craig-mulligan/etcher-website) built from modifications of the resinos site. I think it's a good interim step forward until we chat more about static sites next week with alex n co. 

You can see the src on `master` and the built static files on `gh-pages`
and the hosted files: https://craig-mulligan.github.io/etcher-website/

jviotti:
We've discussed with @craig-mulligan a way to move the website to the main Etcher repository, so I'll close this one for now!

Issue #801, opened 2016-10-31
Etcher CLI broken?


lurch:
I just tried following the `docs/CLI.md` instructions (on Ubuntu 14.04), and they didn't work. So then I tried again in a totally fresh directory, and got the same error:
```bash
$ cd /tmp
$ git clone https://github.com/resin-io/etcher
$ cd etcher
$ scripts/build/linux.sh install x64
# snip long output, but it all seemed to work fine
$ bin/etcher --help
module.js:597
  return process.dlopen(module, path._makeLong(filename));
                 ^

Error: Module version mismatch. Expected 48, got 50.
    at Error (native)
    at Object.Module._extensions..node (module.js:597:18)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at /tmp/etcher/node_modules/lzma-native/index.js:14:14
    at Object.<anonymous> (/tmp/etcher/node_modules/lzma-native/index.js:626:3)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/tmp/etcher/node_modules/etcher-image-stream/lib/handlers.js:22:36)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/tmp/etcher/node_modules/etcher-image-stream/lib/index.js:26:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/tmp/etcher/lib/cli/writer.js:20:21)
    at Module._compile (module.js:570:32)
```

jviotti:
[...]

Issue #355, opened 2016-04-22
Package and distribute the Etcher CLI


jviotti:

jviotti:
As @lurch suggested, the ideas mentioned in http://stackoverflow.com/a/14601786/1641422 might be worth trying.

jviotti:
@jhermsmeier Did an amazing work discovering some nodejs command line options to reduce the memory footprint in the CLI, which we should include by default: https://gist.github.com/jhermsmeier/c21270571f5ab2a0dd1262521fee647f

```
node --max-executable-size=8 --max-old-space-size=16 --max-semi-space-size=1
```

jhermsmeier:
These arguments would need to be passed in the `execArgv` option to [`child_process.fork()`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options) to have an effect:

```js
var child = childProcess.fork( './worker', argv, {
  execArgv: [
    // max size of the old space (in MB)
    '--max-old-space-size=16',
    // max size of a semi-space (in MB), the new space consists of two semi-spaces
    '--max-semi-space-size=1',
    // max size of executable memory (in MB)
    '--max-executable-size=8',
  ],
})
```

Note that forcing garbage collection (scavenge) with `--expose-gc` and calling `gc(true)` every n-th chunk during streaming to collect excess `Buffer`s appears to be more effective in keeping memory consumption down (see [block-write-stream/index.js#L29-L37](https://github.com/jhermsmeier/node-block-write-stream/blob/43517d20846110e8390471f5b4a904ed4d25d245/index.js#L29-L37) and [block-write-stream/index.js#L63-L72](https://github.com/jhermsmeier/node-block-write-stream/blob/43517d20846110e8390471f5b4a904ed4d25d245/index.js#L63-L72)) than setting these limits.

jviotti:
I can confirm this issue was sorted out completely by https://github.com/resin-io/etcher/pull/1409

PR #802, opened 2016-10-31
docs(CLI): describe proper way to install dependencies

jviotti:
The `docs/CLI.md` points to user to `docs/CONTRIBUTING.md` for
instructions in how to install dependencies. This document previously
included the explicit `npm` commands to run, but was then changed to
point to the platform scripts, which means that if the user follows the
current instructions but has a NodeJS version which is not ABI
compatible with the NodeJS version used by the currently used Electron
version, the user will get an incompatibility error.

Fixes: https://github.com/resin-io/etcher/issues/801

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -20,8 +20,11 @@ you can run it locally with the following steps:
 git clone https://github.com/resin-io/etcher
 ```
 
-- Install the dependencies, using the instructions from
-[`RUNNING-LOCALLY.md`][RUNNING-LOCALLY].
+- Install the dependencies by running:
+
+```sh
+npm install
+```
 
 - Run the Etcher CLI from `bin/etcher`.
 
@@ -71,5 +74,4 @@ The Etcher CLI uses certain exit codes to signal the result of the operation.
 These are documented in [`lib/src/exit-codes.js`][exit-codes] and are also
 printed on the Etcher CLI help page.
 
-[RUNNING-LOCALLY]: https://github.com/resin-io/etcher/blob/master/docs/RUNNING-LOCALLY.md
 [exit-codes]: https://github.com/resin-io/etcher/blob/master/lib/src/exit-codes.js


lurch:
I guess this PR probably ought to remove the `[RUNNING-LOCALLY]` section from the bottom of the file too.

lurch:
It's obvious I don't yet quite understand what's going on, but I ran `npm install` in the `/tmp/etcher` directory I was using in #801 , and it did some stuff, but then `bin/etcher --help` still gave the same `Error: Module version mismatch. Expected 48, got 50.`.

So then I deleted my `/tmp/etcher` and started afresh, following **only** the instructions in https://github.com/resin-io/etcher/blob/9f8a12bfa448ce91bc91adb49a09cdb0ee402192/docs/CLI.md , and this time it worked okay??
Does this mean that it's not possible to run both the Etcher GUI and the Etcher CLI from the same git-cloned directory? :confused:  Would it be possible for the scripts that install the GUI dependencies to also install the correct CLI dependencies?

jviotti:
> Does this mean that it's not possible to run both the Etcher GUI and the Etcher CLI from the same git-cloned directory? 😕 Would it be possible for the scripts that install the GUI dependencies to also install the correct CLI dependencies?

If the NodeJS version you have installed in your computer is not ABI compatible with the Electron version Etcher is currently using, then indeed you can't run it from the same source tree.

As mentioned in https://github.com/resin-io/etcher/issues/810#issuecomment-257424273, a nice trick is to call `electron` with `ELECTRON_RUN_AS_NODE=1`, so you can run the CLI without needing to clone in a separate place:

```sh
ELECTRON_RUN_AS_NODE=1 ./node_modules/.bin/electron bin/etcher
```

Alternatively, despite not very convenient in some cases, you can ensure the NodeJS version installed in your computer matches the one Electron bundles, which at the moment of this writing is v6.5.0.

jviotti:
> I guess this PR probably ought to remove the [RUNNING-LOCALLY] section from the bottom of the file too.

You're a human-linter! :D

lurch:
> As mentioned in #810 (comment), a nice trick is to call electron with ELECTRON_RUN_AS_NODE=1, so you can run the CLI without needing to clone in a separate place

So why doesn't this PR update `docs/CLI.md` to do that? (apologies if this is just me still misunderstanding things)

I guess (theoretically at least) the "dependency install scripts" (which I'm supposed to be refactoring, when I get back to that) could have separate targets for `install_cli_dependencies` and `install_gui_dependencies`, with the former being a subset of the latter?

jviotti:
> So why doesn't this PR update docs/CLI.md to do that? (apologies if this is just me still misunderstanding things)
> I guess (theoretically at least) the "dependency install scripts" (which I'm supposed to be refactoring, when I get back to that) could have separate targets for install_cli_dependencies and install_gui_dependencies, with the former being a subset of the latter?

The thing is that the CLI document, and the "CLI" target that will take place in the build scripts are supposed to allow you to run Etcher *without* Electron at all. While you can use the little trick I showed you during development, we'd like to publish the CLI as a tool being completely independent from Electron, since shipping Electron just for a CLI is a waste of size (given it ships Chromium and other stuff), and it doesn't run on older systems (like older OS X versions).

lurch:
Ahhh, okay. Thanks for the clarification - always useful to learn these things :-)
I think I'm starting to see how it fits together now...

Presumably another difference between the GUI and CLI versions, is that the GUI versions (e.g. as downloaded from https://etcher.io in an AppImage etc.) are entirely self-contained, whereas if you want to use the CLI version 'standalone' you _need_ to install NodeJS first?

Hmm, here's an off-the-wall idea: I wonder if `bin/etcher` (or some other wrapper script) could detect whether it's being run from within an Electron-enabled (or whatever the right terminology is) directory, and use that to automatically select whether to run (the equivalents of) `ELECTRON_RUN_AS_NODE=1 ./node_modules/.bin/electron bin/etcher` or `nodejs bin/etcher` ?

jviotti:
> Presumably another difference between the GUI and CLI versions, is that the GUI versions (e.g. as downloaded from https://etcher.io in an AppImage etc.) are entirely self-contained, whereas if you want to use the CLI version 'standalone' you need to install NodeJS first?

We have some thoughts about how to bundle NodeJS and the whole Etcher CLI code in a single executable, which we've solved partly, but still have to find a way to deal with native add-ons in this kind of scenarios.

> Hmm, here's an off-the-wall idea: I wonder if bin/etcher (or some other wrapper script) could detect whether it's being run from within an Electron-enabled (or whatever the right terminology is) directory, and use that to automatically select whether to run (the equivalents of) ELECTRON_RUN_AS_NODE=1 ./node_modules/.bin/electron bin/etcher or nodejs bin/etcher ?

The thing is if you package the application as Electron, it means that the code is inside an `asar` archive, which only Electron can open, so the check becomes irrelevant (it will always be Electron).

We don't want to hardcode calling "node" everywhere neither. We need more experimentation with our custom CLI packaging solution, but we'll get there.

lurch:
> We have some thoughts about how to bundle NodeJS and the whole Etcher CLI code in a single executable

Hah, I guess you effectively want "Electron without Chromium"? ;-)

> The thing is if you package the application as Electron, it means that the code is inside an asar archive, which only Electron can open, so the check becomes irrelevant (it will always be Electron).

I was thinking more of the 'incompatibilty' scenario I encountered earlier https://github.com/resin-io/etcher/pull/802#issuecomment-257415982 i.e. if I've previously run `scripts/build/linux.sh install x64` then have the etcher-cli-wrapper automatically call `ELECTRON_RUN_AS_NODE=1 ./node_modules/.bin/electron bin/etcher`, or if I've previously run `npm install` then have the etcher-cli-wrapper automatically call `nodejs bin/etcher`. But perhaps that's too much of a special-case scenario to come up with a general-purpose workaround for?

lurch:
> We need more experimentation with our custom CLI packaging solution, but we'll get there.

I guess you've probably already seen it, but it looks like http://stackoverflow.com/questions/14314038/how-do-i-deploy-node-js-applications-as-a-single-executable-file has some interesting links and suggestions (although I've obviously got no idea how well any of them work).

jviotti:
> I was thinking more of the 'incompatibilty' scenario I encountered earlier #802 (comment) i.e. if I've previously run scripts/build/linux.sh install x64 then have the etcher-cli-wrapper automatically call ELECTRON_RUN_AS_NODE=1 ./node_modules/.bin/electron bin/etcher, or if I've previously run npm install then have the etcher-cli-wrapper automatically call nodejs bin/etcher. But perhaps that's too much of a special-case scenario to come up with a general-purpose workaround for?

Hm, I see. I guess we can check if `./node_modules/.bin/electron` exists, and if so, use that.

> I guess you've probably already seen it, but it looks like http://stackoverflow.com/questions/14314038/how-do-i-deploy-node-js-applications-as-a-single-executable-file has some interesting links and suggestions (although I've obviously got no idea how well any of them work).

I've tried a couple of "pre-made" solutions (like nexe) but none of them handled native add-ons correctly. The 7zip and `makeself`/`unzipsfx` mentioned in http://stackoverflow.com/a/14601786/1641422 looks interesting though. I'll give it a try.

jviotti:
Lets continue discussion about this in https://github.com/resin-io/etcher/issues/355

Issue #413, opened 2016-05-16
Cannot chose writable space for Live Ubuntu USB


timtjtim:
I can't remember which app does it (perhaps `usb-creator-gtk`) but I remember being able to choose the free space for writing files when booting a Live USB such as Ubuntu.

Is this supported / could this be supported?

jviotti:
Hi @timtjtim ,

I don't think I understand the feature request very well. What do you mean by writable space? USB drives written by Etcher shouldn't be read-only, so you could mount it back and put any other files you want.

miniShrooms:
Hi @jviotti,

I believe @timtjtim is speaking about the option for persistent storage when creating a live USB for Linux OS's like Ubuntu, Mint, and more. What it does is allow the live USB to have files written and stored on it (like a normal OS) and actually allow for the files and changes to to be saved when closing, disconnecting, and rebooting back into the live USB. Additionally, **UNetbootin** is the application that has this feature; and honestly, I'm a big fan of it myself. I'd really like to see a feature like this implemented into Etcher if possible.

timtjtim:
@jviotti Yes, @miniShrooms is correct. That's the idea.

jviotti:
Hey there,

Thanks for the very detailed explanation. I'll play around with `UNetbootin` and try to research how this is implemented, and if it'd be reasonable to include it in Etcher.

m4gn3to:
I'm too interested in persistent storage, so I can install some packages in my usb only once, and also only have to disable the screensaver once 👍 
Thanks

Shou:
[...]

PR #805, opened 2016-10-31
docs: suggest UNetbootin if user requires setting persistence storage

jviotti:
See: https://github.com/resin-io/etcher/issues/413
See: https://github.com/resin-io/etcher/pull/783#issuecomment-256959050

### DIFF
--- a/docs/USER-DOCUMENTATION.md
+++ b/docs/USER-DOCUMENTATION.md
@@ -17,6 +17,14 @@ Images known to require special treatment:
 - Microsoft Windows (use [Windows USB/DVD Download Tool][windows-usb-tool], or
   [Rufus][rufus]).
 
+How can I configure persistent storage?
+---------------------------------------
+
+Some programs, usually oriented at making GNU/Linux live USB drives, include an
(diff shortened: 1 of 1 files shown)


lurch:
Oooh, that's an interesting side-observation I've just made: after reading https://unetbootin.github.io and https://sourceforge.net/p/unetbootin/wiki/howitworks/ it looks like UNetbootin doesn't actually write the ISO directly to the raw device (as etcher would), but instead extracts the individual files from the ISO, and writes them to an already-FAT32-formatted partition.

jviotti:
@lurch Yeah, they also install a boot loader on top, such as GRUB. We discussed implementing this "write" mechanism in Etcher, but we never allocated any serious work for it.

alexandrosm:
@lurch echoes of noobs?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lurch:
Sort of, except a) NOOBS _does_ partition and format the SD card itself (and so therefore has multiple partitions rather than just one) and b) NOOBS deals with tarballs rather than ISO files ;-)

Issue #787, opened 2016-10-29
Way of displaying supported formats


tooomm:
- **Etcher version:** 1.0.0-beta.16 portable
- **Operating system and architecture:** win 10 64bit

Why do you hide the most common image format `iso` behind "many more"?
Every bios or fw update, and all os images for linux and win I run into use it...
![iso_hide](https://cloud.githubusercontent.com/assets/9874850/19829947/2cbcfd04-9dee-11e6-919a-23e69928fa6e.png)
Shouldn't it be the other way around?
Show common ones that people know and use most likely, and append more specific and advanced ones?

Maybe there should even be some ordering. Show some important ones, others go to the tool tip in alphabetical order for example. Right now you get a bit lost in the tooltip. :)

jviotti:
@tooomm Thanks for the feedback. This is a great point. Currently we just take the first 3 formats defined by `etcher-image-stream` and put the rest in the tooltip.

Determining which ones should be presented outside the tooltip might be tricky, since for one side, we want highlight and make more obvious the more exotic formats we support, while still make clear that we support the "obvious" ones, like `.iso`.

I'd love to have input from more people on this one.

lurch:
What about "img, iso, zip, etch and many more" ?
`img` and `iso` are the image formats that people are most used to, `zip` shows that it also transparently supports compressed files, and `etch` shows that it also supports exotic formats ;-)

(AFAICT Etcher was originally developed for writing .img files to SD cards, and that's why IMHO `img` should be listed before `iso`)

Although that _might_ then make the text under the button wider than the button itself (?) and I dunno what @taahirisaacs thinks about that...

tooomm:
>Although that might then make the text under the button wider than the button itself (?) 

It could just say "and more" or "and others" instead. Maybe even just some "..." dots or the (?) help icon are enough.

alexandrosm:
I agree with Andrew's proposal, except that i don't see the reason to
promote .etch yet. For the moment it's a stub of a format we may make more
real in the future.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

jviotti:
Sounds good to me. What do you think about @tooomm 's suggestion about changing the text afterwards? I kinda like "and many more", since it sounds like a more "excited" way of promoting if, but maybe that's just me.

alexandrosm:
"and much more" sounds better in some unspecified way?

-- 

*Alexandros Marinos*

[...]

PR #806, opened 2016-10-31
style(GUI): improve ordering of supported extensions

jviotti:
The new ordering of the main supported extensions will be: `img`, `iso`,
`zip`.

We make use of `_.intersection` in order to ensure that the values we're
hardcoding indeed exist in the list of all supported format extensions,
otherwise they won't appear at all. This ensures we never show
extensions that are not supported.

Partly fixes: https://github.com/resin-io/etcher/issues/787

### DIFF
--- a/lib/gui/pages/main/controllers/image-selection.js
+++ b/lib/gui/pages/main/controllers/image-selection.js
@@ -26,7 +26,11 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
    * @type {String[]}
    * @public
    */
-  this.mainSupportedExtensions = _.slice(SupportedFormatsModel.getAllExtensions(), 0, 3);
+  this.mainSupportedExtensions = _.intersection([
+    'img',
+    'iso',
+    'zip'
+  ], SupportedFormatsModel.getAllExtensions());
 
   /**
    * @summary Extra supported extensions
@@ -37,7 +41,7 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
   this.extraSupportedExtensions = _.difference(
     SupportedFormatsModel.getAllExtensions(),
     this.mainSupportedExtensions
-  );
+  ).sort();
 
   /**
    * @summary Select image


jviotti:
Any suggestions regarding how to better impose the ordering while keeping it sane without relying on  positions would be appreciated.

lurch:
Relying on the hard-coded positions of items within arrays IMHO seems pretty insane ;-)
I can't imagine etcher ever *dropping* support for `img`, `iso` or `zip`, so it probably shouldn't be too bad to hardcode them? If you wanted to be bulletproof you could always have logic saying that if one of the hardcoded values **isn't** present in `SupportedFormatsModel`, then just replace it with values from `SupportedFormatsModel`?

jviotti:
> I can't imagine etcher ever dropping support for img, iso or zip, so it probably shouldn't be too bad to hardcode them?

Yeah, that's a fair assumption.

jviotti:
Fixed!

tooomm:
Will the "other" formats be ordered alphabetically?

lurch:
> Will the "other" formats be ordered alphabetically?

Nah, at the moment they'll still be displayed in the same order they are currently, i.e.
```
etch, gz, bz2, xz, dsk, hddimg, raw
```

TBH I'm not sure if alphabetical ordering really improves anything?
```
bz2, dsk, etch, gz, hddimg, raw, xz
```
(the current ordering groups the 'compressed' and 'raw' formats together)

tooomm:
I guess an ordering by starting letters (=alphabetical) is easier to understand for the user than to have technical knowledge and get the idea of an grouping by format type.

Also, if you're looking for a special file format, it's way easier to find if they are all named alphabetically.

jviotti:
Done! I'm now sorting the extensions.

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -26,18 +26,26 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
...
    */
-  this.extraSupportedExtensions = _.difference(
+  this.extraSupportedExtensions = _.sortBy(_.difference(
     SupportedFormatsModel.getAllExtensions(),
     this.mainSupportedExtensions
-  );
+
+  // Sort alphabetically
+  ), (extension) => {
+    return extension.charCodeAt();
```
lurch: Huh, why are you only sorting on the first character in the string? It means `abb` might sort before `aaa`.
Why not just use 'normal' sort? http://www.w3schools.com/jsref/jsref_sort.asp

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -26,18 +26,26 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
    * @type {String[]}
    * @public
    */
-  this.mainSupportedExtensions = _.slice(SupportedFormatsModel.getAllExtensions(), 0, 3);
+  this.mainSupportedExtensions = _.intersection([
```
WasabiFan: I'm a bit confused by what's going on here. This combines this hard-coded array with the list of _all_ supported extensions, but all of these types are -- as far as I know -- already in that list. If you were going to display them in this order, that would make sense, but right now it sorts them a few lines down, so this doesn't actually do anything special as far as I can tell.

lurch: Re-read the full commit message... ;)

WasabiFan: Oh, never mind; I read this as "union" instead of "intersection". That makes more sense. Although, to be honest, if this code is ever necessary I think Etcher has a serious problem :laughing:

lib/gui/pages/main/controllers/image-selection.js
jviotti: Oh, I completely ignored the fact `.sort()` worked out of the box for strings. I'll fix!

PR #807, opened 2016-10-31
style(GUI): improve wording of supported extensions label

jviotti:
- Change "and many more" to "and much more"

See: https://github.com/resin-io/etcher/issues/787

### DIFF
--- a/lib/gui/pages/main/templates/main.tpl.html
+++ b/lib/gui/pages/main/templates/main.tpl.html
@@ -12,7 +12,7 @@
           <p class="step-footer">
             {{ ::image.mainSupportedExtensions.join(', ') }}, and
             <span class="step-footer-underline"
-              uib-tooltip="{{ image.extraSupportedExtensions.join(', ') }}">many more</span>
+              uib-tooltip="{{ image.extraSupportedExtensions.join(', ') }}">others</span>
           </p>
         </div>
         <div ng-if="main.selection.hasImage()">


lurch:
I dunno - "and many more" sounds like it's adding extra items to the end of a list, which is what it's doing currently. Whereas "and much more" makes it sounds like you're describing a list of features, which isn't really the case here.  @alexandrosm 

&lt;pedant&gt; It's a bit like the difference between 'less' and 'fewer' ;-) &lt;/pedant&gt;

jviotti:
Yeah, as a non-native English speaker, I guess I'm not the best to make a sensible decision on this, so happy to go with anything you choose!

WasabiFan:
I'd tend to agree with @lurch here; "and many more" seems like the better of the two. However, I honestly prefer @tooomm's proposal of "and others" because it is short and simple, while still communicating the idea. It does have a connotation associated with it that the hidden ones are lesser though.

Side-note: even though you prepended "partly" to `fixes: #787`, GitHub will still detect it and close that other issue.

lurch:
I also suspect that the "working" in the title of this PR was meant to be "wording"? ;)

jviotti:
> Side-note: even though you prepended "partly" to fixes: #787, GitHub will still detect it and close that other issue.

Oh, are you sure? I think I've used this in the past (or some other variation), which led the same results. Lets give it a go to double-check. I'll re-open the issue manually if this is indeed the case.

> I also suspect that the "working" in the title of this PR was meant to be "wording"? ;)

Haha, good catch!

WasabiFan:
> Oh, are you sure? I think I've used this in the past (or some other variation), which led the same results. Lets give it a go to double-check. I'll re-open the issue manually if this is indeed the case.

You can look at the issue to see if it says this one will close it:

image

You currently have two PRs which will do so. GitHub's logic just looks for something approximating `(fixes|fix|closes|close|...) #\d+`; otherwise you wouldn't be able to say something like `This fixes #123.` Keep in mind that, because it is you that asked for the closure (even if unintentionally), it won't show up in your notifications.

jviotti:
Oh, I see, thanks a lot for the tip. I changed `Partly fixes:` to `See:`.

jviotti:
Lets go for "and others" in the name of simplicity.

tooomm:
My proposals are popular recently :)

PR #809, opened 2016-10-31
chore: persist a chocolatey cache Appveyor CI builds

jviotti:
Setting a Chocolatey cache and Making sure it gets persisted means that
future builds don't have to download every Chocolatey package from the
internet every time.


lurch:
:chocolate_bar:  :yum:

PR #817, opened 2016-11-01
upgrade: `umount` to v1.1.4

jviotti:
- This new version passes `force` to `diskutil unmountDisk`

Change-Type: patch
Changelog-Entry: Fix `at least one volume could not be unmounted` error in OS X.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5132,9 +5132,9 @@
       "resolved": "https://registry.npmjs.org/umd/-/umd-3.0.1.tgz"
     },
     "umount": {
-      "version": "1.1.3",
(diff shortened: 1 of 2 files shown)


lurch:
From a quick google I found http://apple.stackexchange.com/questions/235161/difference-between-diskutil-unmount-and-diskutil-unmountdisk/235162#235162

I wonder if it'd be better to try `unmountDisk` first, and only use `force` if the previous attempt didn't work?

jviotti:
@lurch `diskutil unmountDisk force` should be exactly the same as `diskutil unmountDisk` if there were no processes making the drive busy, so there should be no harm in passing `force` in the first try.

lurch:
Fair enough :-)

Issue #729, opened 2016-09-29
Change the "unsafe mode" cancel button colours


lurch:
- **Etcher version: 1.0.0-beta.15**
- **Operating system and architecture: Ubuntu 14.04 x64**
- **Do you see any meaningful error information on DevTools? No**

When you click on the "Unsafe mode" in the settings screen, the following dialog pops up:
![screenshot from 2016-09-29 14 05 12](https://cloud.githubusercontent.com/assets/476186/18955077/c5c0c3fe-864d-11e6-86dd-e2665fc5329a.png)

IMHO the way the "CANCEL" button is greyed-out, it looks like it's disabled and clicking on "YES, CONTINUE" is the _only_ option available.

alexandrosm:
great point. And in fact, given we're trying to dissuade the user from
proceeding (to an extent) the cancel button should be

1. named differently ("keep safe mode on"?)
2. be the first option
3. be much brighter

@taahirisaacs, care to offer any design drafts?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

taahirisaacs:
I've tried out 3 options.

I like @alexandrosm idea of having the button named differently.

- option 1: Darker grey button
- option 2: Change of text + dark grey button
- option 3: Change of text + brighter colored button

![option1-warning-001](https://cloud.githubusercontent.com/assets/15251376/19689177/c65692c4-9acc-11e6-9e0e-76648325cb14.png)
![option2-warning-001](https://cloud.githubusercontent.com/assets/15251376/19689179/c6685eb4-9acc-11e6-8044-9d8f81b96a17.png)
![option3-warning-001](https://cloud.githubusercontent.com/assets/15251376/19689178/c660dce8-9acc-11e6-8e67-dda86855a59f.png)

lurch:
Yeah, number 3 looks nice :)

WasabiFan:
I think the wording "Stay in safe mode" or even "Back to safety" sounds a bit better, but it doesn't really make a difference. However, I think the word "no" should be included in the second button. The blue is a bit awkward, but I think it communicates the point the best.

lurch:
Hmmm, and perhaps "Are you sure you want to turn this on?" could be changed to "Are you sure you want to enable unsafe mode?".

Could the "Unsafe mode" tickbox in the parent dialog remain "unticked" until the user explicitly clicks on the "Yes, continue" button?

[...]

PR #822, opened 2016-11-02
style(GUI): improve unsafe mode dialog

jviotti:
- Change dialog message to "Are you sure you want to turn this on? You
  will be able to overwrite your system drives if you're not careful."

- Change dialog button labels from "YES, CONTINUE", and "CANCEL", to
  "ENABLE UNSAFE MODE", and "STAY IN SAFE MODE", respectively.

In order to adapt to changes described above:

- `WarningModalService.display()` now takes an object including
  `message`, `confirmButtonMessage`, and `rejectButtonMessage`
  properties.

screenshot 2016-11-03 09 59 00

Fixes: https://github.com/resin-io/etcher/issues/729

### DIFF
--- a/lib/gui/components/warning-modal/services/warning-modal.js
+++ b/lib/gui/components/warning-modal/services/warning-modal.js
@@ -25,20 +25,31 @@ module.exports = function(ModalService) {
    * @function
    * @public
    *
-   * @param {String} message - danger message
+   * @param {Object} options - options
+   * @param {String} options.message - danger message
+   * @param {String} [options.confirmButtonMessage='Yes, continue'] - confirm button message
+   * @param {String} [options.rejectButtonMessage='Cancel'] - reject button message
    * @fulfil {Boolean} - whether the user accepted or rejected the warning
    * @returns {Promise}
    *
    * @example
-   * WarningModalService.display('Don\'t do this!');
+   * WarningModalService.display({
+   *   message: 'Don\'t do this!'
+   * });
    */
-  this.display = (message) => {
+  this.display = (options) => {
+
+    _.defaults(options, {
+      confirmButtonMessage: 'Yes, continue',
+      rejectButtonMessage: 'Cancel'
+    });
+
     return ModalService.open({
       template: './components/warning-modal/templates/warning-modal.tpl.html',
       controller: 'WarningModalController as modal',
-      size: 'settings-dangerous-modal',
+      size: 'warning-modal',
       resolve: {
-        message: _.constant(message)
+        options: _.constant(options)
       }
     }).result;
   };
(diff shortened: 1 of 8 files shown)


WasabiFan:
I think "stay in safe mode" still looks too much like it is disabled. There needs to be more contrast between background and text.

lurch:
Exactly, the original request in #729 was to change the colours, not necessarily change the button labels ;-)

lib/gui/components/warning-modal/services/warning-modal.js
```diff
@@ -25,20 +25,31 @@ module.exports = function(ModalService) {
    * @function
    * @public
    *
-   * @param {String} message - danger message
+   * @param {Object} options - options
+   * @param {String} options.message - danger message
+   * @param {String} [options.confirmButtonMessage] - confirm button message
+   * @param {String} [options.rejectButtonMessage] - reject button message
```
lurch: Should this also document the default values?

lib/gui/pages/settings/templates/settings.tpl.html
```diff
@@ -46,7 +46,11 @@ <h3 class="subtitle">Advanced</h3>
     <label>
       <input type="checkbox"
         ng-model="settings.currentData.unsafeMode"
-        ng-change="settings.enableDangerousSetting('unsafeMode', 'Are you sure you want to turn this on? You will be able to burn to your system drives.')">
+        ng-change="settings.enableDangerousSetting('unsafeMode', {
+          message: 'Are you sure you want to turn this on? You will be able to burn to your system drives.',
```
lurch: IMHO "burn to" sounds kinda strange, and "write to" might be better?

lurch:
P.S. Now that #821 has been merged, it's actually impossible to exactly replicate the screenshot from the first comment in this thread ;-)

lib/gui/pages/settings/templates/settings.tpl.html
WasabiFan: "overwrite" or "destroy [data on]" may be clearer, Also the phrase "if you are not careful" at the end might make it clearer.

lurch:
https://github.com/resin-io/etcher/issues/729#issuecomment-258049364

lib/gui/pages/settings/templates/settings.tpl.html
lurch: Yeah, "overwrite" might sound better; as both "write to" and "destroy data on" make it sound like it'll possibly only write to **part** of the disk, whereas obviously as soon as it's written over the MBR the whole disk is lost (as far as normal users are concerned).
And :+1: to the " if you are not careful" on the end.

lib/gui/components/warning-modal/services/warning-modal.js
jviotti: Oops! Fixed!

jviotti:
The screenshot has been updated!

PR #840, opened 2016-11-07
minifix(GUI): improve unsafe mode setting dialog

jviotti:
- Change dialog message to "Are you sure you want to turn this on? You
  will be able to overwrite your system drives if you're not careful."

- Remove "CANCEL" button and put a cross at the top right corner.

- Change dialog button label from "YES, CONTINUE" to "ENABLE UNSAFE
  MODE"

In order to adapt to changes described above:

- `WarningModalService.display()` now takes an object including
  `description`, and `confirmationLabel`

screenshot 2016-11-07 16 57 07

Fixes: https://github.com/resin-io/etcher/issues/729

### DIFF
--- a/build/css/main.css
+++ b/build/css/main.css
@@ -6465,8 +6465,14 @@ body {
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+.modal-warning-modal .modal-content {
+  width: 350px; }
+
 .modal-warning-modal .modal-title .glyphicon, .modal-warning-modal .modal-title .tick {
   color: #d9534f; }
+
+.modal-warning-modal .button, .modal-warning-modal .progress-button {
+  text-transform: uppercase; }
 
 /*
  * Copyright 2016 resin.io
--- a/lib/gui/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/components/drive-selector/controllers/drive-selector.js
@@ -67,11 +67,14 @@ module.exports = function($uibModalInstance, DrivesModel, SelectionStateModel, W
       return;
     }
 
-    WarningModalService.display([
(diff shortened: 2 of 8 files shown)


jviotti:
As advised as the correct approach, UX wise, by @konmouz

lurch:
I wonder if the small X in the top-right corner is an obvious-enough cancel method for new and/or nervous users? \*shrug\*

WasabiFan:
Yeah, I think this might be a regression over what was there previously. If you are presenting a dialog telling the user they are about to destroy their data, having them take the "passive" action of _closing_ the dialog doesn't feel right. Does closing the dialog actually preserve my data? Or do I have to take a separate action to reset it?

In the end, so much time has been spent debating this that it really isn't worth continuing, but it's important to make sure that whatever is produced is usable.

dlech:
I also disagree that this is good UX. You are asking "are you sure?" but only give one option. It seems rather strange and uncomfortable to me as a user.

konmouz:
A few things: 
- We chose the 'X' button for consistency purposes. It has already been used in other modals, so makes sense to have it here as well. Anyway, this is a standardised CTA these days.
- @dlech Reforming the question or just informing the user about the current actions might help (e.g. By enabling Unsafe Mode, you will be able to overwrite your system drives if you are not careful.)
- I checked the comments history and indeed this topic is over discussed. For instance 'cancel/dismiss' button works well as well, as long as it is consistent with the rest of the UI.
- UX-wise both solutions work fine. Just thought current one looks a bit cleaner, while with the dual buttons there is a lot of text squeezed in a small space. This can be solved by replacing 'ENABLE UNSAFE MODE' with 'GO AHEAD', 'ENABLE', etc., however it might not be as clear as current.

PR #842, opened 2016-11-07
chore: add `productDescription` to `package.json`

jviotti:
This key is added as the long description of the Debian package.
Omitting this option means that `description` is duplicated two times,
which is of course not the desired outcome.

See: https://github.com/resin-io/etcher/issues/632

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "1.0.0-beta.16",
   "main": "lib/start.js",
   "description": "Flash OS images to SD cards & USB drives, safely and easily.",
+  "productDescription": "Etcher is a powerful OS image flasher built with web technologies to ensure flashing an SDCard or USB drive is a pleasant and safe experience. It protects you from accidentally writing to your hard-drives, ensures every byte of data was written correctly and much more.",
   "homepage": "https://github.com/resin-io/etcher",
   "repository": {
     "type": "git",


WasabiFan:
The description added here doesn't really tell you what Etcher is... It just tells you how it's different from other similar apps.

lurch:
...and the fact that it's cross-platform is probably not relevant to the Debian package?

dlech:
How about using the blurb from the README:

> Etcher is a powerful OS image flasher built with web technologies to ensure flashing an SDCard or USB drive is a pleasant and safe experience. It protects you from accidentally writing to your hard-drives, ensures every byte of data was written correctly and much more.

WasabiFan:
Honestly, I think the explanation from the README is too verbose. It's also unclear what "OS" is.

dlech:
This is the "long" description. :wink:

And not many people will probably be reading this, so we don't need to spend too much time thinking about it.

lurch:
I guess "OS image flasher" could be changed to "disk image flasher"? (which is technically more correct :wink: )

jviotti:
Yeah, I've added the README blurb. That should be more than enough now, and we can revisit later.

lurch:
Could / should this necessitate a version-bump of the .deb package to `1.0.0~beta.16-2` ? Or does bintray handle that automatically?

dlech:
I don't think it is worth doing an extra release for this change.

Issue #850, opened 2016-11-08
Writing continues after GUI is closed


lurch:
- **Etcher version:**  1.0.0-beta16 (built from latest GH source on 2016-11-08)
- **Operating system and architecture:** Ubuntu Linux 14.04 x64
- **Do you see any meaningful error information on DevTools?** n/a

I've got an older USB drive which is quite slow (and which blinks an LED when it's being accessed - which is how I noticed this bug!).
I tried writing a big image to it, which started off okay, and then I closed the Etcher GUI because I couldn't be bothered to wait for it to finish. The GUI all closed down okay with no errors, but to my surprise the LED on the USB drive continued to blink! I checked with `top` and the `electron` process was still running in the background, continuing to write to the drive.

I launched Etcher with `npm start` and nothing got printed to the terminal apart from:
```

> etcher@1.0.0-beta.16 start /home/andrew/github/lurch/etcher
> electron lib/start.js

```

lurch:
Just tested the same behaviour on Windows10, and when closing the application window _all_ the Etcher processes die off, and the LED on the slow USB drive stops flashing (which is the expected behaviour).

jviotti:
So this seems to be an issue in https://github.com/jorangreef/sudo-prompt. I'll investigate a bit and report back my results soon.

lurch:
Were you able to replicate this on Mac?

jviotti:
Turns out yes, but its not very trivial. The process remains alive if we do SIGKILL, but not if we do SIGINT.

PR #855, opened 2016-11-09
fix(GUI): kill CLI processes when child writer dies

jviotti:
Turns out that when the parent process gets SIGKILL, the children
process would still remain alive.

In order to mitigate this, we force the IPC client opened by the child
writer to not attempt to reconnect to the IPC server if this one dies,
so we can catch the "disconnect" event and kill the children as a
result.

Fixes: https://github.com/resin-io/etcher/issues/850
Change-Type: patch
Changelog-Entry: Fix writing process remaining alive after the GUI is closed.

### DIFF
--- a/lib/src/child-writer/writer-proxy.js
+++ b/lib/src/child-writer/writer-proxy.js
@@ -139,10 +139,20 @@ return isElevated().then((elevated) => {
   return new Bluebird((resolve, reject) => {
     ipc.config.id = process.env.IPC_CLIENT_ID;
     ipc.config.silent = true;
+
+    // > If set to 0, the client will NOT try to reconnect.
+    // See https://github.com/RIAEvangelist/node-ipc/
+    //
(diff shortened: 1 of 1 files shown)


lurch:
Yup, this fixes #850 for me.

jviotti:
@lurch I've updated this PR to use the IPC `stopRetrying` approach we discussed before. Do you mind checking if it still works for you too?

lurch:
Nope, the new version of this PR doesn't work, `electron` remains running in the background (and still writing to disk) after closing the Etcher window.

jviotti:
@lurch We've been debugging this with @Shou, and discovered that we we're hitting a context issue in `child.kill()`. We confirmed it works now! Let us know!

lurch:
Yup, updated PR is working again :-)  (writing to disk  stops when I close the Etcher GUI mid-flash)

Issue #859, opened 2016-11-10
Error on apt-key


Cascador:
Hello,

sudo apt-key adv --keyserver pgp.mit.edu --recv-keys 379CE192D401AB61 doesn't work.

The key is missing on pgp.mit.edu ?

Thanks.

lurch:
Works fine for me? What error message do you get?

Cascador:
Problem solved with : sudo apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 379CE192D401AB61

Probably a firewall or proxy problem, see here : http://unix.stackexchange.com/questions/75892/keyserver-timed-out-when-trying-to-add-a-gpg-public-key

Thanks.

lurch:
Thanks for letting us know @Cascador .

@jviotti @dlech should this info be added to https://github.com/resin-io/etcher#debian-and-ubuntu-based-package-repository-gnulinux-x86x64 ?

Cascador:
sudo apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 379CE192D401AB61 works on more cases (especially at work). 

FYI in Docker sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 (https://docs.docker.com/engine/installation/linux/ubuntulinux/)

PR #861, opened 2016-11-10
docs: more explicit key server

dlech:
Use full URL with port for key server. Also use port 80 to work around firewalls and proxies.

Fixes: #859


jviotti:
Oh, I didn't see this and sent a PR for it already: https://github.com/resin-io/etcher/pull/863. I'll close mine so we merge this one instead!

jviotti:
The Appveyor build is timing out. I'll restart to make sure it passes.

Issue #726, opened 2016-09-28
Linux: command line version does not elevate properly


imrehg:
- **Etcher version:** 1.0.0-beta15
- **Operating system and architecture:** ArchLinux x64

Using the command line etcher, does not seem to elevate properly after asking and getting sudo. See the log below. Explanation:
- first run, asks for sudo password, given, complains about privileges, exits
- second run, does not ask for sudo password which means the password was right the first time, should have access, complains about privileges, exits
- run with explicit sudo, does not ask for password which means indeed the password was right the first time, but now flashes card.

```
greg@nostromo:bin 18:04:22 (0) <123>
$./etcher ~/Downloads/resin-watchdog2-1.8.0-1.13.0-7ad55d1c946b.img 
? Select drive /dev/sdf (8.0 GB) - USB3.0 CRW-SD/MS
? This will erase the selected drive. Are you sure? Yes
[sudo] password for greg: 
You don't have enough privileges to run this operation.
Try running this command again prefixing it with `sudo`.

If this is not the case, and you're trying to burn an SDCard, check that the write lock is not set.
greg@nostromo:bin 18:05:15 (0) <124>
$./etcher ~/Downloads/resin-watchdog2-1.8.0-1.13.0-7ad55d1c946b.img 
? Select drive /dev/sdf (8.0 GB) - USB3.0 CRW-SD/MS
? This will erase the selected drive. Are you sure? Yes
You don't have enough privileges to run this operation.
Try running this command again prefixing it with `sudo`.

If this is not the case, and you're trying to burn an SDCard, check that the write lock is not set.
greg@nostromo:bin 18:08:01 (0) <126>
$sudo ./etcher ~/Downloads/resin-watchdog2-1.8.0-1.13.0-7ad55d1c946b.img 
? Select drive /dev/sdf (8.0 GB) - USB3.0 CRW-SD/MS
? This will erase the selected drive. Are you sure? Yes
Flashing [========================] 100% eta 0s  
Validating [========================] 100% eta 0s  
Your flash is complete!
Checksum: 88ab5b93
```

jviotti:
Its strange the CLI is asking you for the `root` password. The CLI has not elevation functionality built-in. I'll investigate. There might be a child process implicitly causing this.

lurch:
[...]

PR #864, opened 2016-11-11
fix(CLI): check CLI elevation early on

jviotti:
The CLI requires elevation permissions to actually be able to write to
drives. For this reason, it makes sense that we introduce such a check
early on.

See: https://github.com/resin-io/etcher/issues/726
Change-Type: patch
Changelog-Entry: Check available permissions in the CLI early on.

### DIFF
--- a/lib/cli/etcher.js
+++ b/lib/cli/etcher.js
@@ -18,6 +18,7 @@
 
 const _ = require('lodash');
 const Bluebird = require('bluebird');
+const isElevated = Bluebird.promisify(require('is-elevated'));
 const visuals = require('resin-cli-visuals');
 const form = require('resin-cli-form');
 const drivelist = Bluebird.promisifyAll(require('drivelist'));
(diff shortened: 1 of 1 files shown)


lib/cli/etcher.js
```diff
@@ -26,28 +27,34 @@ const utils = require('./utils');
...
-    name: 'drive'
-  },
-  {
-    message: 'This will erase the selected drive. Are you sure?',
-    type: 'confirm',
-    name: 'yes',
-    default: false
+isElevated().then((elevated) => {
+  if (!elevated) {
+    throw new Error('This should should be ran as root/administrator permissions');
```
lurch: 'run as' should be 'run with' ?

jviotti: Fixed!

lib/cli/etcher.js
```diff
@@ -26,28 +27,34 @@ const utils = require('./utils');
...
-    name: 'drive'
-  },
-  {
-    message: 'This will erase the selected drive. Are you sure?',
-    type: 'confirm',
-    name: 'yes',
-    default: false
+isElevated().then((elevated) => {
+  if (!elevated) {
+    throw new Error('This should should be ran with root/administrator permissions');
```
lurch: Still need to change `ran` to `run` ;)

Issue #744, opened 2016-10-02
[feature] Confirm exiting application while flashing image


waspinator:
- v1.0.0-beta.15 - 2016-09-26
- macOS x86

While an image is being flashed, it's too easy to close Etcher unintentionally. I believe it would be a helpful feature to show a `confirm()` dialog before quitting while an image is being flashed.

jviotti:
Great idea! I'm sure Electron provides a way to intercept a close event. The `FlashStateModel` component also provides a nice `isFlashing()` method that can be used to either allow the user to close the app directly, or show the confirmation modal.

taahirisaacs:
Created a simple warning modal to alert the user if they close the application while an image is being flashed.

![exit](https://cloud.githubusercontent.com/assets/15251376/20099201/ff230436-a5bf-11e6-817a-9d334245e3b4.png)

lurch:
Looks good. But I think "...sure you want to..." is more idiomatic than "...sure you would like to..."

WasabiFan:
It's worth noting that the implications of this action aren't immediately clear to the user. I think that adding a sentence explaining what will happen would be nice. Also, currently "what will happen" is that Etcher will continue flashing headless :laughing:

lurch:
Hmm, given that the dialog in #840 says "You will be able to overwrite your system drives if you're not careful.", maybe this dialog could also say something like "This may leave your drive in an unusable state." ?

> Also, currently "what will happen" is that Etcher will continue flashing headless :laughing:

Yeah, that's a bug #850

taahirisaacs:
Updated copy on design.

![exit confirmation](https://cloud.githubusercontent.com/assets/15251376/20103212/baec2984-a5d1-11e6-9db0-3227730db291.png)

jviotti:
@taahirisaacs Looks good! Following @konmouz line of thoughts regarding these kinds of modals, I think we can remove the "Cancel" button and just present "Yes, continue". The user can dismiss the modal by clicking somewhere else or in the top right corner cross icon.

WasabiFan:
> I think we can remove the "Cancel" button and just present "Yes, continue".

[...]

PR #868, opened 2016-11-12
feat(GUI): confirm before quitting etcher while writing

Shou:
When Etcher is writing to a drive and the user closes the window,
currently there is no confirmation. This commit changes that and
the user has to confirm the exit.

Other changes include the warning modal taking a `rejectionLabel` option
that shows a reject button when provided.

Closes: #744
Change-Type: minor
Changelog-Entry: Confirm before user quits while writing.

### DIFF
--- a/build/css/main.css
+++ b/build/css/main.css
@@ -6267,6 +6267,8 @@ body {
 
 .modal-text {
   padding: 20px; }
+  .modal-text > p {
+    white-space: pre-line; }
   .modal-text > p:last-child {
     margin-bottom: 0; }
 
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -23,6 +23,8 @@
 /* eslint-disable no-var */
 
 var angular = require('angular');
+const electron = require('electron');
+const EXIT_CODES = require('../src/exit-codes');
 
 /* eslint-enable no-var */
 
@@ -48,6 +50,7 @@ const app = angular.module('Etcher', [
   require('./components/svg-icon/svg-icon'),
   require('./components/update-notifier/update-notifier'),
   require('./components/drive-selector/drive-selector'),
+  require('./components/warning-modal/warning-modal'),
 
   // Pages
   require('./pages/main/main'),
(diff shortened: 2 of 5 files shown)


lib/gui/pages/settings/templates/settings.tpl.html
```diff
@@ -48,7 +48,7 @@ <h3 class="subtitle">Advanced</h3>
         ng-model="settings.currentData.unsafeMode"
         ng-change="settings.enableDangerousSetting('unsafeMode', {
           description: 'Are you sure you want to turn this on? You will be able to overwrite your system drives if you\'re not careful.',
-          confirmationLabel: 'Enable unsafe mode'
+          confirmationLabel: 'Enable unsafe mode', rejectionLabel: 'Cancel'
```
lurch: Was this change (adding the Cancel button back to the "enable unsafe" popup) **supposed** to be included in this PR too? :-/

build/css/main.css
```diff
@@ -6267,6 +6267,8 @@ body {
 
 .modal-text {
   padding: 20px; }
```
WasabiFan: 😖 _the ugly_

lib/gui/pages/settings/templates/settings.tpl.html
jviotti: Yeah, this should be removed from here.

jviotti:
I'd say this is more a feature than a fix.

jviotti:
What happens if I have another modal opened, and then attempt to quit? For example, I could go to settings and attempt to unable "unsafe mode" while a flashing is in process. Will this modal appear on top of the other? Will it close the other one?

Shou:
@jviotti I've changed the title from fix→feat, and to answer the multiple modals question, it overlays the other modals. The way the Bootstrap modals work is, the newer modal is simply inserted before the other modal neighbours so it has precedence.

jviotti:
Sounds good. Lets address the remaining comments and merge.

Shou:
@jviotti I'm not sure what @WasabiFan is addressing but I don't believe it's an addition in this PR as it's identical to the previous commit. I also removed the cancel button from the unsafe mode modal as suggested by lurch—it's an optional argument and will be removed by the `ng-if` if not provided. So we should be good to merge.

WasabiFan:
Nope, my comment isn't an issue; I was just expressing the profound disgust experienced upon discovery of that..._inventive_ CSS style :laughing:

jviotti:
@Shou Damn, I just realized that GitHub is not posting all the comments I made until I "confirm review comments", which is really weird. They should be displaying now :/

lurch:
> I just realized that GitHub is not posting all the comments I made until I "confirm review comments", which is really weird.

Depends if you do "review" or "add single comment". (fairly recent GitHub change I believe)

WasabiFan:
> I just realized that GitHub is not posting all the comments I made until I "confirm review comments", which is really weird.

This was a feature added fairly recently (a few months ago) as @lurch said. It's called "reviews." The idea is that you can group comments into a "review" and then make the whole review public at once with a universal note and status (approve/request changes/comment). https://help.github.com/articles/about-pull-request-reviews/

jviotti:
Yeah, a super-confusing feature UX-wise IMHO :P 

On Tue, Nov 29, 2016 at 05:53:11PM -0800, Wasabi Fan wrote:
> > I just realized that GitHub is not posting all the comments I made until I "confirm review comments", which is really weird.
> 
> This was a feature added fairly recently (a few months ago) as @lurch said. It's called "reviews." The idea is that you can group comments into a "review" and then make the whole review public at once with a universal note and status (approve/request changes/comment). https://help.github.com/articles/about-pull-request-reviews/
> 
> -- 
> https://github.com/resin-io/etcher/pull/868#issuecomment-263759279

-- 
Juan Cruz Viotti
Software Engineer

Shou:
@jviotti I followed the feedback here now, and also I noticed a bug. The update dialog would be affected by the `.modal-footer` `display: flex` I added for `lib/gui/components/modal/styles/_modal.scss`, but it's an unnecessary change if we wrap the warning modal's buttons within a `div.modal-menu` in a similar fashion to the update dialog, which already has a `display: flex` applied. This should be more consistent.

PR #872, opened 2016-11-16
fix(GUI): terminate IPC server on any errors

jviotti:
If the child process (e.g: writer proxy, or CLI) emits either an error,
or an invalid output, the error will be emitted to the GUI, which will
report it to the state machine, however the IPC server will keep
receiving further messages, which will be passed to the GUI, which will
keep trying to update the state, in which case Redux will rightfully
complain that we can't update the flashing state if there is no flash in
process.

As a solution, we make sure the IPC server is terminated on every
possible error code flow, and to do this effectively, we needed to
manually destroy every connected socket before actually closing the
server.

Change-Type: patch
Changelog-Entry: Fix "Can't set the flashing state when not flashing" error.

### DIFF
--- a/lib/src/child-writer/index.js
+++ b/lib/src/child-writer/index.js
@@ -17,6 +17,7 @@
 'use strict';
 
 const EventEmitter = require('events').EventEmitter;
+const _ = require('lodash');
 const childProcess = require('child_process');
 const ipc = require('node-ipc');
 const rendererUtils = require('./renderer-utils');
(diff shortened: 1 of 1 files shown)


lib/src/child-writer/index.js
```diff
@@ -138,7 +155,7 @@ exports.write = (image, drive, options) => {
       }
 
       if (code !== EXIT_CODES.SUCCESS && code !== EXIT_CODES.VALIDATION_ERROR) {
-        return emitter.emit('error', new Error(`Child process exitted with error code: ${code}`));
+        return emitError(new Error(`Child process exitted with error code: ${code}`));
```
lurch: Should only be one 't' in "exited"

PR #882, opened 2016-11-18
fix(GUI): wrap long words in drivelist

Shou:
Drives with long names or descriptions expand outside the drivelist,
this commit fixes that using CSS `word-break: break-all`.

Change-Type: patch
Changelog-Entry: Wrap drive names and descriptions in drivelist.


lurch:
Could you attach a couple of screenshots, showing the before and after difference? :-)

Shou:
@lurch sure!

Before, no word breaking.


After, word breaking applied.

Issue #865, opened 2016-11-11
'disabled' drives have wrong cursor


lurch:
- **Etcher version:** 1.0.0-beta16
- **Operating system and architecture:** Linux x64

When a drive in the select-a-drive list is 'disabled' (whether because the drive is read-only, too small for the selected image, etc.) like shown in https://github.com/resin-io/etcher/issues/830#issuecomment-259515129 , it still displays a hand-cursor when you mouse-hover over it, implying that it's clickable when it isn't.

jviotti:
Tweaking `cursor` should do it.

PR #886, opened 2016-11-21
fix(GUI): show not-allowed cursor over unselectable drives

Shou:
The cursor is incorrectly a pointer on hover over a disabled and
unselectable drive in the drive selector widget. This patch changes the
cursor to `not-allowed`, providing the user with visual feedback that it is
not selectable.

Fixes: https://github.com/resin-io/etcher/issues/865
Change-Type: patch
Changelog-Entry: Use not-allowed cursor over disabled drives in the
drive selector widget.

### DIFF
--- a/build/css/main.css
+++ b/build/css/main.css
@@ -6407,6 +6407,9 @@ body {
   width: 300px;
   height: 320px; }
 
+.modal-drive-selector-modal .list-group-item[disabled] {
+  cursor: not-allowed; }
+
 .component-drive-selector-body .list-group-item-footer {
   margin-top: 8px; }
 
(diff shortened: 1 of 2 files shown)


jviotti:
Lets not use "the drivelist" here. The component that shows the available drives is internally called `drive-selector`, while `drivelist` is actually the module that implements the internal drive scanning routines (which could lead to confusin). In any case, we should keep CHANGELOG entries away from internal naming conventions when possible, so something human readable as "the drive selector widget" is the best way to go.

jviotti:
I'll give this a go when I land :)

Shou:
Is this updated commit message good?

jviotti:
Looks great now!

lurch:
Would be nice if you can edit the PR description to match ;-)

Shou:
@jviotti here we go. Followed your advice above for the preceding commit, and changed the PR body/title to match.

Issue #880, opened 2016-11-18
Support double-click in drivelist dialog?


lurch:
- **Etcher version:** 1.0.0-beta16

UX issue, so pinging @taahirisaacs and @konmouz too...

If you have multiple drives connected before running Etcher, when you get to the "Select drive" dialog you get a list of available drives displayed. Currently you have to single-click on one of the drives, and then click CONTINUE (which closes the dialog).
I wonder if it might also make sense for double-clicking on a drive to automatically select that drive and close the dialog?

jviotti:
This is a very good one. Assigning @Shou... Lets go for it.

konmouz:
That would be great!

PR #889, opened 2016-11-23
feat(GUI): allow double-click to quick-select drives

Shou:
The user can now double-click to swiftly select a drive in the
drive widget instead of having to toggle a drive and then clicking the
continue button to close the modal and finalize the drive selection.

Closes: https://github.com/resin-io/etcher/issues/880
Change-Type: patch
Changelog-Entry: Implement double-click to quick-select drives in the
drive widget.

### DIFF
--- a/lib/gui/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/components/drive-selector/controllers/drive-selector.js
@@ -18,7 +18,7 @@
 
 const _ = require('lodash');
 
-module.exports = function($uibModalInstance, DrivesModel, SelectionStateModel, WarningModalService) {
+module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateModel, WarningModalService) {
 
   /**
    * @summary The drive selector state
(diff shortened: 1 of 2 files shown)


Shou:
I added the changes you suggested in https://github.com/resin-io/etcher/commit/42b708b24fecfc365747c8ebddb9fa17fbf716f6#comments, and a `$timeout` to make click events instant (bringing them to the Angular world). It doesn't seem to be needed for double-clicks, although in retrospective maybe I should add a `$timeout` there as well to avoid future headaches.

jviotti:
@Shou Did you try my `$q` vs `Bluebird` suggestion? As a rule of thumb, manually triggering a digest loop should be use very sparingly, and I believe the issue here is that we are mixing promises types.

Shou:
@jviotti I should've resolved all of that now.

>I don't think the "resolve the selected drive" is accurate, since the promise is resolving undefined.

I removed that with this commit that since it's just copy/pasted from the close method above, so it's probably redundant to mention twice. Instead just saying we're closing should be enough, right?

>Can we move the .then() to the line above, just for consistency reasons with the rest of the code base?

Sure, you mean one long line like this? No worries about a column limit?

```javascript
return shouldChangeDriveSelectionState(drive).then((shouldChangeDriveSelectionState) => {
```

Shou:
I also changed the `shouldChangeDriveSelectionState` function arguments to `canChangeDriveSelectionState` so as not to shadow in this preceding commit—passes the tests now, at least locally.

PR #891, opened 2016-11-23
fix(CLI): always omit first two elements of process.argv

jviotti:
This is a great simplification to the logic previously included in the
Yargs `.parse()` function on the Etcher CLI.

There is no scenario where the CLI script will be executed in a
standalone fashion (e.g: `etcher.js <options>`), so we can always safely
drop the first two elements from `process.argv`.

Change-Type: patch
Changelog-Entry: Fix command line arguments not interpreted correctly when running the CLI with a custom named NodeJS binary.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -18,7 +18,6 @@
 
 const _ = require('lodash');
 const fs = require('fs');
-const path = require('path');
 const yargs = require('yargs');
 const utils = require('./utils');
 const EXIT_CODES = require('../src/exit-codes');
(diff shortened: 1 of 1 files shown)


WasabiFan:
That seems a bit hacky, doesn't it -- special-casing JS files like this? Might it be more versatile it if you made the first binary pop the args array before passing it on?

jviotti:
It is very hacky indeed, but I can't think of a better solution :/ 

> Might it be more versatile it if you made the first binary pop the args array before passing it on?

Do you mean removing the first item of the array (e.g: `unshift`?) If we do so, we end up with `index.js <image>`, which will not make sense on its own. Is this what you mean?

WasabiFan:
Presumably, at some point the modified binary invokes the JS file that was passed, correct? What I'm thinking is that, because you know that the JS file was the first argument passed on the command line, you can omit that argument (and if you aren't already, the "etcher" part as well) so that the JS file only receives one argument: The image file name. There are some nuances here regarding the first element of the argv array -- it probably is the name of the executing script itself -- that you should take into account, but I think you get my point. The idea would be to remove elements from the args list such that you end up with no difference between directly executing the JS file and executing it through the Etcher binary.

jviotti:
I actually have a better approach now. Turns out that there is no scenario where `process.argv.slice(1)` makes sense, since the script will never be called independently (there is always going to be an interpreter behind), so we can ditch the whole section and always do `process.argv.slice(2)`.

Issue #820, opened 2016-11-02
Minify and concatenate JS/CSS/HTML


jviotti:
- Revise that every development-related file is excluded from the final package.
- Concatenate and minify all JavaScript.
- Minify CSS.
- Minify HTML.

From https://github.com/resin-io/etcher/issues/632

lurch:
I suppose the HTML could be minified too?

jviotti:
Correct. I'll add it the issue description.

On Wed, Nov 02, 2016 at 11:05:43AM -0700, Andrew Scheller wrote:
> I suppose the HTML could be minified too?
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/resin-io/etcher/issues/820#issuecomment-257949732

-- 
Juan Cruz Viotti
Software Engineer

alexandrosm:
what kind of improvement are we talking about?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lurch:
I doubt the minification will save much, relative to the size of Etcher; but removing the dev-files from the final packages should save a good chunk.
&lt;/guess&gt;

alexandrosm:
yeah, the minification sounds like pain for little gain. but a test should
tell us :)

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lurch:
So, as well as the `/usr/share/etcher-electron/resources/app.asar` (installed by the etcher-electron .deb) containing the entire `scripts` (2.6MB) directory (as mentioned in https://github.com/resin-io/etcher/issues/632#issuecomment-257945411 ), it also includes other 'useless' stuff like `node_modules/electron-winstaller-fixed` (9.8MB), two copies of `liblzma.dll` (300KB), 7 .exe files (1.1MB), 5 .a files (1.3MB), 80 .o files (535KB), 3 .bz2 files (1.1MB).
**Lots** of scope for improvement ;-)

Also, I've never worked on any NodeJS projects before, but `node_modules` contains 666 directories, which seems a _little_ excessive? \*shrug\* (and `find node_modules/ -type d -iname "node_modules" -print|wc -l` reports that 139 of those have their _own_ `node_modules` directory?!)

jviotti:
> So, as well as the `/usr/share/etcher-electron/resources/app.asar`
[...]

PR #897, opened 2016-11-25
refactor: get rid of `electron-packager` in GNU/Linux

jviotti:
In order to take a whitelisting approach to dependencies, the
`unix/dependencies.sh` script is ran as part of packaging as well, but
passing the `-p` (production) flag.

In the future, the current `install`, `package`, and `installer` targets
will be deplaced by `develop-<type>`, `installer-<type>`.

### DIFF
--- a/scripts/build/linux.sh
+++ b/scripts/build/linux.sh
@@ -75,10 +75,18 @@ if [ "$COMMAND" == "install" ] || [ "$COMMAND" == "all" ]; then
 fi
 
 if [ "$COMMAND" == "package" ] || [ "$COMMAND" == "all" ]; then
+  ./scripts/unix/dependencies.sh \
+    -r "$ARCH" \
+    -v "$ELECTRON_VERSION" \
+    -t electron \
+    -p
+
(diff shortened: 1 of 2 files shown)


lurch:
So, is "rolling our own" a better solution than enhancing the `electron-packager` that we _were_ using?

jviotti:
Yeah, `electron-packager` had many limitations that we were currently
workarounding (like incorrectly setting the binary name for the `*.app`
in OS X, not copying the LICENSE correctly, etc), and we were planning
to get rid of it for quite some time, see
https://github.com/resin-io/etcher/issues/820#issuecomment-258155138 for
one of the major motivations.

On Mon, Nov 28, 2016 at 09:21:05AM -0800, Andrew Scheller wrote:
> So, is "rolling our own" a better solution than enhancing the `electron-packager` that we _were_ using?
> 
> -- 
> You are receiving this because you modified the open/close state.
> https://github.com/resin-io/etcher/pull/897#issuecomment-263333829

-- 
Juan Cruz Viotti
Software Engineer

PR #899, opened 2016-11-25
chore: adopt new build script command convention in GNU/Linux

jviotti:
The GNU/Linux build script (as the rest of the OSes will in the near
future), now accepts the following command:

- `develop-electron`
- `develop-cli`
- `installer-cli`
- `installer-appimage`
- `installer-debian`

Each of the commands is now completely independent from the others.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ install:
   - npm install -g bower
   - npm install -g electron-installer-debian
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-      ./scripts/build/linux.sh install x64;
+      ./scripts/build/linux.sh develop-electron x64;
     fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew install afsctool;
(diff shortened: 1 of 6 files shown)


lurch:
Some notes (even as inline comments in the script) about the intended purpose of each of those "targets" would be useful :-)

jviotti:
@lurch They are being introduced in the relevant documentation sections (for example: how to run Etcher locally), so they are somewhat described there. Once this whole refactoring is done, we'll tie them together in a single Makefile, so I guess we can add a `make help` there. What do you think?

lurch:
Sounds good. But given the way that `make` works, with targets and dependencies etc., I think some of the scripts will need to be adjusted so that intermediate files don't get deleted.
Is it possible to have separate `node_modules` directories for use by the CLI and GUI versions of Etcher?

jviotti:
Yeah, definitely. Regarding separate `node_modules`, I guess we could pass `--prefix` to `npm install` (see https://docs.npmjs.com/files/folders#prefix-configuration) to install directly on the package, which sounds really good in theory, but lets see how that goes :P

PR #907, opened 2016-11-29
chore: don't include `LICENSE` and `version` in OS X package

jviotti:
Including them means we have to nicely display them in the final `.dmg`.
For simplicity reasons, we'll omit it for now.


scripts/darwin/package.sh
```diff
@@ -97,8 +93,10 @@ fi
 APPLICATION_OUTPUT="$ARGV_OUTPUT/$ARGV_APPLICATION_NAME.app"
 mv "$ARGV_OUTPUT/Electron.app" "$APPLICATION_OUTPUT"
 rm "$APPLICATION_OUTPUT/Contents/Resources/default_app.asar"
-cp "$ARGV_LICENSE" "$ARGV_OUTPUT/LICENSE"
-echo "$ARGV_VERSION" > $ARGV_OUTPUT/version
+
+# Don't include these for now
+rm "$ARGV_OUTPUT"/LICENSE*
+rm "$ARGV_OUTPUT/version"
```
lurch: These should probably be `rm -f ` so that they don't fail when the files they're trying to delete don't exist?

lurch:
If the `LICENSE` isn't being included in the package, should there be an 'about' dialog (or something) in Etcher to display the version, copyright info, licence, etc.?

scripts/darwin/package.sh
jviotti: Fixed!

jviotti:
@lurch Yeah, we should. Lets check other OS X applications first and see how they do it. I don't remember seeing any app ship the LICENSE file as a sibling of the `*.app`, so maybe the about section is indeed the way to go.

PR #908, opened 2016-11-29
chore: refactor and enhance dependency checking in build scripts

lurch:
This refactors the way that the build shell-scripts check for dependencies
in order to cut down on code duplication, and also adds additional
dependency-checks that were missing previously.

Change-Type: minor


lurch:
@jviotti I've tested on Linux, but haven't tested on MacOS X.

jviotti:
Looks great! Thanks a lot!

PR #909, opened 2016-11-29
chore: install dependencies directly in package

jviotti:
Previously, the build scripts would override the top level
`node_modules` and `bower_components`. After this commit, the
dependencies are installed directly in the Electron package.

### DIFF
--- a/scripts/unix/dependencies.sh
+++ b/scripts/unix/dependencies.sh
@@ -38,6 +38,7 @@ function usage() {
   echo "    -r <architecture>"
   echo "    -v <target version>"
   echo "    -t <target platform (node|electron)>"
+  echo "    -x <install prefix>"
   echo "    -f force install"
   echo "    -p production install"
   exit 0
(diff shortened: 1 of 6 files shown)


lurch:
Would it also make sense to include the OS and ARCH (and perhaps also GUI / CLI) in the prefix, so that we can have multiple versions installed side-by-side?

jviotti:
Yeah, definitely. I'm doing so on the final Makefile (to not do work in vain on something that will be killed very soon). This is more or less what I have locally:

```
release/electron-$(PLATFORM)-$(ARCH)-app:
	./scripts/unix/create-electron-app.sh -s . -f "$(APPLICATION_FILES)" -o $@

release/electron-$(PLATFORM)-$(ARCH)-app/node_modules:\
	release/electron-$(PLATFORM)-$(ARCH)-app
	./scripts/unix/dependencies.sh -p -r "$(ARCH)" -v "$(ELECTRON_VERSION)" -x $< -t electron

release/electron-$(PLATFORM)-$(ARCH)-app.asar: \
	release/electron-$(PLATFORM)-$(ARCH)-app \
	release/electron-$(PLATFORM)-$(ARCH)-app/node_modules
	./scripts/unix/create-asar.sh -d $< -o $@
```

scripts/unix/dependencies.sh
```diff
@@ -96,9 +97,25 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
...
 fi
 
+if [ -n "$ARGV_PREFIX" ]; then
+  NPM_INSTALL_OPTS="$NPM_INSTALL_OPTS --prefix=$ARGV_PREFIX"
+fi
+
 npm install $NPM_INSTALL_OPTS
 
+# Using `--prefix` might cause npm to create an empty `etc` directory
+if [ -n "$ARGV_PREFIX" ] && [ ! "$(ls -A "$ARGV_PREFIX/etc")" ]; then
```
lurch: Why the use of `$(ls -A ...)` here? Wouldn't a simple `-d` test suffice?

lurch: Ahhh, that makes more sense then.

PR #913, opened 2016-11-30
chore: move GNU/Linux zip generation to `installer-appimage`

jviotti:
Currently, `installer-appimage` creates only the AppImage, and we rely
on code outside the main task to package it up as a Zip. Since Zip is
a required step for AppImage deployment (since the AppImage will lose
its execution permissions if its moved around), we integrate Zip
generation into the AppImage task itself.

### DIFF
--- a/scripts/linux/electron-installer-appimage.sh
+++ b/scripts/linux/electron-installer-appimage.sh
@@ -18,6 +18,7 @@
 
 set -u
 set -e
+set -x
 
 function check_dep() {
   if ! command -v $1 2>/dev/null 1>&2; then
@@ -101,11 +102,11 @@ else
   exit 1
 fi
 
+OUTPUT_FILENAME="$ARGV_APPLICATION_NAME-linux-$ARGV_ARCHITECTURE.AppImage"
+
 # Create AppDir
-OUTPUT_FILENAME="$(basename "$ARGV_OUTPUT")"
 APPDIR_PATH=/tmp/${OUTPUT_FILENAME%.*}.AppDir
 APPDIR_ICON_FILENAME=icon
-mkdir -p "$(dirname "$ARGV_OUTPUT")"
 rm -rf "$APPDIR_PATH"
 mkdir -p "$APPDIR_PATH/usr/bin"
 download_executable \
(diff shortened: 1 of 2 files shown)


jviotti:

scripts/linux/electron-installer-appimage.sh
```diff
@@ -143,12 +144,19 @@ else
...
+mkdir -p "$(dirname "$ARGV_OUTPUT")"
 rm -f "$ARGV_OUTPUT"
 
 APPIMAGEASSISTANT_PATH=/tmp/AppImageAssistant.AppImage
 download_executable \
   "$APPIMAGES_GITHUB_RELEASE_BASE_URL/AppImageAssistant_$APPIMAGES_TAG-$APPIMAGES_ARCHITECTURE.AppImage" \
   $APPIMAGEASSISTANT_PATH
 
-$APPIMAGEASSISTANT_PATH "$APPDIR_PATH" "$ARGV_OUTPUT"
+$APPIMAGEASSISTANT_PATH "$APPDIR_PATH" "$(dirname "$ARGV_OUTPUT")/$OUTPUT_FILENAME"
```
lurch: This script assumes that `OUTPUT_FILENAME` has a specific format, and is always in the same directory as `$(dirname $ARGV_OUTPUT)`. Might be better to instead have an extra command-line option to this script giving the full path to `OUTPUT_FILENAME`?
(And perhaps `OUTPUT_FILENAME` isn't the best name for this parameter, given that it's effectively an 'input' to this script?)

jviotti: They way this script works is that it generates an AppImage (which is located at `$OUTPUT_FILENAME`), it creates a ZIP containing the AppImage, and then removes the standalone AppImage. Thus, the AppImage (`$OUTPUT_FILENAME`) can be thought of as a "temporary file". I agree that putting it as a sibling of the expected output path looks a bit awkward. What about putting it in `$TMPDIR` instead?

lurch: > What about putting it in $TMPDIR instead?

Yeah, I guess that works. And as I mentioned somewhere else, deleting files as you go isn't really very `make`-friendly ;)

jviotti: Done! See https://github.com/resin-io/etcher/pull/936

probonopd:
Please do not put AppImage files into any kind of archives. The reasons are given on https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages#common-mistake.

PR #914, opened 2016-11-30
chore: split `dependencies.sh` into bower and npm scripts

jviotti:
The purpose of this change is to stay true to the Makefile approach,
where each rule is meant to create a single directory/file.

Currently, `unix/dependencies.sh` resulted in `bower_components` and
`node_modules`. Splitting this rule also means that we don't have to
install bower dependencies where we don't need them (e.g: the CLI).

### DIFF
--- a/scripts/unix/dependencies-bower.sh
+++ b/scripts/unix/dependencies-bower.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+###
+# Copyright 2016 resin.io
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
+set -u
+set -e
+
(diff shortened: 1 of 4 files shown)


scripts/unix/dependencies-bower.sh
```diff
@@ -0,0 +1,64 @@
...
+  INSTALL_OPTS="$INSTALL_OPTS --production"
+fi
+
+if [ -n "$ARGV_PREFIX" ]; then
+  pushd "$ARGV_PREFIX"
+  bower install $INSTALL_OPTS
+  popd
+fi
+
+bower install $INSTALL_OPTS
```
lurch: Is this right: `bower install` being run twice?? Or is the second one supposed to be inside an `else` clause?

lurch:
Probably makes sense for `NPM_INSTALL_OPTS` in `dependencies-npm.sh` to be renamed to just `INSTALL_OPTS` ? (so that it matches `dependencies-bower.sh`)

scripts/unix/dependencies-bower.sh
```diff
@@ -0,0 +1,64 @@
...
+  echo ""
+  echo "    -x <install prefix>"
+  echo "    -p production install"
+  exit 0
+}
+
+ARGV_PREFIX=""
+ARGV_PRODUCTION=false
+
+while getopts ":x:fp" option; do
```
jviotti: Good catch! Fixed!

scripts/unix/dependencies-bower.sh
jviotti: Oh, my bad! Fixed!

jviotti:
Done!

PR #916, opened 2016-11-30
chore: exit with error code 1 on build script usage section

jviotti:
We're currently exitting with error code 0 when displaying the usage
information on build scripts, which means that if the user forgets or
mistypes an option argument, the script will carry on its way.


lurch:
Haha, I was planning on making exactly this change myself, and you beat me to it! :+1:

PR #917, opened 2016-11-30
chore: attempt to re-use $TMPDIR in `installer-appimage.sh`

jviotti:
We currently hardcode the temporary directory location to `/tmp`. Some
Linux systems define `$TMPDIR`, so we should attempt to re-use that
first.

### DIFF
--- a/scripts/linux/electron-installer-appimage.sh
+++ b/scripts/linux/electron-installer-appimage.sh
@@ -102,10 +102,15 @@ else
   exit 1
 fi
 
+if [ -z "$TMPDIR" ]; then
+  TMPDIR=$(mktemp -d)
+fi
+mkdir -p "$TMPDIR"
+
 OUTPUT_FILENAME="$ARGV_APPLICATION_NAME-linux-$ARGV_ARCHITECTURE.AppImage"
 
 # Create AppDir
-APPDIR_PATH=/tmp/${OUTPUT_FILENAME%.*}.AppDir
+APPDIR_PATH=$TMPDIR/${OUTPUT_FILENAME%.*}.AppDir
 APPDIR_ICON_FILENAME=icon
 rm -rf "$APPDIR_PATH"
 mkdir -p "$APPDIR_PATH/usr/bin"
@@ -147,7 +152,7 @@ fi
 mkdir -p "$(dirname "$ARGV_OUTPUT")"
 rm -f "$ARGV_OUTPUT"
 
-APPIMAGEASSISTANT_PATH=/tmp/AppImageAssistant.AppImage
+APPIMAGEASSISTANT_PATH=$TMPDIR/AppImageAssistant.AppImage
 download_executable \
   "$APPIMAGES_GITHUB_RELEASE_BASE_URL/AppImageAssistant_$APPIMAGES_TAG-$APPIMAGES_ARCHITECTURE.AppImage" \
   $APPIMAGEASSISTANT_PATH


lurch:
...or we could just use `mktemp -d`, which already handles `$TMPDIR` for us :-)

jviotti:
@lurch I though about it as well, however my main motivation of using `$TMPDIR` is that the Makefile (or parent script) calling the scripts can define its own `$TMPDIR` local to the project, for example (`etcher-release/.tmp`) making the builds completely self-contained. What do you think?

lurch:
Yes, mktemp handles that too...
```bash
$ mkdir -p /tmp/my/awesome/dir
$ mktemp -d
/tmp/tmp.rdwZ9rdEfB
$ TMPDIR=/tmp/my/awesome/dir mktemp -d
/tmp/my/awesome/dir/tmp.pE2P5xs27C
```

jviotti:
Very nice! Done!

jviotti:
@lurch Hm, I don't know if I'm happy with the latest changes to be honest. My use case was that we could set the temporary directory to be `etcher-release/.tmp` or something like that, and we could save things like AppImage related binaries there, and if we detect they are already present, we don't re-download.

The way it is now, a new random directory will be created inside `etcher-release/.tmp`, which makes our "cache" ideas impossible.

What do you think?

jviotti:
Maybe we should only use `mktemp -d` for things that will not be shared, and really need to have non-conflicting unique names?

jviotti:
OK, I've settled with the following approach: if `$TMPDIR` is defined, just use that. If not, use `mktemp -d`.

jviotti:

lurch:
> we could save things like AppImage related binaries there, and if we detect they are already present, we don't re-download.

Yeah, I was thinking of adding something very similar myself :)

> What do you think?

Too late now (given that you've already merged this PR) but in buildroot projects all external tarballs / binaries / etc. get downloaded into a `dl` folder.
However one thing we ought to add is storing the md5sum (or sha1sum) of the downloaded binaries, and then re-downloading the files if they're missing, or if the expected hash doesn't match.

jviotti:
> Too late now (given that you've already merged this PR) but in buildroot projects all external tarballs / binaries / etc. get downloaded into a dl folder.

We can always do a new PR with the changes, so that's not a problem :) How is the `dl` folder determined? Is it hardcoded to be relative to the current directory? Do you pass it as an argument (in our case, it could be determined by the Makefile, and then propagated to all children as command line options)?

> However one thing we ought to add is storing the md5sum (or sha1sum) of the downloaded binaries, and then re-downloading the files if they're missing, or if the expected hash doesn't match.

That's a great idea. Lets see if the AppImages project put the checksum somewhere.

lurch:
> How is the dl folder determined? 

It's just a single `dl` folder in the `buildroot` root directory. I was thinking it might be best to just have a single script which does the downloading (with URL and checksum being input parameters), which other scripts can then call.

> Lets see if the AppImages project put the checksum somewhere.

Might be quicker / easier to just generate it ourselves ;-)  Obviously we'd need a 'checking' script to check that all the current checksums in our scripts are still correct (maybe that could be a Travis step?).

jviotti:
> I was thinking it might be best to just have a single script which does the downloading (with URL and checksum being input parameters), which other scripts can then call.

Sound good, I'll get into it soon!

> Might be quicker / easier to just generate it ourselves ;-) Obviously we'd need a 'checking' script to check that all the current checksums in our scripts are still correct (maybe that could be a Travis step?).

I propose to delegate that responsibility by opening a ticket on the AppImageKit project so they publish checksums to GitHub Releases, and we'll make use of them when we upgrade to `appimagetool`.

PR #918, opened 2016-11-30
chore: improve naming of build scripts

jviotti:
- `darwin/configure-electron.sh` -> `darwin/electron-configure-package-darwin.sh`
- `darwin/installer-zip.sh` -> `darwin/electron-installer-app-zip.sh`
- `darwin/installer-dmg.sh` -> `darwin/electron-installer-dmg.sh`
- `darwin/sign.sh` -> `darwin/electron-sign-app.sh`
- `linux/configure-electron.sh` -> `linux/electron-configure-package-linux.sh`
- `linux/installer-appimage.sh` -> `linux/electron-installer-appimage.sh`
- `linux/installer-deb.sh` -> `linux/electron-installer-debian.sh`
- `unix/create-asar.sh` -> `unix/electron-create-asar.sh`
- `unix/create-electron-app.sh` -> `unix/electron-create-resources-app.sh`
- `unix/download-electron.sh` -> `unix/electron-download-package.sh`


jviotti:
@lurch Looking forward to naming suggestions. I prefixed most with `electron` since they are very electron specific, and by having a prefix in place, we can create a whole new suite of "cli" specific scripts.

jviotti:
In the future, I'm also looking forward to join all scripts in a single directory, since a lot of stuff will be able to get re-used. What do you think? Can you think of any reason to not do so?

lurch:
`electron-create-resources-app.sh` sounds like a strange name?

lurch:
:+1: for re-use :)

jviotti:
> electron-create-resources-app.sh sounds like a strange name?

Yeah, but seems to be the way electron calls it, and despite strange, it will probably be familiar to people working on electron. So when you package Electron applications, you create a directory called `app` (you should compress to `app.asar` for proper releasing) which lives inside a "resources" directory. Therefore its called "the resources app", so its not confused with OS X `*.app`s.

jviotti:
Talking about re-use, I'd be so awesome if we could write both Windows and UNIX scripts in the same language (e.g: bash), since that would allow us to re-use even more, but that doesn't look straightforward. We could require Windows users to build on something like Cygwin, but we rely on "Visual Studio" `cmd.exe` to access command line build utilities, and I'm not sure how easy it'd be to make use of those from outside (since the Visual Studio console is pre-configured with lots of stuff to just work).

Windows 10 ships experimental bash support, which would be the ideal case, but we still need to support older Windows versions, and the bash support is very unstable as far as I've heard.

Something like https://github.com/BYVoid/Batsh looks perfect, but that one is unmaintained, and doesn't build at all in OS X.

Thoughts?

lurch:
> and despite strange, it will probably be familiar to people working with electron

Fair enough.

> Something like https://github.com/BYVoid/Batsh looks perfect

Cool, that looks neat! As you say, shame that it's unmaintained :-(  (and obviously being written in OCaml means it's unlikely to get 3rd-party contributors)

If we wan't to go with "native scripting" then I guess using bash on Linux / OSX and bat on Windows is a pain we'll just have to put up with. OTOH, when we start shifting more stuff over to Makefiles, they're easier to make cross-platform.
Alternatively, given that node and python are build-dependencies, I guess theoretically there's no reason that the build scripts couldn't be rewritten in those?

jviotti:
Yeah, to be honest, I'm very tempted to use the C preprocessor or something like [`gpp`](https://files.nothingisreal.com/software/gpp/gpp.html) and create some macros that translate basic functionality to either bash or batch... Maybe a small side project :P

> Alternatively, given that node and python are build-dependencies, I guess theoretically there's no reason that the build scripts couldn't be rewritten in one of those?

The problem is that some scripts will feel very awkward on something else (for example see the AppImage installer generation script). Bash makes perfect sense for it, while using something else might lead to an unmaintainable mess.

jviotti:
There are [huge delays in OS X builds in Travis](https://www.traviscistatus.com/incidents/nq0sf0srvx8s) at the moment, so I'll merge without waiting for the OS X builds (which can take up to ~10 hours). I tested it myself on my MacBook Pro.

Issue #905, opened 2016-11-29
Add link to CHANGELOG in update prompt


jviotti:
Add a "see what's new" link in the update prompt for users to see the CHANGELOG.

PR #920, opened 2016-12-01
feat(gui): add changelog link to update notifier

Shou:
We add a blue clickable link to the Github changelog in the update dialog
below the modal body text.

Closes: https://github.com/resin-io/etcher/issues/905
Change-Type: patch
Changelog-Entry: Add a changelog link to the update notifier modal.

### DIFF
--- a/lib/gui/components/update-notifier/templates/update-notifier-modal.tpl.html
+++ b/lib/gui/components/update-notifier/templates/update-notifier-modal.tpl.html
@@ -6,6 +6,9 @@
 <div class="modal-body">
   <div class="modal-text">
     <p>A new version of Etcher is available for download</p>
+    <a os-open-external="https://github.com/resin-io/etcher/blob/master/CHANGELOG.md#readme">
+      See what's new
+    </a>
   </div>
 </div>
 
--- a/build/css/main.css
+++ b/build/css/main.css
@@ -6240,6 +6240,8 @@ body {
   padding: 0;
   max-height: 250px;
   overflow: auto; }
+  .modal-body a {
+    color: #5793db; }
   .modal-body .list-group {
     padding: 0 20px; }
   .modal-body .list-group-item {
--- a/lib/gui/components/modal/styles/_modal.scss
+++ b/lib/gui/components/modal/styles/_modal.scss
@@ -44,6 +44,10 @@
   max-height: 250px;
   overflow: auto;
 
+  a {
+    color: $palette-theme-primary-background;
+  }
+
   .list-group {
     padding: 0 20px;
   }


Shou:
@jviotti Is this good?

lurch:
Hmmm, this reminds me of #729 where the "SKIP" button has such a light background colour, it looks like it's disabled and non-clickable?

Maybe @konmouz and @taahirisaacs could review the colour scheme? (and maybe this is another dialog where it doesn't make sense to have both a "SKIP" button and an X in the top-right?)

lurch:
And should the "New update available" dialog also report the version number of the new Etcher release? :)

konmouz:
Yes, let's follow the example below. Just two options without 'x' button. Also, active CTA (blue) should be always on the right.

[modal](https://cloud.githubusercontent.com/assets/15323961/20760236/5a9e8cfa-b717-11e6-8ab5-3b5c9351304d.png)

lib/gui/components/update-notifier/templates/update-notifier-modal.tpl.html
```diff
@@ -6,6 +6,9 @@ <h4 class="modal-title">New Update Available!</h4>
 <div class="modal-body">
   <div class="modal-text">
     <p>A new version of Etcher is available for download</p>
+    <a os-open-external="https://github.com/resin-io/etcher/blob/master/CHANGELOG.md#readme">
```
jviotti: Why the `#readme` hash? Should it be something like https://github.com/resin-io/etcher/blob/master/CHANGELOG.md#change-log?

lib/gui/components/update-notifier/styles/_update-notifier.scss
```diff
@@ -20,3 +20,7 @@
   margin-bottom: 0;
 }
 
+.modal-update-notifier .modal-body a {
```
jviotti: Can you double check if we're using anchors in any other modal? I believe all modal anchors will benefit from this style, so we could move it to `modal/styles/_modal.scss` unless it conflicts with anything else. 

As a side note, have you checked how this behaves on the multiple states an anchor can be at any given time (e.g: `:link`, `:active`, `:focus`, etc)?

jviotti:
> Maybe @konmouz and @taahirisaacs could review the colour scheme? (and maybe this is another dialog where it doesn't make sense to have both a "SKIP" button and an X in the top-right?)
> Yes, let's follow the example below. Just two options without 'x' button. Also, active CTA (blue) should be always on the right.

@Shou Lets make these changes as another PR that is specifically concerned with normalizing modal styles, and addresses any other modal that should be updated to match @konmouz 's example.

jviotti:
> And should the "New update available" dialog also report the version number of the new Etcher release? :)

@Shou Yeah, lets do it!

lib/gui/components/update-notifier/templates/update-notifier-modal.tpl.html
Shou: I decided to go with `#readme` because it's less cramped with the empty space above the header and it becomes more overt to the user that there isn't any more content above where the browser scrolled to.

jviotti: But where does the `#readme` id come from? Is this something defined by GitHub that makes the link present the document in question?

jviotti: screenshot 2016-12-01 12 28 07

Indeed, very interesting. Looks good to me then!

lib/gui/components/update-notifier/styles/_update-notifier.scss
Shou: You mean in terms of styling? It's just a static blue; there's no visual feedback here. It should be easy to add after we've standardized colours like you suggested below though!

jviotti: Oh, so we are not using the pseudo classes mentioned above at all here in Etcher (they have all been normalized to the "normal state"), so just ignore me on this one :)

jviotti: Lets move the anchor style to `modal/styles/_modal.scss` and we should be good to merge.

Shou: Done!

PR #922, opened 2016-12-01
chore: add Makefile to build OS X and GNU/Linux

jviotti:
This Makefile ties together all the build scripts we've been developing
so far. It currently only supports UNIX based operating systems, but
Windows support will be added soon.


jviotti:
@lurch @Shou Do you mind giving this a try (try generating an AppImage or a Debian package), just to confirm if its as user-friendly as we wish? Let me know if there is anything weird along the way!

jviotti:
Damn, the Linux targets do nothing for some reason. I'm investigating...

jviotti:
Hm, looks like the phony targets won't work if they are declared inside conditionals :/

jviotti:
Oh, small typo. Never-mind, it should work now :)

jviotti:
There are [huge delays in OS X builds in Travis](https://www.traviscistatus.com/incidents/nq0sf0srvx8s) at the moment, so I'll merge without waiting for the OS X builds (which can take up to ~10 hours). I tested it myself on my MacBook Pro.

Let me know how it went!

PR #923, opened 2016-12-01
chore: reuse npm/bower dependencies across builds

jviotti:
Currently, we create a "resources app" and install dependencies on it.
This means that to create another "resources app", the whole dependency
installation process needs to be repeated.

As a solution, we cache dependencies at
`release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/` and
copy them to the resources apps as we need them.

This has the additional benefit that the resources app depends on the
dependencies rather than the other way around (dependencies depend on
the resources app), which didn't feel right.

In order to accomodate to these changes, the dependencies scripts are
now in charge of taking their corresponding manifests with them when the
prefix option is passed, given that neither `npm` nor `bower` allow to
customise the output location.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,21 +88,22 @@ APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 # Rules
 # ---------------------------------------------------------------------
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app:
+release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json
+	./scripts/unix/dependencies-npm.sh -p -r "$(TARGET_ARCH)" -v "$(ELECTRON_VERSION)" -x $(dir $@) -t electron
+
+release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components: bower.json
+	./scripts/unix/dependencies-bower.sh -p -x $(dir $@)
+
+release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app: \
+	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules \
+	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components
 	./scripts/unix/electron-create-resources-app.sh -s . -f "$(APPLICATION_FILES)" -o $@
-
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/node_modules:\
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app
-	./scripts/unix/dependencies-npm.sh -p -r "$(TARGET_ARCH)" -v "$(ELECTRON_VERSION)" -x $< -t electron
-
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/bower_components:\
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app
-	./scripts/unix/dependencies-bower.sh -p -x $<
+	for prerequisite in $^; do \
+		cp -rf $$prerequisite $@; \
+	done
 
 release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app.asar: \
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app \
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/node_modules \
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/bower_components
+	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app
 	./scripts/unix/electron-create-asar.sh -d $< -o $@
 
 release/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH): \
--- a/scripts/unix/dependencies-bower.sh
+++ b/scripts/unix/dependencies-bower.sh
@@ -56,8 +56,10 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
 fi
 
 if [ -n "$ARGV_PREFIX" ]; then
+  cp "$PWD/bower.json" "$ARGV_PREFIX"
   pushd "$ARGV_PREFIX"
   bower install $INSTALL_OPTS
+  rm bower.json
   popd
 else
   bower install $INSTALL_OPTS
--- a/scripts/unix/dependencies-npm.sh
+++ b/scripts/unix/dependencies-npm.sh
@@ -97,12 +97,24 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
 fi
 
 if [ -n "$ARGV_PREFIX" ]; then
+  mkdir -p "$ARGV_PREFIX"
   INSTALL_OPTS="$INSTALL_OPTS --prefix=$ARGV_PREFIX"
+  cp "$PWD/package.json" "$ARGV_PREFIX"
+
+  if [ -f "$PWD/npm-shrinkwrap.json" ]; then
+    cp "$PWD/npm-shrinkwrap.json" "$ARGV_PREFIX"
+  fi
 fi
 
 npm install $INSTALL_OPTS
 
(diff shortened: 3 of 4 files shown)


jviotti:

scripts/unix/dependencies-npm.sh
```diff
@@ -97,12 +97,24 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
...
-# Using `--prefix` might cause npm to create an empty `etc` directory
-if [ -n "$ARGV_PREFIX" ] && [ ! "$(ls -A "$ARGV_PREFIX/etc")" ]; then
-  rm -rf "$ARGV_PREFIX/etc"
+if [ -n "$ARGV_PREFIX" ]; then
+  rm -f "$ARGV_PREFIX/package.json"
+  rm -f "$ARGV_PREFIX/npm-shrinkwrap.json"
+
+  # Using `--prefix` might cause npm to create an empty `etc` directory
+  if [ ! "$(ls -A "$ARGV_PREFIX/etc")" ]; then
+    rm -rf "$ARGV_PREFIX/etc"
```
lurch: If this only deletes the directory when it's actually empty, then I guess `rm -rf` could be replaced with `rmdir`.

Makefile
```diff
@@ -88,21 +88,22 @@ APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
...
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app
-	./scripts/unix/dependencies-npm.sh -p -r "$(TARGET_ARCH)" -v "$(ELECTRON_VERSION)" -x $< -t electron
+release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components: bower.json
+	./scripts/unix/dependencies-bower.sh -p -x $(dir $@)
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/bower_components:\
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app
-	./scripts/unix/dependencies-bower.sh -p -x $<
+release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app: \
+	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules \
```
lurch: You need to be careful here - you shouldn't use directories as Makefile dependencies, because when the directory-contents change, the directory's timestamp updates, which causes `make` to think the target is out of date (when it might not be). Instead you should either use "marker files" (buildroot uses `.stamp_blah` files), or order-only dependencies

scripts/unix/dependencies-bower.sh
```diff
@@ -56,8 +56,10 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
 fi
 
 if [ -n "$ARGV_PREFIX" ]; then
+  cp "$PWD/bower.json" "$ARGV_PREFIX"
   pushd "$ARGV_PREFIX"
   bower install $INSTALL_OPTS
+  rm bower.json
```
lurch: I'd be tempted to move this after the popd, so that things are more explicit.

Makefile
lurch: Another problem with having directories as Makefile dependencies (that I've just discovered) is that if you have a script which creates the directory and then executes some process(es) to update the files inside it, and the update-process fails, then the next time you run `make` it'll see that the directory already exists, assume everything is up-to-date, and not do anything.
So a better approach would be to `touch` a "marker file" _after_ the update process has finished (e.g. `.npm_dependencies_installed`), and use _that_ as the dependency, so that if it fails `make` knows that it needs to try running the update-process again next time.

scripts/unix/dependencies-npm.sh
jviotti: Done: https://github.com/resin-io/etcher/pull/933

scripts/unix/dependencies-bower.sh
jviotti: Done! https://github.com/resin-io/etcher/pull/934

Makefile
jviotti: I see, it makes a lot of sense. Let me make sure I completely understand the "marker file" idea:

So the following rule:

```make
release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: 
	./scripts/unix/dependencies-npm.sh -p -r "$(TARGET_ARCH)" -v "$(ELECTRON_VERSION)" -x $(dir $@) -t electron
```

would become something like:

```make
release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies-stamp: 
	./scripts/unix/dependencies-npm.sh -p -r "$(TARGET_ARCH)" -v "$(ELECTRON_VERSION)" -x $(dir $@) -t electron
        touch $@
```

or something like that, and then, the rules that require `node_modules` will declare the stamp file as a dependency.

How do we define that the stamp file is related to a "node_modules" directory somewhere? Does it mean that we will hardcode the path to "node_modules" within the actual targets?

lurch: Nearly. I'd do something more like:
```
release:
        mkdir $@

release/stamp_files: | release
        mkdir $@

release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: | release
        mkdir $@

release/stamp_files/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-npm-dependencies-installed.stamp: | release/stamp_files release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
        ./scripts/unix/dependencies-npm.sh -p -r "$(TARGET_ARCH)" -v "$(ELECTRON_VERSION)" -x release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies -t electron
        touch $@
```
and then have other targets depending on `release/stamp_files/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-npm-dependencies-installed.stamp`. So it's a stamp per Makefile-step, not a stamp per directory (and IMHO it makes sense to keep all the stamp-files in the same directory, so you can see at-a-glance what's been done). And then anywhere where you `rm -rf release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules` you need to remember to `rm release/stamp_files/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies-npm-installed.stamp` too (or create another Makefile rule to clean both in one step). Of course this can be all made to look **much** nicer with the use of Makefile variables (which is something I've started on, but haven't PRed yet, because I have to keep rebasing it! :wink: ).

The bit after the `|` in the rule is an order-only dependency, which means `make` only checks it exists, it doesn't compare timestamps.

In the project for my other client (which uses Perl rather than Bash and Javascript), I actually list the script as a dependency of the rule too, so that if the script gets modified then the rule gets re-run. That may or may not be appropriate here, depending on what each script is actually doing.

jviotti: > Of course this can be all made to look much nicer with the use of Makefile variables (which is something I've started on, but haven't PRed yet, because I have to keep rebasing it! 😉 ).

I probably already did all of the major changes, so happy to wait for that PR!

I don't like the multiple `mkdir -p` rules though. Would it make more sense to rely on the scripts to automatically create the directory structure they need, so we don't have to bother about that in the Makefile? In this case, the rule that `touch`es would be the one to call `mkdir -p` in advance.

I understand that this fits better the Makefile style, but I also think we should try to keep the Makefile as clean as possible (as a file that only delegates to scripts), and I can see how the amount of rules can explode if we add separate ones for each directory we are targeting, although it practice it might not be that much. What do you think?

lurch: The nice thing about having the `mkdir` commands in the `Makfefile`, is that `make` can see if the directory already exists, and so skip running the `mkdir` in the first place. It's also nice to have a central place where all the directories are created, rather than having different `mkdir` commands scattered amongst different Bash scripts? (chances are we have multiple scripts each creating the same directories?)

lurch: Yeah, I guess that could work... although perhaps we'd both be happiest if both the Makefile and the individual scripts created the directories? (just kidding! :wink: )
The `stamps` directory would still need to be created by the Makefile though.

While comparing `dependencies-npm.sh` and `dependencies-bower.sh` recently, I noticed that `dependencies-npm.sh` has a `mkdir -p "$ARGV_PREFIX"` but the other script doesn't. And this is "fine" with the way the scripts are being used currently, because `dependencies-bower.sh` is always called _after_ `dependencies-npm.sh`. But it's a perfect example of my "central place where all the directories are created" comment earlier ;-)

PR #924, opened 2016-12-01
style: convert spaces to tabs in Makefile

jviotti:
Additionally, we setup an EditorConfig rule to automatically switch to
tab style on Makefile.


jviotti:

PR #925, opened 2016-12-01
chore: exit with error code 1 on build script usage section

lurch:
Somehow got missed out when the rest of the scripts got updated


jviotti:
Oh, very good catch!

PR #926, opened 2016-12-02
chore: move electron package caching responsibility to GNU Make

jviotti:
Currently, `scripts/unix/electron-download-package.sh` contains its own
logic to determine if the package was already downloaded.

This commit takes the electron package uncompressing task to the
configure scripts, and reduce `electron-download-package.sh` to a script
that simply downloads the zip to a certain location, since this change
allows us to have a separate Make rule to download the Electron zip (and
thus have Make take care of caching).

After this change, the `electron-configure-package-*.sh` scripts are no
longer routines that modify a certain directory, but scripts that take
the zip as an input and actually create the package directory, which
aligns better with Make's design.

### DIFF
--- a/scripts/linux/electron-configure-package-linux.sh
+++ b/scripts/linux/electron-configure-package-linux.sh
@@ -19,6 +19,12 @@
 set -u
 set -e
 
+OS=$(uname)
+if [[ "$OS" != "Linux" ]]; then
+  echo "This script is only meant to be run in GNU/Linux" 1>&2
+  exit 1
+fi
+
 function check_dep() {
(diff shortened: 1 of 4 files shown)


jviotti:
@lurch Let me know if this is good to merge!

scripts/linux/electron-configure-package-linux.sh
```diff
@@ -26,6 +26,8 @@ function check_dep() {
   fi
 }
 
+check_dep unzip
```
jviotti: Oh, true, fixed!

PR #927, opened 2016-12-02
style: use GNU Make `foreach` instead of a shell for loop

jviotti:
Just for conciseness and style.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,9 +98,7 @@ release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app: \
 	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules \
 	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components
 	./scripts/unix/electron-create-resources-app.sh -s . -f "$(APPLICATION_FILES)" -o $@
-	for prerequisite in $^; do \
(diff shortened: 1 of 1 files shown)


lurch:
...and portability :-)

(in my cross-platform Makefiles I define e.g. `COPY = cp` on Linux, `COPY = copy` on Windows, etc. And I use the presence of the ProgramFiles variable to check if I'm on Windows or not - bit hacky but it works :wink: )

jviotti:
Where were you running `make` from? I've recently created a cross-platform Makefile for `node-static-entry-point` and I had to actually define `CP = cmd.exe /c` for it to work (see https://github.com/resin-io-modules/node-static-entry-point/blob/master/Makefile#L19)

lurch:
> Where were you running make from?

Sorry, I was referring to a project that I've worked on for another client, nothing to do with resin or Etcher ;-)
(and I actually "cheated" a bit by using small parts of https://en.wikipedia.org/wiki/UnxUtils on Windows to give more consistent command-flag handling)

lurch:
...I never needed to add any mentions of `cmd.exe` though, but I was using a MinGW build of GNU make.exe (ah, which I see your `node-static-entry-point` is also using). Weird.

lurch:
(and I eventually ended up needing to use Makefile features which were only in the 4.x version of GNU make, so IIRC I compiled my own mingw version)

PR #928, opened 2016-12-02
feat(gui): show available version in update notifier modal

Shou:
We show the Etcher version string in the update notifier modal, by giving
the version string to the template through options similar to the warning
modal.

- We memoize the version Promise and assume the update version won't change
  during runtime.

Changelog-Entry: Show available Etcher version in the update notifier.

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -80,7 +80,7 @@ app.run(() => {
   ].join('\n'));
 });
 
-app.run((AnalyticsService, UpdateNotifierService, SelectionStateModel) => {
+app.run((AnalyticsService, ErrorService, UpdateNotifierService, SelectionStateModel) => {
   AnalyticsService.logEvent('Application start');
 
   if (UpdateNotifierService.shouldCheckForUpdates() && !process.env.ETCHER_DISABLE_UPDATES) {
@@ -96,9 +96,9 @@ app.run((AnalyticsService, UpdateNotifierService, SelectionStateModel) => {
       if (!isLatestVersion && !SelectionStateModel.hasImage()) {
 
         AnalyticsService.logEvent('Notifying update');
-        UpdateNotifierService.notify();
+        return UpdateNotifierService.notify();
       }
-    });
+    }).catch(ErrorService.reportException);
   }
 
 });
--- a/lib/gui/components/update-notifier/services/update-notifier.js
+++ b/lib/gui/components/update-notifier/services/update-notifier.js
@@ -16,6 +16,7 @@
 
 'use strict';
 
+const _ = require('lodash');
 const semver = require('semver');
 const etcherLatestVersion = require('etcher-latest-version');
 
@@ -25,6 +26,9 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
    * @summary Get the latest available Etcher version
    * @function
    * @private
+   * @description
+   * We assume the received latest version number will not increase
+   * while Etcher is running and memoize it
    *
    * @fulfil {String} - latest version
    * @returns {Promise}
@@ -34,7 +38,7 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
    *   console.log(`The latest version is: ${latestVersion}`);
    * });
    */
-  this.getLatestVersion = () => {
+  this.getLatestVersion = _.memoize(() => {
     return $q((resolve, reject) => {
       return etcherLatestVersion((url, callback) => {
         return $http.get(url).then((response) => {
@@ -50,7 +54,9 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
         return resolve(latestVersion);
       });
     });
-  };
+
+  // Arbitrary identifier for the memoization function
+  }, _.constant('latest-version'));
 
   /**
    * @summary Check if the current version is the latest version
@@ -111,11 +117,18 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
    * UpdateNotifierService.notify();
    */
   this.notify = () => {
-    return ModalService.open({
-      template: './components/update-notifier/templates/update-notifier-modal.tpl.html',
-      controller: 'UpdateNotifierController as modal',
-      size: 'update-notifier'
-    }).result;
+    return this.getLatestVersion().then((version) => {
+      return ModalService.open({
+        template: './components/update-notifier/templates/update-notifier-modal.tpl.html',
+        controller: 'UpdateNotifierController as modal',
+        size: 'update-notifier',
+        resolve: {
+          options: _.constant({
+            version: version
+          })
+        }
+      }).result;
+    });
   };
 
 };
--- a/lib/gui/components/update-notifier/controllers/update-notifier.js
+++ b/lib/gui/components/update-notifier/controllers/update-notifier.js
@@ -16,7 +16,7 @@
 
 'use strict';
 
-module.exports = function($uibModalInstance, SettingsModel) {
+module.exports = function($uibModalInstance, SettingsModel, options) {
 
   // We update this value in this controller since its the only place
   // where we can be sure the modal was really presented to the user.
@@ -31,6 +31,13 @@ module.exports = function($uibModalInstance, SettingsModel) {
    * @public
    */
   this.settings = SettingsModel;
+
+  /**
+   * @summary Modal options
+   * @type Object
+   * @public
+   */
+  this.options = options;
 
   /**
    * @summary Close the modal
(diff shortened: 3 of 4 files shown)


Shou:
Here's what it looks like.

lurch:
Any way to avoid `getLatestVersion()` being called twice?

Shou:
@lurch good call, we could probably just use it once in `lib/gui/app.js` and then pass the version value to `notify`. @jviotti are optional arguments alright, or is this style undesired?
```javascript
this.isLatestVersion = (version = null) => {
  if (version !== null) {
    // do stuff
  } else {
    // do other stuff
  }
}
```

jviotti:
@Shou 

> @lurch good call, we could probably just use it once in lib/gui/app.js and then pass the version value to notify. @jviotti are optional arguments alright, or is this style undesired?

Optional arguments are not a bad thing, however the way you propose them is `isLatestVersion` could lead to a lot of confusion. The function is basically a getter, and passing the value it should itself return as an argument to force the function take another code path would look really weird.

Here's an idea: The result from `isLatestVersion` is unlikely to change while the application is running, and even if to does change, we don't care about it until the next time the application is started. Taking that assumption, you could add memoization to the function, so after the first time, it just keeps returning the same thing.

Shou:
@jviotti here's an update with memoizing; relevant changes on lines 25 and 51.

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -111,11 +125,18 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
...
+        template: './components/update-notifier/templates/update-notifier-modal.tpl.html',
+        controller: 'UpdateNotifierController as modal',
+        size: 'update-notifier',
+        resolve: {
+          options: _.constant({
+            version: version
+          })
+        }
+      }).result;
+    }).catch(ErrorService.reportException);
```
jviotti: Lets move this `.catch()` to the code that actually calls `.notify()` (which I believe is in `lib/gui/app.js` only)

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -16,10 +16,24 @@
...
 
+const _ = require('lodash');
 const semver = require('semver');
 const etcherLatestVersion = require('etcher-latest-version');
 
-module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, ManifestBindService, SettingsModel) {
+module.exports = function($http, $q, ErrorService, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, ManifestBindService, SettingsModel) {
+
+  // Memoize the version Promise
+  const memoizeVersion = (func) => {
```
jviotti: Yeah, I think `_.memoize` *should* work fine with Promises, but I don't know for sure.

jviotti:
Oops, this PR ended up with conflicts because of https://github.com/resin-io/etcher/pull/920

jviotti:
Oh, the CI servers are also failing in GNU/Linux because of https://github.com/resin-io/etcher/pull/931. A rebase should do the trick.

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -34,7 +48,7 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
    *   console.log(`The latest version is: ${latestVersion}`);
    * });
    */
-  this.getLatestVersion = () => {
+  this.getLatestVersion = memoizeVersion(() => {
```
jviotti: I propose to add a JSDoc description to this function explaining that we memoize this function, the reasons we do so.

lib/gui/components/update-notifier/services/update-notifier.js
Shou: Looks like you can just provide a "hash function" like `_.constant('latest-version')` as the second argument and `_.memoize` works fine with Promises. I've done that now.

Shou:
In brief, this commit follows the review advice; memoizing with Lodash, JSDoc description expanded, and moved the `.notify()` catch to `lib/gui/app.js`.

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -50,7 +53,7 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
         return resolve(latestVersion);
       });
     });
-  };
+  }, _.constant('latest-version'));
```
jviotti: Looks great! Just one last minor thing: can we add a comment above this line clarifying that the `latest-version` string is completely arbitrary and has no relation with the actual function name? For a second, I thought that this id meant something special.

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -16,13 +16,16 @@
...
 
+const _ = require('lodash');
 const semver = require('semver');
 const etcherLatestVersion = require('etcher-latest-version');
 
 module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, ManifestBindService, SettingsModel) {
 
   /**
-   * @summary Get the latest available Etcher version
+   * @summary Get the latest available Etcher version and memoize it;
```
jviotti: Ah, lets also keep the summary as it is now, and use `@description` for the rest:

```
@summary Get the latest available Etcher version
@description
foo bar
```

lib/gui/app.js
```diff
@@ -96,7 +96,7 @@ app.run((AnalyticsService, UpdateNotifierService, SelectionStateModel) => {
       if (!isLatestVersion && !SelectionStateModel.hasImage()) {
 
         AnalyticsService.logEvent('Notifying update');
-        UpdateNotifierService.notify();
+        UpdateNotifierService.notify().catch(ErrorService.reportException);
```
jviotti: Can we return this promise and put the `.catch` block in `UpdateNotifierService.isLatestVersion()`? `UpdateNotifierService.isLatestVersion()` might throw something in the future, and we will be already prepared for it!

```js
UpdateNotifierService.isLatestVersion().then((isLatestVersion) => {

  // In case the internet connection is not good and checking the
  // latest published version takes too long, only show notify
  // the user about the new version if he didn't start the flash
  // process (e.g: selected an image), otherwise such interruption
  // might be annoying.
  if (!isLatestVersion && !SelectionStateModel.hasImage()) {

    AnalyticsService.logEvent('Notifying update');
    return UpdateNotifierService.notify();
  }
```

Shou:
@jviotti here we go

lurch:
@jviotti @Shou Is there some flag / setting somewhere we can tweak (possibly even by editing the source code directly to point it at a 'fake' version URL) that we can use to test what the update-dialog looks like, without having to wait for the next update? ;-)

jviotti:
@lurch This is what I usually do (totally hacky):

```diff
diff --git a/lib/gui/app.js b/lib/gui/app.js
index b7077e2..7b73f9e 100644
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -82,6 +82,7 @@ app.run(() => {

 app.run((AnalyticsService, UpdateNotifierService, SelectionStateModel) => {
   AnalyticsService.logEvent('Application start');
+  UpdateNotifierService.notify();

   if (UpdateNotifierService.shouldCheckForUpdates() && !process.env.ETCHER_DISABLE_UPDATES) {
     AnalyticsService.logEvent('Checking for updates');
```

I force the update notifier to be shown by editing the source code a bit. I'm open to discuss better ways though.

lurch:
Hmm, given that we have a `ETCHER_DISABLE_UPDATES` env-var, I wonder if there could be an `ETCHER_PRETEND_THERES_AN_UPDATE` env-var?

jviotti:
I like the idea, however I don't like `ETCHER_PRETEND_THERES_AN_UPDATE` very much. So the update notifier dialog is supposed to be shown is the latest version (returned by `getLatestVersion()`) is different than the current one.

Under than assumption, we could do two things:

- Add an `ETCHER_LATEST_VERSION` variable. That variable will force `getLatestVersion()` to return such version instead of querying the S3 buckets, essentially mocking it.

- Add an `ETCHER_CURRENT_VERSION` that forces Etcher to think the current version is another one. The problem with this one is that we rely on `package.json` to determine the current version, and unifying and controlling access to that resources could be tricky.

I like the first one since there is a clear point of access to that information, and therefore we can mock very easily. What do you think?

lurch:
Yup, I'd go with your first option too. But I'd also give it a longer name, and make clear it's purely a debug/test variable name.
How would 'faking' a new version affect the URL it tries to download?

(in case it's not obvious: `ETCHER_PRETEND_THERES_AN_UPDATE` was intended to be tongue-in-cheek :stuck_out_tongue: )

lurch:
Hmmm, has this PR somehow overwritten the changes from #920 ? :frowning:

Shou:
@lurch oops, I think the link got removed when I rebased to master. Let me fix that.

jviotti:
Ugh, I didn't notice that. @Shou lets open another PR to fix that.

PR #930, opened 2016-12-03
chore: fix the bintray-debian publishing script

lurch:
This change prevents the script erroring with `BINTRAY_USER: unbound variable` when `$BINTRAY_USER` isn't defined, and instead displays the "Please define the following ..." message as expected.

### DIFF
--- a/scripts/publish/bintray-debian.sh
+++ b/scripts/publish/bintray-debian.sh
@@ -34,13 +34,15 @@ if [ "$#" -ne 1 ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jviotti:
Looks cool! I think you'll have to rebase this branch to get the semicolon build fix!

lurch:
Done :-)

(I hoped that simply asking Travis to restart the build would be enough to fix the problem, given that this PR doesn't touch `.travis.yml`, but unfortunately that wasn't the case; a rebase and force-push were required).

lurch:
> a rebase and force-push were required

...and I've just done the same on all PRs which were currently open but failing their CI tests, and a couple more now have passing CI tests :-)

PR #934, opened 2016-12-05
chore: move `bower.json` removal after popd in `dependencies-bower.sh`

jviotti:
The reason for this change is to make things more explicit.

See: https://github.com/resin-io/etcher/pull/923#discussion_r90571316

### DIFF
--- a/scripts/unix/dependencies-bower.sh
+++ b/scripts/unix/dependencies-bower.sh
@@ -59,8 +59,8 @@ if [ -n "$ARGV_PREFIX" ]; then
   cp "$PWD/bower.json" "$ARGV_PREFIX"
   pushd "$ARGV_PREFIX"
   bower install $INSTALL_OPTS
-  rm bower.json
   popd
+  rm "$ARGV_PREFIX/bower.json"
 else
   bower install $INSTALL_OPTS
 fi


lurch:
What makes this a `refactor` rather than a `chore` ? (as #933 was)

jviotti:
@lurch I honestly can't make my mind between these two. The "chore" commit type is supposed to describe "build process" changes, however a refactoring process taking place in the build system could be either a "chore" or a "refactor". Which one do you think it should have more precedence? I'm happy to pick one and stick with it.

lurch:
Well, you're obviously far more familiar with this scheme than I am, but I'd personally been assuming that _anything_ to do with the build/packaging/CI/etc. process is "chore".

jviotti:
@lurch Sounds like a good assumption. Lets go for it. I'll update this PR.

lurch:
Can you change the title to use 'move' instead of 'remove' please :)

jviotti:
Sure, done!

lurch:
Hmmm, I've just had a thought... rather than copying the file and then deleting it, would it be better to create a symlink instead, and then delete the symlink?

jviotti:
Hm, it sounds like a good idea. Let me see if bower is OK with symlinks here.

jviotti:
Yay, it works. I'll update the PR in a minute.

jviotti:
I've created another PR, since adding that change here would require us to update the PR title: https://github.com/resin-io/etcher/pull/935

PR #935, opened 2016-12-05
chore: symlink manifests in dependencies scripts

jviotti:
See: https://github.com/resin-io/etcher/pull/934#issuecomment-264862767

### DIFF
--- a/scripts/unix/dependencies-bower.sh
+++ b/scripts/unix/dependencies-bower.sh
@@ -56,7 +56,7 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
 fi
 
 if [ -n "$ARGV_PREFIX" ]; then
-  cp "$PWD/bower.json" "$ARGV_PREFIX"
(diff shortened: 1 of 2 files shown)


lurch:
Should do the same for `dependencies-npm.sh` too?

lurch:
...and this now needs to be rebased ;)

jviotti:
Yeah, I'll create another one for NPM.

> ...and this now needs to be rebased ;)

Why is that?

jviotti:
Done! https://github.com/resin-io/etcher/pull/937

lurch:
As they're both doing the same thing, wouldn't it be better for #937 to be merged into this PR ?

> Why is that?

I thought this might conflict with #934 , but apparently it doesn't.

jviotti:
> As they're both doing the same thing, wouldn't it be better for #937 to be merged into this PR ?

Sounds good, done! I'll close the other one.

PR #936, opened 2016-12-05
chore: persist temporary artifacts during AppImage builds

jviotti:
Currently, `electron-installer-appimage.sh` deletes temporary stuff it
needs as it goes (like the AppDir, AppImageAssistant, or even the
AppImage after creating a ZIP). Since this is not very Make-friendly,
the script has been split into the following smaller scripts that
perform a single task:

- `electron-create-appimage-linux.sh`
- `electron-create-appdir.sh`

Utilising them in different Makefile rules effectively results in Make
persisting the "temporary artifacts".

See: https://github.com/resin-io/etcher/pull/913#discussion_r90801230

### DIFF
--- a/docs/RUNNING-LOCALLY.md
+++ b/docs/RUNNING-LOCALLY.md
@@ -31,6 +31,7 @@ The following MinGW packages are required:
 - `msys-unzip`
 - `msys-wget`
 - `msys-bash`
+- `msys-coreutils`
 
 ### OS X
 
--- a/scripts/build/electron-create-appdir.sh
+++ b/scripts/build/electron-create-appdir.sh
@@ -18,13 +18,6 @@
 
 set -u
 set -e
-set -x
-
-OS=$(uname)
-if [[ "$OS" != "Linux" ]]; then
-  echo "This script is only meant to be run in GNU/Linux" 1>&2
-  exit 1
-fi
 
 ./scripts/build/check-dependency.sh upx
 
@@ -75,16 +68,11 @@ then
   usage
 fi
 
-TEMPORARY_DIRECTORY=$(mktemp -d)
-OUTPUT_FILENAME="$ARGV_APPLICATION_NAME-linux-$ARGV_ARCHITECTURE.AppImage"
-
 # Create AppDir
-APPDIR_PATH=$TEMPORARY_DIRECTORY/${OUTPUT_FILENAME%.*}.AppDir
+mkdir -p "$ARGV_OUTPUT"
 APPDIR_ICON_FILENAME=icon
-rm -rf "$APPDIR_PATH"
-mkdir -p "$APPDIR_PATH/usr/bin"
 
-cat >"$APPDIR_PATH/$ARGV_APPLICATION_NAME.desktop" <<EOF
+cat >"$ARGV_OUTPUT/$ARGV_APPLICATION_NAME.desktop" <<EOF
 [Desktop Entry]
 Name=$ARGV_APPLICATION_NAME
 Exec=$ARGV_BINARY.wrapper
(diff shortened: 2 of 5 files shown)


scripts/linux/electron-installer-appimage.sh
```diff
@@ -102,17 +101,14 @@ else
   exit 1
 fi
 
-if [ -z "$TMPDIR" ]; then
-  TMPDIR=$(mktemp -d)
-fi
-mkdir -p "$TMPDIR"
+APPIMAGE_TMPDIR=$(mktemp --dry-run -d -t "$ARGV_APPLICATION_NAME-linux-$ARGV_ARCHITECTURE-XXXX")
```
lurch: Creating randomly-named temporary directories is just as make-unfriendly as deleting things!

The idea is to have a (possibly hierarchical) set of consistent (and persistent) dependencies, so that if a target has already been fully-built, then running `make` again won't need to rebuild it; but if one of the build-steps went wrong or got interrupted half-way through, running `make` will just continue from where it left off (which obviously it can't do if directories have a different name each time they're created).

IMHO it probably makes sense to move the directory-creation steps into the `Makefile` too.

jviotti: Yeah, the thing is that this file downloads and creates various other files during its life, which are hard to declare in the Makefile (since they happen within the script). You can think of this script as having an "resource app asar input", and an "appimage output". In order to perform that task, a couple of tools have to be downloaded, and an AppDir needs to be created.

How do you think we can handle that while still being make-friendly?

Here's a suggestion: we have 4 external artifacts during this script:

- `AppRun`
- `desktopintegration`
- `AppImageAssistant`
- The `AppDir`

The `AppRun` and `desktopintegration` files are downloaded straight to the AppDir, and the `AppImageAssistant` is only required to convert the AppDir to an AppImage.

So we can do the following things:

- Create a new script whose mission is to just create an AppDir. This script will download `AppRun` and `desktopintegration` straight to the output directory (the AppDir), so no temporary stuff here.

- We create another script to convert an AppDir to an AppImage. By dividing things this way, the AppDir can now be declared in the Makefile, so thats something less to think about.

The only concern now is `AppImageAssistant`, which needs to be used by the script that goes from AppDir to AppImage. We can always create another script just to download it (and thus declare it in the Makefile), however that means that the Makefile will be responsible to know how to use it correctly, and I feel the Makefile shouldn't care about that.

Another alternative is to be able to pass the path to `AppImageAssistant` to the AppDir->AppImage script, however that feels weird as well.

What do you think?

jviotti: So as far as I can see (their Makefiles are huge, and didn't go through all of it), they have an `INSTALL` utility that gets called by scripts as they need, and the set the output location to the `dl` directory. I like that approach and I'm happy to go with it!

I'll create a generic script that performs downloads (and takes a MD5 argument to verify the files), tries to cache if necessary, and will make all scripts call it.

lurch: > I'll create a generic script that performs downloads (and takes a MD5 argument to verify the files), tries to cache if necessary, and will make all scripts call it.

Awesome. :tada: 

And yeah, the BuildRoot Makefiles are ridiculously huge and complicated - I think they even go as far as using data-files to then auto-generate (and execute) other makefiles - proper mind-bending stuff ;-)

Do you want me to approve this PR, or will it be superseded by your centralised-downloads PR?

jviotti: Haha, sounds fun! I'll update this PR with the changes in a minute (currently doing some testing to ensure I didn't break anything :) )

jviotti:
I've created a `download-tool.sh` script and I'm making use of it at `electron-installer-appimage.sh` for every download need, however notice that the AppDir and the AppImage are still created in the temporary directory. 

How do you want to proceed with that? Should we split the script as I recommended before?

lurch:
I wonder if the `download-tool.sh` should be in a PR of its own, with all the scripts that currently use `wget` updated to use that instead, and then have **this** PR rebased once `download-tool.sh` has hit `master`?

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,82 @@
...
+	test "$(md5sum $file | cut -d ' ' -f1)" = "$hash"
+}
+
+if [ -e "$ARGV_OUTPUT" ] && md5sum_matches "$ARGV_OUTPUT" "$ARGV_MD5_HASH"; then
+  echo "Re-using from cache"
+  exit 0
+fi
+
+wget --no-check-certificate "$ARGV_URL" -O "$ARGV_OUTPUT"
+chmod +x "$ARGV_OUTPUT"
```
lurch: IMHO this `chmod` doesn't belong in this script (since it may well be downloading things that aren't executable), but in the scripts that call this script.

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,82 @@
...
+
+if ! md5sum_matches "$ARGV_OUTPUT" "$ARGV_MD5_HASH"; then
+  echo "Checksum mismatch" 1>&2
+  rm -f "$ARGV_OUTPUT"
+  exit 1
```
lurch: Rather than just exiting straight away, do you think it might be worth re-trying the download?

scripts/unix/download-tool.sh
lurch: (or maybe there could be an extra optional flag to this script, which could be used to signify when the downloaded files should be made executable)

jviotti:
Can we restart the `download-tool.sh` review in https://github.com/resin-io/etcher/pull/957? I've added an option to set execution permissions btw, and also switched to sha256 for the checksums (I'll update this afterwards).

jviotti:
@lurch I've split the AppImage generation script into one that creates an AppDir, one that converts the AppDir to an AppImage, and one that packages the AppImage as a ZIP.

This means that the whole temporary file/directory situation has been reduced to the fact that the AppDir to AppImage script requires AppImageAssistant, which is downloaded to the "download directory" specified as a command line option.

I decided to go for an explicit command line option instead of an environment variable, since it will automatically be documented and enforced by our command line options mechanism, and prevent us from having to write logic to re-use a certain environment variable, but create a temporary one if it doesn't exist, etc.

scripts/build/electron-create-appdir.sh
```diff
@@ -105,65 +88,45 @@ Icon=$APPDIR_ICON_FILENAME
...
-# Generate AppImage
-mkdir -p "$(dirname "$ARGV_OUTPUT")"
-rm -f "$ARGV_OUTPUT"
-
 APPIMAGES_TAG=6
-APPIMAGES_GITHUB_RELEASE_BASE_URL=https://github.com/probonopd/AppImageKit/releases/download/$APPIMAGES_TAG
-APPIMAGEASSISTANT_PATH=$TMPDIR/AppImageAssistant.AppImage
 
 ./scripts/build/download-tool.sh -x \
   -u "https://raw.githubusercontent.com/probonopd/AppImageKit/$APPIMAGES_TAG/desktopintegration" \
```
lurch: To better match the command later in this script, I wonder if it'd be worth pulling out:
```bash
APPIMAGES_GITHUB_RAW_BASE_URL=https://raw.githubusercontent.com/probonopd/AppImageKit
```

jviotti: Makes sense, done!

scripts/build/electron-create-appdir.sh
```diff
@@ -27,14 +26,7 @@ function check_dep() {
   fi
 }
 
-OS=$(uname)
-if [[ "$OS" != "Linux" ]]; then
-  echo "This script is only meant to be run in GNU/Linux" 1>&2
-  exit 1
-fi
```
jviotti: It works fine! UPX detects a linux/ElfAMD format and does the right thing :)

docs/RUNNING-LOCALLY.md
```diff
@@ -31,6 +31,7 @@ The following MinGW packages are required:
 - `msys-unzip`
 - `msys-wget`
 - `msys-bash`
+- `msys-coreutils`
```
lurch: Should this change have been in a different PR? Doesn't really matter though.

WasabiFan:
@lurch BTW, the idea of a GitHub review is that you group all your comments _in one review_ and then mark it with a conclusion. When you're reviewing, instead of hitting the "submit review" button after every comment, you can add all of them to a single review and then conclude your review with a "request changes" or "approve" label. Right now, there's a `lurch reviewed 23 minutes ago` label above each comment because each one is its own review.

lurch:
I'm not specifically doing mini-reviews, I'm just adding comments. Dunno why GitHub is marking them as "reviews".

@jviotti What do you think to @WasabiFan 's suggestion?

WasabiFan:
It marks them as reviews because that's what you're doing -- you are reviewing the code. You aren't, however, actually using the Review functionality in its proper form. The expectation is that all PR reviews will now be conducted as actual Reviews to facilitate a more streamlined workflow.

jviotti:
@lurch Yeah, what @WasabiFan said is spot on, and its what I'm doing at the moment: start a "review" when posting the first message, write everything else I need, and then "publish review". The difference between doing a "review" and simply posting comments is that in the first case, the comments need to be formally "addressed" before we can merge.

docs/RUNNING-LOCALLY.md
jviotti: `msys-coreutils` include the `zip` tool, which we're using in this PR. Of course the file where that is used is Linux specific, but since we proved that we can actually generate the AppDir from any OS (including Windows), I thought it was a good idea to include here anyway.

WasabiFan:
You can also (as of a few days ago) request reviews from specific people near where you assign labels and such; it then tracks reviews that they submit as a unit instead of individual comments.

jviotti:
@lurch Is this good to go (unless I'm missing certain messages that got buried by GitHub)?

PR #939, opened 2016-12-05
chore: make code-signing optional in OS X

jviotti:
Code-signing in OS X now relies on the presence of a
`CODE_SIGN_IDENTITY` variable. If it doesn't exist, all the builds will
gracefully complete without code-signing.

As a consequence of this change, the `electron-installer-dmg.sh` script
has been divided into `electron-create-readwrite-dmg.sh`,
`electron-create-readwrite-dmg.sh`, and `electron-sign-dmg.sh`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@
 # Application configuration
 # ---------------------------------------------------------------------
 
-SIGN_IDENTITY_OSX = Developer ID Application: Rulemotion Ltd (66H43P8FRG)
 ELECTRON_VERSION = $(shell node -e "console.log(require('./package.json').devDependencies['electron-prebuilt'])")
 APPLICATION_NAME = $(shell node -e "console.log(require('./package.json').displayName)")
 APPLICATION_DESCRIPTION=$(shell node -e "console.log(require('./package.json').description)")
@@ -70,6 +69,16 @@ endif
 TARGET_ARCH = $(HOST_ARCH)
 
 # ---------------------------------------------------------------------
+# Code signing
+# ---------------------------------------------------------------------
+
+ifeq ($(TARGET_PLATFORM),darwin)
+ifndef CODE_SIGN_IDENTITY
+$(warning No code-sign identity found (CODE_SIGN_IDENTITY is not set))
+endif
+endif
+
+# ---------------------------------------------------------------------
 # Extra variables
 # ---------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -70,6 +69,16 @@ endif
 TARGET_ARCH = $(HOST_ARCH)
 
 # ---------------------------------------------------------------------
+# Code signing
+# ---------------------------------------------------------------------
+
+ifeq ($(TARGET_PLATFORM),darwin)
+ifndef CODESIGN_IDENTITY
+$(warning No code-sign identity found (CODESIGN_IDENTITY))
```
lurch: Should this warning end with something like `(\$CODESIGN_IDENTITY isn't set)`. I'm not sure how useful `(CODESIGN_IDENTITY)` is supposed to be? ;)

P.S. Just noticed that CODESIGN can either be read as "code-sign" or "co-design" ;-)

scripts/darwin/electron-installer-dmg.sh
```diff
@@ -160,7 +159,9 @@ sync
 afsctool -ci -9 "$VOLUME_APPLICATION"
 
 # TODO: this should be decoupled from this script
-./scripts/darwin/electron-sign-app.sh -a "$VOLUME_APPLICATION" -i "$ARGV_IDENTITY"
+if [ -n "$ARGV_IDENTITY" ]; then
+  ./scripts/darwin/electron-sign-app.sh -a "$VOLUME_APPLICATION" -i "$ARGV_IDENTITY"
+fi
```
jviotti: Yeah, makes sense. Done!

Makefile
jviotti: Done! I've clarified the message and renamed `CODESIGN_IDENTITY` to `CODE_SIGN_IDENTITY`.

Makefile
```diff
@@ -134,17 +143,27 @@ endif
...
 	release/$(APPLICATION_NAME)-darwin-$(TARGET_ARCH)
+ifdef CODESIGN_IDENTITY
 	./scripts/darwin/electron-installer-dmg.sh -p $< -o $@ \
 		-n "$(APPLICATION_NAME)" \
 		-v "$(APPLICATION_VERSION)" \
-		-d "$(SIGN_IDENTITY_OSX)" \
+		-d "$(CODESIGN_IDENTITY)" \
 		-i assets/icon.icns \
 		-b assets/osx/installer.png
+else
```
lurch: Hmm, what happened to the big comment I made here? Did it show up @jviotti ?
Did I stupidly forget to actually click on the "Comment" button? :confused:

jviotti: @lurch I don't know. I thought your addressed this in https://github.com/resin-io/etcher/pull/939#discussion_r90989013

lurch: I wrote a big comment about three different ways of tackling the problem... I spent so much time flipping backwards and forwards between the "Write" and "Preview" tabs, I guess I must have forgotten to actually post the comment - Doh! :slightly_frowning_face:

lurch: Yep :)

PR #940, opened 2016-12-05
chore: group build scripts inside `scripts/build`

jviotti:
This makes it easier to have re-usable functionality that we can
call/source from all scripts, like logic to determine the current
operating system, or to check for a dependency in the `$PATH`.


lurch:
Should the .bat scripts from the windows directory also be moved into this common directory?

jviotti:
> Should the .bat scripts from the windows directory also be moved into this common directory?

The plan is to get rid of all the `.bat` scripts and require users that want to build Etcher to install MinGW. The advantages are:

- MinGW (as opposed to something like Cygwin), makes all the UNIX commands available from `cmd.exe`, so we can call it from the "Visual Studio" console.

- Allows us to re-use the bash scripts, which is a big deal in terms of re-usability.

- Allows us to ditch Batch, which has several limitations. The one that annoy me the most is that there is nothing similar to `set -e`, to make the script exit if a command fails, so the "best practice" seems to be to spam the whole thing with error checks. 

- Its available out of the box in Appveyor.

I'm ignoring `.bat` files in this PR, since I'll delete them completely once I get the Windows build up and running on bash.

lurch:
> The plan is to get rid of all the .bat scripts and require users that want to build Etcher to install MinGW.

Sounds awesome, as long as installing MinGW isn't too much of a big download & install burden?

OffTopic: Would it be possible to extend the MinGW approach, and use that instead of (or as an alternative to) Visual Studio? `RUNNING-LOCALLY.md` says that "Visual Studio Community 2013" is required, but I can't actually find the download link for that anywhere on the Microsoft site? Only "Visual Studio Community 2015"...

lurch:
P.S. This branch now has conflicts.

jviotti:
> Sounds awesome, as long as installing MinGW isn't too much of a big download & install burden?

I think its fine. A lot of developers running Windows have it installed already, and even if you have to, it takes 1/10 the time it takes to install Visual Studio :)

> OffTopic: Would it be possible to extend the MinGW approach, and use that instead of (or as an alternative to) Visual Studio? 

I don't think we'll be able to get rid of Visual Studio any time soon. Its a dependency of gyp, the build system shipped by npm to install native NodeJS modules.

> RUNNING-LOCALLY.md says that "Visual Studio Community 2013" is required, but I can't actually find the download link for that anywhere on the Microsoft site? Only "Visual Studio Community 2015"...

Oh, weird. It seems that they are suddenly redirecting that link to VS2015. I'll find a direct one and update the guide.

jviotti:
I resolved the conflicts!

lurch:
> I resolved the conflicts!

Looks good, but I suspect this will need rebasing again after #939 gets merged ;-)

jviotti:
Yeah, I was expecting this when a PRed so many things at once, but git resolves renames very gracefully, so its not a big deal (luckily)!

lurch:
For those scripts that are still platform-specific (e.g. `scripts/build/electron-sign-app.sh`) I wonder if it might make sense to add the platform into the filename (e.g. `scripts/build/electron-sign-app-darwin.sh` or `scripts/build/electron-sign-darwin-app.sh`) ?

jviotti:
Yeah, makes sense. What do you prefer: prefixing the os in the filename, adding it as a suffix, or something else?

lurch:
Thinking about it, I think my preference would be to add the OS as a suffix. So we'd have:
```
scripts/build/electron-sign-app-darwin.sh
scripts/build/electron-sign-exe-windows.sh
scripts/build/electron-installer-debian-linux.sh
```
etc. (which also fits in well with the `electron-configure-package-linux.sh` filenames you already have!)

jviotti:
Done!

lurch:
Can you try restarting the AppVeyor builds?

jviotti:
@lurch What's your resin email? I'll make sure you have access to the CI servers.

jviotti:
BTW, the Appveyor errors are legitimate ones after the rebase. I'll fix in a second...

scripts/build/windows.bat
```diff
@@ -193,13 +193,13 @@ if not "%arch%"=="%electron_arch%" (
...
 
-call bash.exe scripts\unix\electron-create-asar.sh^
+call bash.exe scripts\build\electron-create-asar.sh^
  -d %package_output%\resources\app^
  -o %package_output%\resources\app.asar
 
 call rimraf %package_output%\resources\app
 
-call bash.exe scripts\windows\electron-sign-exe.sh^
+call bash.exe scripts\build\electron-sign-exe-win32.sh^
```
lurch: It's a good thing that I know this script won't be hanging around for very long, otherwise I'd have asked for all these `win32` strings to be replaced with variables ;-D

PR #941, opened 2016-12-05
chore: use `pushd`/`popd` instead of `--prefix` on `dependencies-npm.sh`

jviotti:
We currently use the `--prefix` option in `dependencies-npm.sh` to
install dependencies in another directory, however that function causes
ENOSELF to be thrown on Windows under Msys. As a solution, we take the
same `pushd`/`popd` approach as we do in `dependencies-bower.sh`.

### DIFF
--- a/scripts/unix/dependencies-npm.sh
+++ b/scripts/unix/dependencies-npm.sh
@@ -98,23 +98,18 @@ fi
 
 if [ -n "$ARGV_PREFIX" ]; then
   mkdir -p "$ARGV_PREFIX"
-  INSTALL_OPTS="$INSTALL_OPTS --prefix=$ARGV_PREFIX"
   ln -s "$PWD/package.json" "$ARGV_PREFIX/package.json"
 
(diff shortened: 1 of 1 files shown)


scripts/unix/dependencies-npm.sh
```diff
@@ -98,23 +98,18 @@ fi
...
-npm install $INSTALL_OPTS
+  pushd "$ARGV_PREFIX"
+  npm install $INSTALL_OPTS
+  popd
 
-if [ -n "$ARGV_PREFIX" ]; then
   rm -f "$ARGV_PREFIX/package.json"
   rm -f "$ARGV_PREFIX/npm-shrinkwrap.json"
-
-  # Using `--prefix` might cause npm to create an empty `etc` directory
```
lurch: How come this piece of code has suddenly been removed?

jviotti: We're not longer using `--prefix`, which was the reason this "etc" directory gets introduced.

PR #942, opened 2016-12-05
chore: fix certificate verification error with wget on MinGW

jviotti:
The `electron-download-package.sh` script yields the following error on
MinGW:

```
cannot verify github.com's certificate
```


lurch:
Is this because the wget used by MinGW is 'old' and doesn't have up-to-date enough ca-certs?

jviotti:
> Is this because the wget used by MinGW is 'old' and doesn't have up-to-date enough ca-certs?

Yeah, it seems to be something like that. It seems that you can fix the error by refreshing the local certificates, or something like that, but I guess that's out of the scope of these scripts.

PR #944, opened 2016-12-05
chore: don't include *.dll files inside the asar archive 

jviotti:
This change is needed to make `electron-create-asar.sh` output valid
asar packages on Windows (Msys).


lurch:
I'd change the title to something like "chore: don't include *.dll files inside the asar archive" ?

IMHO "upack *.dll" makes it sound like you're (somehow) extracting the contents of .dll files ;-)

jviotti:
@lurch Makes sense. I chose "unpack" to be consistent with the asar command line tool option, but I agree it might not make much sense to someone not familiar with it.

PR #946, opened 2016-12-06
chore: move directory creation responsibility to Makefile

jviotti:
This commit adds new rules to create directories and sets order only
pre-requisites to the existing rules as a way to centralise directory
creation, which is currently scattered around all the build scripts.

See: https://github.com/resin-io/etcher/pull/923#discussion_r90881453


Makefile
```diff
@@ -89,36 +89,48 @@ ifeq ($(TARGET_ARCH),x64)
...
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json
+release:
+	mkdir -p $@
+
+release/.tmp: | release
+	mkdir -p $@
+
+release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: | release
+	mkdir -p $@
```
lurch: No need for the `-p` flags, because the Makefile dependency rules mean the directories will always get created in the right order :-)

Makefile
```diff
@@ -89,36 +89,48 @@ ifeq ($(TARGET_ARCH),x64)
...
+
+release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json \
+	| release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
 	./scripts/unix/dependencies-npm.sh -p \
 		-r "$(TARGET_ARCH)" \
 		-v "$(ELECTRON_VERSION)" \
 		-x $(dir $@) \
```
lurch: If you wanted to, you could now replace `$(dir $@)` with `$|` ;)

Makefile
```diff
@@ -147,40 +160,45 @@ ifeq ($(TARGET_PLATFORM),linux)
...
-	release/$(APPLICATION_NAME)-darwin-$(TARGET_ARCH)
+	release/$(APPLICATION_NAME)-darwin-$(TARGET_ARCH) \
+	| release
 	./scripts/darwin/electron-create-readwrite-dmg.sh -p $< -o $@ \
 		-n "$(APPLICATION_NAME)" \
 		-i assets/icon.icns \
 		-b assets/osx/installer.png
 
+release/out: | release
+	mkdir -p $@
```
lurch: *Might* make sense to move this up with all the other `mkdir` commands?

Makefile
```diff
@@ -147,40 +160,45 @@ ifeq ($(TARGET_PLATFORM),linux)
...
 
 release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).dmg: \
-	release/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-rw.dmg
+	release/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-rw.dmg \
+	| release/out
 ifdef CODE_SIGN_IDENTITY
 	./scripts/darwin/electron-sign-dmg.sh \
-		-n "$(APPLICATION_NAME)" \
-		-d $< \
-		-i "$(CODE_SIGN_IDENTITY)"
```
lurch: Umm, mistake?

Makefile
```diff
@@ -89,36 +89,48 @@ ifeq ($(TARGET_ARCH),x64)
 	TARGET_ARCH_DEBIAN = amd64
 endif
 
-TEMPORARY_DIRECTORY = release/.tmp
```
lurch: IMHO there's no reason not to keep referring to this as `$(TEMPORARY_DIRECTORY)` throughout the rest of the `Makefile`?

scripts/unix/electron-create-resources-app.sh
```diff
@@ -50,7 +50,6 @@ if [ -z "$ARGV_SOURCE_DIRECTORY" ] ||
 fi
 
 mkdir -p "$ARGV_OUTPUT"
-
```
lurch: Deleted the wrong line??

scripts/linux/electron-installer-appimage.sh
```diff
@@ -105,7 +105,6 @@ fi
 if [ -z "$TMPDIR" ]; then
   TMPDIR=$(mktemp -d)
```
lurch: Probably need to change this script to pass in `TEMPORARY_DIRECTORY` and then remove this if-block too?

Makefile
jviotti: Very good point! I'll remove them!

Makefile
jviotti: Oh, very nice! I didn't know that was possible. Updating...

Makefile
jviotti: Sure. The intention was to have it near the rules that actually use `release/out`, but I like the suggestion.

Makefile
jviotti: Damn. This is why you shouldn't resolve conflicts late at night :)

scripts/unix/electron-create-resources-app.sh
jviotti: Fixed!

scripts/linux/electron-installer-appimage.sh
jviotti: The intention is to re-use `$TMPDIR` from the environment if its there, otherwise use `mktemp -d`. We are manually passing a custom `TMPDIR` from the Makefile, but this script might get ran in a completely standalone fashion, in which case we would have to document that this script depends on a certain environment setting.

Makefile
jviotti: How do you feel about making the whole `release` prefixes configurable by variables? I imagine we could ship this Makefile for other users to `include`, in which case we could make the "build" directory configurable.

jviotti: This is how it would look like:

```diff
diff --git a/Makefile b/Makefile
index 3649137..6998c78 100644
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,12 @@
 # ---------------------------------------------------------------------
+# Build configuration
+# ---------------------------------------------------------------------
+
+BUILD_DIRECTORY = release
+BUILD_TEMPORARY_DIRECTORY = $(BUILD_DIRECTORY)/.tmp
+BUILD_OUT_DIRECTORY = $(BUILD_DIRECTORY)/out
+
+# ---------------------------------------------------------------------
 # Application configuration
 # ---------------------------------------------------------------------
 
@@ -96,20 +104,20 @@ APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 # Rules
 # ---------------------------------------------------------------------
 
-release:
+$(BUILD_DIRECTORY):
 	mkdir $@
 
-release/.tmp: | release
+$(BUILD_TEMPORARY_DIRECTORY): | $(BUILD_DIRECTORY)
 	mkdir $@
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: | release
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: | $(BUILD_DIRECTORY)
 	mkdir $@
 
-release/out: | release
+$(BUILD_OUT_DIRECTORY): | $(BUILD_DIRECTORY)
 	mkdir $@
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json \
-	| release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json \
+	| $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
 	./scripts/unix/dependencies-npm.sh -p \
 		-r "$(TARGET_ARCH)" \
 		-v "$(ELECTRON_VERSION)" \
@@ -117,33 +125,34 @@ release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: pa
 		-t electron \
 		-s "$(TARGET_PLATFORM)"
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components: bower.json \
-	| release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components: bower.json \
+	| $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
 	./scripts/unix/dependencies-bower.sh -p -x $(dir $@)
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app: \
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules \
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components \
-	| release
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app: \
+	$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules \
+	$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components \
+	| $(BUILD_DIRECTORY)
 	./scripts/unix/electron-create-resources-app.sh -s . -f "$(APPLICATION_FILES)" -o $@
 	$(foreach prerequisite,$^,cp -rf $(prerequisite) $@;)
 
-release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app.asar: \
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app \
-	| release
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app.asar: \
+	$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app \
+	| $(BUILD_DIRECTORY)
 	./scripts/unix/electron-create-asar.sh -d $< -o $@
 
-release/electron-$(ELECTRON_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip: | release
+$(BUILD_DIRECTORY)/electron-$(ELECTRON_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip: \
+	| $(BUILD_DIRECTORY)
 	./scripts/unix/electron-download-package.sh \
 		-r "$(TARGET_ARCH)" \
 		-v "$(ELECTRON_VERSION)" \
 		-s "$(TARGET_PLATFORM)" \
 		-o $@
 
-release/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH): \
-	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app.asar \
-	release/electron-$(ELECTRON_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip \
-	| release
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH): \
+	$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app.asar \
+	$(BUILD_DIRECTORY)/electron-$(ELECTRON_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip \
+	| $(BUILD_DIRECTORY)
 ifeq ($(TARGET_PLATFORM),darwin)
 	./scripts/darwin/electron-configure-package-darwin.sh -p $(word 2,$^) -a $< \
 		-n "$(APPLICATION_NAME)" \
@@ -162,25 +171,25 @@ ifeq ($(TARGET_PLATFORM),linux)
 		-o $@
 endif
 
-release/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-rw.dmg: \
-	release/$(APPLICATION_NAME)-darwin-$(TARGET_ARCH) \
-	| release
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-rw.dmg: \
+	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-darwin-$(TARGET_ARCH) \
+	| $(BUILD_DIRECTORY)
 	./scripts/darwin/electron-create-readwrite-dmg.sh -p $< -o $@ \
 		-n "$(APPLICATION_NAME)" \
 		-i assets/icon.icns \
 		-b assets/osx/installer.png
 
-release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).zip: \
-	release/$(APPLICATION_NAME)-darwin-$(TARGET_ARCH) \
-	| release/out
+$(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).zip: \
+	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-darwin-$(TARGET_ARCH) \
+	| $(BUILD_OUT_DIRECTORY)
 ifdef CODE_SIGN_IDENTITY
 	./scripts/darwin/electron-sign-app.sh -a $</$(APPLICATION_NAME).app -i "$(CODE_SIGN_IDENTITY)"
 endif
 	./scripts/darwin/electron-installer-app-zip.sh -a $</$(APPLICATION_NAME).app -o $@
 
-release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).dmg: \
-	release/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-rw.dmg \
-	| release/out
+$(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).dmg: \
+	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-rw.dmg \
+	| $(BUILD_OUT_DIRECTORY)
 ifdef CODE_SIGN_IDENTITY
 	./scripts/darwin/electron-sign-dmg.sh \
 		-n "$(APPLICATION_NAME)" \
@@ -189,19 +198,19 @@ ifdef CODE_SIGN_IDENTITY
 endif
 	./scripts/darwin/electron-create-readonly-dmg.sh -d $< -o $@
 
-release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-linux-$(TARGET_ARCH).zip: \
-	release/$(APPLICATION_NAME)-linux-$(TARGET_ARCH) \
-	| release/out release/.tmp
-	TMPDIR=release/.tmp ./scripts/linux/electron-installer-appimage.sh -p $< -o $@ \
+$(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-linux-$(TARGET_ARCH).zip: \
+	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-linux-$(TARGET_ARCH) \
+	| $(BUILD_OUT_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
+	TMPDIR=$(BUILD_TEMPORARY_DIRECTORY) ./scripts/linux/electron-installer-appimage.sh -p $< -o $@ \
 		-n "$(APPLICATION_NAME)" \
 		-d "$(APPLICATION_DESCRIPTION)" \
 		-r "$(TARGET_ARCH)" \
 		-b "$(APPLICATION_NAME_LOWERCASE)" \
 		-i assets/icon.png
 
-release/out/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb: \
-	release/$(APPLICATION_NAME)-linux-$(TARGET_ARCH) \
-	| release/out
+$(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb: \
+	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-linux-$(TARGET_ARCH) \
+	| $(BUILD_OUT_DIRECTORY)
 	./scripts/linux/electron-installer-debian.sh -p $< -r "$(TARGET_ARCH)" \
 		-c scripts/build/debian/config.json \
 		-o $(dir $@)
@@ -234,12 +243,12 @@ electron-develop:
 	./scripts/unix/dependencies-bower.sh
 
 ifeq ($(TARGET_PLATFORM),darwin)
-electron-installer-app-zip: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
-electron-installer-dmg: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg
+electron-installer-app-zip: $(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
+electron-installer-dmg: $(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg
 endif
 ifeq ($(TARGET_PLATFORM),linux)
-electron-installer-appimage: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
-electron-installer-debian: release/out/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb
+electron-installer-appimage: $(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
+electron-installer-debian: $(BUILD_OUT_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb
 endif
 
 info:
@@ -249,4 +258,4 @@ info:
 	@echo "Target arch     : $(TARGET_ARCH)"
 
 clean:
-	rm -rf release
+	rm -rf $(BUILD_DIRECTORY)
```

I see a couple of flaws already:

- The directory creation rules hardcode a certain relationship. For example:

```make
$(BUILD_TEMPORARY_DIRECTORY): | $(BUILD_DIRECTORY)
	mkdir $@
```

This implies that `$(BUILD_TEMPORARY_DIRECTORY)` is inside `$(BUILD_DIRECTORY)`, but of course the user could configure this differently, in which case the rule stops making sense. This could be fixed by not using any `|` pre-requisites on the build directories.

- A certain "build directory" might contain any level of subdirectories. For example, I could declare that the "out directory" is `$(BUILD_DIRECTORY)/foo/bar/baz/out`, in which case our rules would fail because they don't include `mkdir -p`.

What do you think? I like how this reduces duplication, but I wonder if its the right thing to do.

scripts/linux/electron-installer-appimage.sh
lurch: Ah! I'd missed the bit where you'd set `TMPDIR=` in the Makefile rule :)
IMHO it'd still be worth renaming this variable from `TMPDIR` to `TEMPORARY_DIRECTORY` though (both in this script and in the Makefile) because it probably makes sense to do things in a subdirectory of `$TMPDIR` (if it's set in the environment) rather than in`$TMPDIR` itself, and "a subdirectory of `$TMPDIR` (if it's set in the environment)" is exactly what `mktemp -d` gives us.

jviotti: Sounds good to me. Do you think we should still rename this in the Makefile, though? Using a custom environment variable means that we have to document it, and we still have to include an `if` conditional to ensure it exists.

If we only rename on the script, then it means we can simply do:

```sh
TEMPORARY_DIR=$(mktemp -d)
```

and get rid of the conditional. Since we still pass `TMPDIR` in the Makefile, `mktemp` will create a directory inside it, and get still get exactly what we wanted.

Makefile
lurch: Obviously you need to be careful (and e.g. use the `word` macro) if there's more than one order-only pre-requisite.

scripts/linux/electron-installer-appimage.sh
lurch: Yeah, that's a nice idea :)

lurch: ...although it _does_ mean that we're back to the situation with mktemp creating randomly-named directories, which I commented on in another issue as being make-unfriendly... :-S

Makefile

scripts/linux/electron-installer-appimage.sh
jviotti: Done!

jviotti:
I've just fixed the conflicts. @lurch Is this good to merge?

Makefile
```diff
@@ -1,4 +1,17 @@
 # ---------------------------------------------------------------------
+# Build configuration
+# ---------------------------------------------------------------------
+
+BUILD_DIRECTORY ?= release
+
+ifndef BUILD_DIRECTORY
```
lurch: This doesn't make sense - `BUILD_DIRECTORY` will **always** be defined.
I guess we could check that `$(dirname $(BUILD_DIRECTORY))` is actually a directory that exists?
FYI: `$(dirname release)` == `.`

Makefile
```diff
@@ -1,4 +1,17 @@
...
+	$(error You should define a build directory (set BUILD_DIRECTORY))
+endif
+
+BUILD_TEMPORARY_DIRECTORY = $(BUILD_DIRECTORY)/.tmp
+BUILD_OUT_DIRECTORY = $(BUILD_DIRECTORY)/out
```
lurch: IMHO `BUILD_OUTPUT_DIRECTORY` or even `BUILD_INSTALLER_DIRECTORY` might be better names here?

To me, BUILD_OUT_DIRECTORY sounds like 'buildout directory'.

jviotti: I agree. I've renamed it to `BUILD_OUTPUT_DIRECTORY`.

Makefile
jviotti: So by using `$(dir $(BUILD_DIRECTORY)) == ./` we ensure its not an absolute path, which doesn't mean anything, since the directory name might or might not exist, and there can still be any level of depth (`foo/bar/baz/release`).

What about changing only the rule that builds `$(BUILD_DIRECTORY)` to `mkdir -p`?

lurch: Nah, that's not quite what I meant (sorry for the confusion). I meant we should check that `$(dir $(BUILD_DIRECTORY))` is a directory that exists, so that if the user does e.g. `BUILD_DIRECTORY=/tmp/some/path make my-target` we should error if `/tmp/some` isn't a pre-existing directory.
Which seems like a reasonable thing to do IMHO?
(of course if `/tmp/some/path` already exists, that's just fine)

Hmmm, interesting thought: `BUILD_DIRECTORY=/tmp make clean` will try to delete all of `/tmp` which probably isn't what we want ;-)  The best fix for that might be to have `make clean` delete specifically-named files / dirs, rather than the whole `$(BUILD_DIRECTORY)`.

jviotti: Oh, I see now. I'll update the PR in a minute.

> Hmmm, interesting thought: BUILD_DIRECTORY=/tmp make clean will try to delete all of /tmp which probably isn't what we want ;-) The best fix for that might be to have make clean delete specifically-named files / dirs, rather than the whole $(BUILD_DIRECTORY).

Check my latest changes. I've updated `clean` to only remove things inside `$(BUILD_DIRECTORY)`, but not the directory itself.

Makefile
```diff
@@ -1,4 +1,18 @@
 # ---------------------------------------------------------------------
+# Build configuration
+# ---------------------------------------------------------------------
+
+BUILD_DIRECTORY ?= release
+
+# See http://stackoverflow.com/a/20763842/1641422
+ifeq ($(wildcard $(dir $(BUILD_DIRECTORY)).),)
+$(error $(dir $(BUILD_DIRECTORY)) does not exist)
```
lurch: These two lines _might_ be slightly more readable if `$(dir $(BUILD_DIRECTORY))` is factored out to e.g. a separate `BUILD_DIRECTORY_PARENT` variable?

Makefile
lurch: Hmmm, that just changes `make clean` to delete everything **in** the BUILD_DIRECTORY, rather than deleting the BUILD_DIRECTORY itself. I'm not sure that's much better :-/

IMHO it _should_ be valid to do `BUILD_DIRECTORY=/tmp make my-target`, but `BUILD_DIRECTORY=/tmp make clean` _shouldn't_ do `rm -rf /tmp/*` (it deletes more than it should), instead it should only delete files that we know the `Makefile` (and associated scripts) created. I'll have a think tomorrow (when I'm less tired) and see if I can come up with a nice solution.


Of course the alternative would be to do:
```Makefile
BUILD_DIRECTORY_PARENT ?= .
BUILD_DIRECTORY = $(BUILD_DIRECTORY_PARENT)/release
```
and then allow the user to (optionally) do `BUILD_DIRECTORY_PARENT=/tmp make my-target`, which means that everything would get built inside `/tmp/release`, and the 'clean' rule could stay as `rm -rf $(BUILD_DIRECTORY)`, but I'm not sure if that feels very nice...

lurch: Yeah, that sounds like a good solution. And obviously if `BUILD_DIRECTORY` ever gets mentioned in the documentation, we'd need to mention that there too. But for now just a comment in the Makefile above the BUILD_DIRECTORY declaration is probably enough.

PR #948, opened 2016-12-06
chore: require MinGW to build Etcher on Windows

jviotti:
Using MinGW allows us to re-use the build scripts we've been developing
for GNU/Linux and OS X, which are much more robust than their `*.bat`
counterparts (due to Batch limitations).

This commit does the following changes:

- Documents MinGW as a requirement in `RUNNING-LOCALLY.md`
- It setups MinGW in Appveyor CI
- Transforms `windows/sign.bat` into `windows/electron-sign-exe.sh`
- Re-uses `dependencies-npm.sh` and `dependencies-bower.sh` in
  `build/windows.bat`


WasabiFan:
Does this mean that one must download and install the whole MinGW package to build Etcher? If so, it would be optimal if you could streamline this to at least mention what specific components are required, and preferably deliver the executables without installation at all.

jviotti:
I've added a list of packages that we require at the moment (in the middle of the transition), but I'll update the list as we go. Luckily very few packages need to be installed, so I'll investigate if we can somehow ship the binaries with Etcher.

scripts/windows/electron-sign-exe.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
```
lurch: I was _going_ to say that `electron-sign-exe.sh` is probably too generic of a name, but then I noticed that we also have `electron-sign-app.sh` and `electron-sign-dmg.sh` so this is probably okay ;-)
(but see the comment I just added to the script-renaming PR)

scripts/windows/electron-sign-exe.sh
```diff
@@ -0,0 +1,81 @@
...
+
+function check_dep() {
+  if ! command -v $1 2>/dev/null 1>&2; then
+    echo "Dependency missing: $1" 1>&2
+    exit 1
+  fi
+}
+
+OS=$(uname -o 2>/dev/null || true)
+if [[ "$OS" != "Msys" ]]; then
```
lurch: If we're checking `uname -o` for "Msys" here, then it probably makes sense to make that same check in the `Makefile`?

scripts/windows/electron-sign-exe.sh
```diff
@@ -0,0 +1,81 @@
...
+TIMESTAMP_SERVER=http://timestamp.comodoca.com
+
+signtool sign \
+  /t "$TIMESTAMP_SERVER" \
+  /d "$ARGV_SIGNATURE_DESCRIPTION" \
+  /f "$ARGV_CERTIFICATE_FILE" \
+  /p "$ARGV_CERTIFICATE_PASSWORD" \
+  "$ARGV_FILE"
+
+signtool verify /pa /v "$ARGV_FILE"
```
lurch: What happens if this fails? Just sets a bad exit code? (causing this script to fail, due to `set -e`)

lurch:
Unless they're very small I'd be :-1: on shipping the mingw binaries with Etcher, but having some kind of externally-hosted `etcher-mingw-dependencies.zip` file linked to from the instructions would IMHO be okay.

WasabiFan:
I'm thinking that you could host the executables (in whatever format you choose) in the Amazon S3 storage you already have. The build script could download it as needed.

scripts/windows/electron-sign-exe.sh
jviotti: Yeah, exactly.

scripts/windows/electron-sign-exe.sh
jviotti: What about creating a separate script that checks the current OS and re-using that in the Makefile and in every script that needs it?

lurch: Is that really worth the overhead of having scripts executing other scripts, when it's just 2 or 3 lines in the first place? If you're going down that route (which I'm not suggesting), you could also make `check_dep` etc. a script of its own.

lurch:
@WasabiFan I didn't suggest that because I didn't know if they'd need to be "installed" or anything first? If it's just a case of download-extract-execute then yeah it probably would make sense to have them auto-downloaded by the build script (just like the AppImage tool is).

Oh! Except of course one of the packages is `wget`, and we need to use `wget` to download the other files... ;-)

WasabiFan:
> Oh! Except of course one of the packages is wget, and we need to use wget to download the other files... ;-)

Nah, Windows has this covered with a much more _elegant_ solution.

```cmd
@powershell -NoProfile -ExecutionPolicy Bypass -Command "(New-Object System.Net.WebClient).DownloadString('http://my/path/to/files') > \my\local\file" 
```

I mean, who _wouldn't_ think of that?

WasabiFan:
More seriously though, PowerShell is the rightful parallel to the UNIX world's shell scripts -- in the next Windows 10 update, it will even become the default shell. If you guys are looking for elegant ways to write native Windows scripts, PowerShell is a great option.

lurch:
> I mean, who _wouldn't_ think of that?

Yeah, just rolls off the tongue :tongue: :laughing: 

Is PowerShell included pre-installed on all version of Windows on which Etcher is supported though :question:

WasabiFan:
> Is PowerShell included pre-installed on all version of Windows on which (building) Etcher is supported though ❓

It's definitely pre-installed on Windows 7 up, however I believe the version in Windows 7 is version 2 and each successive release (8, 8.1, 10 Anniversary Update) incremented the major version, so there might be incompatibilities. Of course, testing with Windows 7 is difficult, because the only people who still have it are people who either are required to use it by their employer or are avoiding all new updates since 2009 -- it's no longer sold to normal consumers (and rightfully so). The main thing to keep in mind is that the default `ExecutionPolicy` for PowerShell prevents you from running remotely-downloaded scripts, so you sometimes have to have the user change that security setting before using it.

lurch:
> Of course, testing with Windows 7 is difficult

The only Windows device that I have (an old Lenovo Thinkpad) is still on Windows 7 :smiley: :stuck_out_tongue:  (I keep it up to date, but I didn't want the "free upgrade" to Windows 10)

With the direction @jviotti is going (which I agree with), we're going to be using Makefiles and Bash as much as possible across all 3 platforms (for maximum code re-use), rather than using bash on Linux & OSX, and PowerShell on Windows. But if we can use PowerShell to "bootstrap" the process on Windows by downloading bash and wget, that might be useful :)

@jviotti Something else that's just occurred to me: it may already be on your radar, but if we're going to be using msys-unzip then I guess we can get rid of the 7zip dependency.

WasabiFan:
> With the direction @jviotti is going (which I agree with), we're going to be using Makefiles and Bash as much as possible across all 3 platforms (for maximum code re-use), rather than using bash on Linux & OSX, and PowerShell on Windows. But if we can use PowerShell to "bootstrap" the process on Windows by downloading bash and wget, that might be useful :)

Yeah, I think that's the best approach right now; use PowerShell as the Windows bootstrapper for whatever else you need to download in the background. It would be nice to use one of the existing build dependencies (e.g. Python) for all of it, but that conversation was had before and the decision was made to stick with Bash and an emulation layer. In the end, I'm just here to provide some user insight when I can and to generally make sure I understand Etcher -- I have no stake in the internals of the thing, so I have no reason to get involved in it :laughing:

> Something else that's just occurred to me: it may already be on your radar, but if we're going to be using msys-unzip then I guess we can get rid of the 7zip dependency.

🎉 🎈 🍾

lurch:
Your help and comments are very much appreciated @WasabiFan  :clap: :grinning:

jviotti:
Sounds good! I'm happy to dive into PowerShell as a bootstrapped for the whole process. Re-writing all the Windows scripts in PowerShell sounds appealing as well, but at least for the time being, being able to re-use all the bash scripts we've been working on would be a big deal for us. Windows 10 also has bash support (although unstable as far as I've heard of), so maybe in that version, we can make use of it.

> @jviotti Something else that's just occurred to me: it may already be on your radar, but if we're going to be using msys-unzip then I guess we can get rid of the 7zip dependency.

Yeah, I noticed that as well :) I'll update the dependencies list once I get to that point.

In any case, is this PR good to merge? I'll move forward with PowerShell for bootstrapping purposes later on.

PR #954, opened 2016-12-07
chore: add `help` Makefile target to show available targets

lurch:
Also sets `help` as the default target, so that typing just `make` displays
the list of possible targets. Also customises the list of targets based on
TARGET_PLATFORM (because you can't build e.g. `electron-installer-dmg` when
you're on Linux).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,14 +189,28 @@ release/out/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)
 # Phony targets
 # ---------------------------------------------------------------------
 
-.PHONY: \
+TARGETS = \
+	help \
 	info \
 	clean \
-	electron-develop \
+	electron-develop
+
+ifeq ($(TARGET_PLATFORM),darwin)
+electron-installer-app-zip: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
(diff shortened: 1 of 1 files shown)


jviotti:
Can we leave the `publish-*` targets outside of this PR, and create a new one for them?

Makefile
```diff
@@ -93,6 +93,9 @@ TEMPORARY_DIRECTORY = release/.tmp
 APPLICATION_NAME_LOWERCASE = $(shell echo $(APPLICATION_NAME) | tr A-Z a-z)
 APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 
+# The first (default) rule
+help:
```
jviotti: Can we use `.DEFAULT_GOAL` instead (https://www.gnu.org/software/make/manual/html_node/Special-Variables.html#Special-Variables)?

lurch: Good catch. Done.

lurch:
> Can we leave the publish-* targets outside of this PR, and create a new one for them?

Sure. Would you like me to PR against current master, or is it okay if I wait for this PR to get merged to master first?

jviotti:
As you wish. If you're willing to wait for a bit, then you'll probably get away without any conflicts.

jviotti:
Lets change the commit and PR title here. "chore: adjust Makefile targets" doesn't say much. Maybe something like `chore: add help Makefile rule that shows available targets` or something like that.

jviotti:
Actually, I can edit the title myself when I merge. Let me know! (I can edit myself to a title you like, or wait for you to do it)

lurch:
Done :)

PR #955, opened 2016-12-07
chore: add publishing Makefile targets

lurch:
Currently `publish-aws` and `publish-debian` for Linux, and `publish-aws`
for OSX. Like all phony targets, these get displayed by `make help`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,13 @@ APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 # Rules
 # ---------------------------------------------------------------------
 
+# See http://stackoverflow.com/a/12528721
+# Note that the blank line before 'endef' is actually important - don't delete it
+define execute-command
+	$(1)
+
+endef
+
 release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json
 	./scripts/build/dependencies-npm.sh -p \
 		-r "$(TARGET_ARCH)" \
@@ -112,7 +119,7 @@ release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app: \
 	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules \
 	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower_components
 	./scripts/build/electron-create-resources-app.sh -s . -f "$(APPLICATION_FILES)" -o $@
-	$(foreach prerequisite,$^,cp -rf $(prerequisite) $@;)
+	$(foreach prerequisite,$^,$(call execute-command,cp -rf $(prerequisite) $@))
 
 release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app.asar: \
 	release/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -193,21 +193,32 @@ TARGETS = \
...
 	info \
 	clean \
-	electron-develop
+	electron-develop \
+	publish-aws
 
 ifeq ($(TARGET_PLATFORM),darwin)
 electron-installer-app-zip: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
 electron-installer-dmg: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg
+publish-aws: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg
```
jviotti: Notice that we usually deploy multiple artifacts per platform. For example in the case of OS X, we deploy both the DMG and a ZIP. In Windows, once we integrate it here, we will deploy executables and standalone ZIPs. How would that work here? Should we add many pre-requisites to the publish rules and iterate through them?

jviotti: Alternatively (and this helps reducing duplication of the publish targets), can we accumulate the things we want to deploy in a variable, like we do with the targets, and create single deployment targets (e.g: an AWS target) that iterates through them?

jviotti: The ZIP is currently being uploaded to S3 (since like v1.0.0-beta.4), but its not promoted anywhere, since its only meant to be used by the auto-update engine (which requires a ZIP). We never got the auto-update feature up and running, so the ZIP is unused at the moment, but its still there.

jviotti:
Lets also update the publishing related docs to mention the new Make targets.

lurch:
> Lets also update the publishing related docs to mention the new Make targets.

Good call :+1:

lurch:
Thanks for prompting me to take another look at this, the updated version is much nicer IMO :)

Makefile
```diff
@@ -201,13 +208,35 @@ electron-installer-dmg: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$
...
 electron-installer-debian: release/out/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb
 TARGETS +=  \
 	electron-installer-appimage \
 	electron-installer-debian
+PUBLISH_AWS_S3 += \
+	release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
+PUBLISH_BINTRAY_DEBIAN += \
+	release/out/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb
+endif
+
```
jviotti: Is this extra blank line necessary?

Makefile
```diff
@@ -201,13 +208,35 @@ electron-installer-dmg: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$
 TARGETS += \
 	electron-installer-dmg \
 	electron-installer-app-zip
+PUBLISH_AWS_S3 += \
+	release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip \
+	release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg
 endif
```
jviotti: Now that the content of each conditional grew so much, maybe a blank line between `endif` and `ifeq` can help making things a bit easier to the eye?

Makefile
```diff
@@ -201,13 +208,35 @@ electron-installer-dmg: release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$
 TARGETS += \
 	electron-installer-dmg \
 	electron-installer-app-zip
+PUBLISH_AWS_S3 += \
```
jviotti: I don't like how we're duplicating the whole installer target in both the phony targets declarations (`electron-installer-appimage`, `electron-installer-dmg`, etc) and when accumulating them in the `PUBLISH_*` lists.

Maybe we can create variables for each installer and re-use them? For example:

```
INSTALLER_ELECTRON_APPIMAGE = release/out/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
INSTALLER_ELECTRON_DEBIAN = release/out/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb
```

Or something like that?

Makefile
lurch: Good catch - fixed

Makefile
lurch: Yeah, that's something else I've already done in my yet-to-be-PRed `Makefile` changes ;-)
I want to get all my little changes in first, before starting with the simplification-via-variables, as the latter touches almost every line.

jviotti: Sounds good! I'll approve!

jviotti:
@lurch Ugh, conflicts! :)

lurch:
Fortunately nice and easy to resolve :)

PR #957, opened 2016-12-07
chore: implement `download-tool.sh` script

jviotti:
This script is in charge of downloading any asset from the internet
required by the build system. It accepts a SHA256 checksum and it uses
it to ensure the downloaded file integrity.


scripts/linux/electron-installer-appimage.sh
```diff
@@ -152,10 +127,43 @@ fi
...
-  "$APPIMAGES_GITHUB_RELEASE_BASE_URL/AppImageAssistant_$APPIMAGES_TAG-$APPIMAGES_ARCHITECTURE.AppImage" \
-  $APPIMAGEASSISTANT_PATH
+
+./scripts/unix/download-tool.sh -x \
+  -u "https://raw.githubusercontent.com/probonopd/AppImageKit/$APPIMAGES_TAG/desktopintegration" \
+  -c "bf321258134fa1290b3b3c005332d2aa04ca241e65c21c16c0ab76e892ef6044" \
+  -o "$APPDIR_PATH/usr/bin/$ARGV_BINARY.wrapper"
+
+if [ "$ARGV_ARCHITECTURE" == "x64" ]; then
+
```
lurch: I wonder if it'd be worth having `ARCHITECTURE_APPIMAGE=x86-64`
...

scripts/linux/electron-installer-appimage.sh
```diff
@@ -152,10 +127,43 @@ fi
...
+    -c "28b9c59facd7d0211ef5d825cc00873324cc75163902c48e80e34bf314c910c4" \
+    -o "$APPDIR_PATH/AppRun"
+
+  ./scripts/unix/download-tool.sh -x \
+    -u "$APPIMAGES_GITHUB_RELEASE_BASE_URL/AppImageAssistant_$APPIMAGES_TAG-x86_64.AppImage" \
+    -c "e792fa6ba1dd81de6438844fde39aa12d6b6d15238154ec46baf01da1c92d59f" \
+    -o "$APPIMAGEASSISTANT_PATH"
+
+elif [ "$ARGV_ARCHITECTURE" == "x86" ]; then
+
```
lurch: ... and `ARCHITECTURE_APPIMAGE=i686`
?  (analogous to what we do elsewhere)

scripts/linux/electron-installer-appimage.sh
```diff
@@ -83,25 +83,6 @@ then
...
-APPIMAGES_TAG=6
-APPIMAGES_GITHUB_RELEASE_BASE_URL=https://github.com/probonopd/AppImageKit/releases/download/$APPIMAGES_TAG
-
-if [ "$ARGV_ARCHITECTURE" == "x64" ]; then
-  APPIMAGES_ARCHITECTURE="x86_64"
-elif [ "$ARGV_ARCHITECTURE" == "x86" ]; then
-  APPIMAGES_ARCHITECTURE="i686"
-else
-  echo "Invalid architecture: $ARGV_ARCHITECTURE" 1>&2
-  exit 1
```
lurch: .....doh! I see you already had it, LOL ;-)   Might be worth adding it back?

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,95 @@
...
+check_dep wget
+
+if command -v sha256sum 2>/dev/null 1>&2; then
+  SHA256SUM=sha256sum
+elif command -v shasum 2>/dev/null 1>&2; then
+  SHA256SUM="shasum -a 256"
+else
+  echo "Dependency missing: sha256sum or shasum" 1>&2
+  exit 1
+fi
```
lurch: Nice :+1:

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,95 @@
...
+
+function usage() {
+  echo "Usage: $0"
+  echo ""
+  echo "Options"
+  echo ""
+  echo "    -u <url>"
+  echo "    -c <checksum>"
```
lurch: Might be worth clarifying `<SHA256-checksum>`  ?

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,95 @@
...
+fi
+
+function checksum_matches() {
+  local file=$1
+  local hash=$2
+	test "$($SHA256SUM $file | cut -d ' ' -f1)" = "$hash"
+}
+
+if [ -e "$ARGV_OUTPUT" ] && checksum_matches "$ARGV_OUTPUT" "$ARGV_CHECKSUM"; then
+  echo "Re-using from cache"
```
lurch: Might be worth refactoring this, so that the file is deleted if it exists but its checksum doesn't match?

scripts/linux/electron-installer-appimage.sh
jviotti: Haha, I thought it didn't make sense anymore given that we had to create conditionals for the checksums. In any case, I came up with something even better. I'm updating the PR in a second...

lurch:
Looks very promising, I'll do some more testing later :)

scripts/linux/electron-installer-appimage.sh
jviotti: Done!

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,95 @@
...
+function check_dep() {
+  if ! command -v $1 2>/dev/null 1>&2; then
+    echo "Dependency missing: $1" 1>&2
+    exit 1
+  fi
+}
+
+check_dep wget
+
+if command -v sha256sum 2>/dev/null 1>&2; then
```
jviotti: The thing is that Electron publishes SHA256, and since we can't hardcode every possible hash of every possible version, we are making use of those, and for simplicity purposes, we can use SHA256 everywhere else.

scripts/unix/download-tool.sh
jviotti: Done!

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,95 @@
...
+fi
+
+function usage() {
+  echo "Usage: $0"
+  echo ""
+  echo "Options"
+  echo ""
+  echo "    -u <url>"
+  echo "    -c <checksum>"
+  echo "    -x set execution permissions"
```
jviotti: Done!

scripts/unix/download-tool.sh
jviotti: Good catch! Done!

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,95 @@
...
+  echo "Re-using from cache"
+  exit 0
+fi
+
+wget --no-check-certificate "$ARGV_URL" -O "$ARGV_OUTPUT"
+
+if ! checksum_matches "$ARGV_OUTPUT" "$ARGV_CHECKSUM"; then
+  echo "Checksum mismatch" 1>&2
+  rm -f "$ARGV_OUTPUT"
+  exit 1
```
jviotti: I propose to not bother with it for the time being. We'll see if it comes up in the future.

scripts/linux/electron-installer-appimage.sh
```diff
@@ -152,10 +127,37 @@ fi
...
+  APPIMAGEASSISTANT_CHECKSUM=0faade0c009e703c221650e414f3b4ea8d03abbd8b9f1f065aef46156ee15dd0
+else
+  echo "Invalid architecture: $ARGV_ARCHITECTURE" 1>&2
+  exit 1
+fi
+
+./scripts/unix/download-tool.sh -x \
+  -u "$APPIMAGES_GITHUB_RELEASE_BASE_URL/AppRun_$APPIMAGES_TAG-$APPIMAGES_ARCHITECTURE" \
+  -c "$APPRUN_CHECKSUM" \
+  -o "$APPDIR_PATH/AppRun"
```
lurch: Very nice, these are much clearer now :-)

scripts/unix/download-tool.sh
```diff
@@ -0,0 +1,99 @@
...
+  echo "    -u <url>"
+  echo "    -c <sha256 checksum>"
+  echo "    -x set executable permissions"
+  echo "    -o <output>"
+  exit 1
+}
+
+ARGV_URL=""
+ARGV_CHECKSUM=""
+ARGV_EXECUTION_PERMISSIONS=false
```
lurch: Only a minor nitpick, but could this be renamed to `ARGV_EXECUTABLE_PERMISSIONS` ?

Although looking at https://en.wikipedia.org/wiki/Chmod perhaps it's be better to rename this to `ARGV_EXECUTE_PERMISSION` ? (and update the usage text similarly)

scripts/unix/package-cli.sh
```diff
@@ -78,13 +77,36 @@ NODE_STATIC_ENTRY_POINT_VERSION=v1.0.0 # NodeJS v6
...
+    NODE_STATIC_ENTRY_POINT_CHECKSUM=49eeacc086df04bb3370b7c25afcf36b6e45083889ca438747ece3dd96602b8d
+  fi
+elif [ "$ARGV_OPERATING_SYSTEM" == "win32" ]; then
+  if [ "$ARGV_ARCHITECTURE" == "x64" ]; then
+    NODE_STATIC_ENTRY_POINT_CHECKSUM=60f167aa3389e86956c4ec3c44de7107b4dc9d634b26dbd5f08e14e85f32c2ea
+  fi
+  if [ "$ARGV_ARCHITECTURE" == "x86" ]; then
+    NODE_STATIC_ENTRY_POINT_CHECKSUM=19f07b2d89a727dc64dd4b4e74b7ee8f4464ddc908b63bd60506d471e2f9f602
+  fi
+else
```
lurch: To match the `electron-download-package.sh` script, would it be better if the https://github.com/resin-io-modules/node-static-entry-point project published a `SHASUMS256.txt` file? :)

scripts/unix/package-cli.sh
```diff
@@ -78,13 +77,36 @@ NODE_STATIC_ENTRY_POINT_VERSION=v1.0.0 # NodeJS v6
...
+  echo "Unsupported operating system: $ARGV_OPERATING_SYSTEM" 1>&2
```
lurch: I'm not sure if checking for a valid operating system should be in here - maybe that's a job better delegated to the `Makefile`?

lurch:
Rather than downloading files directly into their final location, there's still part of me that wonders if there should be a 'central' download directory, outside of the 'release' `BUILD_DIRECTORY` (perhaps optionally configurable by another env-var). And then the download-tool could download files to this central-download-directory first, and then copy them from there to the final 'expected' location (obviously checking sha256sums at each stage, and not copying/downloading files if not necessary). This would mean that you could do a `make clean` (deleting the `release` folder), and the next time you ran `make` it would just pick up the files from the central-download-directory, without having to download them from scratch.
(again, this is an idea from BuildRoot that I think it might be worth copying)

And there could be a `cleandownloads` Makefile target which deletes the central-download-directory, and a `distclean` target which depends on both `clean` and `cleandownloads` .

And this would also obviously have caching benefits in the unlikely event that the same downloaded-file ends up getting used in two separate build-directories (e.g. Linux-x86 and Linux-x64).

scripts/unix/package-cli.sh
jviotti: Yeah we could, although I propose to not pay much (if any) attention to the CLI scripts. They still need a lot of work to function properly, and in the future, the whole packaging logic will be part of Capitano.

scripts/unix/package-cli.sh
jviotti: The thing is that someone might call the script directly (not through the Makefile), so I think we provide safety checks anyway instead of just trusting our current Makefile. If the OS is not supported by this script, then it means that no checksum will be calculated, and the download part later on will fail with a cryptic error.

jviotti:
Sounds good. I imagine a `DOWNLOAD_TOOL_CACHE` environment variable that the script uses when its present (otherwise it just downloads directly to the output location).

Could we open an issue for this (as an "improvement" the download script)? We should make sure we do it well, and I'd like to get this on `master`, since I have some Windows packaging changes waiting for it.

jviotti:
Awesome, thanks a lot! BTW, I'm creating a new GitHub label to target the "build system" component, so feel free to fill that with any suggestions/ideas so we keep track on things to improve.

PR #958, opened 2016-12-08
chore: speed up `make` startup time by using `jq`

lurch:
By changing the `Makefile` to use `jq` instead of `node` for getting settings
from `package.json`, the time to do `make help` drops from over 1.5 seconds to
under 0.05 seconds!


jviotti:
Oh, very, very nice!

PR #959, opened 2016-12-08
chore: create re-usable check-dependency.sh script

jviotti:
We have a little snippet to check if a dependency is installed on the
system that we're literally copy-pasting in every single script in the
build system script collection.

For re-usability purposes, this snippet has been extracted to a
`check-dependency.sh` that is called by every script that needs it.

### DIFF
--- a/scripts/build/check-dependency.sh
+++ b/scripts/build/check-dependency.sh
@@ -16,22 +16,33 @@
 # limitations under the License.
 ###
 
-# See http://www.davidpashley.com/articles/writing-robust-shell-scripts/
-set -u
 set -e
+set +u
+ARGV_DEPENDENCIES=$*
+set -u
 
-./scripts/build/check-dependency.sh aws
-
-if [ "$#" -ne 1 ]; then
-  echo "Usage: $0 <file>" 1>&2
+if [ -z "$ARGV_DEPENDENCIES" ]; then
+  echo "Usage: $0 <dependency...>"
   exit 1
 fi
 
-ETCHER_VERSION=`node -e "console.log(require('./package.json').version)"`
-S3_BUCKET="resin-production-downloads"
+RESULT=""
 
-aws s3api put-object \
-  --bucket $S3_BUCKET \
-  --acl public-read \
-  --key etcher/$ETCHER_VERSION/`basename $1` \
-  --body $1
+for dependency in $ARGV_DEPENDENCIES; do
+  if command -v $dependency 2>/dev/null 1>&2; then
+    RESULT=$dependency
+    break
+  fi
+done
+
+if [ -z "$RESULT" ]; then
+  echo "Dependency missing: $ARGV_DEPENDENCIES" 1>&2
+  exit 1
+fi
+
+# Only print back if more than one command was passed
+# otherwise if the script passes, its clear that the
+# single command checked is the one that exists.
+if [ "$#" -ne 1 ]; then
+  echo $RESULT
+fi
--- a/scripts/build/dependencies-bower.sh
+++ b/scripts/build/dependencies-bower.sh
@@ -19,14 +19,7 @@
 set -u
 set -e
 
-function check_dep() {
-  if ! command -v $1 2>/dev/null 1>&2; then
-    echo "Dependency missing: $1" 1>&2
-    exit 1
-  fi
-}
-
-check_dep bower
+./scripts/build/check-dependency.sh bower
 
 function usage() {
   echo "Usage: $0"
--- a/scripts/build/download-tool.sh
+++ b/scripts/build/download-tool.sh
@@ -19,22 +19,11 @@
 set -u
 set -e
 
-function check_dep() {
-  if ! command -v $1 2>/dev/null 1>&2; then
-    echo "Dependency missing: $1" 1>&2
-    exit 1
-  fi
-}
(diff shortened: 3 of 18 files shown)


scripts/build/dependencies-bower.sh
```diff
@@ -19,14 +19,7 @@
...
 
-function check_dep() {
-  if ! command -v $1 2>/dev/null 1>&2; then
-    echo "Dependency missing: $1" 1>&2
-    exit 1
-  fi
-}
-
-check_dep bower
+./scripts/build/check-dependency.sh bower
```
lurch: Probably better to use `$(dirname $0)/check-dependency.sh bower` ?

scripts/build/check-dependency.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [ -z "$ARGV_DEPENDENCY" ]; then
+  echo "Usage: $0 <dependency>"
+  exit 1
+fi
+
+if ! command -v $ARGV_DEPENDENCY 2>/dev/null 1>&2; then
+  echo "Dependency missing: $ARGV_DEPENDENCY" 1>&2
+  exit 1
+fi
```
lurch: I wonder if you could extend this to optionally take a second (or more?) argument(s), and then return the first of those arguments that actually exists? I.e. the stuff you do in `download-tool.sh` to work out which SHA256SUM command to use.

lurch:
So, following on from this, would it also be worth creating `check-linux.sh` and `check-darwin.sh` scripts?

jviotti:
> So, following on from this, would it also be worth creating check-linux.sh and check-darwin.sh scripts?

I was actually working on a script that can detect the platform or the architecture (depending on boolean options) and re-use that in the Makefile as well. What do you think?

lurch:
Yeah, I guess that works. But a simple `./scripts/build/check-darwin.sh` might also be neater than `if [[ $(./scripts/build/host-detect.sh -o) != "darwin" ]]; then ...`

Of course it probably makes sense for the former to internally use the latter.

jviotti:
> Yeah, I guess that works. But a simple ./scripts/build/check-darwin.sh might also be neater than if [[ $(./scripts/build/host-detect.sh -o) != "darwin" ]]; then ...
> Of course it probably makes sense for the former to internally use the latter.

That's a great pattern. I'll make use of it when implementing that script.

scripts/build/check-dependency.sh
jviotti: Done!

scripts/build/check-dependency.sh
```diff
@@ -0,0 +1,48 @@
...
+  echo "Dependency missing: $ARGV_DEPENDENCIES" 1>&2
+  exit 1
+fi
+
+# Only print back if more than one command was passed
+# otherwise if the script passes, its clear that the
+# single command checked is the one that exists.
+if [ "$#" -ne 1 ]; then
+  echo $RESULT
+fi
```
lurch: Cool :-)

scripts/build/download-tool.sh
```diff
@@ -19,22 +19,11 @@
...
-  SHA256SUM=sha256sum
-elif command -v shasum 2>/dev/null 1>&2; then
-  SHA256SUM="shasum -a 256"
-else
-  echo "Dependency missing: sha256sum or shasum" 1>&2
-  exit 1
+SHA256SUM=$(./scripts/build/check-dependency.sh sha256sum shasum)
+if [ "$SHA256SUM" == "shasum" ]; then
+  SHA256SUM="$SHA256SUM -a 256"
 fi
```
lurch: If you modified `check-dependency.sh` to use `$@` instead of `$*` (see http://stackoverflow.com/questions/255898/how-to-iterate-over-arguments-in-bash-script ), then you could replace these 4 lines with a single:
```bash
SHA256SUM=$(./scripts/build/check-dependency.sh sha256sum "shasum -a 256")
```
:smile:

jviotti: I tried this, however the shell seems to strip the quotes when passing them to `check-dependency.sh`, therefore the options are ignored (unless I'm doing something wrong):

```diff
diff --git a/scripts/build/check-dependency.sh b/scripts/build/check-dependency.sh
index 6df82ec..a5b3346 100755
--- a/scripts/build/check-dependency.sh
+++ b/scripts/build/check-dependency.sh
@@ -16,9 +16,10 @@
 # limitations under the License.
 ###

+set -x
 set -e
 set +u
-ARGV_DEPENDENCIES=$*
+ARGV_DEPENDENCIES=$@
 set -u

 if [ -z "$ARGV_DEPENDENCIES" ]; then
```

Now check this:

```sh
$ ./scripts/build/check-dependency.sh foo "bower -v"
+ set -e
+ set +u
+ ARGV_DEPENDENCIES='foo bower -v'
+ set -u
+ '[' -z 'foo bower -v' ']'
+ RESULT=
+ for dependency in '$ARGV_DEPENDENCIES'
+ command -v foo
+ for dependency in '$ARGV_DEPENDENCIES'
+ command -v bower
+ RESULT=bower
+ break
+ '[' -z bower ']'
+ '[' 2 -ne 1 ']'
+ echo bower
bower
```

I'll merge this, but happy to have another PR with this improvement if we manage to make it work.

PR #962, opened 2016-12-08
chore: implement `electron-configure-package-win32.sh`

jviotti:
This script configures an Electron package on Windows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ BUILD_OUTPUT_DIRECTORY = $(BUILD_DIRECTORY)/out
 # ---------------------------------------------------------------------
 
 ELECTRON_VERSION = $(shell jq -r '.devDependencies["electron-prebuilt"]' package.json)
+COMPANY_NAME = $(shell jq -r '.companyName' package.json)
 APPLICATION_NAME = $(shell jq -r '.displayName' package.json)
 APPLICATION_DESCRIPTION = $(shell jq -r '.description' package.json)
 APPLICATION_VERSION = $(shell jq -r '.version' package.json)
@@ -91,6 +92,15 @@ TARGET_ARCH = $(HOST_ARCH)
 ifeq ($(TARGET_PLATFORM),darwin)
 ifndef CODE_SIGN_IDENTITY
 $(warning No code-sign identity found (CODE_SIGN_IDENTITY is not set))
+endif
+endif
+
+ifeq ($(TARGET_PLATFORM),win32)
+ifndef CODE_SIGN_CERTIFICATE
+$(warning No code-sign certificate found (CODE_SIGN_CERTIFICATE is not set))
+ifndef CODE_SIGN_CERTIFICATE_PASSWORD
+$(warning No code-sign certificate password found (CODE_SIGN_CERTIFICATE_PASSWORD is not set))
+endif
 endif
 endif
 
@@ -184,6 +194,26 @@ ifeq ($(TARGET_PLATFORM),linux)
 		-v "$(APPLICATION_VERSION)" \
 		-l LICENSE \
 		-o $@
+endif
+ifeq ($(TARGET_PLATFORM),win32)
+	./scripts/build/electron-configure-package-win32.sh -p $(word 2,$^) -a $< \
+		-n "$(APPLICATION_NAME)" \
+		-d "$(APPLICATION_DESCRIPTION)" \
+		-v "$(APPLICATION_VERSION)" \
+		-l LICENSE \
+		-c "$(APPLICATION_COPYRIGHT)" \
+		-m "$(COMPANY_NAME)" \
+		-i assets/icon.ico \
+		-w $(TEMPORARY_DIRECTORY) \
+		-o $@
+ifdef CODE_SIGN_CERTIFICATE
+ifdef CODE_SIGN_CERTIFICATE_PASSWORD
+	./scripts/build/electron-sign-exe.sh -f $@/$(APPLICATION_NAME).exe \
+		-d "$(APPLICATION_NAME) - $(APPLICATION_VERSION)"
+		-c $(CODE_SIGN_CERTIFICATE) \
+		-p $(CODE_SIGN_CERTIFICATE_PASSWORD)
+endif
+endif
 endif
 
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-rw.dmg: \
--- a/scripts/build/electron-configure-package-win32.sh
+++ b/scripts/build/electron-configure-package-win32.sh
@@ -1,0 +1,126 @@
+#!/bin/bash
+
+###
+# Copyright 2016 resin.io
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
+set -u
+set -e
+
+OS=$(uname -o 2>/dev/null || true)
+if [[ "$OS" != "Msys" ]]; then
+  echo "This script is only meant to be run in Windows" 1>&2
+  exit 1
+fi
+
+./scripts/build/check-dependency.sh upx
+./scripts/build/check-dependency.sh unzip
+
+function usage() {
+  echo "Usage: $0"
+  echo ""
+  echo "Options"
+  echo ""
+  echo "    -p <electron package>"
+  echo "    -n <application name>"
+  echo "    -d <application description>"
+  echo "    -v <application version>"
+  echo "    -c <application copyright>"
+  echo "    -l <application license file>"
+  echo "    -m <company name>"
+  echo "    -a <application asar (.asar)>"
+  echo "    -i <application icon (.ico)>"
+  echo "    -w <download directory>"
+  echo "    -o <output directory>"
+  exit 1
+}
+
+ARGV_ELECTRON_PACKAGE=""
+ARGV_APPLICATION_NAME=""
+ARGV_APPLICATION_DESCRIPTION=""
+ARGV_VERSION=""
+ARGV_COPYRIGHT=""
+ARGV_LICENSE=""
+ARGV_COMPANY_NAME=""
+ARGV_ASAR=""
+ARGV_ICON=""
+ARGV_DOWNLOAD_DIRECTORY=""
+ARGV_OUTPUT=""
+
+while getopts ":p:n:d:v:c:l:m:a:i:w:o:" option; do
+  case $option in
+    p) ARGV_ELECTRON_PACKAGE="$OPTARG" ;;
+    n) ARGV_APPLICATION_NAME="$OPTARG" ;;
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -10,6 +10,7 @@ APPLICATION_COPYRIGHT = $(shell jq -r '.copyright' package.json)
 APPLICATION_CATEGORY = public.app-category.developer-tools
 APPLICATION_BUNDLE_ID = io.resin.etcher
 APPLICATION_FILES = lib,build,assets
+COMPANY_NAME = Resinio Ltd
```
lurch: Could / should that be included in `package.json`?

Makefile
```diff
@@ -78,6 +79,15 @@ $(warning No code-sign identity found (CODE_SIGN_IDENTITY is not set))
 endif
 endif
 
+ifeq ($(TARGET_PLATFORM),win32)
+ifndef CODE_SIGN_CERTIFICATE
+$(warning No code-sign certificate found (CODE_SIGN_CERTIFICATE is not set))
+endif
+ifndef CODE_SIGN_CERTIFICATE_PASSWORD
+$(warning No code-sign certificate password found (CODE_SIGN_CERTIFICATE_PASSWORD is not set))
+endif
```
lurch: Probably makes sense to only show these warnings when you're actually about to do the code-signing step, rather than showing them all the time to all users.

lurch: And it probably makes sense to nest the `ifndef CODE_SIGN_CERTIFICATE_PASSWORD` inside the `ifndef CODE_SIGN_CERTIFICATE` too

Makefile
```diff
@@ -145,6 +155,26 @@ ifeq ($(TARGET_PLATFORM),linux)
...
 		-o $@
 endif
+ifeq ($(TARGET_PLATFORM),win32)
+	./scripts/build/electron-configure-package-win32.sh -p $(word 2,$^) -a $< \
+		-n "$(APPLICATION_NAME)" \
+		-d "$(APPLICATION_DESCRIPTION)" \
+		-v "$(APPLICATION_VERSION)" \
+		-l LICENSE \
+		-r "$(APPLICATION_COPYRIGHT)" \
+		-c "$(COMPANY_NAME)" \
```
lurch: In the darwin section, you've used `-c "$(APPLICATION_COPYRIGHT)"` - consistency is always good

scripts/build/electron-configure-package-win32.sh
```diff
@@ -0,0 +1,132 @@
...
+
+function check_dep() {
+  if ! command -v $1 2>/dev/null 1>&2; then
+    echo "Dependency missing: $1" 1>&2
+    exit 1
+  fi
+}
+
+check_dep upx
+check_dep unzip
```
lurch: Don't forget to update this if #959 gets merged first

scripts/build/electron-configure-package-win32.sh
```diff
@@ -0,0 +1,132 @@
...
+if [ -d "$ARGV_ASAR.unpacked" ]; then
+  cp -rf "$ARGV_ASAR.unpacked" "$ARGV_OUTPUT/resources/app.asar.unpacked"
+fi
+
+echo "$ARGV_VERSION" > "$ARGV_OUTPUT/version"
+
+RCEDIT_VERSION=v0.7.0
+RCEDIT="$ARGV_DOWNLOAD_DIRECTORY/rcedit.exe"
+
+# TODO: Move to Makefile
```
lurch: Yep ;)

lurch:
Great to see the `win32` build moving to the same Makefile & Bash setup as `linux` and `darwin` :+1:

scripts/build/electron-configure-package-win32.sh
```diff
@@ -0,0 +1,132 @@
...
+  || [ -z "$ARGV_COMPANY_NAME" ] \
+  || [ -z "$ARGV_ASAR" ] \
+  || [ -z "$ARGV_ICON" ] \
+  || [ -z "$ARGV_DOWNLOAD_DIRECTORY" ] \
+  || [ -z "$ARGV_OUTPUT" ]
+then
+  usage
+fi
+
+mkdir -p $(dirname "$ARGV_OUTPUT")
```
jviotti: Done!

scripts/build/electron-configure-package-win32.sh
jviotti: Done!

scripts/build/electron-configure-package-win32.sh
```diff
@@ -0,0 +1,132 @@
...
+  echo "    -p <electron package>"
+  echo "    -n <application name>"
+  echo "    -d <application description>"
+  echo "    -v <application version>"
+  echo "    -r <application copyright>"
+  echo "    -l <application license file>"
+  echo "    -c <company name>"
+  echo "    -a <application asar (.asar)>"
+  echo "    -i <application icon (.ico)>"
+  echo "    -w <download directory>"
```
jviotti: I think I prefer to be explicit about it and ask it as an option rather than relying on the environment at all.

Makefile
lurch: I guess one option would be to have a non-signed version as the "standard" DMG, and then have a special code-signed DMG required by the publish step? That's probably a bit too much of a step though just to silence a couple of warnings ;-)
Or a simple solution could just be a `SILENCE_CODE_SIGN_WARNINGS ?= 0` in the Makefile, so the user could set a `SILENCE_CODE_SIGN_WARNINGS=1` env-var if they wanted to.

**OTOH**, perhaps it's inconsistent behaviour to have the Makefile generate two different versions of the same file, based on the value of an env-var? (makes it hard to tell them apart?)
Consider if:
 1. You run make without the CODE_SIGN env-vars set
 2. Make creates the unsigned package
 3. You run make again _with_ the CODE_SIGN env-vars set
 4. Make doesn't do anything, because it thinks the package is already up to date
 5. You run 'make publish'
 6. Whoops, make has now uploaded the unsigned packages!

What if instead, there was e.g. a `release/out/signed` directory, and `make publish` depended on `release/out/signed/Etcher-1.0.0-beta.17-darwin-x64.dmg`, which in turn depended on the temporary read-write DMG (could it be given a different name and made non-temporary?). And then `release/out/Etcher-1.0.0-beta.17-darwin-x64.dmg` would also depend on the same 'temporary' read-write DMG, and would always be the non-signed package.
Would that work? ;)

scripts/build/electron-configure-package-win32.sh
lurch: Fair enough

Makefile
lurch: ...and (again, for consistency) could you change `-r "$(APPLICATION_COPYRIGHT)"` to `-c "$(APPLICATION_COPYRIGHT)"` ? :smile:

Makefile
lurch: And for those scripts where you have the option to verify the signature, maybe it would make sense to sign the package / exe in the temporary directory, and only move it to `release/out/signed/` after it's been verified.

scripts/build/electron-configure-package-win32.sh
```diff
@@ -0,0 +1,128 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
+set -u
+set -e
+
+function check_dep() {
```
lurch: Need to remove this function :)

lurch:
By re-ordering some of the lines...
```patch
--- a/scripts/build/electron-configure-package-win32.sh
+++ b/scripts/build/electron-configure-package-win32.sh
@@ -95,16 +95,9 @@ fi
 unzip "$ARGV_ELECTRON_PACKAGE" -d "$ARGV_OUTPUT"
 
 mv "$ARGV_OUTPUT/electron.exe" "$ARGV_OUTPUT/$ARGV_APPLICATION_NAME.exe"
-rm -f "$ARGV_OUTPUT/resources/default_app.asar"
-
-cp "$ARGV_ASAR" "$ARGV_OUTPUT/resources/app.asar"
 cp "$ARGV_LICENSE" "$ARGV_OUTPUT/LICENSE"
-
-if [ -d "$ARGV_ASAR.unpacked" ]; then
-  cp -rf "$ARGV_ASAR.unpacked" "$ARGV_OUTPUT/resources/app.asar.unpacked"
-fi
-
 echo "$ARGV_VERSION" > "$ARGV_OUTPUT/version"
+rm -f "$ARGV_OUTPUT/resources/default_app.asar"
 
 RCEDIT_VERSION=v0.7.0
 RCEDIT="$ARGV_DOWNLOAD_DIRECTORY/rcedit.exe"
@@ -126,3 +119,9 @@ RCEDIT="$ARGV_DOWNLOAD_DIRECTORY/rcedit.exe"
   --set-icon "$ARGV_ICON"
 
 upx -9 "$ARGV_OUTPUT/*.dll"
+
+cp "$ARGV_ASAR" "$ARGV_OUTPUT/resources/app.asar"
+
+if [ -d "$ARGV_ASAR.unpacked" ]; then
+  cp -rf "$ARGV_ASAR.unpacked" "$ARGV_OUTPUT/resources/app.asar.unpacked"
+fi
```
...it's possible to get this script to much more closely match `electron-configure-package-linux.sh`. Have I ever mentioned I like consistency? ;-)

scripts/build/electron-configure-package-win32.sh
jviotti: Oops, done!

scripts/build/electron-configure-package-win32.sh
```diff
@@ -0,0 +1,128 @@
...
+  if ! command -v $1 2>/dev/null 1>&2; then
+    echo "Dependency missing: $1" 1>&2
+    exit 1
+  fi
+}
+
```
jviotti: Indeed. In theory, this script could run in the other operating systems on top of Wine, but lets ignore that for now :)

jviotti:
> Have I ever mentioned I like consistency? ;-)

Haha, I think I heard something about that! :P

I applied the diff!

jviotti:
Is this good to merge?

lurch:
There's two reasons I hadn't approved this yet:
 * https://github.com/resin-io/etcher/pull/962#discussion_r91826927
 * And I didn't want to lose [this comment](https://github.com/resin-io/etcher/pull/962#discussion_r91826772) until you'd replied to it

Makefile
jviotti: Oops, I missed this comment. Done!

jviotti:
Oops, I missed those comments, sorry for that!

Makefile
jviotti: I like the `signed/` subdirectory idea. I propose to do it in another PR so we can do it for all the operating systems. Regarding the warning, what about simply not printing the warning, but report that the binary is not going to be signed, in `make info`? 

If you agree on that, lets merge the PR as it is now, and I'll send a new one right away to make that change, so we do it for all operating systems at once.

lurch: As I said [here](https://github.com/resin-io/etcher/pull/968#issuecomment-266475641) I think it'd make sense for the "standard build" e.g. `make electron-installer-dmg` to produce unsigned images (and have them also be snapshot versions by default), and then have a special e.g. `signed-electron-installer-dmg` target to create (production-version?) files in the `signed/` subdirectory, and only do the check-for-certificate-vars step as part of the `signed-electron-installer-dmg` rules, and upgrade it from a warning to an error.
(If that all makes sense? :wink: )

jviotti: Sounds good to me :) Let me know if I can help with any OS X testing (if you don't have access to OS X yet, you might want to ping the operations team to ship you a Mac Mini or something like that anyway)!

PR #963, opened 2016-12-08
chore: run tests in x64 and x86 in Appveyor

jviotti:
This will allow us to later publish development builds for both
architectures from Appveyor. In order to easily do so, the Makefile is
used instead of the current `./scripts/build/windows.bat`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,22 +17,25 @@ cache:
 environment:
   global:
     ELECTRON_NO_ATTACH_CONSOLE: true
+    nodejs_version: 6.1.0
   matrix:
-    - nodejs_version: 6.1.0
+    - TARGET_ARCH: x64
+    - TARGET_ARCH: x86
 
 install:
   - ps: Install-Product node $env:nodejs_version x64
   - npm install -g npm bower rimraf asar
   - choco install nsis -version 2.51
   - choco install upx
-  - choco install 7zip.commandline
+  - choco install jq
   - set PATH=C:\Program Files (x86)\Windows Kits\8.1\bin\x86;%PATH%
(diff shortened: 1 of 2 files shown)


lurch:
Probably worth mentioning in the commit-comment that this also switches away from using `windows.bat` to using the `Makefile`. (unless that was an accidental inclusion in this PR?)

jviotti:
Done!

appveyor.yml
```diff
@@ -17,22 +17,25 @@ cache:
...
+  - choco install jq
   - set PATH=C:\Program Files (x86)\Windows Kits\8.1\bin\x86;%PATH%
   - set PATH=C:\Program Files (x86)\NSIS;%PATH%
   - set PATH=C:\Ruby22\bin;%PATH%
   - set PATH=C:\MinGW\bin;%PATH%
   - set PATH=C:\MinGW\msys\1.0\bin;%PATH%
   - gem install scss_lint
-  - .\scripts\build\windows.bat install x64
+  - make -e info
+  - make -e electron-develop
```
lurch: Rather than use `make -e` here, IMHO it'd be better to update the Makefile to do `TARGET_ARCH ?= $(HOST_ARCH)`
(and then #964 should be updated to match)

jviotti: Sounds good! Done in both PRs.

jviotti:
@lurch Is this good to merge?

lurch:
So, out of curiosity, how come the tests pass on both x64 and x86 on the Windows appveyor build? :)

jviotti:
I believe were seeing an Electron issue on the Linux jobs :/

Issue #888, opened 2016-11-22
Make behavior of protected system drives more obvious to user


WasabiFan:
This came out of a discussion on Gitter, but I wanted to materialize it here for tracking.

> @alexandrosm: @WasabiFan in general I wonder if instead of not showing system drives at all, instead Etcher should just not auto-select them, de-emphasise them on the selection menu, and if the user actively selects it, it gives a warning, but ultimately goes ahead.
@alexandrosm: in other words treat the system drive heuristic as a heuristic instead of as a law as it does now

> @WasabiFan: Yeah, I like that. Or maybe gray them out and then put the "unsafe mode" checkbox on the selection screen (something like "Allow selection of system drives").
This would make it much easier for users to understand why their drives aren't showing up.

In general, the idea would be to make the filtering out of "system" drives visible to the user, so that they can see what is being filtered out and why.

alexandrosm:
One more thing to highlight, I like the idea of moving things out of the
"garbage dump" settings page and into the specific screens they are
relevant on. If unsafe mode makes more sense in the drive selection, let's
move it in there.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

WasabiFan:
Agreed, especially if in doing so that checkbox gets a better name! It would be great to see "settings" become something more like "preferences" (logging, telemetry, debugging, etc.).

jviotti:
Sounds cool. Lets put a "system drive" badge as we do for locked, or small devices, and disable them as well. Once that's done, we can move the checkbox there.

lurch:
Yeah, overlaying badges on different types of drives is also something that we discussed with @taahirisaacs and @konmouz at the [Summit](https://resin.io/blog/summit-2016/). I think we decided that removable drives (that we don't detect as "system") should always be displayed at the top of the list.

IMHO if we're going to be displaying "System" drives to the user (with 'just' a simple tickbox to enable writing to them) then perhaps we also ought to attempt to identify "boot" drives, and **never** allow those to be overwritten? And we could/should also filter "source" drives, as mentioned in #830 

So, there'd then be:
 * Removable drives - always okay to write
 * "System" drives - writable if the user enables "unsafe mode" (or whatever it gets renamed to)
 * "Boot" drives (e.g. where the current Operating System was loaded from - should we also try detecting swap / pagefile drives currently being used?) - we should prevent the user writing to here

And then any of the above drives can also be:
[...]

PR #966, opened 2016-12-09
feat(gui): show system drives by default with warnings

Shou:
Instead of hiding system drives, they are now shown by default but with
added warnings upon attempts to select or flash a system drive.

- Red 'Continue' button in drive-list widget when system drive selected
- Alert icons added to system drives
- DriveInfoModal added
  - Used to warn users before flashing a system drive
  - Used to display drive details
- Unsafe mode removed

See: https://github.com/resin-io/etcher/issues/888
Changelog-Entry: Show system drives by default with warnings.

### DIFF
--- a/lib/gui/components/drive-info-modal/services/drive-info-modal.js
+++ b/lib/gui/components/drive-info-modal/services/drive-info-modal.js
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2016 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const _ = require('lodash');
+
+module.exports = function(ModalService) {
+
+  /**
+   * @summary Display the drive info modal
+   * @function
+   * @public
+   *
+   * @param {Object} options - options
+   * @param {String} options.description - danger message
+   * @param {String} options.confirmationLabel - confirmation button text
+   * @param {String} options.rejectionLabel - rejection button text
+   * @fulfil {Boolean} - whether the user accepted or rejected the warning
+   * @returns {Promise}
+   *
+   * @example
+   * WarningModalService.display({
+   *   description: 'Don\'t do this!',
+   *   confirmationLabel: 'Yes, continue!'
+   * });
+   */
+  this.display = (options = {}) => {
+    return ModalService.open({
+      template: './components/drive-info-modal/templates/drive-info-modal.tpl.html',
+      controller: 'DriveInfoModalController as modal',
+      size: 'drive-info-modal',
+      resolve: {
+        options: _.constant(options)
+      }
+    }).result;
+  };
+
+};
--- a/lib/gui/components/drive-info-modal/templates/drive-info-modal.tpl.html
+++ b/lib/gui/components/drive-info-modal/templates/drive-info-modal.tpl.html
@@ -1,0 +1,59 @@
+<div class="modal-header">
+  <h4 class="modal-title">
+    <span class="glyphicon glyphicon-{{ ::modal.options.icon }}"></span>
+    <span>{{ ::modal.options.title }}</span>
+  </h4>
+  <button class="close" ng-click="modal.reject()">&times;</button>
+</div>
+
+<div class="modal-body">
+  <div class="modal-text"
+    ng-if="modal.options.header && modal.options.description">
+    <h4>{{ ::modal.options.header }}</h4>
+    <p>{{ ::modal.options.description }}</p>
+  </div>
+  <div class="modal-selected-drive">
+    <div class="modal-selected-drive-description">
+      {{ ::modal.selection.getDrive().description }}
+
+    </div>
+    <div class="modal-selected-drive-name">
+      {{ ::modal.selection.getDrive().name }} - {{ ::modal.selection.getDrive().size | gigabyte | number:1 }} GB
+
+
+      <span class="label label-warning"
+          ng-show="modal.selection.isDriveLargeEnough(modal.selection.getDrive()) && !modal.selection.isDriveLocked(modal.selection.getDrive()) && !modal.selection.isDriveSizeRecommended(modal.selection.getDrive())">
+          <i class="glyphicon glyphicon-warning-sign"></i>
+          NOT RECOMMENDED</span>
+
+      <span class="label label-danger"
+        ng-hide="modal.selection.isDriveLargeEnough(modal.selection.getDrive())">
+          <i class="glyphicon glyphicon-resize-small"></i>
+          TOO SMALL FOR IMAGE</span>
+
+      <span class="label label-danger"
+        ng-show="modal.selection.isDriveLocked(modal.selection.getDrive()) && modal.selection.isDriveLargeEnough(modal.selection.getDrive())">
+          <i class="glyphicon glyphicon-lock"></i>
+          LOCKED</span>
+
+      <span class="label label-danger"
+        ng-show="modal.selection.getDrive().system">
+          <i class="glyphicon glyphicon-hdd"></i>
+          SYSTEM DRIVE</span>
+
+    </div>
+  </div>
+</div>
+
+<div class="modal-footer"
+  ng-if="modal.options.confirmationLabel && modal.options.rejectionLabel">
+  <div class="modal-menu">
+    <button class="button button-danger button-block"
+      ng-if="modal.options.confirmationLabel" 
+      ng-click="modal.accept()">{{ ::modal.options.confirmationLabel }}</button>
+    <button class="button button-block"
+      ng-if="modal.options.rejectionLabel" 
+      ng-click="modal.reject()">{{ ::modal.options.rejectionLabel }}</button>
+  </div>
+</div>
+
--- a/lib/gui/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/components/drive-selector/controllers/drive-selector.js
@@ -18,7 +18,7 @@
 
 const _ = require('lodash');
 
-module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateModel, WarningModalService) {
+module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateModel, SettingsModel, WarningModalService) {
 
   /**
    * @summary The drive selector state
--- a/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
+++ b/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
@@ -35,6 +35,11 @@
                 <i class="glyphicon glyphicon-lock"></i>
                 LOCKED</span>
 
+            <span class="label label-danger"
+              ng-show="modal.state.isSystemDrive(drive)">
+                <i class="glyphicon glyphicon-hdd"></i>
+                SYSTEM DRIVE</span>
+
           </footer>
         </div>
         <span class="tick tick--success"
@@ -45,7 +50,15 @@
 </div>
 
 <div class="modal-footer">
+  <span ng-show="modal.state.getDrive().system" class="modal-footer-warning">
+      Etcher has identified this as a system drive
+  </span>
   <button class="button button-primary button-block"
+    ng-class="{ 'label-danger': modal.state.getDrive().system, 'button-primary': !modal.state.getDrive().system }"
     ng-click="modal.closeModal()"
-    ng-disabled="!modal.state.hasDrive()">CONTINUE</button>
+    ng-disabled="!modal.state.hasDrive()">
+      <i class="glyphicon glyphicon-alert"
+        ng-show="modal.state.getDrive().system"></i>
+      CONTINUE
+  </button>
 </div>
--- a/lib/gui/models/selection-state.js
+++ b/lib/gui/models/selection-state.js
@@ -169,6 +169,28 @@ SelectionStateModel.service('SelectionStateModel', function(DrivesModel) {
   };
 
   /**
+   * @summary Check if a drive is a system drive
+   * @function
+   * @public
+   * @param {Object} drive - drive
+   * @returns {Boolean} whether the drive is a system drive
+   *
+   * @example
+   * if (SelectionStateModel.isSystemDrive({
+   *   device: '/dev/disk2',
+   *   name: 'My Drive',
+   *   size: 123456789,
+   *   protected: true,
+   *   system: true
+   * })) {
+   *   console.log('This drive is a system drive!');
+   * }
+   */
+  this.isSystemDrive = (drive) => {
+    return drive.system;
+  };
+
+  /**
    * @summary Toggle set drive
    * @function
    * @public
--- a/lib/gui/models/store.js
+++ b/lib/gui/models/store.js
@@ -37,7 +37,6 @@ const DEFAULT_STATE = Immutable.fromJS({
     speed: 0
   },
   settings: {
-    unsafeMode: false,
     errorReporting: true,
     unmountOnSuccess: true,
     validateWriteOnSuccess: true,
@@ -90,9 +89,17 @@ const storeReducer = (state, action) => {
 
       const newState = state.set('availableDrives', Immutable.fromJS(action.data));
 
-      if (action.data.length === 1) {
+      // Non-system drives
+      const noSystemDrives = _.filter(action.data, (drive) => {
+        return !drive.system;
+      });
 
-        const drive = _.first(action.data);
+      const selectedDevice = newState.getIn([ 'selection', 'drive' ]);
+
+      // There are non-system drives; set one as the selected drive.
+      if (!selectedDevice && noSystemDrives.length >= 1) {
+
+        const drive = _.first(noSystemDrives);
 
         if (_.every([
 
@@ -111,8 +118,6 @@ const storeReducer = (state, action) => {
         }
 
       }
-
-      const selectedDevice = newState.getIn([ 'selection', 'drive' ]);
 
       if (selectedDevice && !_.find(action.data, {
         device: selectedDevice
--- a/lib/gui/modules/drive-scanner.js
+++ b/lib/gui/modules/drive-scanner.js
@@ -32,7 +32,7 @@ const driveScanner = angular.module(MODULE_NAME, [
   require('../models/settings')
 ]);
 
-driveScanner.factory('DriveScannerService', (SettingsModel) => {
+driveScanner.factory('DriveScannerService', () => {
   const DRIVE_SCANNER_INTERVAL_MS = 2000;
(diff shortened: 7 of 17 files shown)


Shou:
Here are some screenshots

Before flashing, if your drive's a system drive


The drive details modal


Drive selector modal
image

WasabiFan:
Some comments/questions:
- When does the "drive details" modal appear? Is that a link below the drive selection step?
- The "All data on your system will be lost" text sounds awkward and isn't actually correct. I think that the word "destroy(ed)" should be included somewhere and that the extreme "all" should be removed or qualified.
- The text alignment in the last image feels wrong; it seems like that text -- the line above the button -- should be centered.
- The button text "change" feels odd... a more direct word might be better.

Shou:
@WasabiFan 
* So the drive details modal appears when you click the drive mount point (below the circled 2nd step); the text is accompanied by an alert triangle when it's a system drive, clicking it will show exactly what's 'wrong' through the information presented by the modal.
* Yeah we may have to tweak the wording a bit here, "all" is technically wrong (in most cases). Maybe "this will result in an unbootable system and data loss"?
* Indeed, I actually centred it locally a jiffy ago.
* I think the idea behind the wording was "change [drive]". The "change" button is going to open (but doesn't yet) the drive selector modal.

lurch:
Ping @konmouz for him to compare this to his designs :)

"this will result in an unbootable system and data loss" isn't technically true, because although it may be a "system drive", it may not necessarily be the "boot drive" - the extra subtlety is dependent on https://github.com/resin-io-modules/drivelist/issues/126

I agree the text "All data on your system will be lost" is misleading, but I think the "All data on your system drive will be lost" proposed in #888 _might_ be okay.

"The "change" button is going to open (but doesn't yet)" - does that mean that this PR isn't ready for merging yet?

Is the drive details modal also available for removable drives? And if so could you include a screenshot of that too please?

lib/gui/components/drive-info-modal/templates/drive-info-modal.tpl.html
```diff
@@ -0,0 +1,59 @@
...
+    <div class="modal-selected-drive-description">
+      {{ ::modal.selection.getDrive().description }}
+
+    </div>
+    <div class="modal-selected-drive-name">
+      {{ ::modal.selection.getDrive().name }} - {{ ::modal.selection.getDrive().size | gigabyte | number:1 }} GB
+
+
+      <span class="label label-warning"
+          ng-show="modal.selection.isDriveLargeEnough(modal.selection.getDrive()) && !modal.selection.isDriveLocked(modal.selection.getDrive()) && !modal.selection.isDriveSizeRecommended(modal.selection.getDrive())">
```
lurch: Now that the logic on this line...

lib/gui/components/drive-info-modal/templates/drive-info-modal.tpl.html
```diff
@@ -0,0 +1,59 @@
...
+          <i class="glyphicon glyphicon-warning-sign"></i>
+          NOT RECOMMENDED</span>
+
+      <span class="label label-danger"
+        ng-hide="modal.selection.isDriveLargeEnough(modal.selection.getDrive())">
+          <i class="glyphicon glyphicon-resize-small"></i>
+          TOO SMALL FOR IMAGE</span>
+
+      <span class="label label-danger"
+        ng-show="modal.selection.isDriveLocked(modal.selection.getDrive()) && modal.selection.isDriveLargeEnough(modal.selection.getDrive())">
```
lurch: ...and the logic on this line is shared between `drive-selector-modal.tpl.html` and `drive-info-modal.tpl.html`, perhaps this logic of categorising drives into "OKAY" (no warning label), "NOT RECOMMENDED", "TOO SMALL FOR IMAGE", "LOCKED" and "SYSTEM DRIVE" ought to be refactored out into the model itself (or some other shared JS file), rather than having the logic embedded inside HTML templates?

lurch: Also @jviotti it looks like the current logic is dictating that if a drive is too small, and it's locked, then the warning shown is "TOO SMALL FOR IMAGE".
IMHO the fact that it's locked, out to be displayed in preference to the fact that it's too small? Moving the categorisation logic outside of the template file will make it easier to make this logic more consistent, without this current mish-mash of some of the warning labels here using `ng-show` and some of them using `ng-hide`.

lib/gui/models/store.js
```diff
@@ -90,9 +89,17 @@ const storeReducer = (state, action) => {
...
+        return !drive.system;
+      });
+
+      const selectedDevice = newState.getIn([ 'selection', 'drive' ]);
 
-        const drive = _.first(action.data);
+      // There are non-system drives; set one as the selected drive.
+      if (!selectedDevice && noSystemDrives.length >= 1) {
+
+        const drive = _.first(noSystemDrives);
```
lurch: Um, is this auto-selecting the first removable drive, if there are multiple drives available?
IMHO Etcher shouldn't do that, it should just stick with the current behaviour of only auto-selecting a drive if it's the **only** removable drive available.

lib/gui/models/store.js
```diff
@@ -90,9 +89,17 @@ const storeReducer = (state, action) => {
 
       const newState = state.set('availableDrives', Immutable.fromJS(action.data));
 
-      if (action.data.length === 1) {
+      // Non-system drives
+      const noSystemDrives = _.filter(action.data, (drive) => {
```
lurch: In preparation for the addition of "Boot drive" detection (amongst other filtering), could this `noSystemDrives` be renamed to `removableDrives` ?

lib/gui/pages/main/controllers/main.js
```diff
@@ -65,4 +66,22 @@ module.exports = function(
     return this.shouldDriveStepBeDisabled() || !SelectionStateModel.hasDrive();
   };
 
+  /**
+   * @summary Open a drive info modal
+   * @function
+   * @public
+   *
+   * @returns {Boolean} whether the flash step should be disabled
```
lurch: Copy'n'Paste error?

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -18,7 +18,7 @@
 
 const _ = require('lodash');
 
-module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateModel, WarningModalService) {
+module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateModel, SettingsModel, WarningModalService) {
```
lurch: Is this change needed?

lurch:
I hope you don't think I'm being overly harsh, I just have a very good eye for detail ;-)

It's nice to see the removal of the "Unsafe Mode" :+1:

lurch:
P.S. Looks like the CI tests are failing because the unit tests need to be updated.

lib/gui/components/drive-selector/controllers/drive-selector.js
Shou: Odd, I think `npm test` should've warned me about that?

lib/gui/pages/main/controllers/main.js
Shou: Indeed! I'll get that fixed.

lib/gui/components/drive-info-modal/templates/drive-info-modal.tpl.html
```diff
@@ -0,0 +1,59 @@
...
+      <span class="label label-danger"
+        ng-show="modal.selection.getDrive().system">
+          <i class="glyphicon glyphicon-hdd"></i>
+          SYSTEM DRIVE</span>
+
+    </div>
+  </div>
+</div>
+
+<div class="modal-footer"
```
jviotti: Why does this `DriveInfoModal` component does two things (displaying information and asking for confirmation)? There should be two different component to do different tasks, however, I think you should re-use the`warning-modal` here instead of creating new one.

lib/gui/components/drive-info-modal/services/drive-info-modal.js
```diff
@@ -0,0 +1,52 @@
...
+   *
+   * @param {Object} options - options
+   * @param {String} options.description - danger message
+   * @param {String} options.confirmationLabel - confirmation button text
+   * @param {String} options.rejectionLabel - rejection button text
+   * @fulfil {Boolean} - whether the user accepted or rejected the warning
+   * @returns {Promise}
+   *
+   * @example
+   * WarningModalService.display({
```
jviotti: The example calls `WarningModalService.display` instead of `DriveInfoModalService.display`. In any case, we should re-use `WarningModalService` (extending it if necessary) instead of copy-pasting it here.

lib/gui/components/drive-info-modal/templates/drive-info-modal.tpl.html
```diff
@@ -0,0 +1,59 @@
...
+    <p>{{ ::modal.options.description }}</p>
+  </div>
+  <div class="modal-selected-drive">
+    <div class="modal-selected-drive-description">
+      {{ ::modal.selection.getDrive().description }}
+
+    </div>
+    <div class="modal-selected-drive-name">
+      {{ ::modal.selection.getDrive().name }} - {{ ::modal.selection.getDrive().size | gigabyte | number:1 }} GB
+
```
jviotti: There are two blank lines here.

lib/gui/components/drive-info-modal/templates/drive-info-modal.tpl.html
```diff
@@ -0,0 +1,59 @@
...
+
+      <span class="label label-warning"
```
jviotti: This whole set of labels are being copy-pasted from the drive selector. If we're showing them in two places now, we should extract this logic so it can be re-used between both, otherwise we risk having them out of sync.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -45,7 +50,15 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
 </div>
 
 <div class="modal-footer">
+  <span ng-show="modal.state.getDrive().system" class="modal-footer-warning">
```
jviotti: I'd rename `modal-footer-warning` to something like `modal-footer-label`, since there is nothing in its definition that is specific to a "warning" element:

```css
.modal-drive-selector-modal .modal-footer-warning {
  padding: 8px;
  display: inline-block;
}
```

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -45,7 +50,15 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
+      Etcher has identified this as a system drive
+  </span>
   <button class="button button-primary button-block"
+    ng-class="{ 'label-danger': modal.state.getDrive().system, 'button-primary': !modal.state.getDrive().system }"
```
jviotti: `button-primary` is both in `ngClass` and in the element's `class` attribute. I believe your intention was to but the "danger" class in `ngClass`, such that is gets added when the drive is a system drive, right?

(BTW, the button doesn't go red here for system drives)

screenshot 2016-12-11 20 32 50

jviotti: We should be re-using the `isSystemDrive()` function you added to `SelectionStateModel` instead of `modal.state.getDrive().system`.

lib/gui/models/selection-state.js
```diff
@@ -169,6 +169,28 @@ SelectionStateModel.service('SelectionStateModel', function(DrivesModel) {
...
+   *   device: '/dev/disk2',
+   *   name: 'My Drive',
+   *   size: 123456789,
+   *   protected: true,
+   *   system: true
+   * })) {
+   *   console.log('This drive is a system drive!');
+   * }
+   */
+  this.isSystemDrive = (drive) => {
```
jviotti: This function should contain unit tests.

lib/gui/models/selection-state.js
```diff
@@ -169,6 +169,28 @@ SelectionStateModel.service('SelectionStateModel', function(DrivesModel) {
...
+    return drive.system;
```
jviotti: Can we cast to boolean, just to be on the safe side?

```js
return Boolean(drive.system);
```

lib/gui/models/store.js
```diff
@@ -90,9 +89,17 @@ const storeReducer = (state, action) => {
...
+        return !drive.system;
+      });
+
+      const selectedDevice = newState.getIn([ 'selection', 'drive' ]);
 
-        const drive = _.first(action.data);
+      // There are non-system drives; set one as the selected drive.
+      if (!selectedDevice && noSystemDrives.length >= 1) {
```
jviotti: This new functionality should be unit tested as well.

lib/gui/modules/drive-scanner.js
```diff
@@ -53,13 +53,7 @@ driveScanner.factory('DriveScannerService', (SettingsModel) => {
...
       });
 
-      if (SettingsModel.get('unsafeMode')) {
-        return drives;
-      }
-
-      return _.reject(drives, (drive) => {
-        return drive.system;
-      });
+      return drives;
```
jviotti: This new functionality should be unit tested.

lib/gui/pages/main/controllers/flash.js
```diff
@@ -26,10 +26,50 @@ module.exports = function(
...
+   * FlashController.flashImageToDrive('rpi.img', {
+   *   device: '/dev/disk2',
+   *   description: 'Foo',
+   *   size: 99999,
+   *   mountpoint: '/mnt/foo',
+   *   system: false
+   * });
+   */
+  this.flashImageToDriveModal = (image, drive) => {
+    if (drive.system) {
```
jviotti: Lets reuse `isSystemDrive` here.

lib/gui/pages/main/controllers/flash.js
```diff
@@ -26,10 +26,50 @@ module.exports = function(
...
+        icon: 'alert',
+        title: 'ATTENTION',
+        header: 'FLASH SYSTEM DRIVE?',
+        description: 'All data on your system will be lost.'
+      }).then((userAccepted) => {
+        if (userAccepted) {
+          this.flashImageToDrive(image, drive);
+        } else {
+
+          // DriveSelectionController.openDriveSelector();
```
jviotti: This comment, and empty `else` clause should be removed.

lib/gui/pages/main/controllers/flash.js
```diff
@@ -26,10 +26,50 @@ module.exports = function(
...
+   * @example
+   * FlashController.flashImageToDrive('rpi.img', {
+   *   device: '/dev/disk2',
+   *   description: 'Foo',
+   *   size: 99999,
+   *   mountpoint: '/mnt/foo',
+   *   system: false
+   * });
+   */
+  this.flashImageToDriveModal = (image, drive) => {
```
jviotti: This variable name implies that `this.flashImageToDriveModal` represents the modal itself, and not a function that opens something else (the `DriveInfoModal` component). Maybe we can rename to something like `askFlashConfirmation` or something like that?

jviotti: This function uses `this.flashImageToDrive()`, however its declared before it. Can we move its declaration below?

lib/gui/pages/main/controllers/main.js
```diff
@@ -65,4 +66,22 @@ module.exports = function(
...
+   *
+   * @example
+   * if (MainController.shouldFlashStepBeDisabled()) {
+   *   console.log('The flash step should be disabled');
+   * }
+   */
+  this.openDriveInfo = () => {
+    return DriveInfoModalService.display({
```
jviotti: How does this function know which drive to open?

jviotti: Oh, I see it reads from `SelectionStateModel` internally. A better design would be to decouple `DriveInfoModal` from `SelectionStateModel` and pass the drive to it from here.

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -67,7 +67,11 @@
           <div class="step-selection-text"
             ng-class="{
               'text-disabled': main.shouldDriveStepBeDisabled()
-            }">{{ main.selection.getDrive().name }} - {{ main.selection.getDrive().size | gigabyte | number:1 }} GB</div>
+            }"
+            ng-click="main.openDriveInfo()">{{ main.selection.getDrive().name }} - {{ main.selection.getDrive().size | gigabyte | number:1 }} GB
+            <i class="glyphicon glyphicon-alert"
+              ng-show="main.selection.getDrive().system"></i>
```
jviotti: Lets reuse `isSystemDrive` here.

lib/gui/pages/main/controllers/main.js
```diff
@@ -19,6 +19,7 @@
 module.exports = function(
   SelectionStateModel,
   DrivesModel,
+  DriveInfoModalService,
```
jviotti: The main page should declare the `drive-info-modal` component as a dependency before attempting to use it.

lib/gui/modules/drive-scanner.js
```diff
@@ -32,7 +32,7 @@ const driveScanner = angular.module(MODULE_NAME, [
   require('../models/settings')
```
jviotti: If this component doesn't depend on `SettingsModel`  anymore, then it should remove it from its dependencies.

lib/gui/pages/main/controllers/flash.js
```diff
@@ -26,10 +26,50 @@ module.exports = function(
   FlashErrorModalService,
   ErrorService,
   OSNotificationService,
-  OSWindowProgressService
+  OSWindowProgressService,
+  DriveInfoModalService
```
jviotti: Make sure the main page declares `DriveInfoModal` as a dependency before using it.

lib/gui/pages/main/controllers/main.js
```diff
@@ -65,4 +66,22 @@ module.exports = function(
...
+   * @public
+   *
+   * @returns {Boolean} whether the flash step should be disabled
+   *
+   * @example
+   * if (MainController.shouldFlashStepBeDisabled()) {
+   *   console.log('The flash step should be disabled');
+   * }
+   */
+  this.openDriveInfo = () => {
```
jviotti: Maybe `showDriveInformation` is a better name?

lib/gui/pages/main/controllers/main.js
```diff
@@ -65,4 +66,22 @@ module.exports = function(
...
+    return DriveInfoModalService.display({
+      title: 'Drive details'
+    });
```
jviotti: Lets use the `ErrorService` here to catch any potential error.

jviotti:
This is very nice, however its a huge PR, and it will be difficult to review appropriately. I propose to split this PR into the following changes, so that we can review it more easily, and get it merged one step at a time:

- Show a "SYSTEM DRIVE" badge if the drive is a system drive. This would include the `isSystemDrive` implementation in `SelectionStateModel`.
- Turn the "Continue" button to red in the drive selector and show the warning label if the drive is a system drive.

(Notice that at this point, we are not even showing the system drives by default, so this changes only apply if the user clicked "unsafe mode") 

- Remove the "unsafe mode" option and add the confirmation modal before "flash".
- Show the "drive information" dialog.

***

You can do them in order and keep rebasing this big one once every step is merged.

jviotti:
I'm closing this PR since drive selection has been re-designed on v2.

PR #968, opened 2016-12-09
chore: add support for snapshot builds

jviotti:
This commit introduces a `RELEASE_TYPE` option to the Makefile, which
can be either `production` or `snapshot`. If its `production`, the
behaviour its same as it was before.

If `RELEASE_TYPE` is `snapshot`, the shorter version of the current
commit hash is appended to the version (replacing the
`APPLICATION_VERSION` in the Makefile, as well as the `package.json`
version inside the `.asar`), and the publish scripts get configured to
deploy to different "snapshot" locations.

This commit also reduces the size of the version tag in the bottom right
corner a bit so that it fits when the commit hash is appended to it.


Makefile
```diff
@@ -21,11 +21,25 @@ BUILD_OUTPUT_DIRECTORY = $(BUILD_DIRECTORY)/out
...
+
+# Add the current commit to the version if release type is "development"
+RELEASE_TYPE ?= development
+PACKAGE_JSON_VERSION = $(shell jq -r '.version' package.json)
+ifeq ($(RELEASE_TYPE),production)
+APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)
+endif
+ifeq ($(RELEASE_TYPE),development)
+CURRENT_COMMIT_HASH = $(shell git log -1 --format="%h")
+APPLICATION_VERSION ?= $(PACKAGE_JSON_VERSION)+$(CURRENT_COMMIT_HASH)
```
lurch: Why is this using `?=` ?

docs/PUBLISHING.md
```diff
@@ -75,7 +75,7 @@ Make sure you set the following environment variables:
 Run the following command:
 
 ```sh
-make publish-bintray-debian
+make publish-bintray-debian RELEASE_TYPE=<production|development>
```
lurch: Better to do `RELEASE_TYPE=<production|development> make publish-bintray-debian` ?
(emphasis that it's an env-var)

scripts/build/electron-create-resources-app.sh
```diff
@@ -19,39 +19,46 @@
...
    [ -z "$ARGV_FILES" ] || \
    [ -z "$ARGV_OUTPUT" ]; then
   usage
 fi
 
 mkdir -p "$ARGV_OUTPUT"
 
-cp "$ARGV_SOURCE_DIRECTORY/package.json" "$ARGV_OUTPUT"
+cat "$ARGV_SOURCE_DIRECTORY/package.json" \
+  | jq ".version = \"$ARGV_APPLICATION_VERSION\"" > "$ARGV_OUTPUT/package.json"
```
lurch: ...or you could do:
```bash
jq ".version = \"$ARGV_APPLICATION_VERSION\"" "$ARGV_SOURCE_DIRECTORY/package.json" > "$ARGV_OUTPUT/package.json"
```

docs/PUBLISHING.md
```diff
@@ -92,14 +92,14 @@ access resin.io's production downloads S3 bucket.
 Run the following command to publish a specific file:
 
 ```sh
-./scripts/publish/aws-s3.sh <file>
+./scripts/publish/aws-s3.sh -f <file> -b <bucket> -v <version> -t <production|development>
```
lurch: Given that `<bucket>` never changes, should this be documented as `... -b resin-production-downloads ...` ?

Or maybe it should just be hardcoded back inside `aws-s3.sh` rather than being a command-line option? (much like the way that `bintray-debian.sh` hardcodes `PACKAGE_COMPONENT=etcher`)

scripts/publish/bintray-debian.sh
```diff
@@ -37,29 +67,25 @@ if [ -z "$BINTRAY_USER" ] || [ -z "$BINTRAY_API_KEY" ]; then
...
+  PACKAGE_DISTRIBUTION=devel
+else
+  echo "Invalid release type: $ARGV_RELEASE_TYPE" 1>&2
   exit 1
 fi
 
-if [ -z $PACKAGE_ARCH ]; then
-  echo "Couldn't infer the architecture from the package file name" 1>&2
-  exit 1
-fi
```
lurch: Any reason not to keep this detect-arch-from-the-filename logic?

lurch:
To me, a 'development build' implies extra debug functionality, so I wonder if what you've termed 'development' here should be instead be renamed to 'snapshot'?

Should TravisCI be updated to build both the 'production' and 'development' releases?

lurch:
I've just done some testing of this on Linux, and IMHO it still needs some additional tweaks to separate out the two builds...

As a test, I did
```bash
make clean
RELEASE_TYPE=development make electron-installer-appimage electron-installer-debian
RELEASE_TYPE=production make electron-installer-appimage electron-installer-debian
cp -a release release-clean-development-production
```
and then
```bash
make clean
RELEASE_TYPE=production make electron-installer-appimage electron-installer-debian
RELEASE_TYPE=development make electron-installer-appimage electron-installer-debian
cp -a release release-clean-production-development
```
This _should_ give me two directories `release-clean-development-production` and `release-clean-production-development` which are largely identical. However running Meld over the two directories (and ignoring the obvious log-file differences) shows these files as being different, when they probably shouldn't be?
```
Etcher-linux-x64/version
electron-linux-x64-app/package.json
```

Also, `release-clean-development-production/out/` doesn't contain `etcher-electron_1.0.0~beta.17_amd64.deb`, and `release-clean-production-development/out/` doesn't contain `etcher-electron_1.0.0-beta.17+e249d15_amd64.deb` :exclamation:

Makefile

docs/PUBLISHING.md
jviotti: The idea behind it is that the `<environment variable>=<value> make <target>` form will not work in Windows under something like `cmd.exe`, since environment variables have to be declared with `set`. The way it is documented now works across all OSes, which means we don't have to write different instructions for Windows.

scripts/build/electron-create-resources-app.sh
jviotti: Ah, much nicer!

docs/PUBLISHING.md
jviotti: I'm trying to not hardcode anything at all anywhere. I propose to decouple `bintray-debian.sh` from `etcher` instead.

scripts/publish/bintray-debian.sh
jviotti: For simplicity purposes. We have this information already on the Makefile, and getting it from there is way more reliable than from the filename (which could be anything, and our detection logic could fail very weirdly on some specially crafted file names).

docs/PUBLISHING.md
lurch: Hmmm, okay.

docs/PUBLISHING.md
lurch: Haha, says the man who once said he wanted as little logic in the `Makefile` as possible :wink:

jviotti:
> To me, a 'development build' implies extra debug functionality, so I wonder if what you've termed 'development' here should be instead be renamed to 'snapshot'?

Oh, nice. I like the word "snapshot" for this. I'll update every reference to point to that.

> Should TravisCI be updated to build both the 'production' and 'development' releases?

The problem is that we still can't make development builds for x86 in Travis CI (apart from the PR in progress), because AppImageAssistant can only inject a valid runtime of the architecture its running in, however this will change very soon on `appimagetool`. Until then, lets keep this functionality on the Makefile, but not make use of it on the CI servers yet.

docs/PUBLISHING.md
jviotti: Hahaha, well, they are variable declarations, which I guess don't count as logic very much :)

jviotti:
> This should give me two directories release-clean-development-production and release-clean-production-development which are largely identical. However running Meld over the two directories (and ignoring the obvious log-file differences) shows these files as being different, when they probably shouldn't be?
> Also, release-clean-development-production/out/ doesn't contain etcher-electron_1.0.0~beta.17_amd64.deb, and release-clean-production-development/out/ doesn't contain etcher-electron_1.0.0-beta.17+e249d15_amd64.deb ❗️

Hm, weird. Let me try that myself. Do you still have those directories in your system? Mind posting a `diff` of the `version` and `package.json` files?

lurch:
Sorry, I deleted them because they were taking up too much disk space. But IIRC in the release-clean-development-production directory, version contained `1.0.0-beta.17+e249d15`, and in the release-clean-production-development directory it contained `1.0.0-beta.17`  (i.e. it looks like they got created on the first run, but then never got updated on the second run).

lurch:
> The problem is that we still can't make development builds for x86 in Travis CI

Huh, what does building both 'production' and 'snapshot' releases on Travis have to do with x86 builds?

jviotti:
> Huh, what does building both 'production' and 'snapshot' releases on Travis have to do with x86 builds?

lol, totally non-sense, sorry about that, it was too late at night for me already :) We've been discussing this a bit with the dev-ops team, and decided on the following way forward: the `resin-io/etcher` repository will only produce un-signed snapshot builds. We'll also maintain a private fork of etcher called `resin-io/etcher-release`, or something like that, that only produces signed production builds, therefore releasing will mean updating the release fork to trigger a release deployment.

The reason behind this is that we can better control who have access to the "production release" repository (because it will have the certificates), while still be able to give more access to Etcher contributors for development purposes.

lurch:
Hmm, what advantage does that have over the current solution of having env-vars specify / point to the certificate files?

jviotti:
> Hmm, what advantage does that have over the current solution of having env-vars specify / point to the certificate files?

Do you mean having the fork vs secure environment variables in Travis CI? If so, the problem is that any contributor we add to Etcher will have write access to the repo, and therefore can create pull-requests that can read the secure environment variable during the builds, and therefore a specially crafted PR is enough to get access to our certificates. The certificates will then be on the fork, and we have very limited access to the private fork.

> EDIT: As I (sort of) discussed here I'd be quite happy with the 'default' make targets to always create unsigned builds, and for there to be 'special' make targets to create signed builds (which I guess means we could 'upgrade' the "code-signing env-vars missing" warning to an error).

Sounds good to me!

jviotti:
@lurch I believe I fixed the errors you were facing when doing snapshot + production builds one after the others. Let me know!

lurch:
"secure environment variables in Travis CI"

Ahh, I didn't realise that that was what the fork-discussion was related to ;) (and that env-vars in Travis are easy to leak)

Would it be better for the etcher-with-code-signing-secrets to have the current etcher repo as a submodule, rather than having etcher-with-code-signing-secrets be a fork of the current etcher repo?
I'm sure you know what you're doing though! :+1: :)

jviotti:
That's certainly an option. I guess we'll figure it out when making it. Having it as a submodule it an option, and it can be done with a small shell script, however the idea is that if we do things right, we can re-use the exact same Travis CI and Appveyor CI configuration files, and they will sign depending on what gets scanned from the environment, so we don't have to maintain any extra code at all.

Makefile
```diff
@@ -154,15 +168,17 @@ $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/bower
 	| $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
 	./scripts/build/dependencies-bower.sh -p -x $|
 
-$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app: \
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH)-app: \
```
lurch: I guess the downside of changing `electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app` to `electron-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH)-app` (with the `$(APPLICATION_VERSION)` in the default `snapshot` builds containing the git hash), is that every time you do a `make` after changing commits, you'll end up with an entirely separate directory, which kinda destroys the artefact-caching that Make is designed to give you :-(  (and unless you run `make clean` regularly, you'll also end up with a plethora of slightly-differently-named directories in `$(BUILD_DIRECTORY)`)

I know you said that you wanted to keep "as much as possible" in the shell scripts and out of the `Makefile`, but in this circumstance I wonder if a better approach might be to revert the directory names back to `electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app` and then have some build-step in the Makefile which force-updates the `version` and `package.json` files as necessary? (in fact, to not screw up the timestamps, it'd probably be better to have a shell script which compared the existing contents of `version`, and didn't update it if it doesn't need to).
I'll have a play and see what I come up with...

jviotti: > every time you do a make after changing commits, you'll end up with an entirely separate directory, which kinda destroys the artefact-caching that Make is designed to give you

Well, if you change commits then the actual package contents is supposed to have change, so I guess that's intended?

> I'll have a play and see what I come up with...

Looking forward to that!

scripts/publish/bintray-debian.sh
```diff
@@ -37,29 +71,25 @@ if [ -z "$BINTRAY_USER" ] || [ -z "$BINTRAY_API_KEY" ]; then
...
+PACKAGE_COMPONENT=$(echo "$ARGV_APPLICATION_NAME" | tr '[:upper:]' '[:lower:]')
```
lurch: Rather than doing the `tr` here, it's probably better to just pass in `$(APPLICATION_NAME_LOWERCASE)` from the Makefile?

And I guess you _could_ pass in `$(APPLICATION_VERSION_DEBIAN)` too?

scripts/publish/aws-s3.sh
```diff
@@ -22,16 +22,54 @@ set -e
...
+then
+  usage
 fi
 
-ETCHER_VERSION=`node -e "console.log(require('./package.json').version)"`
-S3_BUCKET="resin-production-downloads"
+FILENAME=$(basename "$ARGV_FILE")
+
+if [ "$ARGV_RELEASE_TYPE" == "production" ]; then
+  ROOT_KEY=etcher
```
lurch: What was it you were saying about not hard-coding Etcher-specific strings? ;-)

scripts/publish/bintray-debian.sh
jviotti: > Rather than doing the tr here, it's probably better to just pass in $(APPLICATION_NAME_LOWERCASE) from the Makefile?

I thought about that, however that would mean we have to document in the options that we're expecting a lower case version of the name, which would be something easy to miss as a developer. As a consequence, we'd have the transformation "just in case", and if we're having that anyway, we rather pass `APPLICATION_NAME` from the Makefile, which is more readable and in-line with the rest of the scripts rather than `APPLICATION_NAME_LOWERCASE`, since they'd have the same effect anyway.

> And I guess you could pass in $(APPLICATION_VERSION_DEBIAN) too?

I agree with that! I'll update the PR!

scripts/publish/aws-s3.sh
jviotti: Hahaha, good catch! I'll fix in a minute.

jviotti: Done! I'm now accepting a "product name" option here. Let me know if you think we can rename this to something better!

scripts/publish/bintray-debian.sh
lurch: Hmm, the angle I was looking at it from, is that you'd want to pass the "exact" `$PACKAGE_COMPONENT` into the script, rather than having the `bintray-debian.sh` modify it behind your back? The `.deb` filename itself already uses lower-case `etcher` too.

jviotti: Oh, I see your point now. Yeah, makes a lot of sense. Done!

PR #969, opened 2016-12-10
chore: fix `check-dependency.sh` to work with arguments containing spaces

lurch:
It was much easier to just use `"$@"` directly (which is only used in one place anyway), than it was to try and copy the `"$@"` array to a differently-named array variable.

### DIFF
--- a/scripts/build/check-dependency.sh
+++ b/scripts/build/check-dependency.sh
@@ -17,18 +17,18 @@
 ###
 
 set -e
-set +u
-ARGV_DEPENDENCIES=$*
 set -u
 
-if [ -z "$ARGV_DEPENDENCIES" ]; then
+if [ "$#" -lt 1 ]; then
   echo "Usage: $0 <dependency...>"
   exit 1
 fi
 
 RESULT=""
+DEPENDENCIES=""
 
-for dependency in $ARGV_DEPENDENCIES; do
+for dependency in "$@"; do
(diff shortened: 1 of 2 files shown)


scripts/build/check-dependency.sh
```diff
@@ -17,26 +17,30 @@
...
   if command -v $dependency 2>/dev/null 1>&2; then
     RESULT=$dependency
     break
   fi
 done
 
 if [ -z "$RESULT" ]; then
-  echo "Dependency missing: $ARGV_DEPENDENCIES" 1>&2
+  if [ "$#" -eq 1 ]; then
+    echo "Dependency missing:$DEPENDENCIES" 1>&2
```
jviotti: Can we add a space after the colon here and below for consistency?

jviotti:
Awesome! I left a small comment regarding to spacing, but I'm approving anyway.

scripts/build/check-dependency.sh
lurch: Ahhh, there's actually a reason for the no-space-after-the-colon... ;-)
`DEPENDENCIES="$DEPENDENCIES $(echo $dependency | cut -d ' ' -f 1)"`
means that there's always a space at the start of `$DEPENDENCIES` (because it's empty the first time around the loop), so if we added a space after the colon here, we'd end up displaying a double-space, which IMHO looked weird.

The reason we do the `cut` and build the `$DEPENDENCIES` this way, is so that `check-dependency.sh 'foo arg' "bar -l -t wub"` errors with: `No dependency found from: foo bar`.

PR #970, opened 2016-12-10
chore: fix error functions in Makefile

lurch:
`$(error ...)` (and `$(warning ...)`) functions in a Makefile can't be indented - if they are make displays a `*** commands commence before first target. Stop.` error.


jviotti:
Ah, weird it works on OS X's Make :)

PR #971, opened 2016-12-12
chore: add `distclean` target to Makefile

lurch:
This new target also deletes the `node_modules` and `bower_components`
directories, taking the project directory to a totally pristine state.


lurch:
If / when we add download caching, the download cache should also be cleared by this rule.

PR #974, opened 2016-12-13
doc: recommend `make electron-develop` to clean dependency tree

jviotti:
Because of NPM shrinkwrap, we recommend having a clean dependency tree
in `CONTRIBUTING.md` before upgrading a dependency. We currently
recommend maually wiping out `node_modules`, but since the
`electron-develop` make target already does that, is more user friendly
to recommend that instead.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -37,7 +37,7 @@ dependency.
 
 Use the following steps to ensure everything goes flawlessly:
 
-- Delete your `node_modules/` to ensure you don't have extraneous dependencies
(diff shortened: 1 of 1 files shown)


lurch:
Hmm, is this really the best advice?? `electron-develop` does a _lot_ more than just delete `node_modules` (it's always quite a slow process for me).
Maybe it'd be better to recommend `make distclean` ( #971 ) instead?

jviotti:
@lurch The documentation was currently inaccurate (I just detected that), since you actually need to re-install the dependencies before moving to the next step (which `make electron-develop`) does.

lurch:
Aha, cool :)

PR #976, opened 2016-12-14
CLI drive selection

stefan-mihaila:
Still need to write some unit tests, but I wanted to open the PR so I can get some feedback.

screen shot 2016-12-14 at 16 49 08
screen shot 2016-12-14 at 16 49 21


lib/cli/etcher.js
```diff
@@ -18,62 +18,114 @@
...
+  }
+  return choice;
+};
+
+const mustConfirm = !options.robot && !options.yes;
+
+Bluebird.all([
+  isElevated(),
+  imageStream.getImageMetadata(options._[0])
+]).then((results) => {
```
stefan-mihaila: I would've loved to use rest parameters destruction here `then(...[elevated, imageMetadata')`, but `ESLint` complains.

npm-shrinkwrap.json
```diff
@@ -101,6 +101,23 @@
       "from": "any-promise@>=1.1.0 <2.0.0",
       "resolved": "https://registry.npmjs.org/any-promise/-/any-promise-1.3.0.tgz"
     },
+    "appdmg": {
```
stefan-mihaila: I don't know why this changes are here, I just added `inquirer` and removed `resin-cli-form` in `package.json`. @jviotti can you tell me if it should be this way?

jviotti: Did you run `npm run shrinkwrap` after installing the dependencies? That script is exactly meant to remove these kind of things.

npm-shrinkwrap.json
```diff
@@ -1792,6 +1826,11 @@
       "from": "fs-temp@>=1.0.0 <2.0.0",
       "resolved": "https://registry.npmjs.org/fs-temp/-/fs-temp-1.1.0.tgz"
     },
+    "fs-xattr": {
```
jviotti: This one should have been removed by `npm run shrinkwrap` as well.

lib/cli/etcher.js
```diff
@@ -18,61 +18,115 @@
...
-      device: answers.drive
+      device: drive
     });
 
     if (!selectedDrive) {
-      throw new Error(`Drive not found: ${answers.drive}`);
+      throw new Error(`Drive not found: ${drive}`);
+    }
+
+    if (selectedDrive.system && mustConfirm) {
```
jviotti: This should re-use logic from `SelectionStateModel`.

lib/cli/etcher.js
```diff
@@ -18,61 +18,115 @@
...
+      return inquirer.prompt([ {
```
jviotti: Lets try to refactor this so there is a single `inquirer.prompt()` invocation in `etcher.js`. Conditional questions can be made with `when()`.

lurch:
I realize that things are currently in flux ( #888 ), but for this initial version of the PR I wonder if the "labels" (`FIXED DRIVE`, `NON-REMOVABLE DRIVE`, etc.) should be matched up with what the GUI displays?

lib/cli/drive-scanner.js
```diff
@@ -0,0 +1,111 @@
...
+
+'use strict';
+
+const drivelist = require('drivelist');
+const EventEmitter = require('events').EventEmitter;
+const _ = require('lodash');
+const Rx = require('rx');
+
+const interval = new WeakMap();
+const prevPollDrives = new WeakMap();
```
jviotti: What is the reason to use maps if we only get/set from/to a single key (`this`)?

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+const inquirer = require('inquirer');
+const Choices = require('inquirer/lib/objects/choices');
+
+const updateChoices = function(choices) {
+  this.opt.choices = new Choices(choices);
+  this.render();
+};
+
+/**
+ * Prompt for `inquirer` that adds dynamic update functionality to the `list` prompt.
```
jviotti: This should be prefixed with `@summary`.

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @class
+ * @public
+ * @extends inquirer.prompt.prompts.list
+ *
+ * @example
+ *
```
jviotti: Any reason for this while line? If we decide to process the JSDoc from this file, there would be a weird blank line on top.

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+ * inquirer.registerPrompt('dynamicList', DynamicList);
+ *
+ * inquirer.prompt([{
+ *   message: ''What size do you need?'',
+ *   type: 'dynamicList',
+ *   choices: ['Jumbo', 'Large', 'Standard', 'Medium', 'Small', 'Micro'],
+ *   emitter: scanner,  // Event emitter with events `add` and `remove` and methods `start` and `stop`
+ *   name: 'size',
+ *   mapFn: _.identity  // mapFn is used transform one choice from the emitter into a `list` choice
+ *  });
```
jviotti: The indentation here is a bit off.

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+ * @class
+ * @public
+ * @extends inquirer.prompt.prompts.list
+ *
+ * @example
+ *
+ * inquirer.registerPrompt('dynamicList', DynamicList);
+ *
+ * inquirer.prompt([{
+ *   message: ''What size do you need?'',
```
jviotti: There are double single quotes here.

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+ *   type: 'dynamicList',
+ *   choices: ['Jumbo', 'Large', 'Standard', 'Medium', 'Small', 'Micro'],
+ *   emitter: scanner,  // Event emitter with events `add` and `remove` and methods `start` and `stop`
```
jviotti: Here is says that the scanner should emit `add` and `remove`, but the current implementation emits `change`.

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+ * @extends inquirer.prompt.prompts.list
+ *
+ * @example
+ *
+ * inquirer.registerPrompt('dynamicList', DynamicList);
+ *
+ * inquirer.prompt([{
+ *   message: ''What size do you need?'',
+ *   type: 'dynamicList',
+ *   choices: ['Jumbo', 'Large', 'Standard', 'Medium', 'Small', 'Micro'],
```
jviotti: Can we add spaces between the items and the brackets for consistency with the remaining of the codebase?

```js
[ 'Jumbo', 'Large', 'Standard', 'Medium', 'Small', 'Micro' ]
```

lib/cli/drive-scanner.js
```diff
@@ -0,0 +1,111 @@
...
+    })
+    .pausable(new Rx.Subject());
+};
+
+const emitChangeEvent = function(drives) {
+  const prevDrives = prevPollDrives.get(this);
+  if (prevDrives !== null && _.isEqual(drives, prevPollDrives)) {
+    return;
+  }
+  this.emit('change', drives, prevDrives);
```
jviotti: Can we make this function not rely on the current context?

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+ *   emitter: scanner,  // Event emitter with events `add` and `remove` and methods `start` and `stop`
+ *   name: 'size',
+ *   mapFn: _.identity  // mapFn is used transform one choice from the emitter into a `list` choice
+ *  });
+ */
+class DynamicList extends inquirer.prompt.prompts.list {
+
+  /**
+   * This is called by `inquirer`
```
jviotti: See https://github.com/resin-io/resin-cli-visuals/blob/master/lib/widgets/progress.coffee for a JavaScript class JSDoc annotation example. The annotation that contains `@class` should be actually on the constructor. The methods should contain `@method`, access types (e.g: `@public`), etc.

lib/cli/drive-scanner.js
```diff
@@ -0,0 +1,111 @@
...
+
+    drivesStream.get(this).subscribe((drives) => {
+      drives.sort(driveCompareFn);
+      Reflect.apply(emitChangeEvent, this, [ drives ]);
+      prevPollDrives.set(this, drives);
+    });
+  }
+
+  /**
+   * @summary Start scanning drives.
```
jviotti: What is the access type of these methods?

lib/cli/dynamic-list.js
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * inquirer.registerPrompt('dynamicList', DynamicList);
+ *
+ * inquirer.prompt([{
+ *   message: ''What size do you need?'',
+ *   type: 'dynamicList',
+ *   choices: ['Jumbo', 'Large', 'Standard', 'Medium', 'Small', 'Micro'],
+ *   emitter: scanner,  // Event emitter with events `add` and `remove` and methods `start` and `stop`
+ *   name: 'size',
+ *   mapFn: _.identity  // mapFn is used transform one choice from the emitter into a `list` choice
```
jviotti: I think we're missing one piece in this design, which the actual "drive selection" widget. We have an inquirer "dynamic list" and a "drive scanner". The "drive selection" widget would be the one that ties them up, since `etcher.js` is currently dealing with the low-level details of that.

npm-shrinkwrap.json
stefan-mihaila: I thought I did, but... err... nevermind 🙃

jviotti:
@lurch I agree. As a consequence, we need to have a function to return "labels" that can be re-used between the CLI and the GUI.

lib/cli/drive-scanner.js
stefan-mihaila: They're private instance variables. (Different instances of `DriveScanner` would have access to different values) My first try was to use private by convention (`this._interval`), but dangling underscores aren't permitted in `ESLint` so I assumed you don't want them.

stefan-mihaila:
@lurch At the moment there is no label for fixed drives in GUI, is there? Should I remove it in the CLI?
screen shot 2016-12-14 at 18 36 05

jviotti:
That is because we don't show system drives at all on the GUI, but we will in the future. See https://github.com/resin-io/etcher/issues/888. The label will be "SYSTEM DRIVE".

stefan-mihaila:
@jviotti We show them with the "unsafe" or what's its name option enabled in settings.
What labels we want from the CLI in this PR?

jviotti:
Let’s put in this PR the ones that currently appear on the GUI.

--
Juan Cruz Viotti
Software Engineer, Resin.io (https://resin.io)

lib/cli/drive-scanner.js
stefan-mihaila: What do you mean by only one key? Different instances of `DriveScanner` would use different keys in the map, won't them?

lib/cli/drive-scanner.js
stefan-mihaila: I can move it in the class. The reason I extracted it is because I thought it would make the constructor more readable. Once again, I created it having a private method in mind and it made sense to me to be here as one of the `DriveScanner`'s private methods, and it looked like it won't cause damage because this whole module exists just to accomodate the `DriveScanner`.
On the other hand, I admit that I don't know if that's the way to do things in JS or not, maybe it's an anti-pattern. I think I need to read some more on the subject of private instance methods and variables in JS best practices and patterns.

lib/cli/dynamic-list.js
stefan-mihaila: I'll get rid of it

lib/cli/etcher.js
```diff
@@ -18,61 +18,115 @@
...
+    choice.disabled = 'TOO SMALL FOR IMAGE';
+  }
+  return choice;
+};
+
+const mustConfirm = !options.robot && !options.yes;
+
+Bluebird.all([
+  isElevated(),
+  imageStream.getImageMetadata(options._[0])
```
stefan-mihaila: Wasteful in what sense? My thought process was that, doing it like this would be faster (as in more responsive for the user) in the positive case.

jviotti: The `isElevated` function is really cheap to execute. Its code is just:

```
return process.getuid && process.getuid() === 0;
```

So the savings we get from parallelizing this are minimal, and if the user indeed lacks privileges, we have to wait for the `getImageMetadata()` function, which is way more expensive, just to be able to say to the user that he doesn't have permissions to proceed.

My way of thinking is that since `isElevated` is so cheap, doing `isElevated` and `.getImageMetadata()` in sequence is unnoticeable from doing them in parallel, and still provides a "fast failure" for the case where the user lacks the privileges.

stefan-mihaila:
@jviotti I `git push --force` without realising it will screw up all comments in the added files. I tried to revert and push force again, but it looks like it didn't do the trick. At the moment not all your comments are addressed, I will notify you when they are so you can review again.

> Let’s put in this PR the ones that currently appear on the GUI.

Do you want me to merge some other branch here and take over?

lib/cli/etcher.js
```diff
@@ -18,61 +18,115 @@
...
+const DynamicList = require('./dynamic-list');
+const DriveScanner = require('./drive-scanner');
 const EXIT_CODES = require('../src/exit-codes');
 
-isElevated().then((elevated) => {
+inquirer.registerPrompt('dynamicList', DynamicList);
+
+const scanner = new DriveScanner(2000);
+
+const driveToChoice = (drive, minSizeToEnable) => {
```
stefan-mihaila: Not sure if `drive-scanner` should care about image size either...

lib/cli/etcher.js
```diff
@@ -18,61 +18,115 @@
...
+
+  if (drive.system) {
+    displayName += chalk.red(' FIXED DRIVE');
+  }
+
+  const choice = {
+    name: displayName,
+    value: drive.device
+  };
+  if (drive.size < minSizeToEnable) {
```
stefan-mihaila: It permits it actually, as it displays the disabled label iff drive size is strictly lesser than the image size.
Anyway, by reusing logic from the `SelectionStateModel` you meant actually importing it here? Or maybe refactoring some useful methods to a `common` module?

lib/cli/etcher.js
```diff
@@ -18,61 +18,115 @@
...
+  return choice;
+};
+
+const mustConfirm = !options.robot && !options.yes;
+
+Bluebird.all([
+  isElevated(),
+  imageStream.getImageMetadata(options._[0])
+]).then((results) => {
+  const elevated = results[0];
```
stefan-mihaila: Nice one, thanks

lib/cli/etcher.js
stefan-mihaila: Not sure what logic I can reuse for this condition.

jviotti:
> @jviotti I git push --force without realising it will screw up all comments in the added files. I tried to revert and push force again, but it looks like it didn't do the trick. At the moment not all your comments are addressed, I will notify you when they are so you can review again.

You can still reveal the "outdated" comments from the GH UI:

screenshot 2016-12-16 11 26 39

> Do you want me to merge some other branch here and take over?

I mean lets put in the CLI all the labels that exist currently on the GUI.

lib/cli/etcher.js
jviotti: Oops, my bad. `SelectionStateModel.isSystemDrive()` was currently in a branch. I got merged very recently (https://github.com/resin-io/etcher/pull/978), so you can rebase to get it.

lib/cli/etcher.js
jviotti: Oh, I see, my bad.

> Anyway, by reusing logic from the SelectionStateModel you meant actually importing it here? Or maybe refactoring some useful methods to a common module?

It'd be really cool if we could re-use it as it is, however `SelectionStateModel` is an Angular service, and can't be used outside such framework. I propose extracting all the "static" functions that module contains (that don't interact with the application state), such as `isSystemDrive()`, `isDriveLocked()`, etc; as a separate NodeJS module we can also import in the CLI.

Some things to keep in mind:

- These functions contain tests, so the tests will have to be extracted as well.
- The extracted module will still need to be exposed as an Angular service so the Angular components that currently use these functions can still do so (you can collaborate with Shou on this one).

The current approach to re-usable code between the GUI and CLI is to put it on `lib/src` (I know the name is not ideal, we should probably rename to `lib/shared`), so the extracted functionality can live there, and then there can be another small module in `lib/gui` that simply requires that logic and re-exposes it as an Angular service.

In the future, we want to have a re-usable module that handles the "labels" we put to drives (such as system, locked, etc) and also be able to automatically compute a list of them (so we don't hardcode the label names anywhere), so that would be the perfect place to include that.

Let's make those changes as another PR, so its easier to review, and then you can rebase here.

lib/cli/etcher.js
jviotti: Yeah, we should find a way to have all the logic necessary to display the drive options in the drive selection widget, while still have the knowledge about the image in the main file.

lurch:
> all the labels that exist currently on the GUI.

Which, to be fair, is currently a moving target...

jviotti:
Yeah, let's try our best to keep this updated. I expect this PR to be merged soon, so that shouldn't be a major problem.

stefan-mihaila:
@jviotti I've update the PR. One observation is that currently, with "Unsafe mode" enabled, the GUI doesn't prompt the user for any confirmation when flashing an image to a "SYSTEM DRIVE" (non removable actually). It looked to me to dangerous to not prompt the user for confirmation in the CLI (at least in the GUI the user has to activate "unsafe mode", there is no such thing in the CLI)

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+const driveToChoiceFn = (image) => {
+
+  /**
+   * Map a drive object to a `inquirer` choice object
+   * @param {object} drive - Drive object as returned by `drivelist`
+   * @private
+   * @returns {{name: string, value: object}}
+   */
+  const driveToChoice = (drive) => {
+    const size = drive.size / 1000000000;
```
jviotti: We have this exact same logic in `lib/gui/utils/byte-size/filters/gigabyte.js`. Maybe worth moving to `lib/shared`? :)

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
+/*
```
jviotti: I'm not going to comment on each one, but the JSDoc should contain their title preceded by `@summary`, the types should be capitalized (e.g: `Number`, `Object`), every annotation should contain the access type, and the type of annotation (e.g: `@function`), and each function should contain an `@example`.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+
+    const driveInfo = `${drive.device} (${size.toFixed(1)} GB) - ${drive.description}`;
+
+    const systemDriveText = chalk.red(constraints.labels.SYSTEM_DRIVE);
+    const systemDriveDisplay = constraints.isSystemDrive(drive) ? systemDriveText : '';
+
+    const notRecommendedSizeText = chalk.yellow(constraints.labels.NOT_RECOMMENDED);
+    const recommendedSizeDisplay = constraints.isDriveSizeRecommended(drive, image) ? '' : notRecommendedSizeText;
+
+    const displayName = `${driveInfo} ${systemDriveDisplay} ${recommendedSizeDisplay}`;
```
jviotti: If the drive is not a system drive `systemDriveDisplay` will be an empty string. If the drive size is not recommended, this will result in the following string (notice the double spaces after the drive description):

```
/dev/disk2 (2 GB) - Foo Bar  NOT RECOMMENDED
```

jviotti: Same for the converse scenario. If the drive is a system one and the size is recommended, there will be trailing spaces.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+
+    const choice = {
+      name: displayName,
+      value: drive
+    };
+
+    if (!constraints.isDriveLargeEnough(drive, image)) {
+      choice.disabled = constraints.labels.TOO_SMALL;
```
jviotti: Does the `disabled` property have any meaning? Why is it assigned to a string? I can't find it on the docs.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+  const confirmWriteToNotRecommendedDrive = (answers) => {
+    if (!confirmationRequired || !answers.yes) {
+      return false;
+    }
+    return !constraints.isDriveSizeRecommended(answers.drive, image);
+  };
+
+  initialDrives.sort(driveComparer);
+
+  return inquirer.prompt([ {
```
jviotti: Let's open the object in a separate line.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+    type: 'dynamicList',
+    name: 'drive',
+    choices: initialDrives.map(driveToChoiceFn(image)),
+    emitter: scanner,
+    mapFn: driveToChoiceFn(image),
+    sort: true,
+    compareFn: (choice1, choice2) => {
+      return driveComparer(choice1.value, choice2.value);
+    }
+  }, {
```
jviotti: Let's open the objects in separate lines here and below as well.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+    when: confirmWriteToFixedDrive
+  }, {
+    message: (answers) => {
+      return constraints.labels.unrecommendedSizeWarningText(answers.drive, image);
+    },
+    type: 'confirm',
+    name: 'yes',
+    default: false,
+    when: confirmWriteToNotRecommendedDrive
+  } ]).then((answers) => {
```
jviotti: The closing `]` should be on a separate line.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+      return driveComparer(choice1.value, choice2.value);
+    }
+  }, {
+    message: 'This will erase the selected drive. Are you sure?',
+    type: 'confirm',
+    name: 'yes',
+    default: false,
+    when: _.constant(confirmationRequired)
+  }, {
+    message: `You have selected a ${chalk.red('NON REMOVABLE DRIVE')}. Do you want to continue?`,
```
jviotti: I wonder if we should re-use the `SYSTEM_DRIVE` label here.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Map a drive object to a `inquirer` choice object
+   * @param {object} drive - Drive object as returned by `drivelist`
+   * @private
+   * @returns {{name: string, value: object}}
+   */
+  const driveToChoice = (drive) => {
+    const size = drive.size / 1000000000;
+
+    const driveInfo = `${drive.device} (${size.toFixed(1)} GB) - ${drive.description}`;
```
jviotti: Can we extract all the logic to create a descriptive line for a drive (including labels, etc) to a private exported function in this file, so we can unit test it?

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+const _ = require('lodash');
+
+const DynamicList = require('./dynamic-list');
+const scanner = require('./drive-scanner');
+const driveComparer = require('../shared/utils').driveComparer;
+const constraints = require('../shared/drive-constraints');
+
+inquirer.registerPrompt('dynamicList', DynamicList);
+
+const driveToChoiceFn = (image) => {
```
jviotti: This function should contain a JSDoc annotation. Also, let's drop the `Fn` prefix.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Map a drive object to a `inquirer` choice object
+   * @param {object} drive - Drive object as returned by `drivelist`
+   * @private
+   * @returns {{name: string, value: object}}
```
jviotti: I don't think this is valid JSDoc. You can say `@returns {Object} an inquirer choice object`. Users wanting to know more about what that means can look at Inquirer's documentation, at the function example, or at the unit tests.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+      choice.disabled = constraints.labels.LOCKED;
+    }
+
+    return choice;
+  };
+
+  return driveToChoice;
+};
+
+module.exports = function(initialDrives, confirmationRequired, image) {
```
jviotti: Can we not export the function directly, but export it as property, like `exports.display()`. The benefit of the latter case is that we can stub this function on the client's code if needed, while a plain function can't be stubbed.

jviotti: Do you think the `initialDrives` is really required? As far as I can see, the drive scanner will scan right away, so we can safely delay showing the list until we receive the first set of data (which should be just a couple of milliseconds anyway).

jviotti: Can we transform this function signature to take an object instead? For example:

```
{
    image: { ... },
    omitConfirmation: true 
}
```

jviotti: Also, I think its nice to change `confirmationRequired` to the opposite meaning: e.g: `omitConfirmation`. This has the benefit that confirmation is the default (which is a great default), and that you don't have to negate `options.yes` and `robot.isEnabled()` on the client code.

lib/cli/drive-selection-widget.js
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Returns true iff `inquirer` should prompt in the case of a fixed drive selected
+   * @param {object} answers - The answers given so far
+   * @returns {boolean}
+   */
+  const confirmWriteToFixedDrive = (answers) => {
+    if (!confirmationRequired || !answers.yes) {
```
jviotti: This snippet is exactly the same in the function below. Maybe we can extract and re-use? Also, if this piece of code is extracted, the logic would be so simple that I think we could inline it on the inquirer questions below.

jviotti:
> One observation is that currently, with "Unsafe mode" enabled, the GUI doesn't prompt the user for any confirmation when flashing an image to a "SYSTEM DRIVE" (non removable actually). It looked to me to dangerous to not prompt the user for confirmation in the CLI (at least in the GUI the user has to activate "unsafe mode", there is no such thing in the CLI)

Sounds totally reasonable! :+1:

lib/cli/drive-selection-widget.js
jviotti: Actually, this WHOLE logic would be re-used between the CLI and the GUI, and not just the byte to gigabyte function I mentioned above. The GUI displays the exact same thing, but it creates separate HTML elements for the labels.

We can build a generic function that returns an object that looks like this:

```javascript
{
   description: '/dev/disk2 (2 GB) - Foo bar',
   labels: [
      'SYSTEM DRIVE',
      'LOCKED'
   ]
}
```

Then, in the GUI, we can loop between the labels in Angular and create elements, while in the CLI, we can join the label values and append it to the description.

jviotti:
@stefan-mihaila I spotted various pieces of functionality that could be moved to `lib/shared` in other to be re-used between the CLI and the GUI. I think its better to handle those as separate PRs and rebase in order to not grow this into a monster PR :)

jviotti:
I might help you with some of the GUI/CLI `lib/shared` in the mid-time. I'll let you know if I make any of those myself!

lib/cli/drive-selection-widget.js
jviotti: See https://github.com/resin-io/etcher/pull/1021

lib/shared/drive-constraints.js
```diff
@@ -147,3 +147,22 @@ exports.isDriveValid = (drive, image) => {
...
+ * @summary Constant strings used as drive labels
+ * @object
+ * @public
+ */
+exports.labels = {
+  TOO_SMALL: 'TOO SMALL FOR IMAGE',
+  LOCKED: 'LOCKED',
+  SYSTEM_DRIVE: 'SYSTEM DRIVE',
+  NOT_RECOMMENDED: 'NOT RECOMMENDED',
+  unrecommendedSizeWarningText: (drive, image) => {
```
jviotti: See https://github.com/resin-io/etcher/pull/1022

lib/cli/drive-scanner.js
```diff
@@ -0,0 +1,129 @@
...
+ *   console.log(currentDrives , previousDrives);
+ * }
+ *
+ * // Start emitting events
+ * scanner.start()
+ *
+ * // Stop emitting events
+ * scanner.stop()
+ */
+module.exports = Object.create(new EventEmitter(), {
```
stefan-mihaila: `class DriveEmitter extends EventEmitter` was my first choice, but I changed this to be a singleton (https://github.com/resin-io/etcher/pull/976#discussion_r92642941). It's true that I could have made it a singleton just outside this module (as in creating a `DriveEmitter` class and only exporting an instance), but since having multiple instances can be problematic, it seemed right to me to implement it like this.
Regarding composition, looking at [lib/gui/modules/drive-scanner.js](https://github.com/resin-io/etcher/blob/6456a8fe967ffe52f089828c158320d9389cc34d/lib/gui/modules/drive-scanner.js#L114-L114), it uses a nice trick because it returns the emitter. So that's not implemented as a singleton, it's just a function constructor that returns the emitter instead of the implicit `this`, but AngularJS uses it as a singleton (I mean it only creates one instance and uses that instance wherever it's injected). Without this trick, with composition I would also have to use some adaptor methods.

lib/cli/drive-selection-widget.js
stefan-mihaila: Yes, fortunately it works for lists as well as for choices. You can see it used in the `Inquirer.js` examples: https://github.com/SBoudrias/Inquirer.js/blob/master/examples/list.js#L20

lib/cli/drive-scanner.js
```diff
@@ -0,0 +1,129 @@
...
+  stop: {
+    value: () => {
+      previousDrives = null;
+      drivesStream.pause();
+    }
+  }
+});
+
+drivesStream.subscribe((drives) => {
+  if (previousDrives !== null && _.isEqual(drives, previousDrives)) {
```
stefan-mihaila: you're right

lib/cli/drive-scanner.js
```diff
@@ -0,0 +1,129 @@
...
+    return;
+  }
+
+  module.exports.emit('change', drives, previousDrives);
```
stefan-mihaila: yes, we don't use it as an event, so I won't emit it

tests/cli/drive-scanner.spec.js
```diff
@@ -0,0 +1,112 @@
...
+
+const m = require('mochainon');
+const drivelist = require('drivelist');
+
+describe('CLI: DriveScanner', function() {
+
+  describe('given no available drives', function() {
+
+    beforeEach(function() {
+      this.scanner = require('../../lib/cli/drive-scanner');
```
stefan-mihaila: Yes, it doesn't help with anything since it's a singleton anyway.
This brings up a concern. Do tests run in parallel? If I have a test that needs the scanner to not be started to function correctly, and another test starts the scanner, since the scanner is a singleton, will the tests interfere?

stefan-mihaila: Just realised... This can be a problem in code, not in tests. I mean, having a singleton with `start` and `stop` is not a good idea, because it will interact everywhere... Looking into it...

jviotti: The reason for a singleton scanner is that having multiple scanners running at the same time can be *very* troubling, specifically for Windows, which will start throwing cryptic sporadic issues, so a singleton was my way to prevent that from happing at all.

The tests don't run in parallel, so it should be fine. I think that we can keep it as a singleton as long as we have some reliable way to reset the state (if any). Looks like that could be part of `.stop()`.

lib/cli/drive-selection-widget.js
jviotti: Oh, I see. So its some kind of informative message you show to explain why an element is disabled, but it doesn't control whether the option is disabled or not, right?

If so, I think it might be good to add a comment explaining it, since it looks really awkward at first sight (definitely not your fault. The interface exposed by inquirer for this is very weird).

lib/cli/drive-scanner.js
jviotti: > class DriveEmitter extends EventEmitter was my first choice, but I changed this to be a singleton (#976 (comment)). It's true that I could have made it a singleton just outside this module (as in creating a DriveEmitter class and only exporting an instance), but since having multiple instances can be problematic, it seemed right to me to implement it like this.

My thought was that we could use `class DriveEmitter extends EventEmitter` to declare it, but only export `new DriveEmitter()`.

> Regarding composition, looking at lib/gui/modules/drive-scanner.js, it uses a nice trick because it returns the emitter. So that's not implemented as a singleton, it's just a function constructor that returns the emitter instead of the implicit this, but AngularJS uses it as a singleton (I mean it only creates one instance and uses that instance wherever it's injected). Without this trick, with composition I would also have to write some adaptor methods.

NodeJS will not parse a file twice. That means that if you `require()` the same file from multiple modules, they will all get the same instance of it so it should just work, unless I'm missing something?

lib/cli/drive-selection-widget.js
stefan-mihaila: It controls if the option is disabled or not and also displays the informative string. Yes, you're right, it looks awkward, I'll add a comment.

tests/cli/drive-scanner.spec.js
stefan-mihaila: OK, I'm not worried anymore about this in tests, what I'm worried about is that the users of the scanner (us) can't rely on a started scanner to not be stopped in some other part of the application. I'm not saying that it shouldn't be a singleton, just that there's an annoyance with it being one and I was wandering if there's a good way to get around it.

jviotti: Hm I see what you mean. We could either:

- Be careful about our own use with code reviews, etc (error prone)
- Provide some sort of mechanism to prevent this from happening

Any thoughts about the second option? Maybe some sort of lock?

PR #979, opened 2016-12-15
feat(gui): red confirm button for system drives

Shou:
Make the drive-list confirmation button a cautionary red upon selection of
a system drive, and show a descriptive warning label. Also, make the drive
list wider for the label and changes to come.

See: #888
Changelog-Entry: Drive selection cautions user with red confirm button.

### DIFF
--- a/lib/gui/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/components/drive-selector/controllers/drive-selector.js
@@ -18,7 +18,7 @@
 
 const _ = require('lodash');
 
-module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateModel, WarningModalService) {
+module.exports = function($q, $uibModalInstance, DrivesModel, ManifestBindService, SelectionStateModel, WarningModalService) {
 
   /**
    * @summary The drive selector state
@@ -140,6 +140,36 @@ module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateMode
         this.closeModal();
       }
     });
+  };
+
+  this.getButtonStatus = () => {
+    const isDisabled = !this.state.hasDrive() || this.state.getDrive().protected;
+
+    if (isDisabled) {
+      return 'disabled';
+    }
+
+    const isDangerous = this.state.hasDrive() && this.state.isSystemDrive(this.state.getDrive());
+
+    if (isDangerous) {
+      return 'danger';
+    }
+
+    return 'primary';
+  };
+
+  this.getFooterMessage = () => {
+    if (this.getButtonStatus() === 'danger'
+      && this.state.isSystemDrive(this.state.getDrive())) {
+      return ManifestBindService.get('displayName') + ' has identified this as a system drive!';
+
+    }
+
+    return undefined;
+  };
+
+  this.hasFooterMessage = () => {
+    return Boolean(this.getFooterMessage());
   };
 
 };
--- a/lib/gui/components/drive-selector/styles/_drive-selector.scss
+++ b/lib/gui/components/drive-selector/styles/_drive-selector.scss
@@ -14,13 +14,21 @@
  * limitations under the License.
  */
 
-.modal-drive-selector-modal .modal-content {
-  width: 300px;
-  height: 320px;
-}
+.modal-drive-selector-modal {
+  .modal-content {
+    width: 400px;
+    height: 320px;
+  }
 
-.modal-drive-selector-modal .list-group-item[disabled] {
-  cursor: not-allowed;
+  .button,
+  .label,
+  .modal-title {
+    font-weight: bold;
+  }
+
+  .list-group-item[disabled] {
+    cursor: not-allowed;
+  }
 }
 
 .component-drive-selector-body {
@@ -33,5 +41,18 @@
   .list-group-item-text {
     word-break: break-all;
   }
+
+  .list-group-item-text,
+  .list-group-item-footer {
+    display: inline-block;
+  }
+
+  .list-group-item-footer {
+    margin-top: 0;
+  }
+}
+
+.modal-footer-label {
+  @extend .text-center;
 }
 
--- a/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
+++ b/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
@@ -27,17 +27,14 @@
 
             <span class="label label-danger"
               ng-hide="modal.state.isDriveLargeEnough(drive)">
-                <i class="glyphicon glyphicon-resize-small"></i>
                 TOO SMALL FOR IMAGE</span>
 
             <span class="label label-danger"
               ng-show="modal.state.isDriveLocked(drive) && modal.state.isDriveLargeEnough(drive)">
-                <i class="glyphicon glyphicon-lock"></i>
                 LOCKED</span>
 
             <span class="label label-danger"
               ng-show="modal.state.isSystemDrive(drive)">
-                <i class="glyphicon glyphicon-hdd"></i>
                 SYSTEM DRIVE</span>
 
           </footer>
@@ -50,7 +47,17 @@
 </div>
 
 <div class="modal-footer">
-  <button class="button button-primary button-block"
+  <p class="modal-footer-label"
+    ng-show="modal.getButtonStatus() === 'danger'">
+      {{ modal.getFooterMessage() }}
+  </p>
+  <button class="button button-block"
+    ng-class="{ 'label-danger': modal.getButtonStatus() === 'danger',
+                'button-primary': modal.getButtonStatus() === 'primary' }"
     ng-click="modal.closeModal()"
-    ng-disabled="!modal.state.hasDrive()">CONTINUE</button>
+    ng-disabled="!modal.state.hasDrive()">
+      <i class="glyphicon glyphicon-alert"
+        ng-show="modal.getButtonStatus() === 'danger'"></i>
+      Continue
+  </button>
 </div>
--- a/tests/gui/components/drive-selector.spec.js
+++ b/tests/gui/components/drive-selector.spec.js
@@ -1,0 +1,224 @@
+'use strict';
+
+const _ = require('lodash');
+const m = require('mochainon');
+const angular = require('angular');
+require('angular-mocks');
+
+describe('Browser: DriveSelector', function() {
+
+  beforeEach(angular.mock.module(
+    require('../../../lib/gui/components/drive-selector/drive-selector')
+  ));
+
+  describe('DriveSelectorController', function() {
+
+    const defaultDrive = {
+      device: '/dev/disk2',
+      name: 'My Drive',
+      size: 123456789,
+      protected: false,
+      system: false
+    };
+
+    let currentDrive;
+
+    let $controller;
+    let $q;
+    let $rootScope;
+
+    const importSelectionStateModel = () => {
+      beforeEach(angular.mock.module({
+        SelectionStateModel: {
+          getDrive: () => {
+            return currentDrive;
+          },
+          hasDrive: () => {
+            return Boolean(currentDrive);
+          },
+          isSystemDrive: () => {
+            return Boolean(currentDrive.system);
+          }
+        }
+      }));
+    };
+
+    const injectDependencies = () => {
+      beforeEach(angular.mock.inject(function(_$controller_, _$q_, _$rootScope_) {
+        $controller = _$controller_;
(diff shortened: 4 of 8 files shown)


lurch:
Seems like this PR also includes the changes from #978 ?

Also, could we have a screenshot please? :)

Shou:
Yeah, it's dependent on those changes, once #978 gets accepted we'll rebase. And sure! The differences include the warning text now being centred, and the warning icon is on the right instead of left as I think that makes it more consistent (with the info icon in the main window, and the other warning icon that you saw in the big PR). Also, the button being blue sometimes [bug] should have been eradicated now too.

image

lurch:
Just seen https://gitter.im/resin-io/etcher?at=5852f332fb22792b3b6aec54 , oops ;)

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -50,7 +50,17 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
 </div>
 
 <div class="modal-footer">
-  <button class="button button-primary button-block"
+  <div class="modal-footer-warning"
+    ng-show="modal.state.hasDrive() && modal.state.isSystemDrive(modal.state.getDrive())">
```
jviotti: This expression (`modal.state.hasDrive() && modal.state.isSystemDrive(modal.state.getDrive())`) is also used in the `ng-show` from the button icon, and to decide whether the button will be red or not. I wonder if we can extract this to the model controller, as something like `.showFooterWarning()` and potentially unit test it as well.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -50,7 +50,17 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
+      Etcher has identified this as a system drive
```
jviotti: This very nitty-gritty, but you can dynamically get the "display name" as:

```html
<span manifest-bind="displayName"></button>
```

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -50,7 +50,17 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
+  </div>
+  <button class="button button-block"
+    ng-class="{ 'label-danger': modal.state.hasDrive() && modal.state.isSystemDrive(modal.state.getDrive()),
+                'button-primary': modal.state.hasDrive() && !modal.state.isSystemDrive(modal.state.getDrive()) }"
```
jviotti: I'm not 100% sure, but I think we can remove the `modal.state.hasDrive()` check and only apply the button type based on whether the drive is system or not. If we indeed have no drive, `ngDisabled` will take precedence over the button style and do the right thing.

lib/gui/components/drive-selector/styles/_drive-selector.scss
```diff
@@ -35,3 +35,7 @@
   }
 }
 
+.modal-footer-warning {
+  text-align: center;
```
jviotti: You can do `@extend .text-center` here (the `.text-center` class is defined by Bootstrap)

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -50,7 +50,17 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
 </div>
 
 <div class="modal-footer">
-  <button class="button button-primary button-block"
+  <div class="modal-footer-warning"
```
jviotti: What about `modal-footer-label`? Its a warning because we think of it that way, but I imagine this could be used for something else in the future.

Changing it to a label also means that `div` doesn't make much sense anymore, although its not a big deal. If you're motivated to make such change, I guess you can make it something like a `<p>`.

jviotti:
By seeing this screenshot, I feel there should be a bit more space between the red button and the label. What do you think @lurch ?

Shou:
I agree that we should add a padding to the label actually, probably somewhere in the range of 4-8px.

WasabiFan:
Can you also add a period at the end of the sentence?

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
Shou: `<p>` should be perfect here actually—comes with padding for free, too.

lurch:
Probably worth looping in @konmouz ?

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
lurch: If we're having a `.showFooterWarning()`, should we also have a `.getFooterWarning()`, instead of embedding `Etcher has identified this as a system drive` in this file?
Could then extend the model to show other warnings, without having to edit this file.

Or we could have a generic `.getFooterMessage()` (ties into your earlier "I imagine this could be used for something else in the future" comment), and a function to get the 'severity' of the FooterMessage? (which would determine if the button should be red or not)

Or perhaps taking things one step further, the model could have a `.getButtonStatus()` function (independent of `.getFooterMessage()`), which would determine whether the button is disabled (grey), enabled (blue) or warning (red).

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
Shou: The reason I have to check here is because otherwise `.isSystemDrive()` throws an error because it's passed `undefined` when there is no selected drive.

jviotti: Oh I see. In that case, if you create a function in the controller to determine if `danger` should be applies, then you can move that logic there (add an `if` conditional or something), and then simplify these expressions.

jviotti:
@Shou Do you mind posting a screenshot of how the warning label behaves when there are many drives (causing the list to scroll)?

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
jviotti: I like these ideas. Let's try the following:

- `.getButtonStatus()`: This function would return a string, like `primary`, `danger`, `disabled`, etc. The `ngDisabled` directive on the button can use it by comparing it to the string "disabled".

- `.getFooterMessage()`: This function could return `undefined`, or a string representing the message. It can even rely on `.getButtonStatus()` to see what it should show.

- (optional) `.hasFooterMessage()`: returns `true` if the result of `getFooterMessage()` is not `undefined`. This can used in an `ngShow` directive on the footer, just to make things nicer.

The nice thing about extracting all these functions to the controller is that we can unit test them all.

jviotti:
Let's rebase this branch so the contents of the other commit get removed from the "Files changes" tab.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
lurch: ...although I wonder if `.getButtonStatus()` is too GUI-specific, and it should be given a more model-centric name?  (something like `.getSelectionStatus()` ?)

konmouz:
@Shou I think we don't need the system drive icon in the label. Also, warning icon should be on the LHS (continue button). There are a few more changes in fonts, sizes and paddings/margins, should I pass you these details with Zeplin? I think it would be the quickest way. You can use the image below for reference from #888

image

Shou:
@konmouz 
> There are a few more changes in fonts, sizes and paddings/margins, should I pass you these details with Zeplin?

Sure, that'd be great!

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -142,4 +142,34 @@ module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateMode
...
+    if (isDangerous) {
+      return 'danger';
+    }
+
+    return 'primary';
+  };
+
+  this.getFooterMessage = () => {
+    if (this.getButtonStatus() === 'danger'
+      && this.state.isSystemDrive(this.state.getDrive())) {
```
jviotti: Based on:

```
const isDangerous = this.state.hasDrive() && this.state.isSystemDrive(this.state.getDrive());
```

Then if the button status is `danger`, then it follows that the drive is a system drive, therefore I think this conditional could be reduced to:

```
if (this.getButtonStatus() === 'danger') {}
```

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -142,4 +142,34 @@ module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateMode
...
+      return ManifestBindService.get('displayName') + ' has identified this as a system drive!';
+
+    }
+
+    return undefined;
```
jviotti: You can simply omit this line, which will cause `undefined` to be returned implicitly.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -50,7 +47,17 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
 </div>
 
 <div class="modal-footer">
-  <button class="button button-primary button-block"
+  <p class="modal-footer-label"
+    ng-show="modal.getButtonStatus() === 'danger'">
+      {{ modal.getFooterMessage() }}
```
jviotti: Since this function is only being used here, and the logic is quite simple, we should be able to inline it. As a bonus, in my comment on that function, I highlighted that the logic could be reduced to simply checking if the button status is `danger`, which you're already doing in the `ngShow` right above, you can put the message here and that should be it.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -142,4 +142,34 @@ module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateMode
...
+  };
+
+  this.hasFooterMessage = () => {
```
jviotti: This function seems to be completely unused, in which case we can remove it and the tests as well.

tests/gui/components/drive-selector.spec.js
```diff
@@ -0,0 +1,224 @@
...
+
+    let currentDrive;
+
+    let $controller;
+    let $q;
+    let $rootScope;
+
+    const importSelectionStateModel = () => {
+      beforeEach(angular.mock.module({
+        SelectionStateModel: {
```
jviotti: Let's import the real SelectionStateModel and DrivesModel and set available drives for real instead of mocking this logic.

tests/gui/components/drive-selector.spec.js
```diff
@@ -0,0 +1,224 @@
...
+          });
+
+          const result = DriveSelectorController.getButtonStatus();
+          m.chai.expect(result).to.equal('primary');
+
+        });
+      });
+    });
+
+    describe('.getFooterMessage()', function() {
```
jviotti: Now that this function and `hasFooterMessage` are not needed anymore, we can inline the `beforeEach` blocks in `getButtonStatus` instead of using the utility functions.

tests/gui/components/drive-selector.spec.js
```diff
@@ -0,0 +1,224 @@
...
+    describe('.getButtonStatus()', function() {
+
+      importSelectionStateModel();
+      injectDependencies();
+
+      describe('given there is no selected drive', function() {
+
+        it('should return "disabled"', function() {
+
+          const DriveSelectorController = makeDriveSelectorController();
```
jviotti: Instead of creating the controller on each test, you can do it on a `beforeEach` block, which will automatically reset the controller from scratch on each test.

tests/gui/components/drive-selector.spec.js
```diff
@@ -0,0 +1,224 @@
...
+      name: 'My Drive',
+      size: 123456789,
+      protected: false,
+      system: false
+    };
+
+    let currentDrive;
+
+    let $controller;
+    let $q;
```
jviotti: The `$q` module is not being used anywhere on this test, so we should get rid of it.

jviotti:
I'm closing this PR since drive selection has been re-designed on v2.

PR #980, opened 2016-12-18
refactor: extract some checks related to drive/image size to `lib/shared`

stefan-mihaila:
refactor: extract some checks related to drive/image size to `lib/shared`

Change-Type: minor


jviotti:
Why was this closed?

stefan-mihaila:
I closed this in favour of https://github.com/resin-io/etcher/pull/982. This takes a different approach: doesn't separate `SelectionStateModel` into different services, just moves some of its logic in `lib/shared` and calls functions from there. It might actually be more useful, as `isDriveLocked` and `isSystemDrive` are the only static methods extracted in https://github.com/resin-io/etcher/pull/982, but I want `isLargeEnough` and other methods in CLI that depend on the selected image. It also might be easier to solve the `TODO`s in `store` with this one.

Overall, it added some indirection to avoid a little duplication and somewhat didn't feel very clean to me. @jviotti What do you think?

jviotti:
Sounds good! I wanted to document the reason that this was closed here, for future reference!

PR #982, opened 2016-12-18
Extract CLI/GUI shared functionality

stefan-mihaila:
- refactor: `isSystemDrive` is `false` if `system` property undefined

    Change-Type: patch

- refactor: `drive-constraints` - throw when image is undefined

    The default behaviour was to pretend that we're all good if the if
    the image is not specified. We're not using this "feature", and
    it can be dangerous if we forget to pass in the image.

    Change-Type: patch

- refactor: extract additional methods to drive-constraints

    Change-Type: patch

- refactor: Use singleton object for `DriveConstraints`

    Change-Type: patch

- refactor: extract state independent methods from `SelectionStateModel`

    `SelectionStateModel`'s methods `isSystemDrive` and `isDriveLocked`
    don't depend on application state. They have been extracted in a different
    AngularJS service: `DriveConstraintsModel`. The new service's actual
    implementation is in `lib/src`, in order to be reused by the CLI.

    Change-Type: patch

- refactor: rename `lib/src` to `lib/shared`

    Change-Type: patch

- refactor: use `drive-constraints` in `store.js`

    Change-Type: patch


lurch:
I'm too tired to look through this in detail right now, but from a quick glance I wonder if it might be worth splitting this PR into two commits, with the second commit containing the "`lib/src` has been renamed to `lib/shared`" change? Would make the first change easier to view in isolation :)

stefan-mihaila:
@lurch, makes sense, split it in two commits.

jviotti:
Minor observation: refactoring commits should have a "patch" change type :)

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -28,6 +34,13 @@ module.exports = function($q, $uibModalInstance, DrivesModel, SelectionStateMode
   this.state = SelectionStateModel;
 
   /**
+   * @summary Static methods to check a drive's properties
+   * @property
+   * @type Object
+   */
+  this.constraints = DriveConstraintsModel;
```
jviotti: I wonder if we can require this particular module in the places that use this logic instead of completely exposing it here. As a bonus, the Redux store has had TODO comments to re-use these static functions for a while: https://github.com/resin-io/etcher/blob/master/lib/gui/models/store.js#L99

lib/gui/app.js
```diff
@@ -43,6 +43,7 @@ const app = angular.module('Etcher', [
 
   // Models
   require('./models/selection-state'),
+  require('./models/drive-constraints'),
```
jviotti: This is not needed, since the `DriveConstraintsModel` is not being directly used by the `.run()` blocks from this file.

lib/src/drive-constraints.js
```diff
@@ -0,0 +1,70 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const _ = require('lodash');
+
+module.exports = class DriveConstraints {
```
jviotti: This requires using `new` when using this class directly, which is not ideal. 

I think you can declare it the usual NodeJS way (using `exports`) and Angular does the right thing out of the box when creating a service out of it (let me know if its not the case and we'll sort it out somehow):

```
exports.isDriveLocked = (drive) => { ... }
exports.isSystemDrive = (drive) => { ... }
```

jviotti:
As a bonus, the Redux store has had TODO comments to re-use these static functions for a while: https://github.com/resin-io/etcher/blob/master/lib/gui/models/store.js#L99

jviotti:
@lurch I would love to hear you feedback about the name of this new module: "drive-constraints.js". I think its not a bad one, but I feel it can still be improved, since things like "locked" doesn't sound as a "constraint" to me (its more of a "state").

jviotti:
Kudos for changing `lib/src` to `lib/shared` :tada:

lib/gui/models/drive-constraints.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @module Etcher.Models.DriveConstraints
+ */
+
+const angular = require('angular');
+const MODULE_NAME = 'Etcher.Models.DriveConstraints';
+const SelectionStateModel = angular.module(MODULE_NAME, []);
+
+const DriveConstraints = require('../../shared/drive-constraints');
```
jviotti: Can we add a big comment in this file to clarify that this only exposes a file from the CLI/GUI "shared" source code as an Angular module?

lib/gui/models/drive-constraints.js
```diff
@@ -0,0 +1,31 @@
...
+ * @module Etcher.Models.DriveConstraints
+ */
+
+const angular = require('angular');
+const MODULE_NAME = 'Etcher.Models.DriveConstraints';
+const SelectionStateModel = angular.module(MODULE_NAME, []);
+
+const DriveConstraints = require('../../src/drive-constraints');
+
+SelectionStateModel.service('DriveConstraintsModel', DriveConstraints);
```
stefan-mihaila: done

lib/gui/pages/main/main.js
```diff
@@ -47,6 +47,7 @@ const MainPage = angular.module(MODULE_NAME, [
   require('../../modules/analytics'),
   require('../../modules/error'),
   require('../../models/selection-state'),
+  require('../../models/drive-constraints'),
```
stefan-mihaila: You're right, good catch. Removed it from `main.js` and `app.js`

lib/gui/app.js
stefan-mihaila: done

lib/src/drive-constraints.js
stefan-mihaila: If I understand correctly, [the function passed to `angular.service` should be a constructor](https://docs.angularjs.org/api/auto/service/$provide#service). Or you can pass a function returning an instance of the service to [`angular.factory`](https://docs.angularjs.org/api/auto/service/$provide#factory).

Even with `angular.service`, I believe it's actually a singleton, meaning `AngularJS` creates only one instance using the constructor. I'm not 100% on it. Maybe @Shou knows?

[I created a separate commit exporting an object instead of a constructor](https://github.com/resin-io/etcher/commit/f95c10ef1b17f20545ef7021b20febfef3a92aa0). If you like it more, I'll cherry pick it.

lurch:
> Minor observation: refactoring commits should have a "patch" change type :)

@stefan-mihaila I believe your first commit in this PR should still have the "minor" change-type, with just the second commit (where you rename src to shared) being "patch". Unless I've misunderstood @jviotti ?
Or if both commits are indeed "patch", then maybe the whole PR should also be "patch"?

> since things like "locked" doesn't sound as a "constraint" to me (its more of a "state").

I guess there's effectively 3 different "levels" at play here...
* The 'raw hardware' level, i.e. which physical interface a drive is connected to, what its device-name is, etc. - this determines whether a drive is 'system' or not (and in future will also determine if it's a 'boot drive'). Info at this level never changes without rebooting the PC.
* The 'current drive' level, i.e. whether the write-protect tab on an SD is enabled, the drive's raw capacity, what 'name' the drive reports, etc. - this determines whether a drive is 'locked' or not. Info at this level can change when a different SD card is inserted into an SD reader.
* The 'current image' level i.e. name and path of current image, its size, etc - this (in combination with the info from the previous level) determines whether a drive is "recommended size" or "too small". Info at this level can change when the user selects a different image.

So, although "locked" is strictly a "state" (level 2) rather than a "constraint" (level 1), in practice I suspect we'll have to treat levels 1 and 2 as the same 'thing', since inserting a new USB drive (or a new USB SD adaptor) is effectively both levels 1 and 2 changing at the same time.

I'm not sure how these different levels / constraints / restrictions should be refactored into different modules ATM, but obviously both the CLI and GUI ought to share the same eventual top-level logic e.g. like my comment [here](https://github.com/resin-io/etcher/pull/966#pullrequestreview-12338231) is suggesting.

lib/src/drive-constraints.js
```diff
@@ -0,0 +1,70 @@
...
+   *   device: '/dev/disk2',
+   *   name: 'My Drive',
+   *   size: 123456789,
+   *   protected: true
+   * })) {
+   *   console.log('This drive is locked (e.g: write-protected)');
+   * }
+   */
+  isDriveLocked(drive) {
+    return _.get(drive, 'protected', false);
```
lurch: Should this also be forced to `Boolean`, like `isSystemDrive` below?

lib/src/drive-constraints.js
```diff
@@ -0,0 +1,70 @@
...
+   *   name: 'My Drive',
+   *   size: 123456789,
+   *   protected: true,
+   *   system: true
+   * })) {
+   *   console.log('This drive is a system drive!');
+   * }
+   */
+  isSystemDrive(drive) {
+    return Boolean(drive.system);
```
lurch: Should this also default to `false` if the attribute is missing, like `isDriveLocked` above?

stefan-mihaila:
> I believe your first commit in this PR should still have the "minor" change-type, with just the second commit (where you rename src to shared) being "patch". Unless I've misunderstood @jviotti ?
> Or if both commits are indeed "patch", then maybe the whole PR should also be "patch"?

@lurch The first commit is still just a refactor (it neither adds functionality, neither fixes a bug). I'll edit the PR's description to match.

lib/src/drive-constraints.js
stefan-mihaila: This was copied from the existing implementation. Assuming we're giving it drive objects as returned by `drivelist` I don't think it matters, as they seem to always have the `system` property, but I admit that the inconsistency is annoying. Any preference?

lib/src/drive-constraints.js
stefan-mihaila: This was copied from the existing implementation. Assuming we're giving it drive objects as returned by `drivelist` I don't think it matters, as they seem to always have the `protected` property and to be a Boolean, but I admit that the inconsistency is annoying. Any preference?

lib/src/drive-constraints.js
lurch: This is one for @jviotti ;-)

stefan-mihaila:
@lurch @jviotti General question about PRs: Is it better to not `git push --force` until the review is completed and just squash commits after all issues in the ongoing PR are addressed?

stefan-mihaila:
> As a bonus, the Redux store has had TODO comments to re-use these static functions for a while: https://github.com/resin-io/etcher/blob/master/lib/gui/models/store.js#L99

@jviotti To go for the bonus I think it would be better to also extract additional methods, that are dependent on the selected image, like `isDriveLargeEnough`. They should be extracted anyway because they're needed in CLI.

Would it be good to extract them in the same place in the shared class (or object if we change to one) as methods taking an image as a parameter, exposed in the `DriveConstraintsModel` and called in `SelectionStateModel` from the `DriveConstraintsModel` and in the CLI and Redux Store from `lib/shared`?

@lurch If you have any thoughts on this please share them, I only now, at the company call, realised that @jviotti is actually off for the week...

lurch:
> General question about PRs: Is it better to not `git push --force` until the review is completed and just squash commits after all issues in the ongoing PR are addressed?

I think it's a matter of personal preference - and I'm fairly new to the resin.io team myself so I dunno if there's any existing/common conventions. AFAICT @jviotti always squashes and force-pushes his commits (which keeps things 'tidy', but sometimes makes it harder to see what an updated version of the PR has changed compared to the original commit).

> @lurch If you have any thoughts on this please share them

I'm afraid I don't have any experience with Angular / Redux / etc. so wouldn't know what to advise for the best. But in terms of keeping things cleanly separated I wonder if it might be worth creating an additional layer in-between `DriveConstraintsModel` and `SelectionStateModel` (perhaps called `ImageConstraintsModel`?) where the image-specific filtering logic (`isDriveLargeEnough` etc.) is implemented?

Pinging @Shou as this overlaps with some of the GUI-work he's doing.

jviotti:
@stefan-mihaila 

> @lurch If you have any thoughts on this please share them, I only now, at the company call, realised that @jviotti is actually off for the week...

I'll still be active on comments, pull requests, so don't worry :) I'll review the latest changes later today!

WasabiFan:
> AFAICT @jviotti always squashes and force-pushes his commits (which keeps things 'tidy', but sometimes makes it harder to see what an updated version of the PR has changed compared to the original commit).

I've found that it is easier to push commits as you make changes, and then squash using the GitHub UI when you want to merge the PR. That means that reviewers can see what progressive changes have been made and comment threads don't get detached because you edited history.

lib/src/drive-constraints.js
jviotti: Good catch. Feel free to use `Boolean()` if you like.

lib/src/drive-constraints.js
jviotti: I'm always happy to fix inconsistencies! If you can do it (and put some tests that check the default scenario), that'd be awesome!

tests/gui/models/drive-constraints.spec.js
```diff
@@ -0,0 +1,28 @@
...
+
+  describe('DriveConstraintsModel', function() {
+
+    let DriveConstraintsModel;
+
+    beforeEach(angular.mock.inject(function(_DriveConstraintsModel_) {
+      DriveConstraintsModel = _DriveConstraintsModel_;
+    }));
+
+    it('should be an instance of `lib.src.DriveConstraints`', function() {
```
jviotti: Any reasons for the periods though? Referencing the actual file path (`lib/src/drive-constraints.js`) looks cleaner IMHO.

jviotti:
> @stefan-mihaila I believe your first commit in this PR should still have the "minor" change-type, with just the second commit (where you rename src to shared) being "patch". Unless I've misunderstood @jviotti ?
Or if both commits are indeed "patch", then maybe the whole PR should also be "patch"?

Both commits do purely refactoring, so I guess they should be both "patch" (and thus the whole PR is a "patch").

> @lurch @jviotti General question about PRs: Is it better to not git push --force until the review is completed and just squash commits after all issues in the ongoing PR are addressed?

Up to you, really.

> Would it be good to extract them in the same place in the shared class (or object if we change to one) as methods taking an image as a parameter, exposed in the DriveConstraintsModel and called in SelectionStateModel from the DriveConstraintsModel and in the CLI and Redux Store from lib/shared?

Yeah, I like this suggestion. Let's change functions such as `isDriveLargeEnough` to take the image as an argument rather than getting it directly from the state, so the functions become pure, and can be extracted to `drive-constraints.js`. 

Once that's done, we have two way forwards:

- Adapt the existing functions from `SelectionStateModel` to be simply a facade to `DriveConstraintsModel`, so for example, we will have:

```js
DriveConstraintsModel.isDriveLargeEnough = (drive, image) => {
    return drive.size >= image.size;
};
```

And:

```js
SelectionStateModel.isDriveLargeEnough = (drive) => {
    return DriveConstraintsModel.isDriveLargeEnough(drive, this.getImage());
};
```

- Simply modify every code that calls the function to manually fetch and pass the image as an argument.

I propose going for the second option (manually passing the image), since despite not so convenient as the current functions, makes the intention clearer, and we don't have two functions doing basically the same thing.

lurch:
Presumably things like `image.size` are internally-cached, so that if we have 5 drives in our drive-selection-modal, we don't try to recalculate the image size 5 separate times? ;-)

As I mentioned in https://github.com/resin-io/etcher/pull/966#pullrequestreview-12338231 I still think there ought to also be some other place (maybe a layer between `DriveConstraintsModel` and `SelectionStateModel`?) where the higher-level categorisation is done, so that we don't end up with the same logic duplicated in different parts of the GUI and the CLI. But strictly speaking that's probably beyond the scope of this particular PR?

tests/gui/models/drive-constraints.spec.js
stefan-mihaila: I guess that I just wanted to show that I'm talking about a class and where the class lives, so I used the dot-notation from other languages. But I agree that it doesn't look right, I'll change it.

jviotti:
> Presumably things like image.size are internally-cached, so that if we have 5 drives in our drive-selection-modal, we don't try to recalculate the image size 5 separate times? ;-)

Indeed. They are calculated once as soon as you select the image, and saved to the Redux store. When you then call functions that need it, that information is fetched from the Redux store.

> As I mentioned in #966 (review) I still think there ought to also be some other place (maybe a layer between DriveConstraintsModel and SelectionStateModel?) where the higher-level categorisation is done, so that we don't end up with the same logic duplicated in different parts of the GUI and the CLI. But strictly speaking that's probably beyond the scope of this particular PR?

By "higher-level categorisation", you mean a function that gives you the ultimate label names, like "BOOT DRIVE", etc, right? If so I agree, and we should do it right after this PR.

lib/gui/models/drive-constraints.js
```diff
@@ -0,0 +1,37 @@
...
+const angular = require('angular');
+const MODULE_NAME = 'Etcher.Models.DriveConstraints';
+const DriveConstraintsModel = angular.module(MODULE_NAME, []);
+
+const DriveConstraints = require('../../shared/drive-constraints');
+
+// `DriveConstraintsModel.service` expects a constructor as the second argument, but we want
+// to expose an object. Calling `DriveConstraintsModel.factory` with a a function returning
+// the object is the right way to do it.
+DriveConstraintsModel.factory('DriveConstraintsModel', () => {
```
stefan-mihaila: Technically, I could've used `service` here, with a constructor that returns `DriveConstraints` instead of the implicit `this`, but it feels more correct like this, as `AngularJS` won't call this function with `new`.

stefan-mihaila:
@jviotti @lurch Updated PR

stefan-mihaila:
@jviotti I forgot the `TODO`s in `store`. Doing them now.

tests/shared/drive-constraints.spec.js
```diff
@@ -0,0 +1,276 @@
...
+
+      it('should return true if drive is not locked', function() {
+        this.drive.protected = false;
+        m.chai.expect(constraints.isDriveValid(this.drive, this.image)).to.be.true;
+      });
+
+      it('should return false if drive is locked', function() {
+        this.drive.protected = true;
+        m.chai.expect(constraints.isDriveValid(this.drive, this.image)).to.be.false;
+      });
```
jviotti: Can we add a blank line after this one for consistency?

tests/shared/drive-constraints.spec.js
```diff
@@ -0,0 +1,276 @@
...
+    });
+    describe('given drive is not large enough', function() {
```
jviotti: Let's put a blank line above this one as well.

tests/shared/drive-constraints.spec.js
```diff
@@ -0,0 +1,276 @@
...
+        };
+        this.image = {
+          path: 'rpi.img',
+          size: 2000000000
+        };
+      });
+
+      it('should return false', function() {
+        m.chai.expect(constraints.isDriveValid(this.drive, this.image)).to.be.false;
+      });
```
jviotti: Let's put a blank line after this one.

tests/shared/drive-constraints.spec.js
```diff
@@ -0,0 +1,276 @@
...
+    });
+
+    describe('given image is undefined', function() {
```
jviotti: Let's put blank lines between the `it` and `describe` blocks for consistency with the rest of the file:

```js
describe('...', function() {

    it('...', function() {

    });

});
```

tests/shared/drive-constraints.spec.js
```diff
@@ -0,0 +1,276 @@
...
+        const drive = {
+          device: '/dev/disk2',
+          name: 'My Drive',
+          size: 4000000000
+        };
+        m.chai.expect(() => {
+          constraints.isDriveValid(drive);
+        }).to.throw(TypeError);
+      });
+    });
```
jviotti: Let's put a blank line after this one for consistency.

tests/shared/drive-constraints.spec.js
jviotti: I think we can leave this test out, given that the function already documents that it needs an image as the second parameter.

tests/shared/drive-constraints.spec.js
```diff
@@ -0,0 +1,276 @@
...
+        protected: false
+      }, {
+        path: 'rpi.img',
+        size: 1000000001
+      });
+
+      m.chai.expect(result).to.be.false;
+    });
+
+    it('should throw if image is undefined', function() {
```
jviotti: Let's leave this one out as well.

tests/shared/drive-constraints.spec.js
```diff
@@ -0,0 +1,276 @@
...
+      }, {
+        path: 'rpi.img',
+        size: 1000000000,
+        recommendedDriveSize: 2000000001
+      });
+
+      m.chai.expect(result).to.be.false;
+    });
+
+    it('should throw if the image is undefined', function() {
```
jviotti: Let's leave this one out as well.

jviotti:
Very, very nice! Now we can re-use the extracted functions in `store.js`.

jviotti:
> @jviotti I forgot the TODOs in store. Doing them now.

lol, I just sent a comment about that :)

tests/shared/drive-constraints.spec.js
stefan-mihaila: ✔️

tests/shared/drive-constraints.spec.js
stefan-mihaila: ✔️

tests/shared/drive-constraints.spec.js
stefan-mihaila: ✔️

tests/shared/drive-constraints.spec.js
stefan-mihaila: ✔️

stefan-mihaila: ✔️

tests/shared/drive-constraints.spec.js
stefan-mihaila: ✔️

lib/gui/models/store.js
```diff
@@ -94,15 +95,15 @@ const storeReducer = (state, action) => {
 
         const drive = _.first(action.data);
 
-        if (_.every([
-
-          // TODO: Reuse from SelectionStateModel.isDriveValid()
-          state.getIn([ 'selection', 'image', 'size' ], 0) <= drive.size,
+        // Even if there's no image selectected, we need to call `isDriveValid`
+        // to check if the drive is locked, and `{}` works fine with it
+        const image = state.getIn([ 'selection', 'image' ], {
```
jviotti: The second approach indeed works for me.

lib/gui/models/store.js
```diff
@@ -225,8 +226,8 @@ const storeReducer = (state, action) => {
         throw new Error('The drive is write-protected');
       }
 
-      // TODO: Reuse from SelectionStateModel.isDriveLargeEnough()
-      if (state.getIn([ 'selection', 'image', 'size' ], 0) > selectedDrive.get('size')) {
+      const image = state.getIn([ 'selection', 'image' ]);
```
stefan-mihaila: Tried it now, it results in `TypeError: Cannot read property 'toJS' of undefined` when `[ 'selection', 'image' ]` doesn't exits.

jviotti: > Unless .toJS() bails out if [ 'selection', 'image' ] doesn't exist?

It does, ignore me then :)

lib/gui/models/store.js
stefan-mihaila: `state.getIn([ 'selection', 'image' ], {});` will return `{}` when the image is not in the store, and calling `({}).toJS()` will throw a `TypeError`

stefan-mihaila: Oh, yeah, the second one makes sense.

lib/gui/models/store.js
```diff
@@ -268,9 +269,9 @@ const storeReducer = (state, action) => {
       });
 
       return _.attempt(() => {
-        if (_.some([
-          selectedDrive && selectedDrive.get('size', 0) < action.data.size,
-          selectedDrive && selectedDrive.get('size', 0) < action.data.recommendedDriveSize
+        if (selectedDrive && _.some([
```
jviotti: This can be refactored to

```js
if (selectedDrive && !_.every([
  constraints.isDriveLargeEnough(selectedDrive.toJS(), action.data),
  constraints.isDriveSizeRecommended(selectedDrive.toJS(), action.data)
]))
```

So we avoid negating in each part of `_.some`.

lib/gui/models/store.js
jviotti: I wonder if just for clarity, we should let this be `undefined` if it doesn't exist, and move the "empty object" workaround to the places that actually need it:

```js
if (_.every([
    constraints.isDriveValid(drive, image || {}),
    constraints.isDriveSizeRecommended(drive, image || {})
 ]))
```

This would prevent us from accidentally relying on an empty object representing the lack of an image in the store.

lib/gui/models/store.js
stefan-mihaila: nice 1

lib/gui/models/store.js
stefan-mihaila: > This would prevent us from accidentally relying on an empty object representing the lack of an image in the store.

We're not relying on an empty object being the store. If the store returns something, we want to call `toJS()` on that something. If it returns `undefined`, we want to not call `toJS()` and just use `{}`

Your version is more readable, except we still need to call `toJS()` on image if it's not `undefined`.

jviotti:
Awesome! I'll merge as soon as the CI servers complete.

lurch:
@jviotti I know you've approved this PR, but could you not merge it yet please? I've got some comments I'd like to make, which I'll do in the morning when I'm less tired... :-)

stefan-mihaila:
Hey @lurch, I see the PR is not merged. Want to share your comments?

jviotti:
Sure, take your time!

stefan-mihaila:
@lurch @jviotti What's the status on this one?

jviotti:
I guess we're only waiting for @lurch comments.

lib/gui/models/selection-state.js
```diff
@@ -25,7 +25,8 @@ const angular = require('angular');
 const Store = require('./store');
 const MODULE_NAME = 'Etcher.Models.SelectionState';
 const SelectionStateModel = angular.module(MODULE_NAME, [
-  require('./drives')
+  require('./drives'),
+  require('./drive-constraints')
```
jviotti: I just realized we're not using the drive constraints model anywhere here, so I guess it should be removed.

tests/gui/models/drive-constraints.spec.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const m = require('mochainon');
+const angular = require('angular');
+require('angular-mocks');
+
+describe('Browser: SelectionState', function() {
```
jviotti: This file says "SelectionState", however this file is about "DriveConstraints."

Issue #987, opened 2016-12-25
Make website visits coming from the update prompt on etcher identifiable


alexandrosm:
This should be simple to do but v. valuable. We should make the link to the etcher.io website coming from the "update" link be identifiable, perhaps by attaching a `?ref=etcher_update` querystring. That will allow us to separate that traffic in the analytics, and perhaps even offer a slightly different landing page.

lurch:
I guess we could also do something similar for the Etcher logo displayed at the bottom of the main program window, e.g. have it link to `https://etcher.io/?ref=etcher`.
And would we then want to use a link like `https://etcher.io/?ref=github` for links in the README and documentation?

jviotti:
Sounds good, and its a really easy improvement. I assigned @Shou to it.

alexandrosm:
Check the source code :) but the github one, yes!

jviotti:
What do you mean? Editing https://github.com/resin-io/etcher/blob/master/lib/gui/components/update-notifier/templates/update-notifier-modal.tpl.html#L15 should do it.

lurch:
@alexandrosm 

> but the github one, yes!

Does that mean that you **do** still want the links from github.com to etcher.io to have referrer query-URLs? And if so, should this issue be re-opened?

alexandrosm:
actually, I think github has proper referrer, because we can see it in our
analytics. so no real problem there!

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

PR #990, opened 2016-12-29
feat(gui): add referers to etcher.io links

Shou:
We add referers to the two etcher.io links in the interface—one for the
update link, and one for the footer logo link.

Closes: https://github.com/resin-io/etcher/issues/987
Changelog-Entry: Add referers to the etcher.io links

### DIFF
--- a/lib/gui/index.html
+++ b/lib/gui/index.html
@@ -38,7 +38,7 @@
       <svg-icon path="../../../assets/etcher.svg"
         width="83px"
         height="13px"
-        os-open-external="https://etcher.io"></svg-icon>
+        os-open-external="https://etcher.io?ref=etcher_footer"></svg-icon>
 
       <span class="caption">
         IS <span class="caption" os-open-external="https://github.com/resin-io/etcher">AN OPEN SOURCE PROJECT</span> BY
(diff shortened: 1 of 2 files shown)


lib/gui/index.html
```diff
@@ -38,7 +38,7 @@
       <svg-icon path="../../../assets/etcher.svg"
         width="83px"
         height="13px"
-        os-open-external="https://etcher.io"></svg-icon>
+        os-open-external="https://etcher.io?ref=etcher"></svg-icon>
```
alexandrosm: I guess here we'd need to rather make it ref=footer if we're focusing on what part of Etcher referred the user? Maybe it should be etcher_footer and etcher_update?

Shou: Yeah that makes more sense, especially considering we might add more links in the future. I'll incorporate that.

jviotti:
Looks great! I'll merge as soon as the Appveyor builds pass. There seems to be an issue with some of the dependencies that cause a "callback called more than once" weird error on `npm` itself :/

jviotti:
Damn npm issue. I'll merge with the Appveyor tests failing and will debug this when I get back to full-time work.

Issue #876, opened 2016-11-17
Website suggests downloading wrong version of Etcher


Charelene:
- **Etcher version: v1.0.0-beta.16**
- **Operating system and architecture: Ubuntu 16.10 x86**
- **Do you see any meaningful error information on DevTools? nah, not really**
So the thing is, that clicking "Download for Linux" button downloads x64 AppImage instead of x86. It may confuse some people, I guess?

lurch:
This is related to http://stackoverflow.com/questions/1741933/detect-64-bit-or-32-bit-windows-from-user-agent-or-javascript - there's no reliable way to detect whether the host OS is 32-bit or 64-bit (javascript reports the browser arch rather than the host arch). So I guess the Etcher website is just defaultting to 64-bit as that's the most common option.

ping @jviotti and @craig-mulligan

WasabiFan:
It depends on the browser. Some browsers do include the architecture in their user agent string, but it's up to the browser implementation and user settings. I'd expect that browsers often report their own architecture instead of the OS' if they report it at all.

craig-mulligan:
Yea, it just defaults to 64. I guess providing a table of downloads would help here, though we still have to offer a 'default', perhaps the best option is to add a small description below the download button describing what it is defaulting to.

jviotti:
I'm closing this since there is no way to reliably detect the architecture. As a solution, we have a table containing all downloads for all operating systems and architectures in the website.

PR #992, opened 2016-12-29
Add lander

craig-mulligan:
WIP.


jviotti:
Woohoo 🎉 This is VERY exciting! I'll give this a go and a closer look next week :)

www/js/index.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+const _ = require('lodash');
+const platform = require('platform');
+const $ = require('jquery');
+const downloads = $('#downloads').data('downloads');
+
+// Find and download link based on detected OS
+// Defualt is first link the matching is unsuccessful
+const dynamicDownload = _.find(downloads, (d) => {
+  new RegExp(d.OS).test(platform.os.family);
```
stefan-mihaila: Shouldn't we also take architecture into consideration?

lurch: See also #876

jviotti:
This looks really nice, however we need to put a lot of effort in reducing the amount of configuration needed.

 @craig-mulligan do you mind explaining how the current implementation works? As far as I can see, you're using Webpack, and the project maintainer needs to do a build step (`landr deploy`) which builds the website and puts it into `gh-pages`, right?

If so, any reason why you ditched Jekyll templates instead, which wouldn't require the build step?

alexandrosm:
cc @pimterry

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

craig-mulligan:
@jviotti true, there is alot of configuration. I hope to hide that and then only show it if you run smthing like `landr eject` or add the possibility of merging in only things you change (still experimenting) 

Yes, `landr deploy` will build everything + push to gh-pages. 

I moved away from jekyll templates because it means we'd have to commit the templates/js/scss to the users repo, this would mean updates to landr wouldn't update any of that source (like it does now).

I started off thinking webpack could load files using webpack and then spit out a directory like we have here: https://github.com/craig-mulligan/jek-test but then as I mentioned where do you edit the templates?  In the directory we just created or by passing landr some new templates? It essentially creates a double build step which I thought was confusing. But very open to idea's that could make it less so.

jviotti:
Based on today's landr meeting, I'll close this one for now. We'll adopt landr in landr first (which will likely change various things), and then adopt it in Etcher.

Issue #932, opened 2016-12-04
Etcher should bring itself to the foreground when an image is selected


WasabiFan:
- **Etcher version:** Beta 17
- **Operating system and architecture:** Windows 10 x64

When I click the button to open the file selection window, the file picker opens on top of Etcher. I can then switch windows, minimize the main Etcher window, etc -- in my case, I was opening a browser to download the image I wanted after I realized that I didn't have it locally. Once I select a file, the picker goes away but Etcher stays minimized/in the background. I think that Etcher should bring itself to the foreground when this happens so that I don't have to go searching for the main Etcher window.

jviotti:
Sounds like a good addition at first sight, however is this the normal behaviour in other applications in the operating systems we're targeting? Maybe @konmouz can provide some input here.

WasabiFan:
I can't speak to other operating systems, but I definitely expected Etcher to pop back up after selecting the image; I can't say whether that's a universal action or not.

In general for most apps, the file picker behaves as a modal dialog; I can't, for example, put something between the file picker and the Word window that launched it. In that case, there is still only one "window" entry in the taskbar and the dialog is rendered at the same level as its parent window. I also can't manipulate the parent window's position or buttons while the picker is open. This is the same for Skype, Edge and a few other apps I tested, including a test app I just wrote. My guess is that this divergent behavior stems from the way Electron handles dialogs.

I can't imagine a scenario where you would click the "Open" button and _not_ want to proceed with Etcher's process immediately, so it seems like a safe bet.

lurch:
Yeah, I'd classify this as a "modal dialog" bug, not as a "Etcher doesn't pop back to the foreground" bug ;-)

I just checked with `atom`, and when the file-dialog is open in that, and you minimise the parent application, the file-dialog gets hidden too, and only gets redisplayed when the parent app is restored (which is the behaviour I'd expect). This shows that this is (possibly) a bug just with `Etcher`, and not with all `electron`-based apps.

jviotti:
> I just checked with atom, and when the file-dialog is open in that, and you minimise the parent application, the file-dialog gets hidden too, and only gets redisplayed when the parent app is restored 

I tested this in OS X, and I'm unable to minimize/close the parent window until I dismiss the modal, which is the intended behaviour. I'll check in other OSes soon, but I infer from your comments that this is not the case there.

[...]

PR #1009, opened 2017-01-10
fix: make the image selection window a modal (#932)

zvin:
The image selection dialog is attached to the parent window so it is a modal.

### DIFF
--- a/lib/gui/os/dialog/services/dialog.js
+++ b/lib/gui/os/dialog/services/dialog.js
@@ -40,7 +40,8 @@ module.exports = function($q, SupportedFormatsModel) {
    */
   this.selectImage = () => {
     return $q((resolve, reject) => {
-      electron.remote.dialog.showOpenDialog({
+      const currentWindow = electron.remote.getCurrentWindow();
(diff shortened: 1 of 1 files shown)


lib/gui/os/dialog/services/dialog.js
```diff
@@ -39,8 +39,9 @@ module.exports = function($q, SupportedFormatsModel) {
    * });
    */
   this.selectImage = () => {
+    const parent = electron.remote.getCurrentWindow();
```
jviotti: Can we move this line inside the promise, and rename `parent` to `currentWindow` (since otherwise reading `parent` at this point makes it look that the browser window returned by `getCurrentWindow()` is the parent of the current app)? For example:

```js
return $q((resolve, reject) => {
    const currentWindow = electron.remote.getCurrentWindow();
    electron.remote.dialog.showOpenDialog(currentWindow, {
    ...
```

zvin: fixed, thanks

jviotti:
Thanks a lot for the patch!

Issue #986, opened 2016-12-23
" , Unexpected token U in JSON at position 0" - I need help with this error


paulallen8691:
- **Etcher version:**   1.0.0-beta.17
- **Operating system and architecture:** Win10 Home
Version 1607
OS build 14393.576
RAM 12G
64 bit
- **Do you see any meaningful error information on DevTools?**
Not Installed

Can anyone shed some light on the cause of the following error while trying to write to an SD card? ...
I've uninstalled and reinstalled the card reader.
I've tried a different card with the same response.
I don't see anything within this group that talks about this error.
A google search seems to indicate that this error is related to Javascript.
Any help is appreciated..

" , Unexpected token U in JSON at position 0"
![unexpected token](https://cloud.githubusercontent.com/assets/24732963/21454464/ddd928f6-c8e6-11e6-8349-b1b3ffa5a556.jpg)

WasabiFan:
> Do you see any meaningful error information on DevTools? Not Installed

@jviotti and company are probably going to need the output from DevTools to figure this out. There shouldn't be a reason that it is "not installed" -- In Etcher after this error occurs, just press <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>I</kbd> and the pane should pop up, then click on "console" if it isn't already selected and post the contents from there.

This should be a pretty easy bug to fix as long as the maintainers know the log output (assuming the required information is in there).

jviotti:
This is somehow similar to https://github.com/resin-io/etcher/issues/981, where we're receiving standard output from the script as IPC messages. 

Given that the help output from the CLI was shown, it seems that we're having some issues passing the arguments to the child script when spawning it, but I can't see anything obvious at first glance.

I'll investigate further and report back, probably after New Year's.

jviotti:
What is the full path of the image you're trying to flash BTW?

paulallen8691:
G:\Engineering Reference\Raspberry pi\Raspian images\2016-11-25-raspbian-jessie.img

lurch:
Hmmm, I wonder if this might be another problem with filenames being passed to commands unquoted, which might lead to errors when filenames / paths contain spaces.

@paulallen8691 As a test, could you try copying the image you're trying to flash to just `G:\` and see if you get the same error?

jviotti:
@lurch 

> Hmmm, I wonder if this might be another problem with filenames being passed to commands unquoted, which might lead to errors when filenames / paths contain spaces.

[...]

PR #1010, opened 2017-01-10
feat(CLI): replace `--robot` with an `ETCHER_CLI_ROBOT` env var

jviotti:
The `--robot` option of the CLI causes the program to output
machine-parseable strings, which can be easily consumed by the GUI to
update progress and other information.

The problem is that if the CLI fails to parse its command line arguments
when being called from the GUI for whatever reason, the `.fail()` Yargs
handler will be called, which doesn't output error information in the
usual "robot" format, causing the GUI to not understand the error
message.

Since the `.fail()` Yargs handler doesn't have access to the passed
options, we moved the "robot" functionality to an environment variable,
which we can easily check from there.

As a bonus, this PR refactors the whole robot logic into
`lib/shared/robot.js` and adds unit tests to it.

See: https://github.com/resin-io/etcher/issues/986
Change-Type: major
Changelog-Entry: Replace the `--robot` CLI option with an `ETCHER_CLI_ROBOT` environment variable.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -94,8 +94,9 @@ contains certain features to ease communication:
 
 - [Well-documented exit codes.][exit-codes]
 
-- A `--robot` option, which causes the Etcher CLI to output state in a way that
-can be easily machine-parsed.
+- An environment variable called `ETCHER_CLI_ROBOT` option, which when set
+  causes the Etcher CLI to output state in a way that can be easily
+  parser by a machine.
 
 Summary
 -------
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -40,7 +40,6 @@ Options
   --version, -v  show version number
   --drive, -d    drive
   --check, -c    validate write
-  --robot, -r    parse-able output without interactivity
   --yes, -y      confirm non-interactively
   --unmount, -u  unmount on success
 ```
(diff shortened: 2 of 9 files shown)


docs/ARCHITECTURE.md
```diff
@@ -94,8 +94,9 @@ contains certain features to ease communication:
 
 - [Well-documented exit codes.][exit-codes]
 
-- A `--robot` option, which causes the Etcher CLI to output state in a way that
-can be easily machine-parsed.
+- An environment variable called `ETCHER_CLI_ROBOT` option, which when set
+  causes the Etcher CLI to output state in a way that can be easily
+  parser by a machine.
```
lurch: typo: parser -> parsed

docs/CLI.md
```diff
@@ -40,21 +40,20 @@ Options
...
 
 The robot option
 ----------------
 
-The `--robot` option is very particular since it allows other applications to
-easily consume the output of the Etcher CLI in real-time. When using the
-`--robot` option, the `--yes` option is implicit, therefore you need to
-manually specify `--drive`.
+Setting the `ETCHER_CLI_ROBOT` environment variable allows other applications
+to easily consume the output of the Etcher CLI in real-time When using the
```
lurch: Looks like the full-stop was accidentally removed?

lurch:
I wonder if it'd make sense to spin out this "robot" functionality into its own NPM module? Or is it too etcher-specific? (obviously you'd want to make the env-var that it checks for configurable)

docs/ARCHITECTURE.md
jviotti: Fixed! See https://github.com/resin-io/etcher/pull/1029

docs/CLI.md
jviotti: Fixed! See https://github.com/resin-io/etcher/pull/1029

jviotti:
We had a discussion today about creating an "Etcher SDK", which will basically consists of the code that can be re-used between the CLI, the GUI, and any application based on Etcher.

Some things we will include here:

- Drive detection
- Image writing
- Extended formats
- The Redux store

etcetera.

So this would definitely include the robot functionality as well.

lurch:
Drive-detection and image-writing are already handled by various [modules](https://github.com/resin-io-modules) so I wonder if it makes sense to continue that pattern, and simply have the "Etcher SDK" be a collection of additional modules?

PR #1016, opened 2017-01-11
fix: allow undefined drives and images in DriveConstraintsModel

Shou:
Currently the `DriveConstraintsModel` errors when given an `undefined`
drive, or image; this commit changes that behaviour. Comes with tests.

Changelog-Entry: Don't throw errors when drives and/or images are undefined in
the DriveConstraintsModel.

### DIFF
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -16,6 +16,8 @@
 
 'use strict';
 
+const _ = require('lodash');
+
 /**
  * @summary Check if a drive is locked
  * @function
@@ -38,7 +40,7 @@
  * }
  */
 exports.isDriveLocked = (drive) => {
-  return Boolean(drive.protected || false);
+  return Boolean(_.get(drive, 'protected', false));
 };
 
 /**
@@ -60,7 +62,7 @@ exports.isDriveLocked = (drive) => {
  * }
  */
 exports.isSystemDrive = (drive) => {
-  return Boolean(drive.system || false);
+  return Boolean(_.get(drive, 'system', false));
 };
 
 /**
@@ -85,7 +87,7 @@ exports.isSystemDrive = (drive) => {
  * }
  */
 exports.isDriveLargeEnough = (drive, image) => {
-  return drive.size >= (image.size || 0);
+  return _.get(drive, 'size', 0) >= _.get(image, 'size', 0);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


Shou:
@jviotti what do you think of this one?

jviotti:
If the drive passed to these functions is `undefined` then we are in big problems. What about making the functions throw an error, like "Missing drive", instead?

Shou:
Sure, I could transform this PR to making errors more descriptive.

The reason for the change was to simplify interactions with Angular/the GUI where drives/images are sometimes briefly `undefined`, forcing checks wherever used.

jviotti:
> interactions with Angular/the GUI where drives/images are sometimes briefly undefined

Hm, that might be worth discussing in more details then. Can you provide an example? I can't remember of such an scenario, but I might be wrong.

jviotti:
Oh, are you referring to the uncaught error when opening the drive selector modal? I completely missed that! In that case I think the PR makes a lot of sense.

screenshot 2017-01-13 11 18 35

I'm still trying to understand how exactly we end up with an undefined drive here...

jviotti:
This PR definitely fixes the issue \o/

lib/shared/drive-constraints.js
```diff
@@ -85,7 +87,7 @@ exports.isSystemDrive = (drive) => {
  * }
  */
 exports.isDriveLargeEnough = (drive, image) => {
-  return drive.size >= (image.size || 0);
+  return drive ? drive.size >= _.get(image, 'size', 0) : false;
```
jviotti: I believe this can be:

```javascript
return _.get(drive, 'size', 0) >= _.get(image, 'size', 0);
```

lib/shared/drive-constraints.js
```diff
@@ -145,5 +147,5 @@ exports.isDriveValid = (drive, image) => {
  * }
  */
 exports.isDriveSizeRecommended = (drive, image) => {
-  return drive.size >= (image.recommendedDriveSize || 0);
+  return drive ? drive.size >= _.get(image, 'recommendedDriveSize', 0) : false;
```
jviotti: I believe this can be:

```javascript
return _.get(drive, 'size', 0) >= _.get(image, 'recommendedDriveSize', 0);
```

jviotti:
I made the changes suggested above myself in order to merge this and (try to) release today :)

PR #1017, opened 2017-01-11
chore: remove unnecessary modules from `npm-shrinkwrap.json`

jviotti:
Somehow some development dependencies slipped into the shrinkwrap file,
and in some cases, dependencies we don't use anymore didn't remove its
own now unneeded dependencies when running `npm uninstall`.

This PR carefully removes the packages that are not needed anymore,
which are a lot.

See: https://github.com/resin-io/etcher/issues/820

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2,38 +2,6 @@
   "name": "etcher",
   "version": "1.0.0-beta.17",
   "dependencies": {
-    "abbrev": {
-      "version": "1.0.9",
-      "from": "abbrev@>=1.0.0 <2.0.0",
-      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz"
-    },
-    "acorn": {
-      "version": "1.2.2",
-      "from": "acorn@>=1.0.3 <2.0.0",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-1.2.2.tgz"
-    },
-    "acorn-jsx": {
-      "version": "3.0.1",
-      "from": "acorn-jsx@>=3.0.0 <4.0.0",
-      "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
-      "dependencies": {
-        "acorn": {
-          "version": "3.2.0",
-          "from": "acorn@^3.0.4",
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.2.0.tgz"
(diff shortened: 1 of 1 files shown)


alexandrosm:
will this reduce the size of the binary considerably?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

jviotti:
The uncompressed size of the `node_modules` directory was reduced from 179M to 85M, which a **huge** win.

In terms of binary size, the GNU/Linux x86_64 AppImage was reduced from 68M to 54MB :tada:

I just noticed that the fact the development dependencies where included in the shrinkwrap file was caused by an NPM bug (v3). Upgrading to a later version (v4) correctly ignores dependencies of development dependencies, however it hangs at a later stage, causing the `npm shrinkwrap` command to be unusable.... I'll give yarn a try again. Hopefully the issues I encountered before were fixed and we can finally forget about this npm/shrinkwrap hell :/

alexandrosm:
that's nuts! great work juanchi! I assume we could investigate having a
blacklist of some big files chromium/electron includes which we could look
at deleting? (codecs, libraries, etc that are not supre relevant?)

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

jviotti:
I think we can delete some ffmpeg stuff (although not sure its shipped on electron by default due to licensing issues). I'll definitely research if there is something more we can do. If we really want to reduce the Electron framework size by deleting unnecessary stuff, I think we can get REALLY good results by building Electron ourselves as part of an Etcher release, customizing certain build options and applying some patches to ignore certain stuff.

lurch:
From what I can tell (I could be wrong) ffmpeg etc. is included because it's a runtime dependency of the Chrome/Chromium engine. So I wonder of there's scope for further size-reduction by disabling certain Chromium features and building that too? Although I suspect that'd massively increase build times? \*shrug\*
(or maybe that's already what @jviotti meant by building Electron ourselves?)

jviotti:
The Electron teams has pre-built Chromium packages that the Electron building system fetches from the internet to avoid having to re-compile the whole thing. Fiddling with Chromium might not be easy, but its certainly possible, and we can host our own pre-built versions of it so we only re-compile the Electron part.

lurch:
OTOH, hopefully there's more low-hanging-fruit we can pluck to reduce the binary file-sizes, before we start needing to mess about with recompiling Chromium ;)

PR #1021, opened 2017-01-12
refactor: move all byte-size conversion logic to lib/shared/units.js 

jviotti:
This commit extracts the byte-related conversions from the `byte-size`
AngularJS module and the `FlashStateModel` to a re-usable generic
CommonJS module at `lib/shared/units.js`, than can also be used by the
CLI.

### DIFF
--- a/lib/gui/models/flash-state.js
+++ b/lib/gui/models/flash-state.js
@@ -25,6 +25,7 @@ const _ = require('lodash');
 const Store = require('./store');
 const MODULE_NAME = 'Etcher.Models.FlashState';
 const FlashState = angular.module(MODULE_NAME, []);
+const units = require('../../shared/units');
 
 FlashState.service('FlashStateModel', function() {
 
@@ -132,8 +133,8 @@ FlashState.service('FlashStateModel', function() {
         speed: _.attempt(() => {
           if (_.isNumber(state.speed) && !_.isNaN(state.speed)) {
 
-            // Transform bytes to megabytes preserving only two decimal places
-            return Math.floor(state.speed / 1e+6 * 100) / 100;
+            // Preserve only two decimal places
+            return Math.floor(units.bytesToMegabytes(state.speed) * 100) / 100;
 
           }
         })
--- a/lib/shared/units.js
+++ b/lib/shared/units.js
@@ -16,21 +16,32 @@
 
 'use strict';
 
-const units = require('../../../../shared/units');
+/**
+ * @summary Convert bytes to gigabytes
+ * @function
+ * @public
+ *
+ * @param {Number} bytes - bytes
+ * @returns {Number} gigabytes
+ *
+ * @example
+ * const result = units.bytesToGigabytes(7801405440);
+ */
+exports.bytesToGigabytes = (bytes) => {
+  return bytes / 1e+9;
+};
 
-module.exports = () => {
-
-  /**
-   * @summary Convert bytes to gigabytes
-   * @function
-   * @public
-   *
-   * @param {Number} bytes - bytes
-   * @returns {Number} gigabytes
-   *
-   * @example
-   * {{ 7801405440 | gigabytes }}
-   */
-  return units.bytesToGigabytes;
-
+/**
+ * @summary Convert bytes to megabytes
+ * @function
+ * @public
+ *
+ * @param {Number} bytes - bytes
+ * @returns {Number} megabytes
+ *
+ * @example
+ * const result = units.bytesToMegabytes(7801405440);
+ */
+exports.bytesToMegabytes = (bytes) => {
+  return bytes / 1e+6;
 };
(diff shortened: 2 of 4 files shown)


jviotti:
I've updated the PR to include a byte to megabyte conversion we were doing at `FlashStateModel` that I had completely forgot about :)

lurch:
There **must** be a pre-written NPM module that already does this, surely? .... I did a quick search and found https://www.npmjs.com/package/byte-size

And IMHO the PR title should refer to "byte-size conversion" rather than "byte conversion" :)

lurch:
...or https://github.com/resin-io/resin-builder/pull/41#discussion_r88424231

jviotti:
Great find! The API is really awkward though. I'll try to find something else, otherwise we can re-use the functionality but still expose it as a reasonable API in `lib/shared/units.js`.

jviotti:
I couldn't find anything reasonable out there. The "coolest" one is `byte-size`, but it requires us to add a new dependency, plus going from something like:

```javascript
exports.bytesToGigabytes = (bytes) => {
  return bytes / 1e+9;
};
```

to

```javascript
exports.bytesToGigabytes = (bytes) => {
  return parseFloat(byteSyze(bytes, {
    unit: 'GB',
    precision: 8
  }).value);
};
```

Which is way more complex, and we even need to convert to an actual number because the module returns a string (!!).

jviotti:
> And IMHO the PR title should refer to "byte-size conversion" rather than "byte conversion" :)

Good observation, fixed!

lurch:
What's the code-size difference between just doing our own units conversion, and creating a 'friendly' wrapper around `byte-size`?

jviotti:
Not *many* more lines of course, but those new lines are denser, harder to understand, and the simple division is crystal clear about what it does.

I'll merge, but I'm happy to include a library that provides a simpler replacement for these functions without too much hassle. Or maybe resin.io should publish one? :)

lurch:
I just did a bit of testing (using https://runkit.com/npm/byte-size), and it seems that with the `byte-size` NPM module you can't actually explicitly choose which units the input gets converted to, it automatically chooses the 'best' units. So you'll only get GB-sized output, if the input size is actually >= 1GB and < 1TB.

So yeah, the simple division approach is probably easiest for now ;)
(although looking at the code, I think this means the Etcher GUI would currently display a 2TB USB HDD as 2000GB?)

lib/gui/models/flash-state.js
```diff
@@ -132,8 +133,8 @@ FlashState.service('FlashStateModel', function() {
         speed: _.attempt(() => {
           if (_.isNumber(state.speed) && !_.isNaN(state.speed)) {
 
-            // Transform bytes to megabytes preserving only two decimal places
-            return Math.floor(state.speed / 1e+6 * 100) / 100;
+            // Preserve only two decimal places
+            return Math.floor(units.bytesToMegabytes(state.speed) * 100) / 100;
```
lurch: Just out of curiosity, why is this precision-trimming done here, and not later when the value is actually used / displayed?

jviotti:
> So yeah, the simple division approach is probably easiest for now ;)
(although looking at the code, I think this means the Etcher GUI would currently display a 2TB USB HDD as 2000GB?)

Good call. We can have a `bytesToTerabytes()` conversion function and have a small GUI utility that chooses the best based on some heuristics (the GB size, for instance). I'm taking notes...

lib/gui/models/flash-state.js
jviotti: This code is really old, so I guess this was the "best" place to put it back then, but we can definitely rethink it. The precision-trimming is very nice to have in the DevTools output so the flash state output consist of a single line (otherwise DevTools hides the details and shows an arrow to show them, which makes seeing what is doing much harder), but we could also include some generic conversions on the DevTools analytics information output to trim long numbers.

I'm taking notes as well :)

lurch:
> have a small GUI utility that chooses the best based on some heuristics

Um, that's exactly what the `byte-size` NPM module does ;-)  So maybe it'd be better to simply remove the hardcoded `GB` from https://github.com/resin-io/etcher/search?utf8=%E2%9C%93&q=gb and update them to just use the `byte-size` NPM module? (which would also nicely handle the unlikely event of somebody using a very old card with capacity of < 1GB)

And then maybe the same thing could / should be used for https://github.com/resin-io/etcher/search?utf8=%E2%9C%93&q=mb too...

lib/gui/models/flash-state.js
lurch: IMHO it's always best to keep maximum-precision internally (i.e. keep the internal speed as bytes/sec) and then only trim precision and round to 'human-friendly' units at the last stage possible.

jviotti: I agree! I'll refactor soon.

jviotti:
Haha, fair enough :) In this case the extra overhead of the weird `byte-size` API is worth it.

PR #1022, opened 2017-01-12
refactor: extract application messages to `lib/shared/messages.js`

jviotti:
There are certain application messages that should be re-used between
the CLI and the GUI. In order to allow such re-usability, we extract out
the application messages used in JavaScript into
`lib/shared/messages.js` as a collection of Lodash `_.template`
templates.

Notice this file doesn't include application messages included in
Angular templates directly since it'd be hard to refactor all of them.
We plan to move to React soon, which will allow moving the remaining
messages very easily.

### DIFF
--- a/lib/shared/messages.js
+++ b/lib/shared/messages.js
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


lib/shared/messages.js
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * @summary Warning messages
+   * @namespace warning
+   * @memberof messages
+   */
+  warning: {
+
+    unrecommendedDriveSize: _.template([
+      'This image recommends a <%= image.recommendedDriveSize %>',
```
stefan-mihaila: Nice!

PR #1023, opened 2017-01-12
fix(GUI): expand button in modal menu if its the only one

jviotti:
The `.modal-menu` class holds buttons that are shown at the bottom of
the modal. If there are two buttons, the class will show them side by
side with equal sizes, however if there is only one button, it will be
incorrectly resized to 50% of the available space.

screenshot 2017-01-12 19 30 04

Change-Type: patch
Changelog-Entry: Fix alignment of single call to action buttons inside modals.

### DIFF
--- a/lib/gui/components/modal/styles/_modal.scss
+++ b/lib/gui/components/modal/styles/_modal.scss
@@ -104,8 +104,8 @@
 .modal-menu {
   display: flex;
 
-  > .button {
-    flex-basis: 50%;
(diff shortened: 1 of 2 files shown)


lurch:
I added @Shou as a reviewer - hope that's okay?

jviotti:
Yeah, that's perfect! It'd be nice if we could add people as reviewers automatically. I'll see if I can find something on the repo settings.

Issue #1032, opened 2017-01-17
Image name extends beyond tooltip when hovering over text


WasabiFan:
- **Etcher version:** Beta 17
- **Operating system and architecture:** Windows 10 x64
- **Do you see any meaningful error information on DevTools?** N/A

When hovering over the name of an image that's long and does not have natural separators, the text extends beyond the tooltip box.

![image](https://cloud.githubusercontent.com/assets/3310349/22008220/093fd350-dc2e-11e6-9e60-64d21b7c6eba.png)

This can be fixed by adding `word-wrap: break-word` to `.tooltip-inner`.

I can't test this in the newest version of Etcher right now (see #1031) so this may have already been resolved... I recall some talk of replacing this with a modal.

jviotti:
Thanks a lot for reporting! The tooltip is still present on the latest version. We have some plans to provide a new UX for this, but we should definitely fix it in the mid time.

Shou:
This should do it.

PR #1033, opened 2017-01-17
fix(gui): break and wrap bootstrap overflowing tooltip text

Shou:
Instead of letting Bootstrap tooltip text overflow, allow breaks to occur in
words.

Closes: https://github.com/resin-io/etcher/issues/1032
Changelog-Entry: Fix overflowing tooltip text by allowing breaks in words.

### DIFF
--- a/lib/gui/css/main.css
+++ b/lib/gui/css/main.css
@@ -5976,6 +5976,9 @@ body {
 [uib-tooltip] {
   cursor: default; }
 
+.tooltip {
+  word-wrap: break-word; }
+
 /*
  * Copyright 2016 resin.io
  *
--- a/lib/gui/scss/modules/_bootstrap.scss
+++ b/lib/gui/scss/modules/_bootstrap.scss
@@ -35,3 +35,8 @@ body {
 [uib-tooltip] {
   cursor: default;
 }
+
+.tooltip {
+  word-wrap: break-word;
+}
+


lurch:
Is one of these files auto-generated from the other? Can we add some sort of comment (I saw something similar to this mentioned on another resin repo recently) so that github only shows the 'real' difference?

Or if the generated file gets updated during the build, should it be added to `.gitignore`?

Apologies if I'm showing my NodeJS / CSS / Electron noob-ness ;)

Shou:
@lurch yup, `lib/gui/css/main.css` is generated by `npm run sass`, so I think we can safely add it to `.gitignore`, what do you think @jviotti?

jviotti:
The `sass` file is generated with `npm run sass`, which we leave at the responsibility of the developer (e.g: it won't be run during the build, nor during `make electron-develop`).

If we want to auto-generate it, we can definitely put that file on `.gitignore` and make the build system and `electron-develop` call `npm run sass`, but that will mean that people not concerned with developing on `scss` will still need to have all the dependencies to run it, and ruby can be quite a mess to deal with.

Another reason why we're committing the compiled form of this is to spot any compilation issue early one. Subtle scss details (e.g: the `&` position, or when using more complex `scss` stuff like loops, dictionaries, etc) might generate unexpected css that can be hard to debug, and being able to review the CSS is cool (I do review the CSS, for instance, just to double check).

lurch:
So in that case does it makes sense to do something like was discussed [here](https://github.com/resin-io/resin-sdk/pull/236#issuecomment-265349716) ?

Also, would it be worth adding `npm run sass` as a pre-commit git hook? (so that it doesn't forget to be run)

jviotti:
The https://github.com/github/linguist looks perfect. I don't truly understand how to make GitHub not show these on diffs. The documentation provides steps to mark certain files as vendor or documentation using `.gitattributes`, but then just throws a Ruby line for generated files:

```ruby
Linguist::FileBlob.new("underscore.min.js").generated? # => true
```

What are we supposed to do with that? Does it make any sense to you @lurch ?

jviotti:
We could use `git-hooks` (https://github.com/git-hooks/git-hooks/wiki/Get-Started) for it as well, or are you aware of an easier way to include git hooks along with the repo, and have them automatically configured?

lurch:
Looks like that linguist thing only works to ignore minified CSS files, ho hum :-/
https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L98
@Page- might know more about this?

My only previous experience with git hooks was https://github.com/raspberrypi/noobs#addingupdating-translations
@hedss might know more, given what he's been working on recently?

hedss:
@lurch @jviotti 
The way I've done this in the past is to create a tool which sits on-top of `git` and just does transparent calls to it for everything bar `clone`, where it clones a repo than installs the specific client-side hooks into the `.git/hooks` directory. But, it does mean everyone has to alias `git` to the new script, and when people don't... it becomes a bit of a pain to sort out.

`git-hooks` looks pretty damn good, actually, although it forces you to manually install the hooks into a particularly cloned repo. However, sharing schemes and community hooks via symlinks if possible looks ace.

On the `linguist` side, I believe the example in the `README` is there purely for you to add to `linguist`, and isn't an example of something to put in a repo. Having looked, there is a way we could force generated CSS to be assumed minified, see [here](https://github.com/github/linguist/blob/master/lib/linguist/generated.rb#L103).
So, if we create a first line over 110 characters long, `linguist` will assume the CSS is minified and will ignore it. Yes, it's bloody ugly, but it'd work, and looking at the code it could just be a comment.

jviotti:
> So, if we create a first line over 110 characters long, linguist will assume the CSS is minified and will ignore it. Yes, it's bloody ugly, but it'd work, and looking at the code it could just be a comment.

According to the docs, the file will be ignored if the **average** line length is 110 characters.

We can always minify the file for real, but that kills the purpose of me being able to review the generated CSS.

Maybe @lurch can try `git-hooks`, although this is very low priority.

lurch:
Can't promise when I'll get round to looking at git-hooks, but just for future reference: Are they any other semi-manual build-steps in Etcher that _should_ be run before each commit (like this `npm run sass`) ?

lurch:
...hmmm, and I wonder if it'd also be worth setting up a rule in the Makefile for this, so that `lib/gui/css/main.css` gets rebuilt whenever `lib/gui/scss/modules/_bootstrap.scss` (and any of the other files that get merged into it) change? :-)

jviotti:
Ah, I'm happy to move stuff from npm scripts to the `Makefile` whenever it makes sense. This is the only script that needs to be run before a commit as far as I can see.

PR #1038, opened 2017-01-20
chore: support generating armv7l .deb packages

jviotti:
This is the architecture of the Raspberry Pi 3. The following changes
were necessary:

- Detect the architecture using `uname -m`
- Set the Debian equivalent architecture to `armhf`
- Handle the new architecture correctly on `dependencies-npm.sh` and
  `electron-download-package.sh`, referring to it as simply `arm` (which
  is what `node-gyp` expects)

Steps to test this PR:

- `make electron-installer-debian` from a Raspberry Pi 3

This PR also includes `architecture-convert.sh`, which is a script to
find an architecture equivalence for Node and Debian.

### DIFF
--- a/scripts/build/electron-installer-debian-linux.sh
+++ b/scripts/build/electron-installer-debian-linux.sh
@@ -33,7 +33,7 @@ function usage() {
   echo "Options"
   echo ""
   echo "    -p <application directory>"
-  echo "    -r <application architecture>"
+  echo "    -r <architecture>"
   echo "    -c <debian configuration (.json)>"
   echo "    -o <output directory>"
   exit 1
@@ -62,19 +62,12 @@ then
   usage
 fi
 
-if [ "$ARGV_ARCHITECTURE" == "x86" ]; then
-  DEBARCH=i386
-elif [ "$ARGV_ARCHITECTURE" == "x64" ]; then
-  DEBARCH=amd64
-else
-  echo "Unsupported architecture: $ARGV_ARCHITECTURE" 1>&2
-  exit 1
-fi
+DEBIAN_ARCHITECTURE=$(./scripts/build/architecture-convert.sh -r "$ARGV_ARCHITECTURE" -t debian)
 
 cp scripts/build/debian/etcher-electron.sh "$ARGV_DIRECTORY"
 electron-installer-debian \
   --src "$ARGV_DIRECTORY" \
   --dest "$ARGV_OUTPUT" \
   --config "$ARGV_DEBIAN_CONFIGURATION" \
-  --arch "$DEBARCH"
+  --arch "$DEBIAN_ARCHITECTURE"
 rm "$ARGV_DIRECTORY/etcher-electron.sh"
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,9 @@ else
 		ifneq ($(filter %86,$(shell uname -p)),)
 			HOST_ARCH = x86
 		endif
+		ifeq ($(shell uname -m),armv7l)
+			HOST_ARCH = armv7l
+		endif
 	endif
 	ifeq ($(shell uname -s),Darwin)
 		HOST_PLATFORM = darwin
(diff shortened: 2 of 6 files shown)


jviotti:
@lurch Any better suggestions for the `architecture-get.sh` script name?

jviotti:
Ping @lurch

lurch:
Too busy to test this right now (and likely will be for the next couple of weeks, I'm afraid)

scripts/build/electron-installer-debian-linux.sh
```diff
@@ -33,7 +33,7 @@ function usage() {
   echo "Options"
   echo ""
   echo "    -p <application directory>"
-  echo "    -r <application architecture>"
+  echo "    -r <application debian architecture (from dpkg-architecture -L output)>"
```
lurch: Maybe change this to " (from the 'dpkg-architecture -L' list)"?

scripts/build/architecture-get.sh
```diff
@@ -0,0 +1,75 @@
...
+  esac
+done
+
+if [ -z "$ARGV_ARCHITECTURE" ] || [ -z "$ARGV_TYPE" ]; then
+  usage
+fi
+
+RESULT=""
+
+if [ "$ARGV_TYPE" == "electron" ]; then
```
lurch: Is this specifically "electron" architecture, or is it "nodejs" architecture? If it's both, which does it make more sense to use (and why)? :)

lurch:
> Any better suggestions for the `architecture-get.sh` script name?

Something like `get-architecture-name.sh` or `convert-architecture.sh` ?

Seems to be a bit inconsistent that some scripts use `architecture-get.sh` "internally" (i.e. they call it directly), whereas other scripts use it "externally" (i.e. they get passed the output of `architecture-get.sh`). Could that be tidied up?

scripts/build/electron-installer-debian-linux.sh
jviotti: Done!

scripts/build/architecture-get.sh
jviotti: Its the same architecture list that NodeJS used. I thought the deviated in some of them but I guess I overlooked something? I'll rename to node :)

lurch: > I thought the deviated in some of them

I guess it's theoretically possible that some of them might differ between node and electron? Would it be worth making this script support `node`, `electron` and `debian`?

jviotti:
Done! Every script now accepts the type of architecture we define (the ones that the architecture conversion script accepts), and they convert it to any other type of architecture internally if they needed.

I still need to save the debian equivalent in the Makefile to put it on some target names, but its not passed as an option anywhere.

lurch:
> I still need to save the debian equivalent in the Makefile to put it on some target names

Given that the Makefile needs `architecture-convert.sh` anyway, I wonder if it'd make more sense to "flip things around" and *only* pass the already-converted arch-names into the relevant scripts? I guess it'd make the scripts looser coupled, and mean that `architecture-convert.sh` gets mentioned in fewer places.
Or do you think there'll be scripts that might need the arch-name in more than one format?

BTW the commit-message still refers to `architecture-get.sh` ;-)

jviotti:
> Given that the Makefile needs architecture-convert.sh anyway, I wonder if it'd make more sense to "flip things around" and only pass the already-converted arch-names into the relevant scripts? I guess it'd make the scripts looser coupled, and mean that architecture-convert.sh gets mentioned in fewer places.
> Or do you think there'll be scripts that might need the arch-name in more than one format?

I don't think we'll have a use case for scripts needing multiple architecture formats, but it's not impossible neither. We should keep it as flexible as possible in case this does happen though.

The main issue I see with passing the converted architecture to every script is that from a user point of view (think about someone using our scripts independently in the future) it'll be hard to know what architecture type each script supports (error-prone) so we'll have to always clarify it in the help pages. In the other hand, the way it is now we deal with a predictable set of architecture names that all the scripts accept, which is much easier for us (to maintain) and for the user (to use).

I don't think coupling is much of an issue here. The set of scripts are part of a single "framework" (so a consistency in architecture option input sounds like a good idea), plus they are not really re-using each other (apart from a handful of special places). Some base utilities do get re-used, like the check dependency script, and this architecture convert, but I think that's reasonable, since they are more like low-level utilities than "task" scripts like the others.

> BTW the commit-message still refers to architecture-get.sh ;-)

Oops, fixing now :)

lurch:
Fair enough.
I also spotted that one or two of the scripts still refer to 'electron architecture' in their help strings.

lurch:
...and is the "making the script accept the Debian architecture we're already declaring in the Makefile" comment in the commit-message now wrong too?

jviotti:
All fixed! :)

PR #1040, opened 2017-01-23
refactor: integrate `etcher-image-stream` into the `etcher` repository

jviotti:
This is a long lasting task. The `etcher-image-stream` project takes
care of converting any kind of image input into a NodeJS readable
stream, handling things like decompression in betwee, however its a
module that, except for weird cases, there is no benefit on having
separate from the main repository.

In order to validate the assumption above, we've left the module
separate for almost a year, and no use case has emerged to keep it
like that.

This commit joins the code and tests of that module in the main
repository.


lurch:
I guess one of the downsides to this change is that it makes the Etcher unit-tests take longer to run each time?

> In order to validate the assumption above, we've left the module
separate for almost a year, and no use case has emerged to keep it
like that.

I guess with a name like `etcher-image-stream` people might have assumed it was Etcher-specific, and so not looked to see if it could be used standalone? ;)

jviotti:
Yeah, maybe, although I have a hard time thinking for a use case anyway.

> I guess one of the downsides to this change is that it makes the Etcher unit-tests take longer to run each time?

Yeah, but there are some benefits as well:

- We re-use OS X builds (so we don't have to wait for them to kick off twice), even though changes to image stream were not that often
- We get to run the test case on Electron (the test suite runs on Electron itself, where as previously `etcher-image-stream` ran in node). Probably makes little difference in practice, but it might save us some day

In any case, you know my feelings about having to manage multiple repos. I was a big proponent of that architecture but felt too much pain because of it mainly on the CLI/SDK. Lerna might be an option, but I don't feel it solves the problem 100% (I'm not liking the workflow very much). In the ideal world, I'd have the components inside `lib` (like we do have now), but we still deploy them to NPM individually. 

If we adopt that strategy, we will have image stream on the main repo, but still allow external people to use it.

lurch:
I'm afraid much of the NPM ecosystem is still a mystery to me ;-)

Your comments about Electron vs. node are interesting though - for all the other modules that we're currently using in Etcher but still in their separate repos, would it be possible / would it make sense to update their CI tests to test them both in node **and** in Electron?

jviotti:
Yeah, I think we could. The only two modules that are part of the Etcher core and reside in different repositories are `etcher-image-write` and `drivelist`. `etcher-image-write`, in particular, should run its tests on Electron for sure!

jviotti:
Pinging @jhermsmeier ^

Issue #1035, opened 2017-01-19
Warn against flashing Windows images


jviotti:
A lot of users are constantly trying to flash Windows images, which will not work on Etcher given that Windows images require special treatment.

As a solution, let's develop a regex (or collection of regexes) based on all the Windows images file names [available from MSDN](https://msdn.microsoft.com/subscriptions/securedownloads/?#searchTerm=windows&ProductFamilyId=0&Languages=en&FileExtensions=.img,.iso&PageSize=10&PageIndex=0&FileId=0) and use that to show a warning.

See https://github.com/resin-io/etcher/issues/210 for the issue tracking Windows support.

lurch:
Is there a version of that page (or similar) that's viewable without needing a Microsoft login?

jviotti:
I don't think so :(

alexandrosm:
Can we just show a simple warning if the image has the "windows" word in
it? If some other equivalently simple pattern exists in the names, maybe
let's use that, but keep it overall very simple, rather than some
aggressive regex?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

fhdk:
> Is there a version of that page (or similar) that's viewable without needing a Microsoft login?
@lurch I can generate a copy for you

PR #1043, opened 2017-01-24
feat(GUI): warn when atempting to burn Windows images

stefan-mihaila:
Change-Type: minor
Changelog-Entry: Show warning when user tries to flash a Windows image
Closes: https://github.com/resin-io/etcher/issues/1035

### DIFF
--- a/lib/gui/models/supported-formats.js
+++ b/lib/gui/models/supported-formats.js
@@ -107,15 +107,15 @@ SupportedFormats.service('SupportedFormatsModel', function() {
    * @function
    * @public
    *
-   * @param {String} image - image path
+   * @param {String} imagePath - image path
    * @returns {Boolean} whether the image is supported
    *
    * if (SupportedFormatsModel.isSupportedImage('foo.iso.bz2')) {
    *   console.log('The image is supported!');
    * }
    */
-  this.isSupportedImage = (image) => {
-    const extension = path.extname(image).slice(1).toLowerCase();
+  this.isSupportedImage = (imagePath) => {
+    const extension = path.extname(imagePath).slice(1).toLowerCase();
 
     if (_.some([
       _.includes(this.getNonCompressedExtensions(), extension),
@@ -128,7 +128,25 @@ SupportedFormats.service('SupportedFormatsModel', function() {
       return false;
     }
 
-    return this.isSupportedImage(path.basename(image, `.${extension}`));
+    return this.isSupportedImage(path.basename(imagePath, `.${extension}`));
+  };
+
+  /**
+   * @summary Check if an image seems to be a Windows image
+   * @function
+   * @public
+   *
+   * @param {String} imagePath - image path
+   * @returns {Boolean} whether the image seems to be a Windows image
+   *
+   * @example
+   * if (SupportedFormatsModel.looksLikeWindowsImage('path/to/en_windows_7_ultimate_with_sp1_x86_dvd_u_677460.iso')) {
+   *   console.log('Looks like a Windows image');
+   * }
+   */
+  this.looksLikeWindowsImage = (imagePath) => {
+    const regex = /windows|win7|win8|win10|winxp/i;
+    return regex.test(path.basename(imagePath));
   };
 
 });
--- a/lib/gui/os/open-external/open-external.js
+++ b/lib/gui/os/open-external/open-external.js
@@ -21,9 +21,25 @@
  */
 
 const angular = require('angular');
+const url = require('url');
+
 const MODULE_NAME = 'Etcher.OS.OpenExternal';
 const OSOpenExternal = angular.module(MODULE_NAME, []);
 OSOpenExternal.service('OSOpenExternalService', require('./services/open-external'));
 OSOpenExternal.directive('osOpenExternal', require('./directives/open-external'));
+
+OSOpenExternal.run(function(OSOpenExternalService) {
+  document.addEventListener('click', (event) => {
+    const target = event.target;
+    if (target.tagName === 'A' && angular.isDefined(target.href)) {
+
+      // Electron interprets relative URLs as being relative to the current loaded URL (with `webContents.loadURL`) and expands
+      // them to the corresponding absolute URL. If it's a `file://` URL, we don't want it opened in an external browser.
+      if (url.parse(target.href).protocol !== 'file:') {
+        OSOpenExternalService.open(target.href);
+      }
+    }
+  });
+});
 
 module.exports = MODULE_NAME;
--- a/lib/gui/pages/main/controllers/image-selection.js
+++ b/lib/gui/pages/main/controllers/image-selection.js
@@ -17,9 +17,17 @@
 'use strict';
 
 const _ = require('lodash');
+const Bluebird = require('bluebird');
 const messages = require('../../../../shared/messages');
 
-module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsService, ErrorService, OSDialogService) {
+module.exports = function(
+  SupportedFormatsModel,
+  SelectionStateModel,
+  AnalyticsService,
+  ErrorService,
+  OSDialogService,
+  WarningModalService
+) {
 
   /**
    * @summary Main supported extensions
@@ -65,14 +73,36 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
       return;
     }
 
-    SelectionStateModel.setImage(image);
+    Bluebird.try(() => {
+      if (!SupportedFormatsModel.looksLikeWindowsImage(image.path)) {
+        return false;
+      }
 
-    // An easy way so we can quickly identify if we're making use of
-    // certain features without printing pages of text to DevTools.
-    image.logo = Boolean(image.logo);
-    image.bmap = Boolean(image.bmap);
+      AnalyticsService.logEvent('Possibly Windows image', image);
 
-    AnalyticsService.logEvent('Select image', image);
+      // TODO: `Continue` should be on a red background (dangerous action) instead of `Change`.
+      // We want `X` to act as `Continue`, that's why `Continue` is the `rejectionLabel`
+      return WarningModalService.display({
+        confirmationLabel: 'Change',
+        rejectionLabel: 'Continue',
+        description: messages.warning.looksLikeWindowsImage()
+      });
+    }).then((shouldChange) => {
+
+      if (shouldChange) {
+        return this.reselectImage();
+      }
+
+      SelectionStateModel.setImage(image);
+
+      // An easy way so we can quickly identify if we're making use of
+      // certain features without printing pages of text to DevTools.
+      image.logo = Boolean(image.logo);
+      image.bmap = Boolean(image.bmap);
+
+      AnalyticsService.logEvent('Select image', image);
+    }).catch(ErrorService.reportException);
+
   };
 
   /**
--- a/lib/shared/messages.js
+++ b/lib/shared/messages.js
@@ -51,6 +51,13 @@ module.exports = {
     exitWhileFlashing: _.template([
       'You are currently flashing a drive.',
       'Closing Etcher may leave your drive in an unusable state.'
+    ].join(' ')),
+
+    looksLikeWindowsImage: _.template([
+      'It looks like you are trying to burn a Windows image.\n\n',
+      'Unlike other images, Windows images require special processing to be made bootable.',
+      'We suggest you use a tool specially designed for this purpose, such as',
+      '<a href="https://rufus.akeo.ie">Rufus</a>.'
     ].join(' '))
 
   },
--- a/tests/gui/models/supported-formats.spec.js
+++ b/tests/gui/models/supported-formats.spec.js
@@ -123,6 +123,36 @@ describe('Browser: SupportedFormats', function() {
 
     });
 
+    describe('.looksLikeWindowsImage()', function() {
+
+      _.each([
+        'C:\\path\\to\\en_windows_10_multiple_editions_version_1607_updated_jan_2017_x64_dvd_9714399.iso',
+        '/path/to/en_windows_10_multiple_editions_version_1607_updated_jan_2017_x64_dvd_9714399.iso',
+        '/path/to/Win10_1607_SingleLang_English_x32.iso',
+        '/path/to/en_winxp_pro_x86_build2600_iso.img'
+      ], (imagePath) => {
+
+        it(`should return true if filename is ${imagePath}`, function() {
+          const looksLikeWindowsImage = SupportedFormatsModel.looksLikeWindowsImage(imagePath);
+          m.chai.expect(looksLikeWindowsImage).to.be.true;
+        });
+
+      });
+
+      _.each([
+        'C:\\path\\to\\2017-01-11-raspbian-jessie.img',
+        '/path/to/2017-01-11-raspbian-jessie.img'
+      ], (imagePath) => {
+
+        it(`should return false if filename is ${imagePath}`, function() {
+          const looksLikeWindowsImage = SupportedFormatsModel.looksLikeWindowsImage(imagePath);
+          m.chai.expect(looksLikeWindowsImage).to.be.false;
+        });
+
+      });
+
+    });
+
   });
 
 });
--- a/lib/gui/components/warning-modal/services/warning-modal.js
+++ b/lib/gui/components/warning-modal/services/warning-modal.js
@@ -18,7 +18,7 @@
 
 const _ = require('lodash');
 
-module.exports = function(ModalService) {
+module.exports = function($sce, ModalService) {
 
   /**
    * @summary Display the warning modal
@@ -39,6 +39,7 @@ module.exports = function(ModalService) {
    * });
    */
   this.display = (options = {}) => {
+    options.description = $sce.trustAsHtml(options.description);
     return ModalService.open({
       template: './components/warning-modal/templates/warning-modal.tpl.html',
       controller: 'WarningModalController as modal',
--- a/lib/gui/components/warning-modal/templates/warning-modal.tpl.html
+++ b/lib/gui/components/warning-modal/templates/warning-modal.tpl.html
@@ -7,7 +7,7 @@
 </div>
 
 <div class="modal-body">
-  <p>{{ ::modal.options.description }}</p>
+  <p ng-bind-html="modal.options.description"></p>
 </div>
 
 <div class="modal-footer">


stefan-mihaila:
screen shot 2017-01-24 at 15 49 35

lib/gui/models/supported-formats.js
```diff
@@ -131,6 +131,24 @@ SupportedFormats.service('SupportedFormatsModel', function() {
...
+   *
+   * @param {String} image - image path
+   * @returns {Boolean} whether the image seems to be a Windows image
+   *
+   * @example
+   * if (SupportedFormatsModel.seemsWindowsImage('path/to/en_windows_7_ultimate_with_sp1_x86_dvd_u_677460.iso')) {
+   *   console.log('Looks like a Windows image');
+   * }
+   */
+  this.seemsWindowsImage = (image) => {
```
jviotti: Maybe we can refactor this function to be more generic? E.g: to allow us to see if we should display a warning for *any* image that requires special treatments, and not only Windows (for example we can extend it to detect VMWare images, which are also non-bootable without extra treatment)?

I guess we can make this function return an object containing a warning message, a warning title, maybe a documentation link, etc; or `null`/`undefined` if there is nothing to show. Then, the controller making use of this function can check if a warning was returned, and pass all these values to `WarningModalService` is a generic way.

What about something like this?

```js
this.getImageWarning = (imagePath) =>
    const basename = path.basename(imagePath);

    if (/^.+_windows_.+(?:x86|x64|ia64|arm)_dvd_.+$/.test(basename)) {
        return {
            message: messages.warning.possiblyWindowsImage(),
            confirmationLabel: '....',
            ...
        };
    }
};
```

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -65,6 +73,37 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
...
+          AnalyticsService.logEvent('User accepted possibly Windows image', image);
+          this.finalizeImageSelection(image);
+        } else {
+          AnalyticsService.logEvent('User rejected possibly Windows image', image);
+          OSOpenExternalService.open('https://github.com/resin-io/etcher/blob/master/docs/USER-DOCUMENTATION.md#why-is-my-drive-not-bootable');
+        }
+      });
+
+    } else {
+      this.finalizeImageSelection(image);
```
jviotti: After the following PR https://github.com/resin-io/etcher/pull/853 the contents of this function will be reduced to just two lines, so I guess let's just inline it here.

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -65,6 +73,37 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
       return;
     }
 
+    if (SupportedFormatsModel.seemsWindowsImage(image.path)) {
```
jviotti: What about checking the opposite condition, and returning if true, to avoid indenting the whole block below?

```js
if (!SupportedFormatsModel.seemsWindowsImage(image.path)) {
    return this.finalizeImageSelection(image);
}

AnalyticsService.logEvent('Possibly Windows image', image);

...
```

lib/gui/models/supported-formats.js
jviotti: BTW, can we change the argument name to `imagePath`? In `ImageSelectionController` we call the whole image object as `image`, so clarifying it would be really nice.

jviotti:
@konmouz Any UX input here? 

Based on your previous feedback on modals, I think we can only present a single button in the modal (e.g: "Continue") and provide enough information in the dialog such that we don't need to redirect to the documentation (hopefully making that documentation entry no longer necessary).

lib/gui/models/supported-formats.js
stefan-mihaila: I think it's a good idea to do it when we check for other images that require special treatment, but I'm not sure it's better to do it now. The advantage to make it now is that we already have some structure for adding new checks, the disadvantage is that it's harder to see the common patterns for different checks.

Anyways, let's also wait for @konmouz's input regarding the UX. BTW, I think it's also possible to change the `WarningModalService` to accept HTML, and have an anchor element in the displayed message, with an `onclick` event that will open the documentation in an external browser (by using [ngBindHtml](https://docs.angularjs.org/api/ng/directive/ngBindHtml) and [$sce.trustAsHtml](https://docs.angularjs.org/api/ng/service/$sce#trustAsHtml)) if you think it's a better UX experience.

stefan-mihaila:
@jviotti The documentation link is useful for the user because it also provides links to tools that can be used to burn Windows images. I didn't try it, but as I said in https://github.com/resin-io/etcher/pull/1043#discussion_r97748611 it might be possible to have `anchor` elements in the warning modal's text that open browser windows. cc: @konmouz

konmouz:
Can you let me know more about this process? Initially, I am think just a "continue" and "cancel" CTAs and a documentation link within the text. The think is that if you can't flash it directly, why do you see a "continue" button? What is the proper way of writing a windows image?

jviotti:
@konmouz You can flash it anyway (decide to ignore the warning), in which case if the image indeed requires special treatment, it will not work, however I think we should provide a way out in case we're detecting a file incorrectly.

Maybe "Continue" should be "Continue Anyway"?

The X button at the top right corner should de-select the image (not sure if this is the current behaviour, will check).

@stefan-mihaila 

> The documentation link is useful for the user because it also provides links to tools that can be used to burn Windows images. I didn't try it, but as I said in #1043 (comment) it might be possible to have anchor elements in the warning modal's text that open browser windows.

Yeah, that'd be nice. I think there are several entries in that user documentation that only highlight that we failed to provide clear warnings in the app itself. If the app is clear enough about images that might not boot, the entry is unnecessary, which is a great thing. If we can provide links to a tool that flashes Windows images correctly (e.g Rufus), that'd be perfect.

jviotti:
BTW, I downloaded two Windows 10 Developer Preview images recently and the regex didn't catch them:

```
Win10_1607_SingleLang_English_x32.iso
Win10_1607_SingleLang_English_x64.iso
```

Maybe adding `Win([0-9]+)` to the regex would do it?

jviotti:
Yeah, the X button opens the documentation as well, which means its very hard to dismiss the modal to select a new image.

konmouz:
In this situation we don't need an 'x' option. We should use 'Cancel' and 'Continue' like in the example below:
screen shot 2017-01-25 at 15 59 10

jviotti:
OK, cool. Editing the modal title and remove the X will require changes to our modal infrastructure (which I want to avoid on this PR), so let's do the following:

- Let's re-use the modal we have at the moment (with the warning title, and the X), and add "Change" and "Continue" buttons to it. The X button should be the same as "Continue", since if the user closes the modal, its basically saying he doesn't care about that warning.

I'll be doing some work to normalize the modals according to the new designs in the mid-time (I'll chat with you on Flowdock about it so I refactor the modals in a way is future proof for the new design as well).

lib/gui/models/supported-formats.js
```diff
@@ -128,7 +128,25 @@ SupportedFormats.service('SupportedFormatsModel', function() {
...
+   *
+   * @param {String} imagePath - image path
+   * @returns {Boolean} whether the image seems to be a Windows image
+   *
+   * @example
+   * if (SupportedFormatsModel.seemsWindowsImage('path/to/en_windows_7_ultimate_with_sp1_x86_dvd_u_677460.iso')) {
+   *   console.log('Looks like a Windows image');
+   * }
+   */
+  this.seemsWindowsImage = (imagePath) => {
```
lurch: Only a minor nitpick but `seemsWindowsImage` doesn't quite sound right. What do you think to `seemsLikeWindowsImage`, `looksLikeWindowsImage` or `possibleWindowsImage` ?

lib/shared/messages.js
```diff
@@ -52,6 +52,12 @@ module.exports = {
       'You are currently flashing a drive. Closing Etcher may leave',
       'your drive in an unusable state.\n\n',
       'Are you sure you want to close Etcher?'
+    ].join(' ')),
+
+    possiblyWindowsImage: _.template([
+      'It looks like you are trying to write a Windows image.',
+      'Windows images require a special process to be written as bootable images.',
+      'Please check the Etcher documentation for more info.'
```
lurch: I wonder if this message should somehow clarify that Etcher itself isn't capable of doing this "special process"?

(the current message could be mis-interpreted that the Etcher documentation contains info on how to get Etcher to do the magic steps?)

npm-shrinkwrap.json
```diff
@@ -2,11 +2,65 @@
   "name": "etcher",
   "version": "1.0.0-beta.18",
   "dependencies": {
+    "abbrev": {
```
stefan-mihaila: @jviotti I ran `make electron-develop`, `npm install --save angular-bind-html-compile-ci-dev` and `npm run shrinkwrap`.

`npm  --version` is `3.8.6`, `node --version` is `v6.1.0`

Am I missing something?

stefan-mihaila:
Suggestions regarding the warning message are welcome.

screen shot 2017-01-26 at 01 01 10

Also, not sure if is a problem, but the `Change` button is in red (as in dangerous) in order to make `Continue` behave like the `X` button.

This is how it would like after rebase (https://github.com/resin-io/etcher/tree/windows-image-warning-rebased):

screen shot 2017-01-26 at 01 16 09

lib/shared/messages.js
```diff
@@ -52,6 +52,13 @@ module.exports = {
       'You are currently flashing a drive. Closing Etcher may leave',
       'your drive in an unusable state.\n\n',
       'Are you sure you want to close Etcher?'
+    ].join(' ')),
+
+    looksLikeWindowsImage: _.template([
+      'It looks like you are trying to write a Windows image.',
+      'Windows images require special treatment that is not currently built into Etcher in order to be made bootable.',
+      'It is recommended to use a specialised tool such as <a href="#" os-open-external="https://rufus.akeo.ie">Rufus</a>',
```
stefan-mihaila: I guess it's not ideal that this message is now in `shared` as it contains HTML and even an `angular directive`: `os-open-external`. Might be better to extract that part as template variable.

npm-shrinkwrap.json
jviotti: There is an NPM bug I've been hitting, when development dependencies get added (this seems to be the case). Since `angular-bind-html-compile-ci-dev` has no dependencies, I guess the easy way to fix is to delete every addition except for that module?

lib/shared/messages.js
jviotti: Yeah, let's keep it inline for now. I might be wrong, but I though the `<a>` works just fine for these cases. What is its effect?

lurch:
Shouldn't the "Continue" button be the one in red, since that's the "dangerous" option?

I think the message should be made shorter, maybe something like:
> It looks like you've selected a Windows image: _&lt;filename included here&gt;_
Etcher isn't currently able to create bootable Windows disks, and we recommend you instead use a tool specially designed for the purpose, such as **Rufus**

?

jviotti:
> EDIT: Or maybe including the filename would make the dialog too wide.

Yeah, that might make it a bit weird. I think let's shorten the message as you proposed but omit the filename.

stefan-mihaila:
> Shouldn't the "Continue" button be the one in red, since that's the "dangerous" option?

@lurch Yes, it should. The only way is this way is to make "X" behave as "Continue". We can either decide to make "X" behave like "Change" or we need to make some modifications to the Warning Modal.

jviotti:
> @lurch Yes, it should. The only way is this way is to make "X" behave as "Continue". We can either decide to make "X" behave like "Change" or we need to make some modifications to the Warning Modal.

I'm working on some modal refactoring. Let's leave as it is, but include a TODO comment describing the problem. I'll fix it as part of my modal refactoring.

lib/shared/messages.js
stefan-mihaila: @jviotti If you mean using the `href` attribute for the url instead of the `os-open-external` directive: it is rendered correctly (due to `angular-bind-html-compile`, otherwise would've been displayed as text), but it doesn't do anything on click. I'm actually not sure what is its intended use in Electron.

npm-shrinkwrap.json
jviotti: OK, looks not, but I have an alternative :) Can we register a global even handler for anchors and intercept that using `OSOpenExternalService`? This could be a `.run()` block in `lib/gui/os/open-external/open-external.js`.

For example:

```
$(document).on('click', 'a[href^="http"]', function(event) {
    event.preventDefault();
    OSOpenExternalService.open(this.href);
});
```

But we should find a non-jquery way of course.

By doing this it means that you no longer have to compile the HTML inside the modal, and you can use the built-in `ng-bind-html` and avoid the extra dependency altogether.

tests/gui/models/supported-formats.spec.js
```diff
@@ -123,6 +123,35 @@ describe('Browser: SupportedFormats', function() {
 
     });
 
+    describe('.looksLikeWindowsImage()', function() {
+
+      it('should return true if filename is `en_windows_10_multiple_editions_version_1607_updated_jan_2017_x64_dvd_9714399.iso`',
```
jviotti: Since all these tests are very similar, we could have an array of paths, and loop through them creating test cases dynamically. For example:

```js
_.each([
  '/path/to/Win10_1607_SingleLang_English_x32.iso',
  '/path/to/Win10_1607_SingleLang_English_x32.iso',
  '/path/to/Win10_1607_SingleLang_English_x32.iso',
  '/path/to/Win10_1607_SingleLang_English_x32.iso'
], (imagePath) => {

  it(`should return true if filename is ${imagePath}`, function() {
    const looksLikeWindowsImage = SupportedFormatsModel.looksLikeWindowsImage(imagePath);
    m.chai.expect(looksLikeWindowsImage).to.be.true;
  });

});
```

We should also make sure to try this with Windows style paths.

lib/shared/messages.js
stefan-mihaila: I'll reply here to https://github.com/resin-io/etcher/pull/1043/files#r98021509
> OK, looks not, but I have an alternative :) Can we register a global even handler for anchors and intercept that using `OSOpenExternalService`? This could be a `.run()` block in `lib/gui/os/open-external/open-external.js`.

Yes, I thought of something on that lines also, but for some reason (probably because I wanted to see if it works 🙂) I decided to go for `os-open-external`. Now that I think about it again, I like your suggestion more because it separates presentation from behaviour. I'll give it a try.

lurch:
> We can either decide to make "X" behave like "Change" or we need to make some modifications to the Warning Modal.

I guess that implies that clicking on the "X" automatically returns the same result as clicking on one of the dialog buttons. Would it make things too strange (apologies if this is an odd question - I'm very tired) if the "X" returned a different (3rd) value?

konmouz:
we don't need the 'X' anyway but if we can't remove it yet for technical reasons it should dismiss the window and return to the main UI without changing anything.

tooomm:
Regarding the wording.. I think it might be helpful to show the filename. If it's too long, it could just be in a new line for example:

>It looks like you've selected a Windows image:
"filename included here"

I would also rephrase the Rufus part to not make people switch completely to the other tool:
> Currently, Etcher isn't able to create bootable Windows disks. In the meantime we recommend you to use a tool specially designed for this purpose, such as Rufus.

It always helps to not have just one long continuous text, but several paragraphs or linebreaks at least.

stefan-mihaila:
Updated the PR.

I went for "Currently, Etcher isn't able to create bootable Windows disks. In the meantime we recommend you to use a tool specially designed for this purpose, such as Rufus.", as @tooomm suggested as it sounded better in my opinion, but I omitted the filename, because even though it is somewhat useful, it would lead to a horizontal scrollbar in a many cases in the modal.

screen shot 2017-01-30 at 15 34 26

lib/gui/models/supported-formats.js
stefan-mihaila: @jviotti I would leave as it is for now, as we don't check for other images that need special treatment ATM and I don't think it's on our high-priority list; this way we have simpler to test code. Let me know if you think it's important to make it more generic now.

WasabiFan:
I like that phrasing, however "recommend you to" is syntactically problematic.

lurch:
> I like that phrasing, however "recommend you to" is syntactically problematic.

Yeah, removing just the "to" makes it sound much better.

> Also, I'd feel worried clicking the red "change" button... Could that be made a different color? It feels like that action is dangerous.

https://github.com/resin-io/etcher/pull/1043#issuecomment-275420668 ;-)

stefan-mihaila:
Updated message. Sorry for my bad English.

screen shot 2017-01-30 at 18 22 49

lib/gui/app.js
```diff
@@ -36,6 +36,7 @@ const app = angular.module('Etcher', [
   require('angular-ui-router'),
   require('angular-ui-bootstrap'),
   require('angular-if-state'),
+  require('angular-sanitize'),
```
jviotti: Bringing in `angular-sanitize` is probably overkill for this purpose (e.g: we're not dealing with complex whitelists, `$sanitize`, etc, which is what this module shines for). You can make this work using what Angular already provides by calling `$sce.trustAsHtml` on the HTML string.

I guess you can do something like `options.description = $sce.trustAsHtml(options.description)` on `WarningModalService`, or you could expose `this.template = $sce.trustAsHtml(options.description)` on the controller and use like `<p ng-bind-html="modal.template"></p>`.

lib/gui/models/supported-formats.js
```diff
@@ -107,15 +107,15 @@ SupportedFormats.service('SupportedFormatsModel', function() {
...
+   * @param {String} imagePath - image path
    * @returns {Boolean} whether the image is supported
    *
    * if (SupportedFormatsModel.isSupportedImage('foo.iso.bz2')) {
    *   console.log('The image is supported!');
    * }
    */
-  this.isSupportedImage = (image) => {
-    const extension = path.extname(image).slice(1).toLowerCase();
+  this.isSupportedImage = (imagePath) => {
```
jviotti: Amazing catch :D

lib/gui/os/open-external/open-external.js
```diff
@@ -26,4 +26,12 @@ const OSOpenExternal = angular.module(MODULE_NAME, []);
 OSOpenExternal.service('OSOpenExternalService', require('./services/open-external'));
 OSOpenExternal.directive('osOpenExternal', require('./directives/open-external'));
 
+OSOpenExternal.run(function(OSOpenExternalService) {
+  document.addEventListener('click', (e) => {
+    if (e.target.tagName === 'A') {
```
jviotti: Maybe worth checking that `href` is defined and its not `#`?

lib/gui/os/open-external/open-external.js
```diff
@@ -26,4 +26,12 @@ const OSOpenExternal = angular.module(MODULE_NAME, []);
 OSOpenExternal.service('OSOpenExternalService', require('./services/open-external'));
 OSOpenExternal.directive('osOpenExternal', require('./directives/open-external'));
 
+OSOpenExternal.run(function(OSOpenExternalService) {
+  document.addEventListener('click', (e) => {
```
jviotti: Can we rename `e` to `event` for consistency purposes?

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -65,14 +72,42 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
...
 
-    SelectionStateModel.setImage(image);
+    if (!SupportedFormatsModel.looksLikeWindowsImage(image.path)) {
+      SelectionStateModel.setImage(image);
 
-    // An easy way so we can quickly identify if we're making use of
-    // certain features without printing pages of text to DevTools.
-    image.logo = Boolean(image.logo);
-    image.bmap = Boolean(image.bmap);
+      // An easy way so we can quickly identify if we're making use of
```
jviotti: We can unify this whole thing and avoid repeating the image selection logic with something like this:

```javascript
Bluebird.try(() => {
  if (!SupportedFormatsModel.looksLikeWindowsImage(image.path)) {
    return true;
  }

  AnalyticsService.logEvent('Possibly Windows image', image);

  // TODO: `Continue` should be on a red background (dangerous action) instead of `Change`.
  // We want `X` to act as `Continue`, that's why `Continue` is the `rejectionLabel`
  return WarningModalService.display({
    confirmationLabel: 'Change',
    rejectionLabel: 'Continue',
    description: messages.warning.looksLikeWindowsImage()
  });
}).then((shouldContinue) => {

  if (!shouldContinue) {
    return this.reselectImage();
  }

  SelectionStateModel.setImage(image);

  // An easy way so we can quickly identify if we're making use of
  // certain features without printing pages of text to DevTools.
  image.logo = Boolean(image.logo);
  image.bmap = Boolean(image.bmap);

  AnalyticsService.logEvent('Select image', image);
```

Also notice that I catch the whole promise error and report it, so we don't hide uncaught exceptions accidentally.

jviotti:
I left some minor comments, but its looking amazing! I'm testing this on Windows as well!

jviotti:
Works great on Windows as well! :+1:

stefan-mihaila:
@jviotti, regarding https://github.com/resin-io/etcher/pull/1043#discussion_r98485056

 I really like this way of unifying the code, but it wouldn't have worked with using `Continue` as `rejectionLabel` (since we return `true` when `!SupportedFormatsModel.looksLikeWindowsImage(image.path)`, `WarningModalService` must return `true` on `Continue`).

I went for a simple solution of hiding the `X` close button, as @konmouz suggested, since it was easy to implement. This is the only place in the app where it is hidden, so that might be a little inconsistent. It would've been easy to add an option to return `true` on `X` too. Not sure what modifications you want to add to the modals, but I think this will do for this PR.

Now we also have `Continue` in red.

screen shot 2017-01-31 at 16 00 38

lib/gui/os/open-external/open-external.js
```diff
@@ -21,9 +21,23 @@
...
+
 const MODULE_NAME = 'Etcher.OS.OpenExternal';
 const OSOpenExternal = angular.module(MODULE_NAME, []);
 OSOpenExternal.service('OSOpenExternalService', require('./services/open-external'));
 OSOpenExternal.directive('osOpenExternal', require('./directives/open-external'));
 
+OSOpenExternal.run(function(OSOpenExternalService) {
+  document.addEventListener('click', (event) => {
+    const target = event.target;
+    if (target.tagName === 'A' && angular.isDefined(target.href)) {
```
stefan-mihaila: @jviotti regarding https://github.com/resin-io/etcher/pull/1043#discussion_r98483565

Didn't actually knew this, but when `href` is `#`, `target.href` is `file:///Users/stefanmihaila/resin.io/etcher/lib/gui/index.html#` on my machine. For `aaa`, it is `file:///Users/stefanmihaila/resin.io/etcher/lib/gui/aaa`.

I decided to use `OSOpenExternalService.open` just for `http` and `https` external links.

Do you think it's worth adding tests for this? If so, what's a good way to do it? Expose the event handler on `OSOpenExternal` and test it?

lurch:
I guess there's only one disadvantage to losing the 'X', it prevents the following flow:
1. User downloads a Windows ISO
2. User inserts their USB drive
3. User starts up Etcher
4. User selects their USB drive
5. User selects their Windows ISO
6. Error about Windows images pops up
7. User clicks on the X to dismiss the dialog
8. User closes Etcher and goes to download Rufus instead

With the 'X' button missing, I guess it's not clear which button they should click on in the Windows image dialog - they don't want to "Continue" because they know that Etcher won't make them a bootable image, but they don't want to "Change" because they don't have an alternative image that they want Etcher to flash instead.

(I guess perhaps the ideal scenario would be for the 'X' on a dialog to actually return a 3rd status code as I suggested earlier, and in this case when the user clicks 'X' you could just return them to the Etcher GUI (i.e. don't automatically open the file-selector), but make Etcher 'forget' that the user had chosen the Windows ISO).

@konmouz ?  Am I just overcomplicating things?

jviotti:
> I really like this way of unifying the code, but it wouldn't have worked with using Continue as rejectionLabel (since we return true when !SupportedFormatsModel.looksLikeWindowsImage(image.path), WarningModalService must return true on Continue).

I haven't check the code in detail, but can we modify my unification to return `false` and react to it as if it meant `true`, while leaving a TODO comment?

stefan-mihaila:
@lurch It sounds like a reasonable flow to me, didn't think about it.

stefan-mihaila:
> I haven't check the code in detail, but can we modify my unification to return false and react to it as if it meant true, while leaving a TODO comment?

Yes, it works... You prefer it that way?

jviotti:
> @jviotti Yes, it works... You prefer it that way?

Let's go for it so we don't have to deal with modifications to the modals in this PR. I know its a bit ugly, but a nice comment should be enough. I'll send a PR for this very soon (today or tomorrow) so we'll be able to do it right by then.

stefan-mihaila:
> Let's go for it so we don't have to deal with modifications to the modals in this PR.

@jviotti Updated PR

alexandrosm:
I also want to say that "Etcher isn't able to" is not a great warning. This is not a shortcoming of Etcher, but of the image. As such, it should be something more like "Windows images require special processing unlike other images. We suggest you use ...."

stefan-mihaila:
screen shot 2017-02-23 at 16 09 25

There's still the annoying issue of having the `Change` in red so that `x` acts as `Continue`. I can easily solve this by adding another optional parameter to the `WarningModalService`'s `display` method, that when passed will result in having the `x` act as accepting, instead of rejecting (or hide the `x` altogether). @jviotti I know you had something else in mind for the modals, alternatively you can let me know what your idea was and maybe I can take over.

stefan-mihaila:
> I think "special processing" might sound better than "special treatment" - to my mind "processing" sounds more involved (i.e. harder) than just "treatment" ;-)

ok, changed treatment -> processing

jviotti:
> @jviotti I know you had something else in mind for the modals, alternatively you can let me know what your idea was and maybe I can take over.

There are various components that define modals (e.g: the drive selector, the update notifier, the tooltip, the warning modal, etc). Given they way UI Bootstrap modals work, we can't re-use the base structure of the modals (e.g: the footer, header, etc) which means we have to be extra careful to keep them all consistent.

My idea was to build a smarter modal service that takes HTML for just the body of the template, compiles the template, and injects it at runtime inside a predefined modal base template. I also had some ideas about how to solve this problem as part of the unification process.

I think I have a local branch with some work in progress for this, so I'll review it and we can probably coordinate further.

jviotti:
Can we add a `Changelog-Entry` here?

lib/gui/models/supported-formats.js
```diff
@@ -128,7 +128,25 @@ SupportedFormats.service('SupportedFormatsModel', function() {
...
+   * @param {String} imagePath - image path
+   * @returns {Boolean} whether the image seems to be a Windows image
+   *
+   * @example
+   * if (SupportedFormatsModel.looksLikeWindowsImage('path/to/en_windows_7_ultimate_with_sp1_x86_dvd_u_677460.iso')) {
+   *   console.log('Looks like a Windows image');
+   * }
+   */
+  this.looksLikeWindowsImage = (imagePath) => {
+    const re = /windows|win7|win8|win10|winxp/i;
```
jviotti: Can we rename this variable to `regex` or something like that, just to make it ridiculously easy to the eye?

lib/gui/os/open-external/open-external.js
jviotti: I see. I like the `http` and `https` approach. In order to be on the safe side, I suggest checking if the protocol is *not* `file://` instead. Its probably a small detail but I guess in the future we might want to use ftp links, or application specific protocols?

I'd also appreciate if you can add a comment above the `protocol !== 'file:'` explaining the reason.

Regarding testing, I think its fine. There's no much to cover here anyway.

jviotti:
@stefan-mihaila I just noticed that this is not working properly when Etcher is running on Windows.

PR #1044, opened 2017-01-24
fix(GUI): emit an analytics event on ENOSPC

jviotti:
There exists a change the user will run out of space on his drive when
flashing an image format for which we can't guarantee its uncompressed
size (e.g: gzip and bzip2).

When that happens, we show a friendly error message explaining what
happened, however we should also be emitting an analytics event.

Change-Type: patch
Changelog-Entry: Emit an analytics event on `ENOSPC`.


lurch:
"change" -> "chance" ;)

PR #1049, opened 2017-01-25
feat(GUI): swap drive and image selection step order

Shou:
Change the first step to be drive selection and the second step to be
image selection, instead of vice versa as it previously was.

Changelog-Entry: Swap the order of the drive and image selection steps.

### DIFF
--- a/lib/gui/assets/image.svg
+++ b/lib/gui/assets/image.svg
@@ -1,16 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- Generator: Adobe Illustrator 17.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<!-- Generator: Adobe Illustrator 21.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 viewBox="0 0 41 41" enable-background="new 0 0 41 41" xml:space="preserve">
+	 viewBox="0 0 41 41" style="enable-background:new 0 0 41 41;" xml:space="preserve">
 <g>
-	<path fill-rule="evenodd" clip-rule="evenodd" fill="#FFFFFF" d="M20.5,0C31.822,0,41,9.178,41,20.5S31.822,41,20.5,41
-		S0,31.822,0,20.5S9.178,0,20.5,0z"/>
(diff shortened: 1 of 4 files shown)


Shou:
Following the chat discussion.

jviotti:
That was fast! :) I'll give it a go in a minute.

jviotti:
The image SVG looks very bad when we add the disabled color to it. Maybe @taahirisaacs or @konmouz can help us fix it with some SVG editing magic?

This is how it looks when we set a color to it:

screenshot 2017-01-25 11 26 02

Notice that the other ones get colored correctly, so I guess its a issue with this particular SVG.

Here's the file we're using there: https://raw.githubusercontent.com/resin-io/etcher/master/lib/gui/assets/image.svg

jviotti:
The change works like a charm. I'll approve once we sort out that SVG icon glitch.

konmouz:
@jviotti @Shou, please check SVG the code below

`<?xml version="1.0" encoding="utf-8"?>

<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
	 viewBox="0 0 41 41" style="enable-background:new 0 0 41 41;" xml:space="preserve">
<style type="text/css">
	.st0{fill:#555760;}
</style>
<g>
	<path class="st0" d="M20.5,0C9.2,0,0,9.2,0,20.5S9.2,41,20.5,41S41,31.8,41,20.5S31.8,0,20.5,0z M20.5,29c-4.7,0-8.5-3.8-8.5-8.5
		c0-4.7,3.8-8.5,8.5-8.5c4.7,0,8.5,3.8,8.5,8.5C29,25.2,25.2,29,20.5,29z"/>
	<path class="st0" d="M20.5,12.6c-4.3,0-7.9,3.5-7.9,7.9s3.5,7.9,7.9,7.9s7.9-3.5,7.9-7.9S24.8,12.6,20.5,12.6z M20.5,25
		C18,25,16,23,16,20.5s2-4.5,4.5-4.5s4.5,2,4.5,4.5S23,25,20.5,25z"/>
</g>
</svg>`

jviotti:
Hi @konmouz ,

That doesn't seem to work. `.st0` is not a class that exists on the SVG. Maybe its something being added by the SVG editor? I have little knowledge about SVG, so I might be misinterpreting everything.

What we're doing at the moment is use the `fill` property in the `<path>` elements, which works on the "flash" icon. Any clues why it doesn't work on this one? As far as I can see, the flash icon is simply the image CD icon + a flame, so maybe we can re-do the image one based on the flash one?

Here's the flash icon: https://raw.githubusercontent.com/resin-io/etcher/master/lib/gui/assets/flash.svg

konmouz:
This is what the illustrator does automatically. I changed the code to your specs, hope it works.
[image icon.svg.zip](https://github.com/resin-io/etcher/files/730213/image.icon.svg.zip)

jviotti:
Yeah, that one works beautifully. Thanks a lot @konmouz ! @Shou ^

Shou:
This should do. I changed the icon to white too. :)

PR #1055, opened 2017-01-26
refactor: remove extended archives extra functionality

jviotti:
We currently attempt to read extra metadata from ZIP files, such as a
logo, image name, image url, etc. In order to simplify the metadata
story, we decided that this metadata will not live on the image itself,
but rather on a centralised repo, which greatly simplified our custom
archive extraction logic and allows us to make use of these nice
features even when streaming the image directly from the internet.

We'll be working on bringing back this functionality from a centralised
repo in subsequent commits.

Change-Type: major
Changelog-Entry: Remove extended archives metadata extraction logic.


dlech:
Please keep this feature until a replacement is fully ready. We are currently using it. The bmap feature makes flashing way faster. We don't want to loose this!

alexandrosm:
Agreed. We shouldn't drop existing features without replacement

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

jviotti:
We're going to release the next version of Etcher with support for all this features anyway (so I doubt there will be lose of functionality), but happy to apply this along with its replacement if that makes the intention clearer.

WasabiFan:
If you're committing to not publish a release until the replacement is fully-functional and has had at least a couple weeks of general availability, I think that's fine. However you need to make sure that publishers are never left without an option for these features, and preferably there would be a release or two of overlap.

jviotti:
I reverted the commit (https://github.com/resin-io/etcher/commit/fd9d3ce749821f82947bf76ebfff11d36a678266), and will re-apply along with a commit for the replacement.

> However you need to make sure that publishers are never left without an option for these features, and preferably there would be a release or two of overlap.

I guess the tricky point here is that this was an undocumented experimental feature. As far as I'm aware ev3dev is the only publisher making use of this, so in that case I'm happy to keep this functionality until you move to the new approach.

WasabiFan:
> I reverted the commit (fd9d3ce), and will re-apply along with a commit for the replacement.

OK, sounds good.

> I guess the tricky point here is that this was an undocumented experimental feature. As far as I'm aware ev3dev is the only publisher making use of this...

It is a feature that shipped in GA Etcher releases that had incoming technical specs which were referred to by maintainers in chat and on GitHub. With such public content, "undocumented" is a bit of a stretch. I'd still doubt that any other major publishers caught on though.

jviotti:
Yeah, fair enough. I'll reach out to the community to see if any other publisher did it as well before killing the old behaviour.

PR #1057, opened 2017-01-26
refactor(GUI): get rid of the concept of recommended size

jviotti:
This feature makes no sense after
https://github.com/resin-io/etcher/pull/1055. We will consider
cherry picking functionality from this commit back into the code once we
implement the logic to fetch metadata from a central repository.


jviotti:
See https://github.com/resin-io/etcher/pull/1055

PR #1058, opened 2017-01-27
chore: add a `package` Makefile target

jviotti:
This target would easily allow us to create a package for the
application without being wrapped into an installer, which is
particularly useful to provide custom builds, or test asar related
issues, for example.


lurch:
Hah, I suspect this is another one of those `Makefile` improvements that I'd made locally, but hadn't got round to separating out and PR-ing yet ;)

PR #1059, opened 2017-01-27
chore: fix undefined TEMPORARY_DIRECTORY variable

jviotti:
This variable was renamed to `BUILD_TEMPORARY_DIRECTORY`, but the
Makefile silently expands this to nothing and carries on with the build.

Now that the variable is expanded correctly, we must add it as an
order-only prerequisite so it gets created.


lurch:
Just for reference, this mistake occurred in #946

lurch:
LGTM

PR #1061, opened 2017-01-27
test(GUI): fix non-running test in DrivesModel test suite

jviotti:
We have a test to ensure a drive is not auto-selected when its
protected, however the real functionality of this test was not being
tested given that the size the fake drive had was not enough to hold the
image.

### DIFF
--- a/tests/gui/models/drives.spec.js
+++ b/tests/gui/models/drives.spec.js
@@ -210,7 +210,7 @@ describe('Browser: DrivesModel', function() {
               {
                 device: '/dev/sdb',
                 name: 'Foo',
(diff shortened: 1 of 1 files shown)


lurch:
Related to this: presumably Etcher should also not auto-select system drives, when you've enabled 'dangerous mode'? Currently it _does_ auto-select the system drive (when there's only one available), which IMHO doesn't "feel right"?

jviotti:
Good point. I'll send a PR for it.

jviotti:
See https://github.com/resin-io/etcher/pull/1069

Issue #830, opened 2016-11-03
Drive-filtering suggestion


lurch:
After https://github.com/resin-io-modules/drivelist/issues/98 has been implemented (i.e. we can get info about _all_ the mountpoints which are used by each physical drive), we could filter out the drive on which the image we're going to be flashing is stored on.

Currently, this scenario is possible (which would be prevented if this issue is implemented) :
 1. User has a USB Flash drive (e.g. `/dev/sdb`) and a USB SD-reader (e.g. `/dev/sdc`) connected to their computer
 2. User is running out of space on their computer's HDD, so they download e.g. `2016-09-23-raspbian-jessie-lite.zip` to their USB Flash drive
 3. User launches Etcher, and selects the `2016-09-23-raspbian-jessie-lite.zip` file they've stored on their USB Flash drive
 4. User accidentally selects `/dev/sdb` as the drive they want to flash to
 5. User clicks `Flash!`
 6. ???? (I didn't actually try this myself to see what happens)

If this issue was fixed, `/dev/sdb` would automatically be filtered out (until the user selects an image _not_ stored on their USB Flash drive), and Etcher would auto-select `/dev/sdc`.

jviotti:
Very good observation, we should definitely be doing this, so I'll mark it as a bug.

jviotti:
Should we omit the drive altogether, or display a warning label somehow describing that the drive is the place where the image resides?

This is what we do for locked SD Cards for example (although this is an old screenshot the labels are being shown below the drive information at the moment):

![37227a66-2fea-11e6-8ebc-baa15d9b55f9](https://cloud.githubusercontent.com/assets/2192773/20153196/438d635e-a6ca-11e6-81e7-8f31e82a29b3.png)

@konmouz ?

WasabiFan:
Side-note: It would be great if labels like this "locked" one explained _why_ the device is locked. I could easily imagine a user looking at this and thinking that _Etcher_ locked it as opposed to the device being locked by the OS/hardware. Is there a tooltip already there that serves this purpose?

jviotti:
Very good point! Something for @konmouz and @taahirisaacs to help us with.

lurch:
I noticed this evening that an 'unselectable' drive (in my specific case I was experimenting with an image bigger than the drive) in that dialog still displays a hand-cursor when you mouse-hover over it, which it probably shouldn't.

jviotti:
[...]

PR #1066, opened 2017-01-30
fix(GUI): prevent flashing the drive where the source image is located

jviotti:
Currently Etcher will allow you to flash an image to the same drive that
contains the image. As a way to protect against that case we introduce
the concept of a "source drive", which means a drive that contains the
source image.

This commit adds the following logic around this new concept:

- Don't auto-select a source drive
- De-select an already selected drive if an image inside it is selected
- Disable the drive in the drive selector modal
- Add a "Source Drive" badge to the drive in the drive selector modal

screenshot 2017-01-30 17 16 31

Fixes: https://github.com/resin-io/etcher/issues/830
Change-Type: minor
Changelog-Entry: Prevent flashing the drive where the source image is located.

### DIFF
--- a/lib/gui/models/store.js
+++ b/lib/gui/models/store.js
@@ -95,8 +95,8 @@ const storeReducer = (state, action) => {
 
         const drive = _.first(action.data);
 
-        // Even if there's no image selectected, we need to call `isDriveValid`
-        // to check if the drive is locked, and `{}` works fine with it
+        // Even if there's no image selected, we need to call several
+        // drive/image related checks, and `{}` works fine with them
         const image = state.getIn([ 'selection', 'image' ], Immutable.fromJS({})).toJS();
 
         if (_.every([
@@ -268,7 +268,7 @@ const storeReducer = (state, action) => {
 
       return _.attempt(() => {
         if (selectedDrive && !_.every([
-          constraints.isDriveLargeEnough(selectedDrive.toJS(), action.data),
+          constraints.isDriveValid(selectedDrive.toJS(), action.data),
           constraints.isDriveSizeRecommended(selectedDrive.toJS(), action.data)
         ])) {
           return storeReducer(state, {
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -17,6 +17,7 @@
 'use strict';
 
 const _ = require('lodash');
+const pathIsInside = require('path-is-inside');
 
 /**
  * @summary Check if a drive is locked
@@ -63,6 +64,52 @@ exports.isDriveLocked = (drive) => {
  */
 exports.isSystemDrive = (drive) => {
   return Boolean(_.get(drive, 'system', false));
+};
+
+/**
+ * @summary Check if a drive is source drive
+ * @function
+ * @public
+ *
+ * @description
+ * In the context of Etcher, a source drive is a drive
+ * containing the image.
+ *
+ * @param {Object} drive - drive
+ * @param {Object} image - image
+ * @returns {Boolean} whether the drive is a source drive
+ *
+ *
+ * @example
+ * if (constraints.isSourceDrive({
+ *   device: '/dev/disk2',
+ *   name: 'My Drive',
+ *   size: 123456789,
+ *   protected: true,
+ *   system: true,
+ *   mountpoints: [
+ *     {
+ *       path: '/Volumes/Untitled'
+ *     }
+ *   ]
+ * }, {
+ *   path: '/Volumes/Untitled/image.img',
+ *   size: 1000000000
+ * })) {
+ *   console.log('This drive is a source drive!');
+ * }
+ */
+exports.isSourceDrive = (drive, image) => {
+  const mountpoints = _.get(drive, 'mountpoints', []);
+  const imagePath = _.get(image, 'path');
+
+  if (!imagePath || _.isEmpty(mountpoints)) {
+    return false;
+  }
+
+  return _.some(_.map(mountpoints, (mountpoint) => {
+    return pathIsInside(imagePath, mountpoint.path);
+  }));
 };
 
 /**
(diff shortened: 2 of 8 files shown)


jviotti:
I've set an down arrow icon to the badge since I guess it represents that something is "inside it", but let me know if you think another icon [from here](http://getbootstrap.com/components/#glyphicons-glyphs) fits better.

lib/shared/drive-constraints.js
```diff
@@ -66,6 +68,66 @@ exports.isSystemDrive = (drive) => {
...
+  const mountpoints = _.get(drive, 'mountpoints', []);
+
+  if (!image || !image.path || _.isEmpty(mountpoints)) {
+    return false;
+  }
+
+  // NodeJS only allows manipulation Windows paths
+  // from Windows and UNIX paths from UNIX, so we
+  // have to manually set/unset `path.sep` to make
+  // this piece of code truly cross-platform.
```
lurch: Huh? Presumably this is only for the unit-tests though, since Etcher running on Windows will never be passed a Unix-style path, and vice-versa? If that's the case, wouldn't it be better for the unit-tests to be modified to check current-platform paths, rather than 'polluting' this module with mucking around with `path.sep`?

tests/shared/drive-constraints.spec.js
```diff
@@ -254,50 +430,140 @@ describe('Shared: DriveConstraints', function() {
...
-          device: '/dev/disk2',
-          name: 'My Drive',
-          size: 4000000000
-        };
-        this.image = {
-          path: 'rpi.img',
-          size: 2000000000
-        };
-      });
+    it('should return false if the drive is not large enough, is locked, and is a source drive', function() {
```
lurch: Is there any way these eight separate tests (i.e. all the combinations of three separate conditions) could be run in some kind of auto-generated loop, rather than laboriously being typed out individually? Obviously if we added another couple of extra conditions to "what makes a drive valid" we wouldn't really want to have to type out 32 individual tests.

tests/shared/drive-constraints.spec.js
```diff
@@ -92,6 +92,182 @@ describe('Shared: DriveConstraints', function() {
...
+        });
+
+        m.chai.expect(result).to.be.false;
+      });
+
+    });
+
+    describe('given UNIX paths', function() {
+
+      it('should return true if the image lives directly inside a mount point of the drive', function() {
```
lurch: Probably also worth adding a unit-test for a mountpoint being `/` and the image then being at `/image.img` ?

lib/gui/models/store.js
```diff
@@ -95,8 +95,8 @@ const storeReducer = (state, action) => {
 
         const drive = _.first(action.data);
 
-        // Even if there's no image selectected, we need to call `isDriveValid`
-        // to check if the drive is locked, and `{}` works fine with it
+        // Even if there's no image selectected, we need to call several
```
lurch: Typo: selected

lib/shared/drive-constraints.js
```diff
@@ -114,7 +176,11 @@ exports.isDriveLargeEnough = (drive, image) => {
  * }
  */
 exports.isDriveValid = (drive, image) => {
-  return !this.isDriveLocked(drive) && this.isDriveLargeEnough(drive, image);
+  return _.every([
+    !this.isDriveLocked(drive),
+    this.isDriveLargeEnough(drive, image),
+    !this.isSourceDrive(drive, image)
+  ]);
```
lurch: Nice :)

lurch:
I wonder how it'd look with `glyphicon-folder-open`, indicating that it's the drive containing the folder the image is opened from?

tests/shared/drive-constraints.spec.js
jviotti: Done!

tests/shared/drive-constraints.spec.js
jviotti: Haha, exhaustive testing FTW :) Let me see what I can do.

lib/gui/models/store.js
jviotti: Fixed. I wonder if there exists a tool to check grammar in comments. Have you heard about such a tool?

lib/shared/drive-constraints.js
jviotti: Yeah, that was my initial instinct, however it feels weird that some tests only execute in certain OSes. In any case, I agree that polluting the function is worse than conditionally executing tests, so I'll do that instead!

lib/gui/models/store.js
lurch: I've never used it myself, but in other repos I've seen people submitting typo PRs found by a tool called [codespell](https://github.com/lucasdemarchi/codespell).

(And of course just to nitpick, that fixes spelling, not grammar ;) )

jviotti:
@lurch Can you take another look?

jviotti:
This is how the folder icon looks like:
screenshot 2017-01-31 09 58 38

lib/gui/models/store.js
jviotti: Oh, beautiful. It works great! I'll include it in our linting phase :)

jviotti: See this: https://github.com/resin-io/etcher/pull/1067

lurch:
Personally I think the open-folder icon looks nice, but we should get @konmouz 's opinion before merging :)

konmouz:
@lurch and @jviotti looks good to me!

tests/gui/models/selection-state.spec.js
```diff
@@ -485,6 +487,41 @@ describe('Browser: SelectionState', function() {
...
         });
 
+        it('should de-select a previously selected source drive', function() {
+          const imagePath = _.attempt(() => {
+            if (process.platform === 'win32') {
+              return 'E:\\bar\\foo.img';
+            }
+
+            return '/mnt/bar/foo.img';
+          });
```
lurch: It's a shame this differs slightly from the similar function in `tests/gui/models/drives.spec.js`

Issue #1034, opened 2017-01-17
Error parsing Info.plist at launch


jspencer:
- **Etcher version: v1.0.0-beta.18**
- **Operating system and architecture: macOS 10.12.3 Beta (16D30a)**

The following error appears at launch:

```
Error: 2017-01-17 09:27:25.354 diskutil[12682:4102330] There was an error parsing the Info.plist for the bundle at URL <0x7fb6d4c09800>: NSCocoaErrorDomain - 3840
2017-01-17 09:27:25.508 diskutil[12719:4102412] There was an error parsing the Info.plist for the bundle at URL <0x7fb0781010b0>: NSCocoaErrorDomain - 3840
2017-01-17 09:27:25.683 diskutil[12758:4102505] There was an error parsing the Info.plist for the bundle at URL <0x7fb9d4c037c0>: NSCocoaErrorDomain - 3840
2017-01-17 09:27:25.806 diskutil[12799:4102581] There was an error parsing the Info.plist for the bundle at URL <0x7fa011413e60>: NSCocoaErrorDomain - 3840
2017-01-17 09:27:25.928 diskutil[12836:4102642] There was an error parsing the Info.plist for the bundle at URL <0x7ff683c033b0>: NSCocoaErrorDomain - 3840

    at childProcess.execFile (/Applications/Etcher.app/Contents/Resources/app.asar/node_modules/drivelist/lib/scripts.js:90:23)
    at ChildProcess.exithandler (child_process.js:209:7)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Socket.<anonymous> (internal/child_process.js:334:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at Pipe._handle.close [as _onclose] (net.js:493:12)
```

I'd be happy to help debug the problem. I'm not sure if it's related to the beta OS I am running.

jviotti:
Thanks a lot for reporting this! Can you try running the following bash script in your system and report back what you get? https://github.com/resin-io-modules/drivelist/blob/master/scripts/darwin.sh

jspencer:
Sure thing. Here's the output.

Silicates:~ jspencer$ ./darwin.sh 
[...]

PR #1068, opened 2017-01-31
upgrade: `drivelist` to v5.0.9

jviotti:
This version doesn't throw an error if there is `stderr` output from the
scripts but the scripts exit with code zero.

Instead, the module prints the `stderr` output to a `debug` channel, so
we enable debug information by setting a `localStorage` value in order
to see what's going on.

Fixes: https://github.com/resin-io/etcher/issues/1034
Change-Type: patch
Changelog-Entry: Ignore `stderr` output from drive detection scripts if they exit with code zero.

### DIFF
--- a/lib/gui/index.html
+++ b/lib/gui/index.html
@@ -7,6 +7,10 @@
     <link rel="stylesheet" type="text/css" href="css/desktop.css">
     <link rel="stylesheet" type="text/css" href="css/angular.css">
 
+    <!-- Enable debug information from all modules that use `debug` -->
+    <!-- See https://github.com/visionmedia/debug#browser-support -->
+    <script>window.localStorage.debug = '*';</script>
+
     <script>
       window._trackJs = {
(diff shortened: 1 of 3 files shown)


lurch:
Just having a quick look at this again, could you document somewhere how to "enable debug information by setting a localStorage value" please? :)

jviotti:
@lurch I'm enabling debug mode by default, and including a comment with a link to the docs that describe how I did it in more detail. Do you think I should expand the comment, or move it somewhere else?

lurch:
Sorry, brain-fart ;-)
I misread "we enable debug information by ..." as meaning "debug information _can_ be enabled by ...", I misread `window.localStorage.debug = '*'` as `window.localStorage.debug = ''` (i.e. initialising it to off), and I saw `#browser-support` in the URL and thought that that meant it only applied to web-browsers and not Etcher.

... I clearly didn't get enough sleep last night! :-D

PR #1071, opened 2017-01-31
style(GUI): move update notifier checkbox inside modal body

jviotti:
This is a small step towards unifying all the modal skeletons that we
use through the application.

By moving the checkbox to the modal body we can declare the footer and
the header once in a single place.

screenshot 2017-01-31 15 58 56


jviotti:
Thanks for the heads up! I'll rebase in a minute :)

jviotti:
Done!

PR #1076, opened 2017-02-02
chore: get rid of commitizen

jviotti:
No one on the team is using this and the adaptor we're using proved to
be suboptimal. We might consider this again once we develop a good
Commitizen adapter integrated more closely with Versionist.


lurch:
> No one on the team is using this

I _tried_ using it once ;-)  https://github.com/resin-io/etcher/pull/777#issuecomment-256514312

PR #1077, opened 2017-02-02
chore: add an HTML linter and fix its warnings

jviotti:
This PR creates a script at `scripts/html-lint.js` that runs a NodeJS
HTML linter package that understands AngularJS on top of all our HTML
templates.

The following error was detected and fixed:

```
1 errors at lib/gui/pages/finish/templates/success.tpl.html
  [4:63] Duplicate attribute “class”.
```

This script will run as part of `npm run lint`, but you can call it
independently as `npm run htmllint`.

screenshot 2017-02-02 13 37 44


lurch:
LGTM. Perhaps @emirotin and @craig-mulligan would find this interesting too?

PR #1078, opened 2017-02-02
chore: publish snapshot builds to S3 

jviotti:
This commit makes Appveyor and Travis CI publish snapshot builds to S3
when a pull request is merged, by making use of the `publish-aws-s3`
Makefile target.

The changes required for such type of deployment are the followings:

- Set `S3_BUCKET` to `resin-nightly-downloads` when doing snapshot
  builds

- Add deploy sections to `.travis.yml` and `appveyor.yml` that run `make
  publish-aws-s3`

- Don't change `PRODUCT_NAME` when doing snapshot builds (given we'll be
  publishing to a different S3 bucket)

- Install `awscli` in Appveyor CI and Travis CI

- Make GNU/Linux Docker containers inherit `AWS_ACCESS_KEY_ID` and
  `AWS_SECRET_ACCESS_KEY` from the environment (so `awscli` is
  configured correctly inside them)

- Add a boolean option to `aws-s3.sh` publish script to prepend a
  timestamp to the S3 path, so we can more easily distinguish files
  inside the `resin-nightly-downloads` bucket

- Print the published URL from `aws-s3.sh` for convenience purposes, so
  we can click it when skimming through CI builds logs

- Add the `-R` and `-L` options when recursively copying `node_modules`
  during a snapshot build to prevent weird Appveyor errors related to
  hard links. The options listed before make sure that we recursively
  resolve every link while copying

- Move from `wget` to `curl` to avoid certificate check failures


scripts/build/download-tool.sh
```diff
@@ -78,7 +78,7 @@ if [ -f "$ARGV_OUTPUT" ]; then
   fi
 fi
 
-wget --no-check-certificate "$ARGV_URL" -O "$TEMP_OUTPUT"
+wget --no-verbose --no-check-certificate "$ARGV_URL" -O "$TEMP_OUTPUT" 2>&1
```
jhermsmeier: Not necessarily related, but is there a specific reason we'd want `--no-check-certificate` here (and below)? Considering we're doing production builds with this, we should error on broken / invalid certs, imo.

lurch: When we're doing local builds, seeing the verbose output is actually useful. Is there some envvar we can check for to see if we're running on a CI server, and use that to change the command-args as appropriate?
(and why is the `2>&1` necessary?)

scripts/build/electron-download-package.sh
```diff
@@ -60,7 +60,7 @@ ELECTRON_ARCHITECTURE=$(./scripts/build/architecture-convert.sh -r "$ARGV_ARCHIT
 ELECTRON_GITHUB_REPOSITORY=https://github.com/electron/electron
 ELECTRON_DOWNLOADS_BASEURL="$ELECTRON_GITHUB_REPOSITORY/releases/download/v$ARGV_ELECTRON_VERSION"
 ELECTRON_FILENAME="electron-v$ARGV_ELECTRON_VERSION-$ARGV_OPERATING_SYSTEM-$ELECTRON_ARCHITECTURE.zip"
-ELECTRON_CHECKSUM=$(wget --no-check-certificate -O - "$ELECTRON_DOWNLOADS_BASEURL/SHASUMS256.txt" | grep "$ELECTRON_FILENAME" | cut -d ' ' -f 1)
+ELECTRON_CHECKSUM=$(wget --no-verbose --no-check-certificate -O - "$ELECTRON_DOWNLOADS_BASEURL/SHASUMS256.txt" | grep "$ELECTRON_FILENAME" | cut -d ' ' -f 1)
```
lurch: See previous comment ;)

scripts/publish/aws-s3.sh
```diff
@@ -31,20 +31,23 @@ function usage() {
   echo "    -b <s3 bucket>"
   echo "    -v <version>"
   echo "    -p <product name>"
+  echo "    -d include date in S3 key"
```
lurch: I wonder if it'd be more flexible to actually make this `-d <string>  include string in S3 key` (and have it be optional), and then explicitly pass in the date string from higher-up-the-stack, as that'd then allow us to re-use the same functionality for publishing debug builds with custom keys?

scripts/publish/aws-s3.sh
```diff
@@ -59,8 +62,20 @@ fi
...
+
 aws s3api put-object \
   --bucket "$ARGV_BUCKET" \
   --acl public-read \
-  --key "$ARGV_PRODUCT_NAME/$ARGV_VERSION/$FILENAME" \
+  --key "$S3_KEY" \
   --body "$ARGV_FILE"
+
+# Escape plus signs when printing the final URL
+URL="$(echo "https://$ARGV_BUCKET.s3.amazonaws.com/$S3_KEY" | sed 's/\+/%2B/g')"
```
lurch: I wonder if we could get this URL stored 'somewhere' so it'd be easy for us (and / or others?) to get the URL for the latest snapshot build(s) without having to trawl through build-server logs?

scripts/build/docker/run-command.sh
```diff
@@ -82,6 +82,8 @@ fi
 # The `-t` and TERM setup is needed to display coloured output.
 docker run -t \
   --env "TERM=xterm-256color" \
+  --env "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID" \
```
lurch: Presumably this is using the secret / encrypted environment variables thing?

.travis.yml
```diff
@@ -34,8 +34,8 @@ before_install:
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      pip install codespell;
-      npm install -g bower;
+      pip install codespell awscli;
+      npm install -g bower asar;
```
lurch: Why is `asar` suddenly now an extra requirement, when it wasn't before?

.travis.yml
```diff
@@ -56,6 +56,16 @@ script:
...
 
+deploy:
+  provider: script
+  script: if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+            ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "make publish-aws-s3";
+          else
+            make publish-aws-s3;
+          fi
+  on:
+    branch: master
```
lurch: Does this mean that snapshot builds will only be created on merges to master, and not for every PR?
(and if so, I think that's the right behaviour)

lurch: Hmmm, if somebody attempted to change this setting within a PR, is Travis actually clever enough to not 'act' on that setting until it got merged into the master-version? (and same question for Appveyor too)

Makefile
```diff
@@ -139,13 +140,7 @@ endif
...
 TARGET_ARCH_DEBIAN = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t debian)
 
-ifeq ($(RELEASE_TYPE),production)
-	PRODUCT_NAME = etcher
-endif
-ifeq ($(RELEASE_TYPE),snapshot)
-	PRODUCT_NAME = etcher-snapshots
-endif
-
+PRODUCT_NAME = etcher
```
lurch: So I guess the longer `APPLICATION_VERSION` is now the only differentiation between locally-built `snapshot` and `production` builds?

Makefile
```diff
@@ -372,11 +367,20 @@ endif
 
 ifdef PUBLISH_AWS_S3
 publish-aws-s3: $(PUBLISH_AWS_S3)
+ifeq ($(RELEASE_TYPE),production)
 	$(foreach publishable,$^,$(call execute-command,./scripts/publish/aws-s3.sh \
 		-f $(publishable) \
 		-b $(S3_BUCKET) \
 		-v $(APPLICATION_VERSION) \
 		-p $(PRODUCT_NAME)))
+else
```
lurch: Should this explicitly be `ifeq ($(RELEASE_TYPE),snapshot)` rather than `else` ?  (in case we decide to add other RELEASE_TYPEs in future)

scripts/build/docker/run-command.sh
WasabiFan: It's worth noting that, as it is, I see no reason that I couldn't hijack your keys and publish whatever I wanted to your S3 repo. Even if the environment vars are encrypted, I could still open a PR that does `echo $AWS_ACCES_KEY_ID` and your secrets are spilled. In fact, it might automatically show up in the build log as commands are run...

jviotti: @WasabiFan As @lurch pointed out, the environment variables are not accessible from pull requests originating from forks: https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings

lurch: Should the `RELEASE_TYPE` variable also be transmitted from Travis to Docker using the same mechanism?

.travis.yml
```diff
@@ -34,8 +34,8 @@ before_install:
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      pip install codespell;
-      npm install -g bower;
+      pip install codespell awscli;
+      npm install -g bower asar;
       brew install afsctool;
       brew install jq;
```
jviotti: Fixed!

scripts/build/download-tool.sh
jviotti: It fails on MinGW for some reason (I didn't investigate in depth, but I will).

.travis.yml
jviotti: @lurch `asar` is required to create the package, while previously we were only running the test suite.

.travis.yml
```diff
@@ -56,6 +56,16 @@ script:
       ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "xvfb-run --server-args=$XVFB_ARGS npm test && ./scripts/ci/ensure-staged-sass.sh";
     fi
 
+deploy:
+  provider: script
+  script: if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+            ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "make publish-aws-s3";
+          else
```
jviotti: I fixed the `else` inconsistency.

> (and given that we invariably run the same command (here make publish-aws-s3) in both the osx and linux sections, I wonder if the repetition could be eliminated? I'm not very familiar myself with the syntax / logic allowed in Travis config files)

Yeah, I know its ugly, but I can't think of anything, apart from adding the command to a global environment variable at the top of the file and re-use that, but sounds ugly as well.

.travis.yml
jviotti: That's a very good question. I'll get in touch with their support.

Makefile
```diff
@@ -26,17 +26,18 @@ APPLICATION_COPYRIGHT = $(shell jq -r '.copyright' package.json)
...
 RELEASE_TYPE ?= snapshot
 PACKAGE_JSON_VERSION = $(shell jq -r '.version' package.json)
 ifeq ($(RELEASE_TYPE),production)
 APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)
+S3_BUCKET = resin-production-downloads
 endif
 ifeq ($(RELEASE_TYPE),snapshot)
 CURRENT_COMMIT_HASH = $(shell git log -1 --format="%h")
 APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)+$(CURRENT_COMMIT_HASH)
+S3_BUCKET = resin-nightly-downloads
```
jviotti: Yeah, I agree. I didn't choose the name though :(

Makefile
jviotti: Yeah, exactly.

Makefile
jviotti: Sounds good. Fixed!

scripts/build/download-tool.sh
jviotti: Sounds good. Most CI servers define a `CI` env var that we can use to check it. Regarding the `stderr` piping, it was part of my PowerShell debugging attempts, and it looks like I forgot to remove it.

scripts/build/electron-download-package.sh
jviotti: Fixed!

scripts/build/docker/run-command.sh
jviotti: Good catch. Fixed!

Makefile
```diff
@@ -192,7 +187,7 @@ $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_A
 	./scripts/build/electron-create-resources-app.sh -s . -o $@ \
 		-v $(APPLICATION_VERSION) \
 		-f "$(APPLICATION_FILES)"
-	$(foreach prerequisite,$^,$(call execute-command,cp -rf $(prerequisite) $@))
+	$(foreach prerequisite,$^,$(call execute-command,cp -RLrf $(prerequisite) $@))
```
jviotti: Interesting. My BSD implementation of `cp` claims that `-r` is for backwards compatibility purposes and that is similar to `-R`, but not identical. Anyway, let's try just `-R` and see how it goes on the build servers.

scripts/publish/aws-s3.sh
jviotti: Yeah, I'd love that, but I can't think of any place for them. I guess we could make a small HTTP server that accepts requests for them and provides a nice way to explore the bucket?

scripts/publish/aws-s3.sh
jviotti: Sounds good, done!

scripts/build/download-tool.sh
jhermsmeier: Oh, that's probably because MinGW doesn't ship root certs and doesn't point to existing ones either. Maybe `curl` works (I doubt it)? If nothing does, we could add a `.js` script that uses Node's `http(s)` to download stuff, as that'll have everything available & will work on any platform.

.travis.yml
lurch: Yeah, fair enough.

Makefile
lurch: Hmmm, how annoying! I found http://www.unix.com/man-page/FreeBSD/1/cp/ which says "In FreeBSD, `-r` is a synonym for `-RL` and works the same unless modified by other flags" and https://en.wikipedia.org/wiki/Cp_(Unix) which says "Using `cp -R /home/nick/clients/ /home/nick/customers` on a GNU system it behaves as expected; however, on a BSD system, it copies all the _contents_ of the "clients" directory, instead of the directory _clients_ itself. The same happens in both GNU and BSD systems if the path of the source directory ends in `.` or `..` (with or without trailing slash)."

scripts/build/docker/run-command.sh
lurch: ...and do you now also need to add the `CI` variable to Docker too?

scripts/publish/aws-s3.sh
```diff
@@ -30,21 +30,24 @@ function usage() {
   echo "    -f <file>"
   echo "    -b <s3 bucket>"
   echo "    -v <version>"
-  echo "    -p <product name>"
+  echo "    -n <product name>"
+  echo "    -p S3 key prefix"
```
lurch: Out of curiosity, why the backwards-incompatible switching of the `-p` flag to a different option? :-/

scripts/publish/aws-s3.sh
lurch: Or even something as simple as a cgi-script we can call from this script with `wget` (using a secret CI var for authentication) that updates a static (publicly viewable) html list of snapshot builds that we can view via a web-browser.

scripts/build/download-tool.sh
jviotti: I looks like `curl` behaves, but let's see in the CI services :)

Makefile
jviotti: Also add the MinGW's version of `cp` into the mix :P The problem is gone on Windows if I set `-RL` but not if I pass `-r`, so it must mean different things over there.

scripts/build/electron-download-package.sh
jviotti: Maybe you're seeing an outdated diff?

scripts/publish/aws-s3.sh
jviotti: Can you create a ticket for it so it doesn't get lost, until we can allocate a Friday hack to it? :P

scripts/build/docker/run-command.sh
jviotti: Done!

scripts/publish/aws-s3.sh
jviotti: So we can use p for **p**refix. I guess we can also keep `-p` for product name, and use `-x` for the prefix. What do you think?

lurch: I think it'd be better to not change the meanings of existing flags.
You could add `-k` for **k**ey prefix? ;)

scripts/publish/aws-s3.sh
lurch: #1185 :-)

Makefile
lurch: Sorry, I wasn't implying that one option is better than another - we should just use whatever works :-)

scripts/build/download-tool.sh
```diff
@@ -78,7 +78,8 @@ if [ -f "$ARGV_OUTPUT" ]; then
   fi
 fi
 
-wget --no-check-certificate "$ARGV_URL" -O "$TEMP_OUTPUT"
+echo "Downloading $ARGV_URL"
+curl --continue-at - --retry 100 --location --output "$TEMP_OUTPUT" "$ARGV_URL"
```
lurch: I don't think we need `--continue-at` and `--retry`?
And as we're using `-L` in `electron-download-package.sh`, perhaps we should switch to the short-option flags `-L -o` instead of `--location --output`? (either that, or update `electron-download-package.sh` to match here)

jviotti: > I don't think we need --continue-at and --retry?

My internet connection refuses to successfully download anything without those flags :P 

> And as we're using -L in electron-download-package.sh, perhaps we should switch to the short-option flags -L -o instead of --location --output? (either that, or update electron-download-package.sh to match here)

I'll normalize them!

scripts/publish/aws-s3.sh
jviotti: Sounds good. Fixed!

Makefile
```diff
@@ -373,11 +368,21 @@ endif
 
 ifdef PUBLISH_AWS_S3
 publish-aws-s3: $(PUBLISH_AWS_S3)
+ifeq ($(RELEASE_TYPE),production)
 	$(foreach publishable,$^,$(call execute-command,./scripts/publish/aws-s3.sh \
 		-f $(publishable) \
 		-b $(S3_BUCKET) \
 		-v $(APPLICATION_VERSION) \
 		-p $(PRODUCT_NAME)))
```
lurch: See how much nice the diff looks now? ;-)

scripts/build/download-tool.sh
lurch: > My internet connection refuses to successfully download anything without those flags :P

Then maybe you should allow a `$CURL_EXTRA_FLAGS` or something? I don't think they should be in the repo version of the code.

scripts/publish/aws-s3.sh
```diff
@@ -31,20 +31,23 @@ function usage() {
   echo "    -b <s3 bucket>"
   echo "    -v <version>"
   echo "    -p <product name>"
+  echo "    -k <S3 key prefix>"
```
lurch: Should we somehow indicate in these usage-strings which flags are optional, and which are compulsory?

(Traditionally in unix-cli-land, it's only the optional arguments which have flags, and all compulsory arguments are specified by position. But obviously with these scripts having so many arguments, having compulsory positional-only arguments would be a bit unwieldy!)

scripts/build/download-tool.sh
jviotti: The thing is that `wget` does handle all the problems by default. In comparison, `curl` will bail out at the first minimal error. Since `wget` did all these before, it sounds reasonable that we should do it here as well.

scripts/publish/aws-s3.sh
jviotti: I guess we can adhere to the `-k [S3 key prefix]` convention, where if the value is surrounded by `[]` (instead of `<>`), then its optional. I don't know how "standard" this is, but its what most node CLIs do.

scripts/build/download-tool.sh
lurch: Fair enough \*shrug\*. But in that case, I think that everywhere we use `curl` we ought to call it with the same options. But I wonder if a retry of 100 is a little excessive?

scripts/publish/aws-s3.sh
lurch: I think `-k [S3 key prefix]` would indicate that you can call it with `-k` or `-k some_prefix`. Using that style of notation, we'd actually want `[-k <S3 key prefix>]` but I dunno if that's readable / understandable? But maybe that'd be sufficient, given that it's probably only ever Etcher developers who will see those messages? ;-)

jviotti: I suggest to keep it as `-k [s3 key prefix]` for now, as there are other scripts using similar notation that we would have to update as well, and fall beyond the scope of this PR. I want to split the scripts to a separate repository sometime soon (since the resin desktop project needs them), so we can do a full cleanup there.

scripts/build/download-tool.sh
jviotti: This is the only place where `curl` will download "big" files (beyond text files with a few kilobytes), which is the reason I only included those options here. Do you think we should still add those options everywhere?

> But I wonder if a retry of 100 is a little excessive?

A retry of 10 was still failing very often for me. Maybe better to be on the safe side, at least for me? :P

lurch: Awww, poor @jviotti :(  Still on dial-up?
Yeah, okay then ;-)

scripts/publish/aws-s3.sh
lurch: > and fall beyond the scope of this PR

Yeah, after we'd agreed what syntax to use, I was planning to submit a separate PR updating all the other scripts to the same 'standard' :)

jviotti: Sounds great!

docs/PUBLISHING.md
```diff
@@ -93,7 +93,7 @@ access resin.io's production downloads S3 bucket.
 Run the following command to publish a specific file:
 
 ```sh
-./scripts/publish/aws-s3.sh -f <file> -b <bucket> -v <version> -t <production|snapshot>
+./scripts/publish/aws-s3.sh -f <file> -b <bucket> -v <version> -p <product name>
```
lurch: Good catch. Should this document also be updated to cover publishing snapshot builds, or are we going to leave that **only** to the CI servers?

jviotti: This document describes `make publish-aws-s3 RELEASE_TYPE=<production|snapshot>` right below, so I guess that covers the snapshot part already.

lurch: But in that case, should it also document the `-k` flag to `aws-s3.sh`?
Given the way the check-for-upgrade code works, I think it's _very_ important to ensure that snapshot builds don't accidentally end up in the 'release' S3 bucket?

jviotti: I don't think I follow. What does documenting `-k` has to do with preventing snapshot builds from getting to master?

If you run `make publish-aws-s3`, it will always do the right thing based on the value of `RELEASE_TYPE`.

Someone could still publish somewhere else by using `./scripts/publish/aws-s3.sh`, so I guess the best thing to do would be to only describe `make publish-aws-s3`. Is that what you mean?

jviotti: See https://github.com/resin-io/etcher/pull/1197

lurch: Well, I guess it's a bit of a moot point given that you're the only one with the S3 keys anyway :laughing:

jviotti: I have the code signing keys, but these S3 buckets are company wide, and many people have access to them.

PR #1080, opened 2017-02-02
docs: add msys-zip as a required Windows dependency

jviotti:
We need this dependency to create ZIP packages on Windows.


lurch:
Does this mean that the requirement for 7zip will (later) be removed?

jviotti:
Exactly! Check https://github.com/resin-io/etcher/pull/1079

lurch:
I will, when I'm less busy ;)

PR #1081, opened 2017-02-02
docs: explain the "robot" module in a high level way

jviotti:
This commit adds a README file at `lib/shared/robot` explaining in
detail how the "robot" mechanism works, and why it is needed.


lib/shared/robot/README.md
```diff
@@ -0,0 +1,89 @@
...
+    Print messages in	a machine-parsable format.  This  is  intended
+    to  ease	writing	 frontends  that  want	to  use	 xz instead of
+    liblzma, which may be the	case with various scripts.  The	output
+    with  this  option  enabled  is  meant  to  be  stable across xz
+    releases.	 See the section ROBOT MODE for	details.
+```
+
+To enable the "robot" option, we standardised the presence of an
+`ETCHER_CLI_ROBOT` environment variable. You can check if the mode is enabled
+by using the `.isEnabled()` function.
```
sam3d: Where is the method `.isEnabled()` called from? Is it the instance of the `robot` module, and if so should this be clarified here?

I know it says on line 11 about the robot module, and so from context it should be relatively easy to tell what it's referring to. I'm just wondering whether it would be better to have it here also or maybe make it read `robot.isEnabled()` function (or perhaps present a code sample).

lib/shared/robot/README.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+That object can be stringified as `{"percentage":50,"stage":"validation"}` and
+printed to `stdout`.
+
+The contents and structure of this messages is what the "robot" module is
+mainly concerned with. Each "message" consists of a type (a "command" in robot
+parlance) and and arbitrary data object.
+
+This is how a valid robot message looks like:
```
sam3d: > This is _what_ a valid robot message looks like:

lib/shared/robot/README.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+The command content and the data associated with it are application specific,
+however the robot module defines a single command called "error", which is used
+to transmit a JavaScript Error object, along with its metadata (stacktrace,
+code, description, etc) as a string.
+
+You don't have to worry about the internal details of how an Error object is
+encoded/decoded, given that the robot module exposes two high level utility
+functions: `.printError()` and `.recomposeErrorMessage()`.
```
sam3d: This example is clearer because of the below code sample and the additional "the robot module exposes two high level utility functions" which is pretty unambiguous. Maybe consider moving this sort of description further up?

lib/shared/robot/README.md
jviotti: Done! I've clarified that this is a static function of the robot module, and also provides a simple example.

lib/shared/robot/README.md
jviotti: Fixed!

lib/shared/robot/README.md
jviotti: I added an example of how to send/parse generic messages near to the top of the document. Do you think that solves this issue?

jviotti:
@sam3d Thanks a lot for the review! Having fresh eyes take a look at this is incredibly valuable :)

> Also, why are there line breaks at 80 characters? I could be wrong but I was under the impression it was a paragraph per line, feel free to ignore that one though.

Is more of a matter of convention, since its easier to edit the text this way (you don't have to deal with long lines and you can navigate the paragraph more naturally). GitHub will still render it the same way as if it didn't have text wrap.

lib/shared/robot/README.md
sam3d: I totally do, looking perfect! 😄

sam3d:
No worries @jviotti, I know they were relatively minor changes but thank you so much for being receptive!

And vis a vis the tex wrapping, that's what I thought would be the case but just wasn't sure what convention you guys used for markdown around here.

lib/shared/robot/README.md
```diff
@@ -0,0 +1,123 @@
+The "robot" mechanism
+=====================
+
+As discussed in `lib/child-writer`, communication between the main Etcher
```
jhermsmeier: I think it'd make sense to link to `lib/child-writer` here, enabling quick navigation (i.e. for people not familiar with the directory structure)

lib/shared/robot/README.md
```diff
@@ -0,0 +1,123 @@
...
+This is what a valid robot message looks like:
+
+```json
+{
+  "command": "progress",
+  "data": {
+    "percentage": 50
+  }
+}
+```
```
jhermsmeier: What do you think about describing the general data structure before showing method usage? I think that would give more context to the usage examples.

jviotti: Makes sense, done!

lib/shared/robot/README.md
jviotti: Fixed!

lib/shared/robot/README.md
```diff
@@ -0,0 +1,130 @@
...
+by using the `.isEnabled()` static function that the robot module provides:
+
+```js
+if (robot.isEnabled()) {
+  console.log('The robot option is enabled');
+}
+```
+
+The current protocol that we use is based on JSON. The writer process
+stringifies a JSON object as a string, and prints it. The client then gets the
```
stefan-mihaila: `stringifies a JSON object as a string` sounds redundant to me

lib/shared/robot/README.md
```diff
@@ -0,0 +1,130 @@
...
+line, parses it as JSON, and accesses the object.
+
+For example, the writer process may have a ficticious internal object that
```
stefan-mihaila: minor typo: ficticious -> fictitious

lib/shared/robot/README.md
```diff
@@ -0,0 +1,130 @@
...
+  stage: 'validation'
+}
+```
+
+That object can be stringified as `{"percentage":50,"stage":"validation"}` and
+printed to `stdout`.
+
+The contents and structure of this messages is what the "robot" module is
+mainly concerned with. Each "message" consists of a type (a "command" in robot
+parlance) and and arbitrary data object.
```
stefan-mihaila: `and an arbitrary`

lib/shared/robot/README.md
jhermsmeier: I'm confused – isn't it still at the same place it was before? Or am I looking at a stale diff?

jviotti: I put the structure description about 10 lines above this. Do you think we should move it below?

lib/shared/robot/README.md
jviotti: Haha, true. Fixed!

lib/shared/robot/README.md
jviotti: Fixed!

lib/shared/robot/README.md
jviotti: Fixed!

lib/shared/robot/README.md
jhermsmeier: Oh, I meant like right up top, before the first code samples even. And I think I am indeed seeing a stale diff (?!?!).

jviotti: Oh, I got it! Check it out now!

jhermsmeier: Neat, looking good! 👍

Issue #1082, opened 2017-02-04
Duplicated error messages


stefan-mihaila:
I noticed this while working on https://github.com/resin-io/etcher/issues/1027. It is caused by a `$q` [issue](https://github.com/angular/angular.js/issues/7992) that has been [solved](https://github.com/angular/angular.js/pull/13662) in `angular 1.6`. The issue is that angular's `$exceptionHandler` is called even for `$q` promises that have a rejection handler.

For example, [here we have a handler for the promise](https://github.com/resin-io/etcher/blob/master/lib/gui/pages/main/controllers/image-selection.js#L96), but [the angular exception handler](https://github.com/resin-io/etcher/blob/master/lib/gui/app.js#L195) is triggered also.

![screen recording 2017-02-03 at 11 37 pm](https://cloud.githubusercontent.com/assets/2217136/22617872/bf8cba1c-ead7-11e6-9427-6ccd61b42659.gif)

stefan-mihaila:
I tried to upgrade `angular` to `1.6.1`, `angular-moment` to `1.0.1` and `angular-ui-router` to `0.4.2`, but I get a very cryptic error message. I think it `angular 1.6` doesn't play well with some other dependency.

```javascript
Uncaught Error: TrackJS Caught: [$injector:modulerr] Failed to instantiate module Etcher due to:
Error: [$injector:modulerr] Failed to instantiate module Etcher.Pages.Main due to:
Error: [$injector:modulerr] Failed to instantiate module {} due to:
Error: [ng:areq] Argument 'module' is not a function, got Object
http://errors.angularjs.org/1.6.1/ng/areq?p0=module&p1=not%20a%20function%2C%20got%20Object
    at /Users/stefanmihaila/resin.io/etcher/node_modules/angular/angular.js:68:12
    at assertArg (/Users/stefanmihaila/resin.io/etcher/node_modules/angular/angular.js:1972:11)
    at assertArgFn (/Users/stefanmihaila/resin.io/etcher/node_modules/angular/angular.js:1982:3)
    at /Users/stefanmihaila/resin.io/etcher/node_modules/angular/angular.js:4745:11
    at forEach (/Users/stefanmihaila/resin.io/etcher/node_modules/angular/angular.js:357:20)
    at loadModules (/Users/stefanmihaila/resin.io/etcher/node_modules/angular/angular.js:4720:5)
    at /Users/stefanmihaila/resin.io/etcher/node_modules/angular/angular.js:4737:40
[...]

PR #1065, opened 2017-01-30
doc: workaround to npm shrinkwrap development dependencies bug

jviotti:
There are some npm versions that for some strange reason will attempt to
put *every* single development dependency in the `npm-shinkwrap.json`
file when editing the dependencies list.

This commit describes a workaround I've found to ensure only production
dependencies are added to the file.

I also renamed our `npm run shrinkwrap` script to `npm run
clean-shrinkwrap` since in some recent versions of npm, running `npm
shrinkwrap` will trigger our custom script instead of the original
implementation.

Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

PR #1083, opened 2017-02-04
upgrade: angular to v1.6.1; angular-ui-bootstrap to v2.5.0

stefan-mihaila:
Angular 1.5 has a bug that causes angular's `$exceptionHandler`
to be called for rejected `$q` promises even if they have
a rejection handler. This bug caused duplicate error messages
in Etcher.

A consequence of upgrading to Angular 1.6 is that `$q` promises
without a rejection handler will throw `Possibly unhandled rejection`
errors. To avoid these errors, this commit moves code responsible
for opening a tooltip from the template to the controller and handles
the rejection.

Other packages upgraded:
- angular-moment to v1.0.1
- angular-ui-router to v0.4.2
- angular-mocks to v1.6.1

Change-type: patch
Changelog-Entry: Fix duplicate error messages
Fixes: https://github.com/resin-io/etcher/issues/1082
See: https://github.com/angular/angular.js/issues/7992
See: https://github.com/angular/angular.js/pull/13662

### DIFF
--- a/lib/gui/modules/error.js
+++ b/lib/gui/modules/error.js
@@ -21,6 +21,7 @@
  */
 
 const angular = require('angular');
+const _ = require('lodash');
 
 const MODULE_NAME = 'Etcher.Modules.Error';
 const error = angular.module(MODULE_NAME, [
@@ -41,6 +42,10 @@ error.service('ErrorService', function(AnalyticsService, OSDialogService) {
    * ErrorService.reportException(new Error('Something happened'));
    */
   this.reportException = (exception) => {
+    if (_.isUndefined(exception)) {
+      return;
+    }
+
     OSDialogService.showError(exception, exception.description);
     AnalyticsService.logException(exception);
   };
--- a/lib/gui/pages/main/main.js
+++ b/lib/gui/pages/main/main.js
@@ -25,9 +25,12 @@
 const angular = require('angular');
 const MODULE_NAME = 'Etcher.Pages.Main';
 
+require('angular-moment');
+
 const MainPage = angular.module(MODULE_NAME, [
+  'angularMoment',
+
   require('angular-ui-router'),
-  require('angular-moment'),
   require('angular-middle-ellipses'),
   require('angular-seconds-to-date'),
 
--- a/lib/gui/pages/main/controllers/main.js
+++ b/lib/gui/pages/main/controllers/main.js
@@ -22,6 +22,7 @@ module.exports = function(
   FlashStateModel,
   SettingsModel,
   TooltipModalService,
+  ErrorService,
   OSOpenExternalService
 ) {
 
@@ -31,7 +32,6 @@ module.exports = function(
   this.state = FlashStateModel;
   this.settings = SettingsModel;
   this.external = OSOpenExternalService;
-  this.tooltipModal = TooltipModalService;
 
   /**
    * @summary Determine if the image step should be disabled
@@ -63,6 +63,23 @@ module.exports = function(
    */
   this.shouldFlashStepBeDisabled = () => {
     return !SelectionStateModel.hasImage() || this.shouldImageStepBeDisabled();
+  };
+
+  /**
+   * @summary Display a tooltip with the selected image details
+   * @function
+   * @public
+   *
+   * @returns {Promise}
+   *
+   * @example
+   * MainController.showSelectedImageDetails()
+   */
+  this.showSelectedImageDetails = () => {
+    return TooltipModalService.show({
+      title: 'Image File Name',
+      message: SelectionStateModel.getImagePath()
+    }).catch(ErrorService.reportException);
   };
 
 };
--- a/lib/gui/pages/main/templates/main.tpl.html
+++ b/lib/gui/pages/main/templates/main.tpl.html
@@ -62,10 +62,7 @@
         <div ng-if="main.selection.hasImage()">
           <div class="step-selection-text">
             <span
-              ng-click="main.tooltipModal.show({
-                title: 'Image File Name',
-                message: main.selection.getImagePath()
-              })"
+              ng-click="main.showSelectedImageDetails()"
               class="step-image step-name"
               ng-class="{ 'text-disabled': main.shouldImageStepBeDisabled() }"
               ng-bind="main.selection.getImageName() || main.selection.getImagePath() | basename | middleEllipses:20"
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3,9 +3,9 @@
   "version": "1.0.0-beta.18",
   "dependencies": {
     "angular": {
-      "version": "1.5.7",
-      "from": "angular@>=1.5.3 <2.0.0",
-      "resolved": "https://registry.npmjs.org/angular/-/angular-1.5.7.tgz"
+      "version": "1.6.1",
+      "from": "angular@>=1.6.1 <2.0.0",
+      "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.1.tgz"
     },
     "angular-if-state": {
       "version": "1.0.0",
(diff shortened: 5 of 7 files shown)


stefan-mihaila:
Everything seems to work / look fine, but I am somewhat concerned about the `angular-ui-bootstrap` upgrade because there's [a big list of Breaking Changes in version v2.0](https://github.com/angular-ui/bootstrap/blob/master/CHANGELOG.md#breaking-changes) that I don't really understand. `angular-ui-bootrap` `v1.3.2` seemed to work with `Angular` `v1.6` for us, but it might have some issues, as [Angular 1.6 is officially supported starting with v2.5.0](https://github.com/angular-ui/bootstrap/blob/master/CHANGELOG.md#250-2017-01-28).

lib/gui/app.js
```diff
@@ -188,13 +188,18 @@ app.config(($urlRouterProvider) => {
   $urlRouterProvider.otherwise('/main');
 });
 
-app.config(($provide) => {
+app.config(($provide, $qProvider) => {
+
+  // Prevent unhandled rejections from $uibModalInstance.dismiss() from generating errors
+  $qProvider.errorOnUnhandledRejections(false);
```
stefan-mihaila: "Possibly unhandled rejection" errors happen with both `v1.3.2` and `v2.5.0` of `angular-ui-bootstrap` after upgrading to `Angular 1.6`

jviotti:
I'll take a look at the UI Bootstrap breaking change list. Amazing that the shrinkwrap file came out correctly. Did you use the workaround I documented recently?

stefan-mihaila:
@jviotti Actually it did not come up correctly, I adjusted it by hand. Sorry, I've somehow missed the workaround, but would happily try it out and see if it works.

lib/gui/app.js
```diff
@@ -198,6 +198,11 @@ app.config(($provide) => {
   });
 });
 
+app.config(($qProvider) => {
+  // Prevent errors generated by $uibModalInstance.dismiss() unhandled rejections
+  $qProvider.errorOnUnhandledRejections(false);
```
jviotti: This sounds weird. Why shouldn't we error on unhandled rejections? If there was an unhandled rejection, we should have handled it.

If this only related to the modals, can we handle it explicitly instead of disabling it globally here?

jviotti:
I've regenerated the shrinkwrap file based on my workaround, and this is what I get:

```diff
diff --git a/npm-shrinkwrap.json b/npm-shrinkwrap.json
index 6c588af..c48e18e 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3,9 +3,9 @@
   "version": "1.0.0-beta.18",
   "dependencies": {
     "angular": {
-      "version": "1.5.7",
-      "from": "angular@>=1.5.3 <2.0.0",
-      "resolved": "https://registry.npmjs.org/angular/-/angular-1.5.7.tgz"
+      "version": "1.6.1",
+      "from": "angular@1.6.1",
+      "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.1.tgz"
     },
     "angular-if-state": {
       "version": "1.0.0",
@@ -25,9 +25,9 @@
       "resolved": "https://registry.npmjs.org/angular-middle-ellipses/-/angular-middle-ellipses-1.0.0.tgz"
     },
     "angular-moment": {
-      "version": "1.0.0-beta.6",
-      "from": "angular-moment@>=1.0.0-beta.6 <2.0.0",
-      "resolved": "https://registry.npmjs.org/angular-moment/-/angular-moment-1.0.0-beta.6.tgz"
+      "version": "1.0.0",
+      "from": "angular-moment@1.0.0",
+      "resolved": "https://registry.npmjs.org/angular-moment/-/angular-moment-1.0.0.tgz"
     },
     "angular-seconds-to-date": {
       "version": "1.0.0",
@@ -35,14 +35,14 @@
       "resolved": "https://registry.npmjs.org/angular-seconds-to-date/-/angular-seconds-to-date-1.0.0.tgz"
     },
     "angular-ui-bootstrap": {
-      "version": "1.3.3",
-      "from": "angular-ui-bootstrap@>=1.3.2 <2.0.0",
-      "resolved": "https://registry.npmjs.org/angular-ui-bootstrap/-/angular-ui-bootstrap-1.3.3.tgz"
+      "version": "2.5.0",
+      "from": "angular-ui-bootstrap@2.5.0",
+      "resolved": "https://registry.npmjs.org/angular-ui-bootstrap/-/angular-ui-bootstrap-2.5.0.tgz"
     },
     "angular-ui-router": {
-      "version": "0.2.18",
-      "from": "angular-ui-router@>=0.2.18 <0.3.0",
-      "resolved": "https://registry.npmjs.org/angular-ui-router/-/angular-ui-router-0.2.18.tgz"
+      "version": "0.4.2",
+      "from": "angular-ui-router@0.4.2",
+      "resolved": "https://registry.npmjs.org/angular-ui-router/-/angular-ui-router-0.4.2.tgz"
     },
     "ansi-escapes": {
       "version": "1.4.0",
```

Throwing it here so you can double check the manual edits went correctly :)

stefan-mihaila:
@jviotti Can you also link to the workaround please? Sorry, I've missed it and don't know how to find it :-)

lib/gui/app.js
stefan-mihaila: I think it is only related to the modals. I'll take another look tomorrow and handle it wherever it's necessary. For now I know that when closing by using the 'X' (https://github.com/resin-io/etcher/blob/master/lib/gui/components/tooltip-modal/controllers/tooltip-modal.js#L36) is an issue.

jviotti:
Sure, here it is: https://github.com/resin-io/etcher/pull/1065. Notice that the workaround doesn't apply if you already edited the shrinkwrap file. What I did to get the output above is:

- Go back to `master`
- Run `make electron-develop` to get the current dependencies
- Upgrade the dependencies by doing `npm install --save angular@1.6.1 angular-moment@1.0.0 angular-ui-bootstrap@2.5.0 angular-ui-router@0.4.2` and `npm install --save-dev angular-mocks@1.6.1`
- I then ran `npm prune --production`
- And finally `npm shrinkwrap`

lib/gui/app.js
jviotti: Cool, I'll take a look as well!

jviotti: Do you also get those transition errors when you start the application without setting this?

screenshot 2017-02-06 15 29 08

That might be hiding yet another issue.

jviotti: Regarding the transition issues, this comment (https://github.com/angular-ui/ui-router/issues/3246#issuecomment-270973117) says it might be due to an application error. Maybe worth digging further?

stefan-mihaila:
@jviotti Updated PR. The transition error was caused by the fact that I omitted upgrading `angular-if-state`'s version of `angular-ui-router` in the previous version of the PR. https://github.com/resin-io/etcher/pull/1083/files#diff-36018a6a9e1de1798516799b84841318R16

lib/gui/modules/error.js
```diff
@@ -41,6 +41,10 @@ error.service('ErrorService', function(AnalyticsService, OSDialogService) {
    * ErrorService.reportException(new Error('Something happened'));
    */
   this.reportException = (exception) => {
+    if (angular.isUndefined(exception)) {
```
jviotti: This is really minor, but I guess we should use Lodash where possible instead of Angular functions given that we will switch from it in the near future.

lib/gui/pages/main/main.js
```diff
@@ -25,9 +25,11 @@
 const angular = require('angular');
 const MODULE_NAME = 'Etcher.Pages.Main';
 
+require('angular-moment');
+
 const MainPage = angular.module(MODULE_NAME, [
   require('angular-ui-router'),
-  require('angular-moment'),
+  'angularMoment',
```
jviotti: Maybe this is a great opportunity to start a convention where angular modules that don't return their own module string are put at the top of the dependency array?

For example:

```js
require('angular-moment');

const MainPage = angular.module(MODULE_NAME, [
    'angularMoment',

    require('...'),
    require('...'),
    require('...'),
    require('...')
]);
```

jviotti:
Maybe we can include an explanation on how the upgrade fixes the duplicated error messages, and why we needed to move the tooltip open code to the controller in the commit body and the PR description?

jviotti:
I confirm everything works fine for me! :+1:

stefan-mihaila:
updated

PR #1084, opened 2017-02-07
style(GUI): minor fixes in comments

stefan-mihaila:


lurch:
This doesn't modify the docs, it modifies the code comments. IMHO they're very different things ;-)

lurch:
Dunno, @jviotti is much more familiar with the commit conventions than I am.

jviotti:
There are some commit types that can be ambiguous. @lurch comment makes sense: this modifies the annotations, not the docs, however one can also argue that annotations are also documentation.

Maybe we can decide on one and document it on the commit guidelines? I don't really mind any of these.

My proposal is to go with "style". Thoughts?

stefan-mihaila:
I'm good with `style`. Updated.

jviotti:
Cool, I sent a PR updating the commit guidelines to clarify that.

lurch:
@jviotti How come the `boolen` typo wasn't picked up by the spell-checker thing you added recently?

jviotti:
@lurch I investigated this and turns out code spell works with dictionaries that declare pairs of malformed word and correct word, for example:

```
abandonned->abandoned
abbout->about
aberation->aberration
abilties->abilities
abilty->ability
abondon->abandon
abondoned->abandoned
abondoning->abandoning
abondons->abandons
aborigene->aborigine
abortificant->abortifacient
abotu->about
abouta->about a
aboutit->about it
aboutthe->about the
abreviate->abbreviate
abreviated->abbreviated
```

This obviously means that not every possible bad written combination is there, and therefore a lot of stuff gets ignored (code spell only catches usual typos errors).

I guess the thing to do here is to add every typo we find to our own custom dictionary file that is used in conjunction with the new one, in order to prevent regressions, though that still means we have to check typos ourselves :(

lurch:
> turns out code spell works with dictionaries that declare pairs of malformed word and correct word

How bizarre :-/

PR #1087, opened 2017-02-08
chore: start codespell dictionary file in the root of the project

jviotti:
codespell works by having a dictionary of pairs of malformed words and
their correct equivalents. This means that codespell will not catch most
issues out of the box, but we can train it based on the common spelling
issues we encounter along the way.

The plan is to add a pair to `dictionary` every time we encounter a typo
during a code review, and we ensure that error will not happen again
anymore.

See: https://github.com/resin-io/etcher/pull/1084

### DIFF
--- a/dictionary
+++ b/dictionary
@@ -1,0 +1,1 @@
+boolen->boolean
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "jslint": "eslint lib tests scripts bin versionist.conf.js",
     "scsslint": "scss-lint lib/gui/scss",
     "htmllint": "node scripts/html-lint.js",
-    "codespell": "codespell.py lib tests",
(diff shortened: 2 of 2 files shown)


lurch:
Should we periodically attempt upstreaming these dictionary improvements?

jviotti:
Sounds good. I guess first lets try how useful this approach is, or if we should go looking for other more automatic tools.

lurch:
From just having had a look at https://github.com/lucasdemarchi/codespell/blob/master/codespell_lib/_codespell.py#L205 it seems that `--dictionary` by itself actually replaces the usage of `codespell`'s built-in dictionary. It looks like we might need to instead change the command we're using to:
```
codespell.py --dictionary=- --dictionary=dictionary lib tests
```

jviotti:
Amazing catch! Thanks a lot @lurch , I'll update the PR in a minute.

PR #1086, opened 2017-02-08
refactor(gui): revert drive/image selection step order swap

Shou:
We revert the drive and image selection step order swap back from drive
selection first and image second, to image first and drive second.

See: https://github.com/resin-io/etcher/pull/1049
Changelog-Entry: Revert the drive/image selection step order swap.

PR #1089, opened 2017-02-09
feat(gui): consolidate image metadata into a single popover

Shou:
We combine the old image information modal and tooltip into a single
popover that appears on hover and disappears on outside click.

Changelog-Entry: Consolidate the image metadata modal and popup into a
single popover.
Depends: #1086

### DIFF
--- a/lib/gui/pages/main/styles/_main.scss
+++ b/lib/gui/pages/main/styles/_main.scss
@@ -106,3 +106,7 @@
 .page-main .step-size {
   color: $palette-theme-dark-disabled-foreground;
 }
+
+.page-main .popover-content {
+  -webkit-user-select: auto;
+}
--- a/lib/gui/pages/main/styles/_main.scss
+++ b/lib/gui/pages/main/styles/_main.scss
@@ -26,7 +26,7 @@
   color: $palette-theme-dark-foreground;
 }
 
-.page-main .text-disabled {
+.page-main .text-disabled > span {
   color: $palette-theme-dark-disabled-foreground;
 }
 
--- a/lib/gui/pages/main/templates/main.tpl.html
+++ b/lib/gui/pages/main/templates/main.tpl.html
@@ -1,6 +1,45 @@
 <div class="page-main row around-xs">
+  <div class="col-xs" ng-controller="ImageSelectionController as image">
+    <div class="box text-center relative" os-dropzone="image.selectImage($file)">
+
+      <svg-icon
+        class="center-block"
+        path="{{ main.selection.getImageLogo() || '../../../assets/image.svg' }}"></svg-icon>
+
+      <div class="space-vertical-large">
+        <div ng-hide="main.selection.hasImage()">
+          <button
+            class="button button-primary button-brick"
+            ng-click="image.openImageSelector()">Select image</button>
+
+          <p class="step-footer">
+            {{ ::image.mainSupportedExtensions.join(', ') }}, and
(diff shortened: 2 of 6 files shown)


Shou:
image

lib/gui/scss/modules/_bootstrap.scss
```diff
@@ -36,7 +36,16 @@ body {
   cursor: default;
 }
 
-.tooltip {
+.tooltip,
+.popover {
   word-wrap: break-word;
 }
 
+.page-main .popover-title {
```
jviotti: We shouldn't add page specific styles in the `_bootstrap.scss` file, which is meant to hold Bootstrap specific styles.

By seeing this, I think the `.page-main` prefixes are unnecessary (the styles apply to the popovers independently of the page), so just dropping `.page-main` is probably enough.

lib/gui/pages/main/styles/_main.scss
```diff
@@ -106,3 +106,7 @@
 .page-main .step-size {
   color: $palette-theme-dark-disabled-foreground;
 }
+
+.page-main .popover-content {
+  -webkit-user-select: auto;
```
jviotti: Can you explain why this was needed? Also, it seems that this should be moved to `_bootstrap.scss` (without `.page-main`) so it gets applied to all popovers on the app.

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -19,12 +19,13 @@
           </p>
         </div>
         <div ng-if="main.selection.hasImage()">
-          <div class="step-selection-text">
+          <div class="step-selection-text"
+            uib-popover="{{ main.selection.getImagePath() }}"
+            popover-title="{{ main.selection.getImagePath() | basename }} - {{ main.selection.getImageSize() | gigabyte | number:1 }}GB"
+            popover-trigger="{'mouseenter': 'none', 'outsideClick': 'outsideClick'}">
```
jviotti: Since we might want all popovers across the app to behave the same way, I wonder if we can set the triggers globally using the `$uibTooltipProvider` provider (https://angular-ui.github.io/bootstrap/#!#popover). They seem to offer a `setTriggers(obj)` that we could use in `lib/gui/app.js` for example.

jviotti:
Pinging @konmouz to provide design input.

lib/gui/pages/main/styles/_main.scss
Shou: I think if the user wants to see the full path of the image, they may want to copy that path wholly or partially too, so allowing selections here. I assumed it was the case for the modal to match its features, but I just now checked that it isn't, so this isn't necessary but could be nice.

lib/gui/pages/main/templates/main.tpl.html
Shou: That's a good call

Shou: It seems like `$uibTooltipProvider.setTriggers` only affects tooltips and not popovers, which is odd.

jviotti: Hm, really?

> Through the $uibTooltipProvider, you can change the way **tooltips and popovers** behave by default

That'd definitely be weird. I can think of two things:

- Check if they define a `$uibPopoverProvider` and the simply made a typo on the docs
- Check if we're running the latest version, and try upgrading if not

PR #1102, opened 2017-02-14
chore: update `angular-ui-router` in `npm-shrinkwrap.json`

jviotti:
Looks like this dependency was not updated correctly on the shrinkwrap
file. Probably happened in the linked pull request, where the shrinkwrap
file was hand edited at some point (see PR discussion).

See: https://github.com/resin-io/etcher/pull/1083

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -13,9 +13,9 @@
       "resolved": "https://registry.npmjs.org/angular-if-state/-/angular-if-state-1.0.0.tgz",
       "dependencies": {
         "angular-ui-router": {
-          "version": "0.4.2",
-          "from": "angular-ui-router@>=0.4.2 <0.5.0",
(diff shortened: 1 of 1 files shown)


lurch:
Huh? #1083 seemed to upgrade `angular-ui-router` to `0.4.2`, but this PR seems to downgrade it back to `0.3.2` ??
(I don't quite understand how the shrinkwrap stuff works, so I may well be misunderstanding this, in which case an explanation would be great :) )

jviotti:
@lurch This refers to the `angular-ui-router` which is a dependency of `angular-if-state`, but not the top level one we use here (I know its a bit confusing :P), so it should be fine.

lurch:
Ahhh, another case of NPM dependency hell ;-)
I'd missed the "dependency rather than top-level" part, thanks for clarifying.

PR #1107, opened 2017-02-14
docs: add developer documentation to the lib/image-stream module

jviotti:

### DIFF
--- a/lib/image-stream/README.md
+++ b/lib/image-stream/README.md
@@ -1,0 +1,61 @@
+Etcher Image Stream
+===================
+
+This module is in charge of creating a readable stream from any image source
+(e.g: a file, a URL, etc) along with some metadata (like size), and handling
+any necessary transformations (like decompression) that must be applied before
+plugging the stream to [`etcher-image-write`][etcher-image-write].
+
+Given that this module contains the logic to handle image formats, the module
+becomes the most reliable source of truth for the list of supported ones.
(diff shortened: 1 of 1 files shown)


lib/image-stream/README.md
```diff
@@ -0,0 +1,57 @@
...
+data, and useful metadata like the final image size.
+
+Archive Hooks
+-------------
+
+This module supports reading "archive images", which are defined by handlers
+(like `application/zip`). In order to avoid duplication on how to handle
+archives, archive support is implemented by "archive hooks".
+
+Archive hooks are CommonJS modules that expose two functions:
```
jhermsmeier: Could you also add how, and on what the archive hook modules would be instantiated here (i.e. are they called on the stream, or file descriptor, ...)? Or is that too much information, which should rather be somewhere else?

jviotti: Sounds like a good idea! I added an explanation for the handlers, and added proper function signatures to clarify the return type and arguments for the archive hook.

Notice that this module is currently coupled with the concept of a file name, which will be changed soon once we support streaming.

I have some designs for this in mind that I'll put on the relevant GH issue.

jviotti:
@jhermsmeier Do you think this is good to merge now?

jhermsmeier:
@jviotti yup, sorry for the delay!

PR #1110, opened 2017-02-16
docs: Windows 8.1 SDK is needed to build Etcher

jviotti:
Otherwise, `lzma-native` fails.


lurch:
Is it specifically always the Windows 8.1 SDK that's required, or is it the SDK corresponding to the version of Windows that you're currently building on?

jviotti:
It seems to be the Windows 8.1 SDK (I had to install it on Windows 10, for example).

jviotti:
I'm closing this. Even with the SDK it doesn't run. Looks like the Visual Studio installer now de-selects a whole Windows 8.1 section by default, which is needed to run Etcher. I'll create a new PR describing the necessary item users need to select.

Issue #1100, opened 2017-02-14
Improve analytics events


jviotti:
For debugging and UX design purposes (e.g: to make effective design decisions based on real user behaviour). These are the events we have right now:

**For our users:** you can always disable error reporting and analytics from the settings page (its optional), but we should probably make that more obvious (something to discuss in another issue)

- Application start
- Checking for updates
- Notifying update
- Select image
- Re-select image
- Select Drive
- Reselect drive
- Flash
- Restart
- Restart after failure
- Done
- Validation Error
- Out of Space
- Flash Error
- Invalid Image

These are the events we want to have:

### Global

- [x] Application Start
- [x] Application attempted to close while flashing
- [x] Application attempted to close while not flashing
- [x] Show exit while flashing warning
- [x] Cancel flashing warning
- [x] Agree flashing warning

### Updates

- [x] Checking for Updates (include current version)
- [x] Notifying Update (include the latest version)
- [x] Update accepted
- [x] Update dismissed
- [x] Update dismissed and put to sleep
- [x] No Update

### Image

- [x] Open image selector
- [x] Image selector dismissed
- [x] Select image (include image data)
- [x] Change image
- [x] Show image info modal
- [x] Show image info tooltip
- [x] Multiple images inside archive

### Drive

- [ ] No drives (only after you select the image)
- [x] Auto-select drive (include drive)
- [ ] Auto-deselect drive (include drive and image)
- [x] Select drive (include drive, and if using unsafe mode)
- [x] Open drive modal
- [x] Drive modal dismissed
- [x] Toggle drive selection (include drive)

### Header

- [ ] Click help icon
- [ ] Click on settings

### Footer

- [x] Click on Etcher logo
- [x] Click on “an open source project by”
- [x] Click on Resin.io logo
- [x] Click version

### Settings

- [x] Click on back button
- [x] Toggle setting (include setting name and new value)
- [x] Accept setting warning (include setting name)
- [x] Dismiss setting warning (include setting name)

### Flash

- [ ] Click flash (include drive and image, is validation enabled, is unmounting enabled)
- [ ] Show elevation dialog
- [ ] No elevation needed
- [ ] Elevation successful
- [ ] Elevation rejected
- [ ] Flashing process started
- [ ] Flashing half through
- [ ] Flashing is about to complete
- [ ] Flashing error
- [ ] Validation started
- [ ] Validation half through
- [ ] Validation is about to complete
- [ ] Validation failed
- [ ] Out of Space error
[...]

PR #1111, opened 2017-02-17
feat(GUI): improve analytics events

stefan-mihaila:
This PR adds more events to our current analytics.
Will further improve in a future PR.

See: https://github.com/resin-io/etcher/issues/1100


lib/gui/app.js
```diff
@@ -85,22 +86,41 @@ app.run(() => {
...
       // process (e.g: selected an image), otherwise such interruption
       // might be annoying.
-      if (!isLatestVersion && !SelectionStateModel.hasImage()) {
-
-        AnalyticsService.logEvent('Notifying update');
+      if (!SelectionStateModel.hasImage()) {
         return UpdateNotifierService.notify();
       }
+
+      AnalyticsService.logEvent('Not notifying update because image already selected');
```
jviotti: I think we should try to shorten this event name.

lib/gui/app.js
```diff
@@ -85,22 +86,41 @@ app.run(() => {
...
+
       // In case the internet connection is not good and checking the
       // latest published version takes too long, only show notify
       // the user about the new version if he didn't start the flash
       // process (e.g: selected an image), otherwise such interruption
       // might be annoying.
-      if (!isLatestVersion && !SelectionStateModel.hasImage()) {
-
-        AnalyticsService.logEvent('Notifying update');
+      if (!SelectionStateModel.hasImage()) {
```
jviotti: Can we remove the negation from this condition, putting the event below here, and the actual `.notify()` call below?

lib/gui/app.js
```diff
@@ -85,22 +86,41 @@ app.run(() => {
...
         return UpdateNotifierService.notify();
       }
+
+      AnalyticsService.logEvent('Not notifying update because image already selected');
+
     }).catch(ErrorService.reportException);
+
+  } else {
```
jviotti: Let's avoid this `else` by checking the case where we shouldn't notify instead.

```js
if (!shouldCheckForUpdates || process.env.ETCHER_DISABLE_UPDATES) { 
   // Not checking for updates
   return;
}

// Checking for updates
```

lib/gui/app.js
```diff
@@ -164,25 +187,38 @@ app.run(($window, WarningModalService, ErrorService, FlashStateModel, OSDialogSe
...
+    AnalyticsService.logEvent('Close attempt while flashing');
+
     return OSDialogService.showWarning({
       confirmationLabel: 'Yes, quit',
       rejectionLabel: 'Cancel',
       title: 'Are you sure you want to close Etcher?',
       description: messages.warning.exitWhileFlashing()
     }).then((confirmed) => {
       if (confirmed) {
+        AnalyticsService.logEvent('Close confirmed while flashing');
```
jviotti: Did you double check if this event is triggered correctly? We're killing the process right afterwards and it might be that this event has no change to be emitted.

lib/gui/app.js
```diff
@@ -164,25 +187,38 @@ app.run(($window, WarningModalService, ErrorService, FlashStateModel, OSDialogSe
...
       }
 
+      AnalyticsService.logEvent('Close rejected while flashing');
       popupExists = false;
     }).catch(ErrorService.reportException);
   });
 });
 
+app.run(($rootScope, AnalyticsService) => {
+  $rootScope.$on('$stateChangeSuccess', (evt, toState, toParams, fromState) => {
```
jviotti: Very nice one! Minor comment: can we rename `evt` to `event` for consistency reasons?

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -25,7 +25,9 @@ module.exports = function(
   DrivesModel,
   SelectionStateModel,
   WarningModalService,
-  DriveConstraintsModel) {
+  DriveConstraintsModel,
+  AnalyticsService
```
jviotti: You have to add this to the module declaration as well, at `lib/gui/components/drive-selector/drive-selector.js`

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -117,12 +120,19 @@ module.exports = function(
    * @function
    * @public
    *
+   * @param {String} method - how the modal is closed, used for logging
```
jviotti: This looks weird, I think we should avoid thinking how the modal was closed here. We should send a generic "drive selector closed" event, while emitting all the other low level events from the modal service instead (in a generic way so that it applies to all modals).

lib/gui/components/modal/services/modal.js
```diff
@@ -18,7 +18,7 @@
 
 const _ = require('lodash');
 
-module.exports = function($uibModal, $q) {
+module.exports = function($uibModal, $q, AnalyticsService) {
```
jviotti: Remember to add this to `lib/gui/components/modal/modal.js` as well.

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,29 @@ module.exports = function($uibModal, $q) {
...
+                method: error
+              });
               resolve();
 
             // For some annoying reason, UI Bootstrap Modal rejects
             // the result reason if the user clicks on the backdrop
             // (e.g: the area surrounding the modal).
-            } else if (error !== 'backdrop click') {
+            } else {
+              AnalyticsService.logEvent('Modal closed', {
```
jviotti: I don't think this corresponds to a "Click on X", but to any JavaScript error occurred on the modal. I might be wrong but I think a click on X causes the modal to go straight to `.then()`, while resolving it with an undefined value.

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -118,6 +126,9 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
    */
   this.notify = () => {
     return this.getLatestVersion().then((version) => {
+      AnalyticsService.logEvent('Notifying update', {
```
jviotti: All the update related events are handled at `app.js`. It feels weird that this one is here. Let's either move all update events to this module, or move this to `app.js`.

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,29 @@ module.exports = function($uibModal, $q) {
       close: modal.close,
       result: $q((resolve, reject) => {
         modal.result
-          .then(resolve)
+          .then((value) => {
+            AnalyticsService.logEvent('Modal closed', {
+              method: 'User accepted',
+              value: value
+            });
```
jviotti: Can we add a blank line after this?

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,29 @@ module.exports = function($uibModal, $q) {
...
+            resolve(value);
+          })
           .catch((error) => {
-
             // Bootstrap doesn't 'resolve' these but cancels the dialog;
             // therefore call 'resolve' here applied to 'false'.
             if (error === 'escape key press' || error === 'backdrop click') {
+              AnalyticsService.logEvent('Modal closed', {
+                method: error
+              });
```
jviotti: Can we add a blank line after this?

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,29 @@ module.exports = function($uibModal, $q) {
...
               resolve();
 
             // For some annoying reason, UI Bootstrap Modal rejects
             // the result reason if the user clicks on the backdrop
             // (e.g: the area surrounding the modal).
-            } else if (error !== 'backdrop click') {
+            } else {
+              AnalyticsService.logEvent('Modal closed', {
+                method: 'Click on X'
+              });
```
jviotti: Can we add a blank line after this?

lib/gui/components/update-notifier/controllers/update-notifier.js
```diff
@@ -48,6 +48,9 @@ module.exports = function($uibModalInstance, SettingsModel, options) {
    * UpdateNotifierController.closeModal();
    */
   this.closeModal = () => {
+    AnalyticsService.logEvent('Close update modal', {
+      sleepUpdateCheck: this.sleepUpdateCheck
+    });
```
jviotti: Can we add a blank line after this?

lib/gui/models/selection-state.js
```diff
@@ -78,10 +79,22 @@ SelectionStateModel.service('SelectionStateModel', function(DrivesModel) {
...
+
     Store.dispatch({
       type: Store.Actions.SELECT_IMAGE,
       data: image
     });
+
+    AnalyticsService.logEvent('Select image', image);
+
+    if (this.getDrive()) {
+      AnalyticsService.logEvent('Auto-selected drive', this.getDrive());
```
jviotti: This module shouldn't have the knowledge of how drives are auto-selected. Also, notice that a drive might get auto-selected even if no image is set, meaning that this will not catch that.

These events should live in the Redux store, which is where this logic is implemented. The Redux store is not AngularJS, so I guess let's leave this out from the PR and address it once we make the analytics module framework agnostic.

lib/gui/modules/analytics.js
```diff
@@ -53,11 +70,18 @@ analytics.config(($mixpanelProvider) => {
 
     /* eslint-enable camelcase */
 
-    electron: app.getVersion(),
+    electron: process.versions.electron,
     node: process.version,
     arch: process.arch,
-    version: packageJSON.version
+    version: packageJSON.version,
+    OSPlatform: os.platform(),
```
jviotti: Why do some keys start with uppercase letter, while all the other ones with a lower case letter?

lib/gui/modules/analytics.js
```diff
@@ -39,6 +40,22 @@ const analytics = angular.module(MODULE_NAME, [
...
+ * We need this because node's os.arch() returns the process architecture
+ * See: https://github.com/nodejs/node-v0.x-archive/issues/2862
+ *
+ * @return {String} OS architecture
+ */
+const getOSArch = () => {
+  if (os.arch() !== 'ia32') {
+    return os.arch();
+  }
+  return arch() === 'x86' ? 'ia32' : 'x64';
```
jviotti: I don't think we should use this module, given it doesn't seem to handle arm correctly (based on the docs), which is a target we will be releasing for GNU/Linux soon.

lib/gui/modules/analytics.js
```diff
@@ -39,6 +40,22 @@ const analytics = angular.module(MODULE_NAME, [
...
+ * @summary Get OS architecture
+ *
+ * @description
+ * We need this because node's os.arch() returns the process architecture
+ * See: https://github.com/nodejs/node-v0.x-archive/issues/2862
+ *
+ * @return {String} OS architecture
+ */
+const getOSArch = () => {
+  if (os.arch() !== 'ia32') {
```
jviotti: So if the NodeJS architecture is x64, we assume that's the host architecture and simply return it? What if its not the case?

lib/gui/modules/analytics.js
```diff
@@ -39,6 +40,22 @@ const analytics = angular.module(MODULE_NAME, [
...
+/**
+ * @summary Get OS architecture
+ *
+ * @description
+ * We need this because node's os.arch() returns the process architecture
+ * See: https://github.com/nodejs/node-v0.x-archive/issues/2862
+ *
+ * @return {String} OS architecture
+ */
+const getOSArch = () => {
```
jviotti: Hm, I didn't know detecting the host architecture would be so tricky. Based on my other comments on this same function, what about something like this?

```js
const getHostArchitecture = () => {
    return _.includes([ 'ia32', 'x64' ], process.arch) ? arch() : process.arch;
};
```

So if the node architecture is not x86 nor x64, we don't try to disambiguate and return that (e.g: in case of `arm`), otherwise, we attempt to find the right one using the `arch` module, which only works for those architectures.

I think its the best we can do :/

jviotti: Also, let's rename `getOSArch` to `getHostArchitecture`.

lib/gui/modules/analytics.js
```diff
@@ -39,6 +40,22 @@ const analytics = angular.module(MODULE_NAME, [
   require('../models/settings')
 ]);
 
+/**
+ * @summary Get OS architecture
```
jviotti: We should add `@function` and `@private` here, plus provide an `@example`.

jviotti: Maybe the summary should make it clear that we're talking about the host OS architecture?

lib/gui/modules/analytics.js
```diff
@@ -53,11 +70,18 @@ analytics.config(($mixpanelProvider) => {
...
     /* eslint-enable camelcase */
 
-    electron: app.getVersion(),
+    electron: process.versions.electron,
     node: process.version,
     arch: process.arch,
-    version: packageJSON.version
+    version: packageJSON.version,
+    OSPlatform: os.platform(),
+    OSArch: getOSArch(),
```
jviotti: Let's use the word "host" here as well.

lib/gui/os/open-external/services/open-external.js
```diff
@@ -31,6 +31,9 @@ module.exports = function() {
    * OSOpenExternalService.open('https://www.google.com');
    */
   this.open = (url) => {
+    AnalyticsService.logEvent('Open external link', {
+      url: url
```
jviotti: We can simply put `url` here.

lib/gui/os/open-external/services/open-external.js
```diff
@@ -31,6 +31,9 @@ module.exports = function() {
...
+    });
```
jviotti: Can we put a blank line below this?

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -105,8 +107,10 @@ module.exports = function(SupportedFormatsModel, SelectionStateModel, AnalyticsS
    * ImageSelectionController.reselectImage();
    */
   this.reselectImage = () => {
+    AnalyticsService.logEvent('Reselect image', {
+      previousImage: SelectionStateModel.getImage()
+    });
```
jviotti: Can we add a blank line below this?

lib/gui/pages/main/controllers/main.js
```diff
@@ -76,6 +77,9 @@ module.exports = function(
    * MainController.showSelectedImageDetails()
    */
   this.showSelectedImageDetails = () => {
+    AnalyticsService.logEvent('Show selected image tooltip', {
+      imagePath: SelectionStateModel.getImagePath()
+    });
```
jviotti: Can we add a blank line below this?

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -86,4 +90,12 @@ module.exports = function(WarningModalService, SettingsModel) {
     });
   };
 
+  this.set = (key, value) => {
+    AnalyticsService.logEvent('Setting change', {
+      setting: key,
+      value
+    });
```
jviotti: Can we add a blank line below this?

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -86,4 +90,12 @@ module.exports = function(WarningModalService, SettingsModel) {
     });
   };
 
+  this.set = (key, value) => {
+    AnalyticsService.logEvent('Setting change', {
```
jviotti: Maybe "Set setting" instead?

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -86,4 +90,12 @@ module.exports = function(WarningModalService, SettingsModel) {
     });
   };
 
+  this.set = (key, value) => {
```
jviotti: This function should include a JSDoc annotation, like its `.get()` counterpart.

lib/gui/app.js
stefan-mihaila: I test it and it seems to make it to mixpanel

lib/gui/components/modal/services/modal.js
stefan-mihaila: I double checked now, and seems it corresponds to a click on `x` in the case of the `UpdateNotifier` and the `TooltipModal`, because for these two we call `$uibModalInstance.dismiss()` when clicking on `x`. The `WarningModalService` on `x` resolves the modal with `false`, while the `DriveSelectorService` resolved the modal with `undefined`.

lib/gui/modules/analytics.js
stefan-mihaila: It's just that `OS` is upper case, the same way `JSON` in `packageJSON` is. Can change it to lower case if you prefer.

jviotti: Let's use lowercase, given that this is the beginning of the variable name, e.g: `osPlatform`.

lib/gui/modules/analytics.js
```diff
@@ -39,25 +40,40 @@ const analytics = angular.module(MODULE_NAME, [
...
+/**
+ * @summary Get host architecture
+ * @function
+ * @private
+ *
+ * @description
+ * We need this because node's os.arch() returns the process architecture
+ * See: https://github.com/nodejs/node-v0.x-archive/issues/2862
+ *
+ * @return {String} OS architecture
```
jviotti: This should be `returns`.

lib/gui/modules/analytics.js
```diff
@@ -39,25 +40,40 @@ const analytics = angular.module(MODULE_NAME, [
...
+ */
+const getHostArchitecture = () => {
```
jviotti: Can we add an `@example`?

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,32 @@ module.exports = function($uibModal, $q) {
...
-          .catch((error) => {
+          .then((value) => {
+            AnalyticsService.logEvent('Modal closed', {
+              method: 'User accepted',
+              value: value
+            });
 
+            resolve(value);
+          })
+          .catch((error) => {
```
jviotti: Can we add a blank line below this?

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,32 @@ module.exports = function($uibModal, $q) {
...
             // Bootstrap doesn't 'resolve' these but cancels the dialog;
             // therefore call 'resolve' here applied to 'false'.
             if (error === 'escape key press' || error === 'backdrop click') {
+              AnalyticsService.logEvent('Modal closed', {
+                method: error
+              });
+
               resolve();
 
             // For some annoying reason, UI Bootstrap Modal rejects
```
jviotti: After the edit to the `else if`, this comment no longer makes sense here, but should be moved to the top `if`.

lib/gui/components/update-notifier/update-notifier.js
```diff
@@ -26,7 +26,8 @@ const UpdateNotifier = angular.module(MODULE_NAME, [
   require('../modal/modal'),
   require('../../models/settings'),
   require('../../utils/manifest-bind/manifest-bind'),
-  require('../../os/open-external/open-external')
+  require('../../os/open-external/open-external'),
+  require('../../modules/analytics')
```
jviotti: I think this is not needed anymore.

lib/gui/pages/main/controllers/flash.js
```diff
@@ -83,7 +84,7 @@ module.exports = function(
       } else {
         FlashErrorModalService.show(messages.error.genericFlashError());
         ErrorService.reportException(error);
-        AnalyticsService.logEvent('Flash error');
+        AnalyticsService.logEvent('Flash error', error);
```
jviotti: Can we pass `error` as a property instead? E.g:

```js
AnalyticsService.logEvent('Flash error', {
    error: error
});
```

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -18,7 +18,7 @@
 
 const os = require('os');
 
-module.exports = function(WarningModalService, SettingsModel) {
+module.exports = function(WarningModalService, SettingsModel, AnalyticsService) {
```
jviotti: I think you have to include this in the settings module.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -105,6 +107,7 @@ module.exports = function(
    * });
    */
   this.toggleDrive = (drive) => {
+    AnalyticsService.logEvent('Toggle drive', drive);
```
jviotti: Can we pass `drive` as a property? For example:

```js
AnalyticsService.logEvent('Toggle drive', {
    drive: drive
});
```

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -123,6 +126,8 @@ module.exports = function(
   this.closeModal = () => {
     const selectedDrive = SelectionStateModel.getDrive();
 
+    AnalyticsService.logEvent('Drive selector closed', selectedDrive);
```
jviotti: Let's pass this as a property as well.

lib/gui/os/open-external/services/open-external.js
```diff
@@ -31,6 +31,8 @@ module.exports = function() {
    * OSOpenExternalService.open('https://www.google.com');
    */
   this.open = (url) => {
+    AnalyticsService.logEvent('Open external link', url);
```
jviotti: Can we pass this as a property?

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,32 @@ module.exports = function($uibModal, $q) {
       close: modal.close,
       result: $q((resolve, reject) => {
         modal.result
-          .then(resolve)
-          .catch((error) => {
+          .then((value) => {
```
jviotti: Can we put the `.then()` on the same line as the promise, to avoid the indentation to the whole section below? For example:

```js
modal.result.then((value) => {
  AnalyticsService.logEvent('Modal closed', {
    method: 'User accepted',
    value: value
  });

  resolve(value);
}).catch((error) => {
  // Bootstrap doesn't 'resolve' these but cancels the dialog;
  // therefore call 'resolve' here applied to 'false'.
  if (error === 'escape key press' || error === 'backdrop click') {
    AnalyticsService.logEvent('Modal closed', {
      method: error
    });

    resolve();

  // For some annoying reason, UI Bootstrap Modal rejects
  // the result reason if the user clicks on the backdrop
  // (e.g: the area surrounding the modal).
  } else {
    AnalyticsService.logEvent('Modal closed', {
      method: 'rejected'
    });

    return reject(error);
  }

});
```

lib/gui/app.js
```diff
@@ -164,25 +189,38 @@ app.run(($window, WarningModalService, ErrorService, FlashStateModel, OSDialogSe
...
 
+      AnalyticsService.logEvent('Close rejected while flashing');
       popupExists = false;
     }).catch(ErrorService.reportException);
   });
 });
 
+app.run(($rootScope, AnalyticsService) => {
+  $rootScope.$on('$stateChangeSuccess', (event, toState, toParams, fromState) => {
+    AnalyticsService.logEvent('Navigate', {
```
jviotti: I get the following when the application starts:

```
Navigate ({"to":"main","from":""})
```

Maybe worth checking that `from` is defined, and if not, ignore the event?

lib/gui/components/modal/services/modal.js
```diff
@@ -56,18 +56,32 @@ module.exports = function($uibModal, $q) {
...
+            AnalyticsService.logEvent('Modal closed', {
```
jviotti: Maybe better to say: `Modal accepted` and omit `method`?

jviotti:
I found an issue: if I enable unsafe mode I get an event saying that such setting is now true, but then if I untick it, I get no event.

lib/gui/components/modal/services/modal.js
jviotti: This event feels weird. This is what I get if I enable a dangerous setting:

```
tracker.js:10 Mon Feb 20 2017 16:17:53 GMT-0400 (BOT) Dangerous settings modal ({"setting":"unsafeMode"})
tracker.js:10 Mon Feb 20 2017 16:17:54 GMT-0400 (BOT) Modal closed ({"method":"User accepted","value":true})
```

Let's mimic what we do for the other settings:

```
Set setting ({"setting":"errorReporting","value":true})
```

Maybe simply replacing `Set setting` with `Set dangerous setting`.

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -78,6 +78,10 @@ module.exports = function(WarningModalService, SettingsModel) {
     // Keep the checkbox unchecked until the user confirms
     this.currentData[name] = false;
 
+    AnalyticsService.logEvent('Dangerous settings modal', {
```
jviotti: Can we include the verb "open" here?

lib/gui/components/modal/services/modal.js
jviotti: I guess we can keep the "Modal closed" but simply add a new "Set dangerous setting" one and that should do it!

jviotti:
We use the words "closed" and "dismissed" in different parts of the app. Can we unify both?

jviotti:
If I re-select an image, no image selection event is triggered. For example:

- Select an image
- Then click change
- Find a new image
- Accept the dialog

No event is triggered.

lib/gui/components/drive-selector/controllers/drive-selector.js
jviotti: We should include whether the selection is true or false.

lib/gui/pages/main/controllers/flash.js
```diff
@@ -59,7 +59,8 @@ module.exports = function(
 
     AnalyticsService.logEvent('Flash', {
       image: image,
-      device: drive.device
+      device: drive.device,
```
jviotti: Can we include the whole drive and the whole image here?

stefan-mihaila:
@jviotti Updated. Made a few changes, let me know if it still doesn't feel right.

lib/gui/app.js
```diff
@@ -134,7 +156,7 @@ app.run(($timeout, DriveScannerService, DrivesModel, ErrorService, DriveSelector
   DriveScannerService.on('drives', (drives) => {
 
     // Safely trigger a digest cycle.
-    // In some cases, AngularJS doesn't aknowledge that the
+    // In some cases, AngularJS doesn't acknowledge that the
```
lurch: Is this another new typo to add to the codespell dictionary? See #1087

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -89,7 +89,7 @@
           percentage="main.state.getFlashState().percentage"
           striped="{{ main.state.getFlashState().type == 'check' }}"
           ng-attr-active="{{ main.state.isFlashing() }}"
-          ng-click="flash.flashImageToDrive(main.selection.getImagePath(), main.selection.getDrive())"
+          ng-click="flash.flashImageToDrive(main.selection.getImage(), main.selection.getDrive())"
```
lurch: I wonder if the specific changes related to this change should be refactored into their own mini-PR, and then rebase this PR once that mini-PR has been merged?
(Or perhaps I'm just being overly pedantic :-/  )

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -71,6 +71,12 @@ module.exports = function(WarningModalService, SettingsModel) {
    */
   this.enableDangerousSetting = (name, options) => {
     if (!this.currentData[name]) {
+
+      AnalyticsService.logEvent('Set setting', {
+        setting: name,
+        value: false
+      });
```
lurch: For symmetry with the other `'Set setting'` should this include `dangerous: false`? Or isn't that needed with the analytics service we're using?

lurch: (or ignore me if I'm just misunderstanding things)

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -80,10 +86,37 @@ module.exports = function(WarningModalService, SettingsModel) {
...
+   * The logging is done here for now and not in the model to avoid
+   * a circular dependency
+   *
+   * @param {String} key - setting key
+   * @param {Object} value - setting value
+   */
+  this.set = (key, value) => {
+    AnalyticsService.logEvent('Set setting', {
+      setting: key,
+      value
```
lurch: Again, apologies if I'm miles off the mark, but to match up with the other similar changes here should this be `value: value` ?

stefan-mihaila: You're not at all miles off.

In ES6
```javascript
const x = 42;
const o = { x }
```
is just short-hand for
```javascript
const x = 42;
const o = { x: x }
```

Now that I think about it, any guide-styles for short-hand properties, @jviotti? I think at least I should group them at the beginning, as in https://github.com/airbnb/javascript#objects--grouped-shorthand

lib/gui/pages/settings/controllers/settings.js
stefan-mihaila: I think it's a good idea

lib/gui/pages/main/templates/main.tpl.html
stefan-mihaila: I don't really have something against it. Or I can put them in a different commit. But I want to see if @jviotti agrees with the approach of giving the whole `image` to the `flashImageToDrive`.

lib/gui/pages/settings/controllers/settings.js
lurch: Ahhhh, I didn't know that! Now I do, thanks :-)

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -74,7 +74,8 @@ module.exports = function(WarningModalService, SettingsModel, AnalyticsService)
 
       AnalyticsService.logEvent('Set setting', {
         setting: name,
-        value: false
+        value: false,
+        dangerous: false
       });
 
       this.model.set(name, false);
```
lurch: Hmmm, I'm now wondering if the 'prototype' for `this.set` should be updated to `(key, value, dangerous=false)` and then all three of these `'Set setting'` logEvents could be handled just once from within `this.set` ?

stefan-mihaila: @jviotti Any thoughts on that?

lib/gui/pages/main/templates/main.tpl.html
jviotti: I guess its fine given that the reason for passing the whole image instead of just the path only makes sense because of analytics purposes.

lib/gui/pages/settings/controllers/settings.js
jviotti: Sounds like a good idea. Let's give it a go!

lib/gui/modules/analytics.js
```diff
@@ -120,17 +142,22 @@ analytics.service('AnalyticsService', function($log, $window, $mixpanel, Setting
    */
   this.logEvent = (message, data) => {
 
-    if (SettingsModel.get('errorReporting') && isRunningInAsar()) {
+    // Mixpanel expects a flat object
+    const flattenedData = data ? flatten(data, {
```
jviotti: We should extract this logic (and the start casing one) into a separate function and write unit tests for it.

lib/gui/app.js
```diff
@@ -85,23 +86,44 @@ app.run(() => {
...
 
-  if (UpdateNotifierService.shouldCheckForUpdates() && !process.env.ETCHER_DISABLE_UPDATES) {
-    AnalyticsService.logEvent('Checking for updates');
+  const shouldCheckForUpdates = UpdateNotifierService.shouldCheckForUpdates();
 
-    UpdateNotifierService.isLatestVersion().then((isLatestVersion) => {
+  if (!shouldCheckForUpdates || process.env.ETCHER_DISABLE_UPDATES) {
+    AnalyticsService.logEvent('Not checking for updates', {
+      shouldCheckForUpdates,
+      ETCHER_DISABLE_UPDATES: process.env.ETCHER_DISABLE_UPDATES
```
jviotti: The mixpanel context transformation logic is causing the underscores to get removed:

```
Wed Feb 22 2017 16:51:49 GMT-0400 (BOT) Not checking for updates ({"Should Check For Updates":true,"ETCHER DISABLE UPDATES":"true"})
```

We should either:

- Make the transformation keep things like environment variable names intact, or
- Describe this property in a way that is not tied to the actual environment variable name, like "updatesDisabled"

I think I prefer the second option.

lib/gui/modules/analytics.js
```diff
@@ -120,17 +142,22 @@ analytics.service('AnalyticsService', function($log, $window, $mixpanel, Setting
...
-      // Clone data before passing it to `mixpanel.track`
-      // since this function mutates the object adding
-      // some custom private Mixpanel properties.
-      $mixpanel.track(message, _.clone(data));
+    const startCaseFlattenedData = flattenedData ? _.mapKeys(flattenedData, (value, key) => {
+      return _.startCase(key);
+    }) : undefined;
 
+    if (SettingsModel.get('errorReporting') && isRunningInAsar()) {
+      $mixpanel.track(message, _.mapKeys(flattenedData, _.startCase));
```
jviotti: Why are we duplicating `_.mapKeys` here? Maybe I'm missing something, but I think you can simply pass `startCaseFlattenedData` here.

lib/gui/pages/main/controllers/flash.js
```diff
@@ -57,12 +60,15 @@ module.exports = function(
...
     AnalyticsService.logEvent('Flash', {
-      image: image,
-      device: drive.device
+      image,
+      drive,
+      unmount: SettingsModel.get('unmountOnSuccess')
     });
 
-    ImageWriterService.flash(image, drive).then(() => {
+    ImageWriterService.flash(imagePath, drive).then(() => {
```
jviotti: Can we do `image.path` here, since its the only place where the `imagePath` variable is used?

jviotti:
Looking great! I left a few small comments, and I'll do a more in-depth review once the final comments are addressed, the Mixpanel context transformer unit tested, and the settings `.set()` function refactored!

lib/gui/app.js
lurch: I guess if we want to know _why_ "updatesDisabled", then perhaps it should be renamed to something like "disableUpdatesEnvvar" ?

jviotti: Yeah, sounds good to me. Let's write "disableUpdatesEnvironmentVariable" though, I really hate abbreviations in code :P

stefan-mihaila:
@jviotti Take another look when you have a chance

lib/gui/app.js
jviotti: Yeah, let's create another PR adding this to the dictionary @stefan-mihaila

stefan-mihaila: Created https://github.com/resin-io/etcher/pull/1145. I also fixed the typo in that PR, will remove it from this one.

lib/gui/components/update-notifier/update-notifier.js
stefan-mihaila: `AnalyticsService` is still used in `UpdateNotifierController`.

tests/shared/utils.spec.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+'use strict';
+
+const m = require('mochainon');
+const utils = require('../../lib/shared/utils');
+
+describe('Shared: Utils', function() {
+
+  describe('.makeFlatStartCaseObject(object)', function() {
```
jviotti: There are many things to test here. Some from the top of my head:

- Environment variable keys

For example: `ETCHER_DISABLE_UPDATES` should remain the same.

- Environment variables inside objects

For example:

```json
{
   "foo": {
      "FOO_BAR_BAZ": 3
   }
}
```

Should be transformed to `"Foo FOO_BAR_BAZ": 3`.

- Top level keys
- Keys that start with a number
- Keys that are already in start case (e.g: should not be modified)
- Keys containing spaces
- Keys whose values are arrays
- Arrays containing objects

There are also a set of classic values you can test for:

- An empty object
- Something that is not an object, like a string, or an array
- `null`

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -53,34 +53,45 @@ module.exports = function(WarningModalService, SettingsModel) {
...
    * });
    */
-  this.enableDangerousSetting = (name, options) => {
-    if (!this.currentData[name]) {
-      this.model.set(name, false);
-      return this.refreshSettings();
+  this.toggle = (setting, warningOptions) => {
+
+    const value = this.currentData[setting];
+    const dangerous = Boolean(warningOptions);
```
jviotti: Determining if a setting is dangerous or not based on if the options cast to boolean doesn't look nice.

Maybe let's introduce a new boolean option called `requireConfirmation`, or `showConfirmationModal` or something like that, and only read the description and label if that option is set.

What do you think?

lib/shared/utils.js
```diff
@@ -0,0 +1,50 @@
...
+'use strict';
+
+const _ = require('lodash');
+const flatten = require('flat').flatten;
+
+/**
+ * @summary Create a flattened copy of the object with all keys transformed in start case
+ *
+ * @param {Object} object - object to transform
+ * @return {Object}
```
lurch: @jviotti Perhaps the http://eslint.org/docs/rules/valid-jsdoc section of https://github.com/resin-io/etcher/blob/master/.eslintrc.yml needs tweaking? ;-)

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -53,72 +53,48 @@ module.exports = function(WarningModalService, SettingsModel, AnalyticsService)
...
-      AnalyticsService.logEvent('Set setting', {
-        setting: name,
-        value: false,
-        dangerous: false
-      });
+    AnalyticsService.logEvent('Toggle setting', {
+      setting,
+      value,
+      dangerous
+    });
```
lurch: With `logEvent` coming before `WarningModalService.display` I guess we can't know whether the user _actually_ selected to enable the dangerous option or not. I dunno if that's desired behaviour or not?

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -53,72 +53,48 @@ module.exports = function(WarningModalService, SettingsModel, AnalyticsService)
...
+   * @summary Toggle setting
    * @function
    * @public
    *
-   * @param {String} name - setting name
-   * @param {Object} options - options
-   * @param {String} options.description - modal description
-   * @param {String} options.confirmationLabel - modal confirmation label
+   * @description
+   * If warningOption is given, it should be an object having `description` and `confirmationLabel`;
```
lurch: warningOption -> warningOptions

lib/shared/utils.js
jviotti: Good idea. See https://github.com/resin-io/etcher/pull/1147. I didn't enable the rule to require a return description though, since it requires them also for functions that return promises that resolve to `undefined`, which looks weird. I'll see if I can find something else for that.

lib/gui/pages/settings/controllers/settings.js
lurch: I think it's okay, because for any non-dangerous settings, `warningOptions` won't be supplied. Perhaps it would be better if the code was more explicit, like `const dangerous = (warningOptions == undefined);` (or whatever the correct JS / nodeJS syntax is ;) )

npm-shrinkwrap.json
```diff
@@ -482,6 +492,11 @@
       "from": "is-arrayish@>=0.2.1 <0.3.0",
       "resolved": "https://registry.npmjs.org/is-arrayish/-/is-arrayish-0.2.1.tgz"
     },
+    "is-buffer": {
+      "version": "1.1.4",
+      "from": "is-buffer@>=1.1.4 <1.2.0",
+      "resolved": "https://registry.npmjs.org/is-buffer/-/is-buffer-1.1.4.tgz"
+    },
```
lurch: Any idea what caused this change?

jviotti: @lurch The `flat` package seems to depend on it: https://github.com/hughsk/flat/blob/master/package.json#L17

lurch: Ahhh. I thought npm-shrinkwrap showed dependencies-of-other-dependencies as indented under the parent dependency? Or am I just mis-remembering / confused?

jviotti: NPM tries to flatten the dependency structure whenever it can. It will only fallback to nested dependencies if there is a conflict it can't resolve.

lurch: Ahhh! Thanks for plugging another gap in my knowledge :-)

lib/gui/pages/settings/controllers/settings.js
stefan-mihaila: We log in `services/modal.js` if the modal was accepted or rejected

lurch: Ahh. Would needing to check for two events (which setting was being changed, and whether the warning was accepted or rejected) make the analytics harder to analyse though?

stefan-mihaila: @lurch It is a very good question. It probably does. I guess we don't have very precise requirements at the moment, and we'll probably improve the events later if they don't fit our use cases

lib/gui/pages/settings/controllers/settings.js
stefan-mihaila: @jviotti Not sure what I think about it, I agree that it's not ideal, but I think adding another parameter that controls whether a third one should be given or not is not ideal either. No strong opinions on this one from my side.
@lurch I went for the `lodash` `const dangerous = !_.isUndefined(warningOptions);`

stefan-mihaila:
@jviotti updated

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -53,34 +54,45 @@ module.exports = function(WarningModalService, SettingsModel) {
...
-   * @param {String} name - setting name
-   * @param {Object} options - options
-   * @param {String} options.description - modal description
-   * @param {String} options.confirmationLabel - modal confirmation label
+   * @description
+   * If warningOptions is given, it should be an object having `description` and `confirmationLabel`;
+   * these will be used to present a user confirmation modal before enabling the setting.
+   *
+   * @param {String} setting - setting key
+   * @param {Object} warningOptions - (optional) WarningModalService options
```
jviotti: You can define a optional parameter as follows:

```
@param {Object} [warningOptions] - WarningModalService options
```

jviotti: Maybe worth elaborating on the options instead of simply referring to `WarningModalService`? 

For example:

```
@param {Object} [options] - options
@param {String} [options.description] - warning modal description
@param {String} [options.confirmationLabel] - warning modal confirmation label
```

lib/shared/utils.js
```diff
@@ -0,0 +1,82 @@
...
+
+  // Transform primitives to objects
+  if (!_.isObject(object)) {
+    return {
+      Value: object
+    };
+  }
+
+  if (_.isArray(object)) {
+    return object.map((x) => {
```
jviotti: Can use Lodash's `_.map` for consistency purposes?

lib/shared/utils.js
```diff
@@ -0,0 +1,82 @@
...
+
+const _ = require('lodash');
+const flatten = require('flat').flatten;
+const deepMapKeys = require('deep-map-keys');
+
+/**
+ * @summary Create a flattened copy of the object with all keys transformed in start case
+ *
+ * @param {Object} object - object to transform
+ * @param {Object} objectFromPrimitive - whether should transform primitives into objects
```
jviotti: I don't see this being declared in the argument list. Is it still valid?

lib/shared/utils.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+'use strict';
+
+const _ = require('lodash');
+const flatten = require('flat').flatten;
+const deepMapKeys = require('deep-map-keys');
+
+/**
+ * @summary Create a flattened copy of the object with all keys transformed in start case
```
jviotti: Let's add `@function` and `@public` here.

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -53,34 +54,45 @@ module.exports = function(WarningModalService, SettingsModel) {
...
+    if (!value || !dangerous) {
+      return this.model.set(setting, value);
     }
 
     // Keep the checkbox unchecked until the user confirms
-    this.currentData[name] = false;
+    this.currentData[setting] = false;
 
-    WarningModalService.display(options).then((userAccepted) => {
+    WarningModalService.display(warningOptions).then((userAccepted) => {
```
jviotti: I just realised we should be adding a `.catch(ErrorService.reportException)` to this promise, and add the error module to the dependencies.

stefan-mihaila:
updated

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -53,72 +54,50 @@ module.exports = function(WarningModalService, SettingsModel, AnalyticsService)
...
    * @function
    * @public
    *
-   * @param {String} name - setting name
-   * @param {Object} options - options
-   * @param {String} options.description - modal description
-   * @param {String} options.confirmationLabel - modal confirmation label
+   * @description
+   * If warningOptions is given, it should be an object having `description` and `confirmationLabel`;
+   * these will be used to present a user confirmation modal before enabling the setting.
```
lurch: Maybe this could also mention that no confirmation modal is shown when disabling a 'dangerous' setting?

jviotti:
@stefan-mihaila Oops, conflicts! :)

stefan-mihaila:
@jviotti rebased ;-)

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -153,6 +162,11 @@ module.exports = function(
     return shouldChangeDriveSelectionState(drive).then((canChangeDriveSelectionState) => {
       if (canChangeDriveSelectionState) {
         SelectionStateModel.setDrive(drive.device);
+
+        AnalyticsService.logEvent('Drive selected', {
+          method: 'double click'
```
lurch: Only a minor point, but it looks a bit weird specifying `method: 'double click'` here when AFAICT there aren't any other 'Drive selected' events?

lib/gui/components/modal/services/modal.js
```diff
@@ -55,23 +59,33 @@ module.exports = function($uibModal, $q) {
...
             // Bootstrap doesn't 'resolve' these but cancels the dialog;
-            // therefore call 'resolve' here applied to 'false'.
-            if (error === 'escape key press' || error === 'backdrop click') {
-              resolve();
+            // therefore call 'resolve' here.
+            resolve();
+
+          } else {
+            AnalyticsService.logEvent('Modal rejected', {
+              value: error
```
lurch: ?? Just a few lines previously you used `method: error`, but here you're using `value: error`?
Should they be made consistent, or could you add a code-comment explaining why they're not consistent? ;)  (assuming I'm not missing anything obvious)

lib/gui/pages/main/controllers/flash.js
```diff
@@ -36,11 +36,14 @@ module.exports = function(
    * @function
    * @public
    *
-   * @param {String} image - image path
+   * @param {String} image - image
```
lurch: Is this now an `{Object}` rather than `{String}` ?

lib/gui/app.js
```diff
@@ -83,23 +84,44 @@ app.run(() => {
...
-      // latest published version takes too long, only show notify
-      // the user about the new version if he didn't start the flash
-      // process (e.g: selected an image), otherwise such interruption
-      // might be annoying.
-      if (!isLatestVersion && !SelectionStateModel.hasImage()) {
+  AnalyticsService.logEvent('Checking for updates', {
+    currentVersion: packageJSON.version
+  });
 
-        AnalyticsService.logEvent('Notifying update');
```
stefan-mihaila: I initially moved it in `UpdateNotifierService.notify()`, then removed it from there so all updates related logging is in `app.js`, but never added it back... It would get logged as an open modal (`Open modal ({"Template":"./components/update-notifier/templates/update-notifier-modal.tpl.html"})`), but an explicit log might be useful as well. I'll add it back.

lib/gui/components/modal/services/modal.js
stefan-mihaila: Here it's the value with which the modal was rejected (`reject(error)`). On the `method: error`, the modal is rejected with the value `undefined`, but I thought it would be useful to know that the rejection is caused by the user hitting escape or clicking outside the modal. I can add an explicit `value: undefined` there too.

On a different note, I do feel that the events I came up are very unstructured. I didn't really had any better ideas on how to structure them, if you want to think about them a little and help out with better ways of making them more consistent, I'm open to ideas @lurch & @jviotti

lib/gui/components/drive-selector/controllers/drive-selector.js
stefan-mihaila: Changed it to `AnalyticsService.logEvent('Drive selected (double click)')`

lib/gui/components/modal/services/modal.js
lurch: I suppose the best way to answer that question is to ask what kind of things we want to analyse after all the statistics have been gathered? (and therefore whether we're gathering the "right" statistics).

But that's something I have no idea about.

lurch: Hmmm, perhaps it'd be worth copy'n'pasting the list from https://github.com/resin-io/etcher/issues/1100 onto a new page on the [wiki](https://github.com/resin-io/etcher/wiki) and then for each event that we log, make a note of what string we use to log the event, and any extra data that gets logged along with the event? I think being able to see them all at-a-glance in one place might be really useful?

PR #1114, opened 2017-02-19
docs: document supported operating systems and versions

jviotti:

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,17 @@ was written correctly and much more.
 
 ![Etcher](https://raw.githubusercontent.com/resin-io/etcher/master/screenshot.png)
 
+Supported Operating Systems
+---------------------------
+
+- Linux (most distros)
+- macOS 10.9 and later
+- Microsoft Windows 7 and later
+
+Note that Etcher will run on any platform officially supported by
+[Electron][electron]. Read more in their
+[documentation][electron-supported-platforms].
+
 Installers
 ----------
 
@@ -70,6 +81,8 @@ Etcher is free software, and may be redistributed under the terms specified in
 the [license].
 
 [etcher]: https://etcher.io
+[electron]: http://electron.atom.io
+[electron-supported-platforms]: http://electron.atom.io/docs/tutorial/supported-platforms/
 [SUPPORT]: https://github.com/resin-io/etcher/blob/master/SUPPORT.md
 [CONTRIBUTING]: https://github.com/resin-io/etcher/blob/master/docs/CONTRIBUTING.md
 [CLI]: https://github.com/resin-io/etcher/blob/master/docs/CLI.md


README.md
```diff
@@ -20,6 +20,13 @@ was written correctly and much more.
 
 ![Etcher](https://raw.githubusercontent.com/resin-io/etcher/master/screenshot.png)
 
+Supported Operating Systems
+---------------------------
+
+- GNU/Linux
```
jhermsmeier: Maybe we should list some common distros here, like `GNU/Linux (Ubuntu, Fedora, Arch, ...)` or something, so that people won't think their Linux distro doesn't cut it?

README.md
```diff
@@ -20,6 +20,13 @@ was written correctly and much more.
...
+- macOS 10.9 and later
```
jhermsmeier: Curiosity question: What dependency do we have that only allows for macOS 10.9+? Electron itself is 10.8+, right?

README.md
jviotti: Etcher should run on virtually all modern distributions. How do you we can clarify that? I don't want to list distributions because it might give the impression that we only support the ones that we list.

README.md
jviotti: Electron only supports 10.9+, see https://github.com/electron/electron/blob/master/docs/tutorial/supported-platforms.md#macos

lurch:
Perhaps it'd be worth explaining that Etcher does/should run on all platforms on which Electron runs, and then using http://electron.atom.io/docs/tutorial/supported-platforms/ as a "read more" link?

README.md
lurch: I realise that "GNU/Linux" is the official nomenclature, but in common parlance just "Linux" is much more common, and is probably more familiar to the people that we're aiming Etcher at?

(the typical Ubuntu user might know that Ubuntu is a version of Linux, but they might think that "GNU/Linux" is some other version of Linux?)

README.md
jhermsmeier: Oh, thanks for clarifying, I guess the should update [their website](http://electron.atom.io/docs/development/build-instructions-osx/) then

lurch: @jhermsmeier Or you could PR a fix? "See something that needs fixing? Propose a change on the source."

jhermsmeier: @lurch yeah, might look into that

README.md
jhermsmeier: I concur, just "Linux" seems to be the least confusing.

jviotti: Done! I also added "most distros" in parenthesis.

README.md
```diff
@@ -20,6 +20,17 @@ was written correctly and much more.
...
 ![Etcher](https://raw.githubusercontent.com/resin-io/etcher/master/screenshot.png)
 
+Supported Operating Systems
+---------------------------
+
+- Linux (most distros)
+- macOS 10.9 and later
+- Microsoft Windows 7 and later
+
+Notice that Etcher will run on any platform officially supported by
```
lurch: Notice -> Note

jviotti: Fixed!

PR #1118, opened 2017-02-20
feat(GUI): update image step icon

Shou:
We replace the old image step with the more general 'plus' icon.

Changelog-Entry: Update the old image step icon with 'plus' icon.


Shou:
image

konmouz:
@Shou very nice! Don't forget the flash icon as well (svg below).

[flash icon.svg.zip](https://github.com/resin-io/etcher/files/787545/flash.icon.svg.zip)

PR #1119, opened 2017-02-20
feat(gui): new flashing step icon

Shou:
We update the flashing step's icon to the new lightning strike icon.

Changelog-Type: Update flashing step's icon to a lightning strike.


Shou:
@jviotti @konmouz here's the flashing icon update.

image

PR #1121, opened 2017-02-21
chore: ensure `npm run sass` was ran and the results committed

jviotti:
We recently encountered a UI regression caused by forgetting to commit
the generated CSS files.

The solution is to make the CI servers run `npm run sass` and check if
there are unstaged files afterwards. If so, the tests halt.

In order to avoid duplication between Travis CI and Appveyor CI, this
logic has been encoded into a bash script at `scripts/ci`.

See: https://github.com/resin-io/etcher/pull/1120

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,10 @@ install:
 script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       npm test;
+      ./scripts/ci/ensure-staged-sass.sh;
     fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-      ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "xvfb-run --server-args=$XVFB_ARGS npm test";
+      ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "xvfb-run --server-args=$XVFB_ARGS npm test && ./scripts/ci/ensure-staged-sass.sh";
     fi
 
 notifications:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,6 +43,7 @@ build: off
 test_script:
   - node --version
   - npm --version
+  - bash .\scripts\ci\ensure-staged-sass.sh
   - cmd: npm test
 
 notifications:
--- a/scripts/ci/ensure-staged-sass.sh
+++ b/scripts/ci/ensure-staged-sass.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -e
+set -u
+
+npm run sass
+
+# From http://stackoverflow.com/a/9393642/1641422
+if [[ -n $(git status -s) ]]; then
+  echo "There are unstaged sass changes. Please commit the result of:" 1>&2
+  echo ""
+  echo "    npm run sass" 1>&2
+  echo ""
+  exit 1
+fi


lurch:
...or as I suggested elsewhere, we could advise all maintainers to set up a pre-commit git hook to do this automatically? ;-)
But yeah, the approach used here is obviously much more robust.

P.S. Travis has just failed with:
```
Error: Missing binding /home/travis/build/resin-io/etcher/node_modules/node-sass/vendor/linux-x64-46/binding.node

Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 4.x
```

jviotti:
> ...or as I suggested elsewhere, we could advise all maintainers to set up a pre-commit git hook to do this automatically? ;-)
But yeah, the approach used here is obviously much more robust

I think we should provide a mechanism to install git hooks for developers that would like to adopt it (count me in!), but as you mentioned, we should still provide a robust barrier for people not using them.

I'll look into the CI errors :)

jviotti:
Ah, so the dependencies are installed in the Docker container and not in the outside world, so the sass binary is not found.

jviotti:
screenshot 2017-02-21 11 44 54

Working fine on Appveyor CI (I'm forcing the error for testing purposes, and will rebase afterwards)

jviotti:
screenshot 2017-02-21 11 46 41

Working for Linux!

lurch:
> Working for Linux!

Hmmmm https://travis-ci.org/resin-io/etcher/jobs/203841922

WasabiFan:
Why do you commit the CSS at all? For all the disadvantages of having committed build artefacts, I can't think of any significant benefits.

lurch:
I think I asked something similar once, and IIRC the answer was to make https://github.com/resin-io/etcher/blob/master/docs/RUNNING-LOCALLY.md easier?  (or perhaps I'm misremembering)
And I think @jviotti said somewhere he likes verifying the actual CSS changes?

jviotti:
Yeah, I had a couple of "incidents" with transpilers in the past that made me very cautious about them. I double check the output since then.

WasabiFan:
If the issue is that you want to check the output, how about this: you stop checking in the compiled code, and instead rig the CI builds to print the diff of the output? That makes it easy to check the diff yet avoids all the pains of forcing manual recompilation. I can't say how precisely that CI configuration works, though.

jviotti:
The issue of that approach is the context switch. I'll have to check the PR diff AND the CI server output. I guess its nicer to have it all together.

jviotti:
I'm puzzled about why the Linux build fails. It seems that it considers the shell script to be outside of the conditional even though its inside of it :/

screenshot 2017-02-21 15 48 41

jviotti:
I forced an update to make Travis CI start again. Maybe we're hitting some Travis weirdness?

jviotti:
BTW, I'm thinking we could use a similar approach to ensure the shrinkwrap file is generated correctly...

jviotti:
This was a Travis issue indeed. I triggered a new build and now it worked fine.

jviotti:
I've rebased, so now all the tests should pass.

lurch:
> BTW, I'm thinking we could use a similar approach to ensure the shrinkwrap file is generated correctly...

Yeah, as long as it's reliable and we don't get it flagging up a load of false positives...?

lurch:
> so now all the tests should pass.

...if Travis ever wakes up :-(

jviotti:
> Yeah, as long as it's reliable and we don't get it flagging up a load of false positives...?

I guess we can experiment with it and see how it goes.

> ...if Travis ever wakes up :-(

Looks like all the builds finished successfully, just that GitHub didn't realize for some reason. I'll force yet another build...

lurch:
The Linux x86 build on Travis had failed with a timeout problem, so I restarted it and now all the CI tests are passing :+1:

PR #1122, opened 2017-02-21
fix(GUI): update notifier transformRequest error at startup

jviotti:
Etcher checks what is the latest version by making an HTTP request to S3
in order to see if it should present the update notifier modal at
startup.

If the user is not connected to the internet, or is on an unstable
connection, the request might fail completely or timeout, leading to a
weird error message being presented to the user:

```json
{
  "data": null,
  "status": -1,
  "config": {
    "method": "GET",
    "transformRequest": [null],
    "transformResponse": [null],
    "url": "https://resin-production-downloads.s3.amazonaws.com",
    "headers": {
      "Accept": "application/json, text/plain, */*"
    }
  },
  "statusText": ""
}
```

To solve the error, we tweak `UpdateNotifierService.getLatestVersion()`
to resolve the current version if there was a problem making the request
(e.g: the response didn't come through at all), which after some
experimentation, we can check by comparing the `status` property with
`-1`.

This error happens in all operating systems, and currently is trending
on TrackJS.

Change-Type: patch
Changelog-Entry: Fix `transformRequest` error at startup when not connected to the internet, or when on an unstable connection.

### DIFF
--- a/lib/gui/components/update-notifier/services/update-notifier.js
+++ b/lib/gui/components/update-notifier/services/update-notifier.js
@@ -23,6 +23,14 @@ const etcherLatestVersion = require('etcher-latest-version');
 module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, ManifestBindService, SettingsModel) {
 
   /**
+   * @summary The current application version
+   * @constant
+   * @private
+   * @type {String}
+   */
+  const CURRENT_VERSION = ManifestBindService.get('version');
+
+  /**
    * @summary Get the latest available Etcher version
    * @function
    * @private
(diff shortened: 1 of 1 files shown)


lurch:
Does this also cover the case of Etcher being run on a computer with no network connection at all? (yes kids, such things do exist!) Or is that already covered by some other error-flow?

jviotti:
> Does this also cover the case of Etcher being run on a computer with no network connection at all? (yes kids, such things do exist!) Or is that already covered by some other error-flow?

Amazing catch. I never tested Etcher without internet connection, and that also triggers the `transformRequest` bug. I'll update the PR message to reflect this as well.

lurch:
So what happens when TrackJS isn't able to log that Etcher doesn't have an internet connection...? :rofl:

jviotti:
I confirm the TrackJS module reacts fine to those cases!

PR #1133, opened 2017-02-24
chore: don't use symbolic links on Windows builds

jviotti:
Cygwin can't resolve them correctly, causing cryptic errors during
build.

### DIFF
--- a/scripts/build/dependencies-bower.sh
+++ b/scripts/build/dependencies-bower.sh
@@ -49,7 +49,7 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
 fi
 
 if [ -n "$ARGV_PREFIX" ]; then
-  ln -s "$PWD/bower.json" "$ARGV_PREFIX/bower.json"
+  cp "$PWD/bower.json" "$ARGV_PREFIX/bower.json"
   pushd "$ARGV_PREFIX"
   bower install $INSTALL_OPTS
   popd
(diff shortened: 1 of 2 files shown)


lurch:
LOL, effectively reverting #935 ;-)
(why is Windows always such a pain?)

I guess this means Cygwin isn't using https://msdn.microsoft.com/en-us/library/windows/desktop/aa363878(v=vs.85).aspx ?

jhermsmeier:
Windows users need explicit permissions set to be able to make symlinks (see http://superuser.com/a/105381), which is probably why this is failing.
The really annoying thing about this permission is, that even if given to a user who's also an Administrator, symlinking will require elevated privileges.

jviotti:
Haha, weird Windows stuff. At least copying will always work.

jviotti:
@jhermsmeier MinGW works fine though, even without permissions.

lurch:
I guess that must mean MinGW is using some other "symlink" mechanism, so maybe that's why Cygwin is getting confused? \*shrug\*
@jviotti Does it work if you always use just-cygwin or always use just-mingw, and only get confused if you try mixing the two?

jviotti:
@lurch Symlinks created by Cygwin don't work at all (even if I just use Cygwin).

lurch:
Any difference if you try running Cygwin with elevated privileges?

jviotti:
Not really, looks like Cygwin symbolic links are simply plain files containing:

```
!<symlink> <path to real file>
```

I found http://stackoverflow.com/questions/18654162/enable-native-ntfs-symbolic-links-for-cygwin/18659632, which explains how to configure Cygwin to use native NTFS symbolic links, but I still think we should merge this, in order to avoid having to explain this mess.

lurch:
> but I still think we should merge this, in order to avoid having to explain this mess.

Agreed. \*sigh\*

PR #1134, opened 2017-02-24
chore: add missing `:` after options in publish scripts 

jviotti:
Not including the colon means that the option parsing mechanism is not
expecting an argument for the `-t` and the `-p` option (for
`bintray-debian.sh` and `aws-s3.sh` respectively), and thus `$OPTARG`
will be undefined.

### DIFF
--- a/scripts/publish/aws-s3.sh
+++ b/scripts/publish/aws-s3.sh
@@ -39,7 +39,7 @@ ARGV_BUCKET=""
 ARGV_VERSION=""
 ARGV_PRODUCT_NAME=""
 
-while getopts ":f:b:v:p" option; do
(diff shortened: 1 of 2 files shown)


lurch:
I guess this means that https://github.com/resin-io/etcher/blob/master/scripts/publish/aws-s3.sh#L42 needs updating too?

jviotti:
Good catch! I'll send a new PR for it :)

lurch:
Oh, rather than rolling it into this PR? (the original mistake in both scripts was made in the same PR #968 )

jviotti:
I'm fine with both approaches. I updated this PR and will close the other one :)

PR #1135, opened 2017-02-24
chore: fix various code-signing Windows issues

jviotti:
- Add missing backslash that caused the command to not be interpreted
  completely

- Update `electron-sign-exe.exe` to `electron-sign-exe-win32.exe` in
  Makefile

- Use hyphen options instead of Windows slash style, which confuses bash


jviotti:
I remember catching these same issues the previous time I released, but looks like I forgot to PR the fixes last time

PR #1140, opened 2017-02-27
chore: add missing `:` after `p` option in aws-s3.sh

jviotti:
Not including the colon means that the option parsing mechanism is not
expecting an argument for the `-p` option, and thus `$OPTARG` will be
undefined.

See: https://github.com/resin-io/etcher/pull/1134


jviotti:
The changes introduced by this PR were added to https://github.com/resin-io/etcher/pull/1134

PR #1141, opened 2017-02-27
chore: define `S3_BUCKET` in Makefile

jviotti:
We use that variable in the Makefile but we're not defining it, which
means that maintainers publishing Etcher need to remember to pass that
as an option.


lurch:
Was there a decision made yet on whether to use a different bucket for snapshot builds?

jviotti:
Not yet, I'm afraid.

Issue #884, opened 2016-11-20
Make meaning clear of "connect a drive" text when no drives are found


WasabiFan:
- **Etcher version:** Beta 15, 16
- **Operating system and architecture:** Windows 10 64-bit
- **Do you see any meaningful error information on DevTools?** N/A

When you don't have a drive connected, Etcher looks like this:

![image](https://cloud.githubusercontent.com/assets/3310349/20459891/0215bb82-ae87-11e6-83e5-dc2b019c396f.png)

Once you connect one, that changes to this:

![image](https://cloud.githubusercontent.com/assets/3310349/20459896/3fd7b9f2-ae87-11e6-9cdc-bd956a11d1d8.png)

Not only is the first one unclear -- I can't tell what I am actually supposed to do there -- but it barely changes once you have things connected. I think that the "no drives connected" state should be changed to make the lack of drives clear and to make the rectangle look less like a button.

WasabiFan:
Interesting... it looks like this was a regression since beta 7, when the rectangle was red while no drives were available. I don't think that the old behavior was optimal, but it was better than it is now.

jviotti:
Yeah, we should definitely get this fixed for once and for all. @taahirisaacs , @konmouz mind giving us a way forward here?

lurch:
It's also been niggling me slightly that the `Change` buttons (for both Image and Drive) look nothing like buttons; but I dunno if that's something that needs "fixing" or not...

konmouz:
Please check [here](https://github.com/resin-io/etcher/issues/888) regarding 'connect a drive' issue.

konmouz:
@Shou @jviotti, we still have the issue with 'connect a drive' on the latest version. This looks like a CTA while it is just a message. Ideally, this message appears within the drives' modal window (screenprint from Zeplin). If this is tricky, 'connect a drive' should have an orange background instead of blue.

<img width="923" alt="screen shot 2017-02-27 at 16 51 41" src="https://cloud.githubusercontent.com/assets/15323961/23370639/458c61b6-fd0d-11e6-8efd-1278b8f7fb96.png">

jviotti:
Yeah, let's make the current dialog open by default if there is no drive and keep improving from there. /cc @Shou

Shou:
So to clarify we open the drive selector modal on startup and display the message there if there is no drive?

lurch:
No, I think automatically opening the drive-selector modal on startup would confuse / scare people.

[...]

PR #1146, opened 2017-02-28
feat(GUI): allow opening drive selector modal when no drives

Shou:
We change the drive selector button text to be a static 'select drive'
and allow the user to open the dialog even if there are no drives,
displaying the 'no drives' warning inside the modal.

Closes: https://github.com/resin-io/etcher/issues/884
Changelog-Entry: Allow opening of drive selector modal when there are no
drives.

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -29,7 +29,6 @@ const messages = require('../shared/messages');
 
 /* eslint-enable no-var */
 
-const _ = require('lodash');
 const Store = require('./models/store');
 
 const app = angular.module('Etcher', [
@@ -50,7 +49,6 @@ const app = angular.module('Etcher', [
   // Components
   require('./components/svg-icon/svg-icon'),
   require('./components/update-notifier/update-notifier'),
-  require('./components/drive-selector/drive-selector'),
   require('./components/warning-modal/warning-modal'),
 
   // Pages
@@ -130,7 +128,7 @@ app.run((AnalyticsService, OSWindowProgressService, FlashStateModel) => {
   });
 });
 
-app.run(($timeout, DriveScannerService, DrivesModel, ErrorService, DriveSelectorService) => {
+app.run(($timeout, DriveScannerService, DrivesModel, ErrorService) => {
   DriveScannerService.on('drives', (drives) => {
 
     // Safely trigger a digest cycle.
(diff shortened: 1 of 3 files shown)


Shou:
image

jviotti:
I'm restarting the builds. We're facing some OS X timeouts.

lib/gui/app.js
```diff
@@ -130,7 +129,7 @@ app.run((AnalyticsService, OSWindowProgressService, FlashStateModel) => {
   });
 });
 
-app.run(($timeout, DriveScannerService, DrivesModel, ErrorService, DriveSelectorService) => {
+app.run(($timeout, DriveScannerService, DrivesModel, ErrorService) => {
```
jviotti: Looks like we can remove the `require('./components/drive-selector/drive-selector'),` from the top dependencies as well.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -51,6 +51,13 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
           ng-show="modal.constraints.isDriveValid(drive, modal.state.getImage())"
           ng-disabled="!modal.state.isCurrentDrive(drive.device)"></span>
     </li>
+    <li class="list-group-item"
+      ng-show="modal.drives.getDrives().length === 0">
```
jviotti: There is a function called `modal.drives.hasAvailableDrives()` than we can re-use here.

jviotti:
I love how this UX feels. What are your thoughts @konmouz ? The flow is the following:

If there are no drives, we show the "Select a drive" button, and when the user clicks on it, the modal appears automatically saying that you have no drive. If you do have one drive, the drive is auto selected.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -51,6 +51,13 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
+      <div>
+        <b>Connect a drive!</b>
+        <div>No removable drive detected.</div>
```
jviotti: I'm not sure about this (I'm not a native English speaker), but should't this be "No removable drives detected."?

jviotti: @lurch ?

lurch: I guess it could be either, given that the title is specifically asking "Connect a drive".
But in the general case, you'd use 0 drives, 1 drive, 2 drives, 3 drives, ...

konmouz:
@jviotti I agree that the flow is smooth and nice. Looking forward to test it.

Shou:
I just pushed the requested changes.

PR #1148, opened 2017-02-28
chore: add ESLint Lodash plugin

jviotti:
This is one step towards enforcing the linter until we barely have to
check for style issues on code reviews.

### DIFF
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -127,9 +127,10 @@ return isElevated().then((elevated) => {
 
         // Translate the current arguments to point to the AppImage
         // Relative paths are resolved from `/tmp/.mount_XXXXXX/usr`
-        const translatedArguments = _.map(_.tail(process.argv), (argv) => {
-          return argv.replace(path.join(process.env.APPDIR, 'usr/'), '');
-        });
+        const translatedArguments = _.chain(process.argv)
+          .tail()
+          .invokeMap('replace', path.join(process.env.APPDIR, 'usr/'), '')
+          .value();
 
         return commandPrefix
           .concat([ process.env.APPIMAGE ])
--- a/lib/cli/errors.js
+++ b/lib/cli/errors.js
@@ -43,18 +43,14 @@ exports.HUMAN_FRIENDLY = {
    * @memberof HUMAN_FRIENDLY
    * @returns {String} message
    */
-  EPERM: () => {
-    return 'You\'re not authorized to perform this operation';
-  },
+  EPERM: _.constant('You\'re not authorized to perform this operation'),
 
   /**
    * @property {Function} EACCES
    * @memberof HUMAN_FRIENDLY
    * @returns {String} message
    */
-  EACCES: () => {
-    return 'You\'re don\'t have access to this resource';
-  }
+  EACCES: _.constant('You don\'t have access to this resource')
 
   /* eslint-enable new-cap */
 
--- a/lib/gui/models/store.js
+++ b/lib/gui/models/store.js
@@ -76,6 +76,30 @@ const ACTIONS = _.fromPairs(_.map([
 }));
 
 /**
+ * @summary Find a drive from the list of available drives
+ * @function
+ * @private
+ *
+ * @param {Object} state - application state
+ * @param {String} device - drive device
+ * @returns {(Object|Undefined)} drive
+ *
+ * @example
+ * const drive = findDrive(state, '/dev/disk2');
+ */
+const findDrive = (state, device) => {
+
+  /* eslint-disable lodash/prefer-lodash-method */
+
+  return state.get('availableDrives').find((drive) => {
+    return drive.get('device') === device;
+  });
+
+  /* eslint-enable lodash/prefer-lodash-method */
+
+};
+
+/**
  * @summary The redux store reducer
  * @function
  * @private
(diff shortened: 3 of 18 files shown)


lib/cli/errors.js
```diff
@@ -43,18 +43,14 @@ exports.HUMAN_FRIENDLY = {
...
 
   /**
    * @property {Function} EACCES
    * @memberof HUMAN_FRIENDLY
    * @returns {String} message
    */
-  EACCES: () => {
-    return 'You\'re don\'t have access to this resource';
-  }
+  EACCES: _.constant('You\'re don\'t have access to this resource')
```
lurch: You're -> You

lib/gui/models/store.js
```diff
@@ -267,10 +271,15 @@ const storeReducer = (state, action) => {
...
       const selectedDevice = state.getIn([ 'selection', 'drive' ]);
+
+      /* eslint-disable lodash/prefer-lodash-method */
+
       const selectedDrive = state.get('availableDrives').find((drive) => {
         return drive.get('device') === selectedDevice;
       });
 
+      /* eslint-enable lodash/prefer-lodash-method */
+
```
lurch: Would it be worth refactoring these two similar-looking code-blocks into a shared helper function?

versionist.conf.js
```diff
@@ -16,6 +16,8 @@
 
 'use strict';
 
+const _ = require('lodash');
```
lurch: Does it make sense for `versionist.conf.js` to depend on `lodash`?

lib/cli/errors.js
jviotti: Fixed!

versionist.conf.js
jviotti: There is nothing preventing the file from re-using dependencies from the project it lives in. It might be even desirable in some cases, where the hooks defined here have to make HTTP requests, etc, so I guess its fine.

lib/gui/models/store.js
jviotti: That's a great idea. I'll refactor into a private function.

jhermsmeier:
As an aside; What's the reasoning behind forcing lodash method usage (and using lodash in general)?

jviotti:
@jhermsmeier JavaScript does have a subset of the functions Lodash defines (like `.filter`, `.map`, etc), however it lacks a lot of very nifty functions that Lodash provides, like `.invokeMap`, `.mapKeys`,  `.difference`, etc.

Re-defining those utilities from the basic provided ones makes no sense (we rather re-use them from Lodash), so the remaining way to go is to use JavaScript native functions when possible and Lodash high level utilities when necessary, however that means we're mixing two different things, which reads a bit weird (e.g: JavaScript functions are methods that operate on the object itself, while Lodash functions accept objects as arguments), so its more convenient to just do Lodash all the way.

lib/child-writer/writer-proxy.js
```diff
@@ -103,9 +103,10 @@ return isElevated().then((elevated) => {
...
-        const translatedArguments = _.map(_.tail(process.argv), (argv) => {
-          return argv.replace(path.join(process.env.APPDIR, 'usr/'), '');
-        });
+        const translatedArguments = _.chain(process.argv)
+          .tail()
+          .invokeMap('replace', path.join(process.env.APPDIR, 'usr/'), '')
+          .value();
 
         return commandPrefix
         .concat([ process.env.APPIMAGE ])
```
jhermsmeier: Looks like there's an indent missing here?

lib/gui/utils/byte-size/byte-size.js
```diff
@@ -26,6 +26,11 @@
 const angular = require('angular');
 const MODULE_NAME = 'Etcher.Utils.ByteSize';
 const ByteSize = angular.module(MODULE_NAME, []);
+
+/* eslint-disable lodash/prefer-lodash-method */
```
jhermsmeier: Would it make sense to loosen up the eslint rule, instead of having to pepper the code with these comments (since the number of exceptions is quite high with this one)?

lurch: I guess this is the classic cost/benefit trade-off...

jviotti: This is happening 4 times across the code base:

- 2 of them because of the Angular way to declare filters (e.g: `module.filter`)
- 1 of them because of the RxJS `.map` function
- 1 of them because of the ImmutableJS `.find` function

I rather deal with these 4 cases than risking some inconsistencies because of not catching something during a code review. I do have some ideas to make this better though:

- We only use ImmutableJS stuff in the store, to I guess we can disable the rule for that whole file.
- We'll move from Angular soon, which means half of the annotations will disappear. Angular filters are just some syntax sugar, so we could even not make filters and we should be OK

What do you think?

lurch: Personally, I agree that being strict with exceptions, is better than being lax and allowing accidental inconsistencies.

jhermsmeier: Aye, makes sense. Happy to keep it like it is then, since it's going to go away in the near future anyways.

lib/child-writer/writer-proxy.js
jviotti: True, let me investigate if we can make the linter catch this as well.

jviotti: I couldn't find any :( I'll fix this in a separate PR.

jviotti: https://github.com/resin-io/etcher/pull/1150

jviotti:
I'll merge this PR after https://github.com/resin-io/etcher/pull/1149, so that conflicts are a bit easier to resolve (this PR touches 17 files while the other touches 49).

PR #1149, opened 2017-03-01
chore: revise ESLint built-in configuration

jviotti:
There are a lot of new rules since the last time I revised the ESLint
rules documentation.

I've updated the main `.eslintrc.yml` to include some newer additions,
plus I added another ESLint configuration file inside `tests`, so we can
add some stricted rules to the production code while relaxing them for
the test suite (due to the fact that Mocha is not very ES6 friendly and
Angular tests require a bit of dark magic to setup).

This is a summary of the most important changes:

- Disallow "magic numbers"

These should now be extracted to constants, which forces us to think of
a good name for them, and thus make the code more self-documenting (I
had to Google up the meaning of some existing magic numbers, so I guess
this will be great for readability purposes).

- Require consistent `return` statements

Some functions relied on JavaScript relaxed casting mechanism to work,
which now have explicit return values. This flag also helped me detect
some promises that were not being returned, and therefore risked not
being caught by the exception handlers in case of errors.

- Disallow redefining function arguments

Immutability makes functions easier to reason about.

- Enforce JavaScript string templates instead of string concatenation

We were heavily mixing boths across the codebase.

There are some extra rules that I tweaked, however most of codebase
changes in this commit are related to the rules mentioned above.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -9,9 +9,6 @@ rules:
 
   # Possible Errors
 
-  comma-dangle:
-    - error
-    - never
   no-cond-assign:
     - error
   no-console:
@@ -59,6 +56,8 @@ rules:
     - error
   no-sparse-arrays:
     - error
+  no-template-curly-in-string:
+    - error
   no-unexpected-multiline:
     - error
   no-unreachable:
@@ -82,8 +81,12 @@ rules:
     - error
   block-scoped-var:
     - error
+  class-methods-use-this:
+    - error
   complexity:
     - off
+  consistent-return:
+    - error
   curly:
     - error
   default-case:
@@ -125,6 +128,8 @@ rules:
     - error
   no-floating-decimal:
     - error
+  no-global-assign:
+    - error
   no-implicit-coercion:
     - error
   no-implicit-globals:
@@ -138,6 +143,8 @@ rules:
   no-lone-blocks:
     - error
   no-loop-func:
+    - error
+  no-magic-numbers:
     - error
   no-multi-spaces:
     - error
@@ -155,11 +162,18 @@ rules:
     - error
   no-octal-escape:
     - error
+  no-param-reassign:
+    - error
   no-proto:
     - error
   no-redeclare:
     - error
+  no-restricted-properties:
+    - error
+    - property: __proto__
   no-return-assign:
+    - error
+  no-return-await:
     - error
   no-script-url:
     - error
@@ -172,6 +186,8 @@ rules:
   no-throw-literal:
     - error
   no-unmodified-loop-condition:
+    - error
+  no-unused-expressions:
     - error
   no-unused-labels:
     - error
@@ -205,12 +221,18 @@ rules:
 
   # Variables
 
+  init-declarations:
+    - error
+    - always
   no-catch-shadow:
     - error
   no-delete-var:
     - error
   no-label-var:
     - error
+  no-restricted-globals:
+    - error
+    - event
   no-shadow:
     - error
   no-shadow-restricted-names:
@@ -218,6 +240,8 @@ rules:
   no-undef:
     - error
   no-undef-init:
+    - error
+  no-undefined:
     - error
   no-unused-vars:
     - error
@@ -257,6 +281,13 @@ rules:
     - 1tbs
   camelcase:
     - error
+  capitalized-comments:
+    - error
+    - always
+    - ignoreConsecutiveComments: true
+  comma-dangle:
+    - error
+    - never
   comma-spacing:
     - error
     - before: false
@@ -272,6 +303,12 @@ rules:
     - self
   eol-last:
     - error
+  func-call-spacing:
+    - error
+    - never
+  func-name-matching:
+    - error
+    - always
   func-names:
     - error
     - never
@@ -280,6 +317,11 @@ rules:
     - expression
   id-blacklist:
     - error
+  id-length:
+    - error
+    - min: 2
+      exceptions:
+        - "_"
   indent:
     - error
     - 2
@@ -293,6 +335,12 @@ rules:
     - error
     - before: true
       after: true
+  line-comment-position:
+    - error
+    - position: above
+  linebreak-style:
+    - error
+    - unix
   lines-around-comment:
     - error
     - beforeBlockComment: true
@@ -305,6 +353,9 @@ rules:
       allowObjectEnd: false
       allowArrayStart: true
       allowArrayEnd: false
+  lines-around-directive:
+    - error
+    - always
   max-len:
     - error
     - code: 130
@@ -312,13 +363,20 @@ rules:
       ignoreComments: false
       ignoreTrailingComments: false
       ignoreUrls: true
+  max-params:
+    - off
   max-statements-per-line:
     - error
     - max: 1
+  multiline-ternary:
+    - error
+    - never
   new-cap:
     - error
   new-parens:
     - error
+  newline-per-chained-call:
+    - off
   no-array-constructor:
     - error
   no-bitwise:
@@ -327,9 +385,13 @@ rules:
     - error
   no-inline-comments:
     - error
+  no-lonely-if:
+    - error
   no-mixed-operators:
     - error
   no-mixed-spaces-and-tabs:
+    - error
+  no-multi-assign:
     - error
   no-multiple-empty-lines:
     - error
@@ -344,7 +406,13 @@ rules:
     - error
   no-plusplus:
     - error
+  no-restricted-syntax:
+    - error
+    - WithStatement
+    - ForInStatement
   no-spaced-func:
+    - error
+  no-tabs:
     - error
   no-trailing-spaces:
     - error
@@ -363,6 +431,9 @@ rules:
     - always
   object-property-newline:
     - error
+  one-var-declaration-per-line:
+    - error
+    - always
   one-var:
     - error
     - never
@@ -372,6 +443,9 @@ rules:
   operator-linebreak:
     - error
     - before
+  padded-blocks:
+    - error
+    - classes: always
   quote-props:
     - error
     - as-needed
@@ -384,6 +458,7 @@ rules:
         FunctionDeclaration: true
         ClassDeclaration: true
         MethodDefinition: true
+        ArrowFunctionExpression: true
   semi:
     - error
     - always
@@ -401,7 +476,14 @@ rules:
     - never
   space-infix-ops:
     - error
+  space-unary-ops:
+    - error
+    - words: true
+      nonwords: false
   spaced-comment:
+    - error
+    - always
+  template-tag-spacing:
     - error
     - always
   unicode-bom:
@@ -447,12 +529,24 @@ rules:
     - error
   no-var:
     - error
+  object-shorthand:
+    - error
+    - always
   prefer-const:
     - error
   prefer-reflect:
     - error
   prefer-spread:
     - error
+  prefer-numeric-literals:
+    - error
+  prefer-rest-params:
+    - error
+  prefer-template:
+    - error
+  prefer-arrow-callback:
+    - error
+    - allowNamedFunctions: false
   require-yield:
     - error
   rest-spread-spacing:
@@ -460,6 +554,8 @@ rules:
   template-curly-spacing:
     - error
     - never
+  symbol-description:
+    - error
   yield-star-spacing:
     - error
     - before: true
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -61,7 +61,7 @@ exports.write = (image, drive, options) => {
 
   const argv = cli.getArguments({
     entryPoint: rendererUtils.getApplicationEntryPoint(),
-    image: image,
+    image,
     device: drive.device,
     validateWriteOnSuccess: options.validateWriteOnSuccess,
     unmountOnSuccess: options.unmountOnSuccess
@@ -77,6 +77,14 @@ exports.write = (image, drive, options) => {
   ipc.config.silent = true;
   ipc.serve();
 
+  /**
+   * @summary Safely terminate the IPC server
+   * @function
+   * @private
+   *
+   * @example
+   * terminateServer();
+   */
   const terminateServer = () => {
 
     // Turns out we need to destroy all sockets for
@@ -90,6 +98,16 @@ exports.write = (image, drive, options) => {
     ipc.server.stop();
   };
 
+  /**
+   * @summary Emit an error to the client
+   * @function
+   * @private
+   *
+   * @param {Error} error - error
+   *
+   * @example
+   * emitError(new Error('foo bar'));
+   */
   const emitError = (error) => {
     terminateServer();
     emitter.emit('error', error);
@@ -97,7 +115,7 @@ exports.write = (image, drive, options) => {
 
   ipc.server.on('error', emitError);
   ipc.server.on('message', (data) => {
-    let message;
+    let message = null;
 
     try {
       message = robot.parseMessage(data);
@@ -112,7 +130,7 @@ exports.write = (image, drive, options) => {
       return emitError(robot.recomposeErrorMessage(message));
     }
 
-    emitter.emit(robot.getCommand(message), robot.getData(message));
+    return emitter.emit(robot.getCommand(message), robot.getData(message));
   });
 
   ipc.server.on('start', () => {
@@ -144,9 +162,13 @@ exports.write = (image, drive, options) => {
         });
       }
 
-      if (code !== EXIT_CODES.SUCCESS && code !== EXIT_CODES.VALIDATION_ERROR) {
-        return emitError(new Error(`Child process exited with error code: ${code}`));
+      // We shouldn't emit the `done` event manually here
+      // since the writer process will take care of it.
+      if (code === EXIT_CODES.SUCCESS || code === EXIT_CODES.VALIDATION_ERROR) {
+        return null;
       }
+
+      return emitError(new Error(`Child process exited with error code: ${code}`));
     });
   });
 
--- a/lib/child-writer/renderer-utils.js
+++ b/lib/child-writer/renderer-utils.js
@@ -40,9 +40,12 @@ exports.getApplicationEntryPoint = () => {
     return path.join(process.resourcesPath, 'app.asar');
   }
 
+  const ENTRY_POINT_ARGV_INDEX = 1;
+  const relativeEntryPoint = electron.remote.process.argv[ENTRY_POINT_ARGV_INDEX];
+
   // On GNU/Linux, `pkexec` resolves relative paths
   // from `/root`, therefore we pass an absolute path,
   // in order to be on the safe side.
-  return path.join(CONSTANTS.PROJECT_ROOT, electron.remote.process.argv[1]);
+  return path.join(CONSTANTS.PROJECT_ROOT, relativeEntryPoint);
 
 };
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -37,8 +37,32 @@ const packageJSON = require('../../package.json');
 // and `stderr` to the parent process using IPC communication,
 // taking care of the writer elevation as needed.
 
-const EXECUTABLE = process.argv[0];
-const ETCHER_ARGUMENTS = process.argv.slice(2);
+/**
+ * @summary The Etcher executable file path
+ * @constant
+ * @private
+ * @type {String}
+ */
+const executable = _.first(process.argv);
+
+/**
+ * @summary The first index that represents an actual option argument
+ * @constant
+ * @private
+ * @type {Number}
+ *
+ * @description
+ * The first arguments are usually the program executable itself, etc.
+ */
+const OPTIONS_INDEX_START = 2;
+
+/**
+ * @summary The list of Etcher argument options
+ * @constant
+ * @private
+ * @type {String[]}
+ */
+const etcherArguments = process.argv.slice(OPTIONS_INDEX_START);
 
 return isElevated().then((elevated) => {
 
@@ -153,7 +177,7 @@ return isElevated().then((elevated) => {
       ipc.of[process.env.IPC_SERVER_ID].on('error', reject);
       ipc.of[process.env.IPC_SERVER_ID].on('connect', () => {
 
-        const child = childProcess.spawn(EXECUTABLE, ETCHER_ARGUMENTS, {
+        const child = childProcess.spawn(executable, etcherArguments, {
           env: {
 
             // The CLI might call operating system utilities (like `diskutil`),
@@ -169,6 +193,18 @@ return isElevated().then((elevated) => {
         child.on('error', reject);
         child.on('close', resolve);
 
+        /**
+         * @summary Emit an object message to the IPC server
+         * @function
+         * @private
+         *
+         * @param {Buffer} data - json message data
+         *
+         * @example
+         * emitMessage(Buffer.from(JSON.stringify({
+         *   foo: 'bar'
+         * })));
+         */
         const emitMessage = (data) => {
 
           // Output from stdout/stderr coming from the CLI might be buffered,
--- a/lib/cli/etcher.js
+++ b/lib/cli/etcher.js
@@ -29,6 +29,9 @@ const robot = require('../shared/robot');
 const messages = require('../shared/messages');
 const EXIT_CODES = require('../shared/exit-codes');
 
+const ARGV_IMAGE_PATH_INDEX = 0;
+const imagePath = options._[ARGV_IMAGE_PATH_INDEX];
+
 isElevated().then((elevated) => {
   if (!elevated) {
     throw new Error(messages.error.elevationRequired());
@@ -50,10 +53,10 @@ isElevated().then((elevated) => {
     override: {
       drive: options.drive,
 
-      // If `options.yes` is `false`, pass `undefined`,
+      // If `options.yes` is `false`, pass `null`,
       // otherwise the question will not be asked because
       // `false` is a defined value.
-      yes: robot.isEnabled(process.env) || options.yes || undefined
+      yes: robot.isEnabled(process.env) || options.yes || null
 
     }
   });
@@ -76,7 +79,7 @@ isElevated().then((elevated) => {
       throw new Error(`Drive not found: ${answers.drive}`);
     }
 
-    return writer.writeImage(options._[0], selectedDrive, {
+    return writer.writeImage(imagePath, selectedDrive, {
       unmountOnSuccess: options.unmount,
       validateWriteOnSuccess: options.check
     }, (state) => {
@@ -109,6 +112,7 @@ isElevated().then((elevated) => {
       console.log(`Checksum: ${results.sourceChecksum}`);
     }
 
+    return Bluebird.resolve();
   }).then(() => {
     process.exit(EXIT_CODES.SUCCESS);
   });
@@ -121,6 +125,7 @@ isElevated().then((elevated) => {
     }
 
     errors.print(error);
+    return Bluebird.resolve();
   }).then(() => {
     if (error.code === 'EVALIDATION') {
       process.exit(EXIT_CODES.VALIDATION_ERROR);
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -25,6 +25,33 @@ const EXIT_CODES = require('../shared/exit-codes');
 const packageJSON = require('../../package.json');
 
 /**
+ * @summary The minimum required number of CLI arguments
+ * @constant
+ * @private
+ * @type {Number}
+ */
+const MINIMUM_NUMBER_OF_ARGUMENTS = 1;
+
+/**
+ * @summary The index of the image argument
+ * @constant
+ * @private
+ * @type {Number}
+ */
+const IMAGE_PATH_ARGV_INDEX = 0;
+
+/**
+ * @summary The first index that represents an actual option argument
+ * @constant
+ * @private
+ * @type {Number}
+ *
+ * @description
+ * The first arguments are usually the program executable itself, etc.
+ */
+const OPTIONS_INDEX_START = 2;
+
+/**
  * @summary Parsed CLI options and arguments
  * @type Object
  * @public
@@ -34,7 +61,7 @@ module.exports = yargs
   // Don't wrap at all
   .wrap(null)
 
-  .demand(1, 'Missing image')
+  .demand(MINIMUM_NUMBER_OF_ARGUMENTS, 'Missing image')
 
   // Usage help
   .usage('Usage: $0 [options] <image>')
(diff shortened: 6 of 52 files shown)


lurch:
Looks like you've been 'relaxying' your spelling too. LOL

jviotti:
lol, fixed!

lib/gui/models/store.js
```diff
@@ -91,7 +105,8 @@ const storeReducer = (state, action) => {
 
       const newState = state.set('availableDrives', Immutable.fromJS(action.data));
```
lurch: I've never written any Redux code before, but should the first parameter to these `.set` calls also be constants, rather than just strings, to cut down on the possibility of any typos?

lib/gui/models/store.js
```diff
@@ -354,7 +369,7 @@ module.exports = _.merge(redux.createStore(
       // otherwise we will be basically erasing the property
       // we aim the keep serialising the in future.
```
lurch: "aim to keep" ?

lib/gui/models/supported-formats.js
```diff
@@ -115,7 +115,7 @@ SupportedFormats.service('SupportedFormatsModel', function() {
    * }
    */
   this.isSupportedImage = (imagePath) => {
-    const extension = path.extname(imagePath).slice(1).toLowerCase();
+    const extension = _.join(_.tail(path.extname(imagePath)), '').toLowerCase();
```
lurch: IMHO this is _much_ more unreadable. What about using `replace('.', '')` instead?

Also, `.${extension}` is used again at the bottom of the function, so I guess you could maybe do:
```
const extname = path.extname(imagePath);
const extension = extname.replace('.', '').toLowerCase();
```
and then replace `.${extension}` with `extname`. Should there also be some check for `extension` being empty?

lib/gui/modules/analytics.js
```diff
@@ -129,11 +129,8 @@ analytics.service('AnalyticsService', function($log, $window, $mixpanel, Setting
 
     }
 
-    if (data) {
-      message += ` (${JSON.stringify(data)})`;
-    }
-
-    this.logDebug(message);
+    const debugMessage = data ? `${message} (${JSON.stringify(data)})` : message;
```
lurch: IMHO the old version was more readable, but I guess this is the price you pay for banning string concatenation ;)

lib/gui/modules/drive-scanner.js
```diff
@@ -34,16 +34,17 @@ const driveScanner = angular.module(MODULE_NAME, [
 
 driveScanner.factory('DriveScannerService', (SettingsModel) => {
   const DRIVE_SCANNER_INTERVAL_MS = 2000;
+  const FIRST_SCAN_DELAY_MS = 0;
```
lurch: It would be nice if this was also prefixed with `DRIVE_SCANNER_` to match the other constant here.

lib/gui/modules/drive-scanner.js
```diff
@@ -52,7 +53,9 @@ driveScanner.factory('DriveScannerService', (SettingsModel) => {
 
         return drive;
       });
+    })
 
+    .map((drives) => {
```
lurch: I'm not familiar with the coding style used here (so I can't quite tell what this is doing) but how come the old code had two `map` functions, but the new code has three?

lib/gui/os/dialog/services/dialog.js
```diff
@@ -113,20 +113,28 @@ module.exports = function($q, SupportedFormatsModel) {
    * });
    */
   this.showWarning = (options) => {
+    const BUTTONS = [
+      options.confirmationLabel,
+      options.rejectionLabel
+    ];
+
+    const BUTTON_CONFIRMATION_INDEX = _.indexOf(BUTTONS, options.confirmationLabel);
+    const BUTTON_REJECTION_INDEX = _.indexOf(BUTTONS, options.rejectionLabel);
```
lurch: This looks horrible :-(

Feels like there must be a nicer way to do this...?

lib/gui/os/dialog/services/dialog.js
```diff
@@ -113,20 +113,28 @@ module.exports = function($q, SupportedFormatsModel) {
...
+
     return $q((resolve) => {
       electron.remote.dialog.showMessageBox(currentWindow, {
         type: 'warning',
         buttons: [
```
lurch: Given you've constructed the array above, could this just be `buttons: BUTTONS` ?

lib/gui/os/dialog/services/dialog.js
```diff
@@ -113,20 +113,28 @@ module.exports = function($q, SupportedFormatsModel) {
...
-        defaultId: 1,
-        cancelId: 1,
+        defaultId: BUTTON_REJECTION_INDEX,
+        cancelId: BUTTON_REJECTION_INDEX,
         title: 'Attention',
         message: options.title,
         detail: options.description
       }, (response) => {
-        return resolve(response === 0);
+        return resolve(response === BUTTON_CONFIRMATION_INDEX);
```
lurch: It means changing an integer-comparison to a string-comparison, but I guess `return resolve(BUTTONS[response] === options.confirmationLabel);` would work here?

lib/gui/os/dialog/services/dialog.js
```diff
@@ -113,20 +113,28 @@ module.exports = function($q, SupportedFormatsModel) {
    * });
    */
   this.showWarning = (options) => {
+    const BUTTONS = [
+      options.confirmationLabel,
```
lurch: Should we be setting defaults for `.confirmationLabel` and `.rejectionLabel` in case they don't exist?

lib/gui/os/window-progress/services/window-progress.js
```diff
@@ -47,11 +44,14 @@ module.exports = function() {
    * OSWindowProgressService.set(85);
    */
   this.set = (percentage) => {
-    if (percentage > 100 || percentage < 0) {
+    const PERCENTAGE_MINIMUM = 0;
+    const PERCENTAGE_MAXIMUM = 100;
```
lurch: Lol

lib/gui/pages/main/controllers/flash.js
```diff
@@ -111,9 +111,12 @@ module.exports = function(
       return 'Flash!';
     }
 
-    if (flashState.percentage === 0 && !flashState.speed) {
+    const PERCENTAGE_MINIMUM = 0;
```
lurch: Does the same constant in multiple files imply that it out to be abstracted out?

lib/gui/pages/main/controllers/flash.js
```diff
@@ -111,9 +111,12 @@ module.exports = function(
...
+    const PERCENTAGE_MAXIMUM = 100;
+
+    if (flashState.percentage === PERCENTAGE_MINIMUM && !flashState.speed) {
       return 'Starting...';
-    } else if (flashState.percentage === 100) {
+    } else if (flashState.percentage === PERCENTAGE_MAXIMUM) {
       if (isChecking && SettingsModel.get('unmountOnSuccess')) {
         return 'Unmounting...';
       }
```
lurch: There's multiple separate `if ...` blocks here. Should they be combined into a single `if ... else if ... else if ... else ...` block?

lib/gui/pages/settings/settings.js
```diff
@@ -25,7 +25,8 @@ const MODULE_NAME = 'Etcher.Pages.Settings';
 const SettingsPage = angular.module(MODULE_NAME, [
   require('angular-ui-router'),
   require('../../components/warning-modal/warning-modal'),
-  require('../../models/settings')
+  require('../../models/settings'),
+  require('../../modules/error')
```
lurch: Hmmm? How did the code work without this `require('../../modules/error')` before?

lib/image-stream/archive.js
```diff
@@ -170,11 +170,12 @@ exports.extractImage = (archive, hooks) => {
   return hooks.getEntries(archive).then((entries) => {
 
     const imageEntries = _.filter(entries, (entry) => {
-      const extension = path.extname(entry.name).slice(1);
+      const extension = _.join(_.tail(path.extname(entry.name)), '');
```
lurch: See my earlier `extension` comments

lib/shared/drive-constraints.js
```diff
@@ -20,6 +20,14 @@ const _ = require('lodash');
 const pathIsInside = require('path-is-inside');
 
 /**
+ * @summary The default size for things such as images and drives
+ * @constant
+ * @private
+ * @type {Number}
+ */
+const DEFAULT_SIZE = 0;
```
lurch: Would `UNKNOWN_SIZE` be better here?

lib/child-writer/index.js
```diff
@@ -144,9 +162,13 @@ exports.write = (image, drive, options) => {
         });
       }
 
-      if (code !== EXIT_CODES.SUCCESS && code !== EXIT_CODES.VALIDATION_ERROR) {
-        return emitError(new Error(`Child process exited with error code: ${code}`));
+      // We shouldn't emit the `done` event manually here
+      // since the writer process will take care of it.
+      if (code === EXIT_CODES.SUCCESS || code === EXIT_CODES.VALIDATION_ERROR) {
```
jviotti: There are various ways the CLI process can fail, which are grouped into two categories:

- The CLI finished its task (flash & validation)
- The CLI crashed in the middle

If the CLI finished its task, then there are two possible ways out:

- The flash completed successfully (`EXIT_CODES.SUCCESS`)
- The flash completed, but the validation failed (`EXIT_CODES.VALIDATION_ERROR`)

If the CLI completes successfully, then it will emit its own `done` event, therefore we should not emit ourselves here.

lib/child-writer/renderer-utils.js
```diff
@@ -40,9 +40,12 @@ exports.getApplicationEntryPoint = () => {
     return path.join(process.resourcesPath, 'app.asar');
   }
 
+  const ENTRY_POINT_ARGV_INDEX = 1;
+  const RELATIVE_ENTRY_POINT = electron.remote.process.argv[ENTRY_POINT_ARGV_INDEX];
```
jviotti: Sounds like a good convention! I'll fix.

lib/child-writer/writer-proxy.js
```diff
@@ -37,8 +37,32 @@ const packageJSON = require('../../package.json');
...
+ */
+const OPTIONS_INDEX_START = 2;
+
+/**
+ * @summary The list of Etcher argument options
+ * @constant
+ * @private
+ * @type {String[]}
+ */
+const ETCHER_ARGUMENTS = process.argv.slice(OPTIONS_INDEX_START);
```
jviotti: Fixed!

lib/child-writer/writer-proxy.js
```diff
@@ -169,6 +193,18 @@ return isElevated().then((elevated) => {
         child.on('error', reject);
         child.on('close', resolve);
 
+        /**
+         * @summary Emit an object message to the IPC server
+         * @function
+         * @private
+         *
+         * @param {Buffer} data - message data
```
jviotti: It should be JSON. I clarified it on the param.

lib/cli/etcher.js
```diff
@@ -53,7 +56,7 @@ isElevated().then((elevated) => {
       // If `options.yes` is `false`, pass `undefined`,
```
jviotti: Good catch!

lib/cli/etcher.js
```diff
@@ -120,7 +124,7 @@ isElevated().then((elevated) => {
       return robot.printError(error);
     }
 
-    errors.print(error);
+    return errors.print(error);
```
jviotti: Haha, fixed!

lib/cli/options.js
```diff
@@ -25,6 +25,25 @@ const EXIT_CODES = require('../shared/exit-codes');
 const packageJSON = require('../../package.json');
 
 /**
+ * @summary The required number of CLI arguments
+ * @constant
+ * @private
+ * @type {Number}
+ */
+const REQUIRED_NUMBER_OF_ARGUMENTS = 1;
```
jviotti: Yeah, I like how that reads.

lib/cli/options.js
```diff
@@ -25,6 +25,25 @@ const EXIT_CODES = require('../shared/exit-codes');
...
+
+/**
+ * @summary The first index that represents an actual option argument
+ * @constant
+ * @private
+ * @type {Number}
+ *
+ * @description
+ * The first arguments are usually the program executable itself, etc.
```
jviotti: Not really. Consider that `argv` looks like this:

```sh
node bin/cli.js path/to/image.js <options>
```

The first two ones are not options.

Maybe the name is confusing? My intention is that `OPTIONS_INDEX_START` equals the first index of `argv` that is a option to the program.

lib/cli/options.js
```diff
@@ -25,6 +25,25 @@ const EXIT_CODES = require('../shared/exit-codes');
...
+ */
+const OPTIONS_INDEX_START = 2;
```
jviotti: See https://github.com/resin-io/etcher/pull/1149#discussion_r104035788. This is not related to the minimum number of arguments.

lib/cli/options.js
```diff
@@ -64,20 +83,21 @@ module.exports = yargs
...
     }
 
-    process.exit(1);
+    process.exit(EXIT_CODES.GENERAL_ERROR);
   })
 
   // Assert that image exists
   .check((argv) => {
-    fs.accessSync(argv._[0]);
+    const IMAGE_PATH_ARGV_INDEX = 0;
```
jviotti: Done!

lib/gui/components/modal/services/modal.js
```diff
@@ -60,17 +60,11 @@ module.exports = function($uibModal, $q) {
           .catch((error) => {
 
             // Bootstrap doesn't 'resolve' these but cancels the dialog;
```
jviotti: Fixed!

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -60,7 +60,9 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_TIME, M
           // The error status equals -1 if the request couldn't
           // be made successfully, for example, because of a
           // timeout on an unstable network connection.
-          if (error.status === -1) {
+          const ERROR_CODE_UNSUCCESSFULL_REQUEST = -1;
```
jviotti: > IMHO this constant should either be moved to the top of this file, or to the constants.js that I saw used elsewhere. 

The reason I put it here is that this error code is only relevant in the context of the `etcherLatestVersion` function. I'm worried that moving it somewhere else (even if its the top of the file), will make it seem that it applies somewhere else.

> And the comment above should be changed to not refer directly to -1.

Good idea!

> And there's only one 'l' in 'unsuccessful' ;)

Ah, fixed!

lib/gui/etcher.js
```diff
@@ -41,12 +41,13 @@ electron.app.on('ready', () => {
   // Prevent flash of white when starting the application
   // https://github.com/atom/electron/issues/2172
   mainWindow.webContents.on('did-finish-load', () => {
+    const FLASH_TIMEOUT_MS = 100;
```
jviotti: Makes sense. I've changed it to `WEBVIEW_LOAD_TIMEOUT_MS`, since the timeout is necessary because the Chromium WebView takes a bit to load completely.

lib/gui/models/drives.js
```diff
@@ -64,19 +64,34 @@ Drives.service('DrivesModel', function() {
...
+   *
+   * The problem is that every time you call `myImmutableObject.toJS()`
+   * you will get a new object, whose reference is different from
+   * the one you previously got, even if the data is exactly the same.
+   *
+   * @param {Function} func - function that returns an ImmutableJS list
+   * @returns {Function} memoized function
+   *
+   * @example
+   * const memoizedFunction = memoizeImmutableListReference(() => {
```
jviotti: Fixed!

lib/gui/models/store.js
```diff
@@ -75,8 +75,22 @@ const ACTIONS = _.fromPairs(_.map([
   return [ message, message ];
 }));
 
-const storeReducer = (state, action) => {
-  state = state || DEFAULT_STATE;
+/**
+ * @summary The redux store reduced
```

lib/gui/models/store.js
```diff
@@ -75,8 +75,22 @@ const ACTIONS = _.fromPairs(_.map([
...
+ * @param {Object} currentState - application state
+ * @param {Object} action - dispatched action
+ * @returns {Object} new application state
+ *
+ * @example
+ * const newState = storeReducer(DEFAULT_STATE, {
+ *   type: ACTIONS.REMOVE_DRIVE
+ * });
+ */
+const storeReducer = (currentState, action) => {
```
jviotti: Oh, fixed! For some reason I though I could only use a default value for the last argument (?)

lib/gui/models/store.js
```diff
@@ -75,8 +75,22 @@ const ACTIONS = _.fromPairs(_.map([
...
+  const state = currentState || DEFAULT_STATE;
```
jviotti: See https://github.com/resin-io/etcher/pull/1149#discussion_r104038893, I got confused about default arguments :)

lib/gui/models/store.js
jviotti: These are property names declare for the default state at the top of the file. We *could* reduce the chances of typos by making constants but I wonder if its to much complexity for trying to mitigate the fact that JavaScript is not a strongly typed language. Maybe moving to TypeScript is the solution?

lib/gui/models/store.js
```diff
@@ -91,7 +105,8 @@ const storeReducer = (state, action) => {
 
       const newState = state.set('availableDrives', Immutable.fromJS(action.data));
 
-      if (action.data.length === 1) {
+      const DRIVE_AUTOSELECT_REQUIRED_LENGTH = 1;
+      if (action.data.length === DRIVE_AUTOSELECT_REQUIRED_LENGTH) {
```
jviotti: Looks great, and much more obvious!

lib/gui/models/store.js

lib/gui/models/supported-formats.js
jviotti: Sounds good. I used `_.replace`, but decided to not store another variable with the extension + the period, since its tricky to find a good name for each, so that its not confusing.

lib/gui/modules/analytics.js
jviotti: I rewrote it without using a one-liner. Its hopefully clearer now.

lib/gui/modules/drive-scanner.js
jviotti: Done!

lib/gui/modules/drive-scanner.js
jviotti: `.flatMap` is a utility from RxJS to make all the `drivelist` function calls originate from the same stream (otherwise RxJS creates a new one for each).

Regarding the other `.map()`, I divided it into two to hopefully make the transformations clearer. The first one updates the descriptions, and the other one applies the system filtering. Happy to bring it back to one if that was better.

lib/gui/os/dialog/services/dialog.js
jviotti: Sounds like a good idea!

lib/gui/os/dialog/services/dialog.js
jviotti: Haha, looks like I missed this completely. I created the `BUTTONS` array but was only using it to get the indexes :)

lib/gui/os/dialog/services/dialog.js
jviotti: Yeah, I agree. The Electron dialog module makes heavy use of number indexes, which is terrible to deal with: https://electron.atom.io/docs/api/dialog/#dialogshowmessageboxbrowserwindow-options-callback

I guess we can remove the need for `BUTTON_CONFIRMATION_INDEX ` using your suggestion from https://github.com/resin-io/etcher/pull/1149#discussion_r104024047, and only keep the rejection one, which is at least nicer. What do you think?

lib/gui/os/dialog/services/dialog.js
jviotti: Hm, but that might fail if both the confirmation and rejection label are the same, which would be terrible on its own, but maybe its better to be on the safe side?

lib/gui/os/window-progress/services/window-progress.js
jviotti: Haha, the drawbacks of disallowing magic numbers :P

lib/gui/pages/main/controllers/flash.js
jviotti: Done!

lib/gui/pages/main/controllers/flash.js
jviotti: Yeah, but where? `lib/shared/percentage.js`? :P I guess we can start a file with constants in `lib/shared/constants.js`. What do you think?

lib/gui/pages/settings/settings.js
jviotti: Coincidence. Another module probably required the error module before this module started, so it all just worked. The Angular module system is awful though, and doesn't give any warning for these kind of things. React should make things better.

lib/image-stream/archive.js
jviotti: Fixed!

lib/shared/drive-constraints.js
jviotti: Yep, fixed!

lib/gui/components/update-notifier/update-notifier.js
```diff
@@ -29,7 +29,18 @@ const UpdateNotifier = angular.module(MODULE_NAME, [
...
 
-UpdateNotifier.constant('UPDATE_NOTIFIER_SLEEP_TIME', 7 * 24 * 60 * 60 * 100);
+/**
+ * @summary The update notifier sleep time
+ * @constant
+ * @private
+ * @type {Number}
+ *
+ * @description
+ * Seven days: 7 * 24 * 60 * 60 * 100
```
jviotti: Crap, this went undetected for so long! :O Amazing catch!

lib/gui/components/update-notifier/update-notifier.js
```diff
@@ -29,7 +29,18 @@ const UpdateNotifier = angular.module(MODULE_NAME, [
   require('../../os/open-external/open-external')
 ]);
 
-UpdateNotifier.constant('UPDATE_NOTIFIER_SLEEP_TIME', 7 * 24 * 60 * 60 * 100);
+/**
+ * @summary The update notifier sleep time
```
jviotti: Great catch! Done! I also realized we should move the days to milliseconds conversion to `units.js`.

lib/gui/components/update-notifier/services/update-notifier.js
lurch: > this error code is only relevant in the context of the `etcherLatestVersion` function.

In that case maybe the same module that exports the `etcherLatestVersion` function could also export appropriate constants for the error codes? :)

lib/gui/models/store.js
lurch: Hmmm, good point. You know javascript better than I do ;-)

lib/gui/components/update-notifier/services/update-notifier.js
jviotti: Hm, I investigated this one in more detail and it turns out its a bit tricky. The originator of this `-1` error code when the request couldn't be made is `$http`. `etcherLatestVersion` accepts a  function that given a URL allows the client to determine how to perform an HTTP request to it (given we use the module in the browser as well).

Since `etcherLatestVersion` accepts a function to make HTTP requests, then hardcoding the `-1` knowledge there looks a bit wrong, since different HTTP request functions might use different error codes for different things.

That leaves us with a few options:

- Expose a set of error constants from `etcherLatestVersion` and update the `.catch` block of the `$http` promise to compare against `-1` and translate that to an error number `etcherLatestVersion` understands, but that means we still have to declare this `-1` constant here

- Make `etcherLatestVersion` ignore every error coming from the HTTP requester function. This means we will swallow errors that we should be seeing

- Leave it as it is now

What do you think?

jviotti:
@lurch Do I have the green sign for this one? :)

lib/gui/components/modal/services/modal.js
```diff
@@ -60,17 +60,11 @@ module.exports = function($uibModal, $q) {
...
-            // therefore call 'resolve' here applied to 'false'.
             if (error === 'escape key press' || error === 'backdrop click') {
-              resolve();
-
-            // For some annoying reason, UI Bootstrap Modal rejects
-            // the result reason if the user clicks on the backdrop
-            // (e.g: the area surrounding the modal).
-            } else if (error !== 'backdrop click') {
-              return reject(error);
+              return resolve();
```
Shou: Yeah, this works.

lib/gui/os/dropzone/directives/dropzone.js
```diff
@@ -41,7 +41,7 @@ module.exports = ($timeout) => {
       osDropzone: '&'
     },
     link: (scope, element) => {
-      const domElement = element[0];
+      const domElement = _.first(element);
```
jhermsmeier: Since `element` seems to be a node list, shouldn't it be `elements`?

.eslintrc.yml
```diff
@@ -447,19 +529,33 @@ rules:
...
     - error
+  prefer-numeric-literals:
+    - error
+  prefer-rest-params:
+    - error
+  prefer-template:
+    - error
+  prefer-arrow-callback:
+    - error
+    - allowNamedFunctions: false
```
jhermsmeier: Does this mean `.method( function() {})` or `.on( 'event', function() {})` will be disallowed?

lib/gui/os/dropzone/directives/dropzone.js
jviotti: @jhermsmeier Its a jQuery lite object, in which you can access the native DOM element by accessing its first element (which I agree its weird)

.eslintrc.yml
jviotti: It means that something like `var foo = function foo() { ... }` is not allowed.

jhermsmeier: Ah I see, `prefer-arrow-callback`'s semantics weren't quite clear to me

lib/gui/os/dropzone/directives/dropzone.js
jhermsmeier: Oh, right – in that case renaming it to `$element` could clarify it. What do you think?

jviotti: Sounds good, fixed!

jviotti:
The conflict hell just began :P

lib/cli/options.js
lurch: > Maybe the name is confusing? My intention is that OPTIONS_INDEX_START equals the first index of argv that is a option to the program.

Yeah, I understood that, I just think the `@description` is confusing. Perhaps it'd simply be clearer if the `@description` here (and at https://github.com/resin-io/etcher/pull/1149/files#diff-1db67373c40ea3d2756fbc0768723000R54 ) was just removed. I don't think it adds anything.

lib/cli/options.js
lurch: Oh? If `MINIMUM_NUMBER_OF_ARGUMENTS` is `1`, doesn't that mean that the program-usage looks like:
```
program_name compulsory_argument [optional_arguments, ...]
```
and so the fact that there's a single compulsory_ argument, means that the optional_arguments start at index 2, i.e. `MINIMUM_NUMBER_OF_ARGUMENTS + 1` ?

Whereas if `MINIMUM_NUMBER_OF_ARGUMENTS` is `2`, the program-usage would be:
```
program_name compulsory_argument1 compulsory_argument2 [optional_arguments, ...]
```
and `OPTIONS_INDEX_START` would then be `3`, which is still `MINIMUM_NUMBER_OF_ARGUMENTS + 1` ?

lib/gui/components/update-notifier/services/update-notifier.js
lurch: Yeah, I guess you could leave it as it is now, but maybe add a comment explaining that the `-1` actually relates to the `$http` object, because currently that's kinda unclear. I guess  `$http` doesn't have any useful constants you could import here?
`ERROR_CODE_UNSUCCESSFUL_REQUEST` is kinda generic-sounding ;)  What about `ERROR_HTTP_REQUEST_UNSUCCESSFUL`?

(P.S. Sorry for coming back to this so late)

lib/gui/os/dialog/services/dialog.js
```diff
@@ -149,19 +159,19 @@ module.exports = function($q, SupportedFormatsModel) {
    * OSDialogService.showError('Foo Bar', 'An error happened!');
    */
   this.showError = (error, description) => {
```
lurch: I guess this could use the same `this.showError = (error, description) => {` as https://github.com/resin-io/etcher/pull/1149#discussion_r104005634 ?

lib/gui/os/window-progress/services/window-progress.js
```diff
@@ -47,11 +44,14 @@ module.exports = function() {
...
+
+    if (percentage > PERCENTAGE_MAXIMUM || percentage < PERCENTAGE_MINIMUM) {
```
lurch: Only a minor nit, but I think comparing the minimum value first reads more naturally? i.e.
```js
if (percentage < PERCENTAGE_MINIMUM || percentage > PERCENTAGE_MAXIMUM) {
```

lib/gui/models/store.js
lurch: >  We could reduce the chances of typos by making constants but I wonder if it's too much complexity

\*shrug\*

> trying to mitigate the fact that JavaScript is not a strongly typed language

Even a strongly-typed language doesn't prevent making typos in strings-used-as-constants ;-P

> Maybe moving to TypeScript is the solution?

I'm still getting to grips with ES, I'm in no hurry to switch to TS ;-D

lib/gui/os/dialog/services/dialog.js
lurch: Given that we've _only_ got two buttons here, could you just do:
```js
const BUTTON_CONFIRMATION_INDEX = 0;
const BUTTON_REJECTION_INDEX = 1;
```
rather than using `indexOf`? As you say below, this would currently go wrong if `confirmationLabel` and `rejectionLabel` were the same.

lib/gui/pages/main/controllers/flash.js
lurch: `lib/shared/constants.js` sounds good :+1:

lib/gui/app.js
```diff
@@ -164,7 +167,7 @@ app.run(($window, WarningModalService, ErrorService, FlashStateModel, OSDialogSe
     // Don't open any more popups
     popupExists = true;
 
-    return OSDialogService.showWarning({
+    OSDialogService.showWarning({
       confirmationLabel: 'Yes, quit',
       rejectionLabel: 'Cancel',
```
lurch: Now that `showWarning` applies default button labels, this line could theoretically be removed ;-)
(I'm not suggesting we do though, because it probably makes the code more unreadable)

PR #1151, opened 2017-03-01
chore: fix JSDoc warnings

jviotti:
- The type in `@type` should be surrounded by curly braces
- We should use `@type` in some cases where we used `@property`
- `@module` descriptions should go above such tag
- Namespaces should not declare their access type

### DIFF
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -26,7 +26,7 @@ const packageJSON = require('../../package.json');
 
 /**
  * @summary Parsed CLI options and arguments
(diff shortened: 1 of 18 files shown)


jviotti:
Ah, turns out that module descriptions should be *before* `@module`. I'll update the PR.

lurch:
Is this a continuation of the other recent 'improved linting' PRs, or is this the result of some other linting process? (and if the latter, does it / should it get run with the other linting tests?)

Issue #1098, opened 2017-02-12
Confusing ENOMEM error when spawning the GNU/Linux drive detection script


keunes:
- **Etcher version:** 1.0.0-beta.18
- **Operating system and architecture:** Ubuntu 14.04

Running Etcher to get Ubuntu 16.04 on a 4GB USB key I get the following error:

```
Error: spawn ENOMEM
    at exports._errnoException (util.js:1026:11)
    at ChildProcess.spawn (internal/child_process.js:313:11)
    at exports.spawn (child_process.js:392:9)
    at Object.exports.execFile (child_process.js:155:15)
    at Object.module.(anonymous function) [as execFile] (ELECTRON_ASAR.js:224:18)
    at Object.exports.run (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/drivelist/lib/scripts.js:84:16)
    at Object.exports.list (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/drivelist/lib/drivelist.js:56:11)
    at Object.tryCatcher (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/bluebird/js/release/util.js:16:23)
    at Object.ret [as listAsync] (eval at makeNodePromisifiedEval (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/bluebird/js/release/promisify.js:184:12), <anonymous>:13:39)
    at isElevated.then.then (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/lib/cli/etcher.js:70:20)
    at tryCatcher (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/bluebird/js/release/promise.js:504:31)
    at Promise._settlePromise (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/bluebird/js/release/promise.js:561:18)
    at Promise._settlePromise0 (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/bluebird/js/release/promise.js:606:10)
    at Promise._settlePromises (/tmp/.mount_ZZ2t46/usr/bin/resources/app.asar/node_modules/bluebird/js/release/promise.js:685:18)
[...]

PR #1154, opened 2017-03-02
refactor: unify error related tasks

jviotti:
The current error handling logic is a mess. We have code that tries to
fetch information about errors in different places throughout the
application, and its incredibly hard to ensure certain types of error
get decent human friendly error messages.

This commit groups, improves, and tests all error related functions in
`lib/shared/errors.js`.

Here's a summary of the changes, in more detail:

- Move the `HUMAN_FRIENDLY` object to `shared/errors.js`
- Extend `HUMAN_FRIENDLY` with error descriptions
- Add `ENOMEM` to `shared/errors.js`
- Group CLI and `OSDialogService` mechanisms for getting an error title
  and an error description
- Move error serialisation routines from `robot` to `shared/errors.js`
- Create and use `createError()` and `createUserError()` utility
  functions
- Add user friendly descriptions to many errors
- Don't report user errors to TrackJS

Fixes: https://github.com/resin-io/etcher/issues/1098
Change-Type: minor
Changelog-Entry: Make errors more user friendly throughout the application.


lib/shared/errors.js
```diff
@@ -0,0 +1,366 @@
...
+      return `No such file or directory: ${error.path}`;
+    },
+
+    /**
+     * @property {Function} description
+     * @memberof ENOENT
+     * @param {Error} error - error object
+     * @returns {String} message
+     */
+    description: _.constant('The file you\'re trying to access doesn\'t exist')
```
jviotti: @jhermsmeier Some error messages depend on the error content. Check `ENOENT` a bit above for an example. The purpose of making them functions is that you can interpolate the error contents if needed, plus create if conditionals based on the environment, etc, which might be useful (we used to use that for the Resin CLI for example).

I guess we could make errors that don't need such feature strings, and check the type in the helper function. What do you think?

jviotti:
@jhermsmeier Any updates on this one?

lib/shared/errors.js
jhermsmeier: Sorry for the belated response – I was sure I commented last week, but must have screwed it up or forgotten.

> I guess we could make errors that don't need such feature strings, and check the type in the helper function. What do you think?

I think we can keep it as it is, but to cut down on the verbosity of this, what would you think about adding a class / function which handles the title / description setting?

```js
class ErrorDetails {
  constructor( options ) {
    
    /**
     * @property {Function} title
     * @param {Error} error - error object
     * @returns {String} message
     */
    this.title = typeof options.title !== 'function'
      ? _.constant( options.title )
      : options.title
    
    /**
     * @property {Function} description
     * @param {Error} error - error object
     * @returns {String} message
     */
    this.description = typeof options.description !== 'function'
      ? _.constant( options.description )
      : options.description
    
  }
}
```

Which would reduce declaring errors to:

```js
ENOENT: new ErrorDetails({
  title( error ) {
    return `No such file or directory: ${error.path}`
  },
  description: 'The file you\'re trying to access doesn\'t exist'
})
```

lurch:
Only a minor niggle, but I wouldn't classify this as `Fixes: #1098` as it only improves the error-reporting, it doesn't reduce the memory-usage of Etcher ;)

lib/shared/robot/index.js
```diff
@@ -128,12 +125,7 @@ exports.buildErrorMessage = (error) => {
     error = new Error(error);
```
lurch: `stacktrace` is still present in the code-comments above here.

lurch:
With all refactoring done here, I wonder what the 'overlap' now is between `shared/errors.js` and `shared/messages.js`? Should they be merged, or left separate? Are there any new strings that ought to be added to `shared/messages.js`?

lib/shared/errors.js
```diff
@@ -0,0 +1,366 @@
...
+     */
+    title: _.constant('You\'re not authorized to perform this operation'),
+
+    /**
+     * @property {Function} description
+     * @memberof EPERM
+     * @param {Error} error - error object
+     * @returns {String} message
+     */
+    description: _.constant('Please ensure you have to necessary permissions for this task')
```
jviotti: Oh, good catch. Fixed!

lib/shared/errors.js
```diff
@@ -0,0 +1,366 @@
...
+     */
+    title: _.constant('You don\'t have access to this resource'),
+
+    /**
+     * @property {Function} description
+     * @memberof EACCES
+     * @param {Error} error - error object
+     * @returns {String} message
+     */
+    description: _.constant('Please ensure you have to necessary permissions for this task')
```
jviotti: Done!

lib/shared/errors.js
```diff
@@ -0,0 +1,366 @@
...
+ *
+ * @param {String} string - string
+ * @returns {Boolean} whether the string is blank
+ *
+ * @example
+ * if (isBlank('   ')) {
+ *   console.log('The string is blank');
+ * }
+ */
+const isBlank = _.flow([ _.trim, _.isEmpty ]);
```
jviotti: The thing is that this code will run on error objects that we don't control, and they could really be anything, so I guess it's better to be defensive.

lib/shared/errors.js
```diff
@@ -0,0 +1,366 @@
...
+
+/**
+ * @summary Get the title of an error
+ * @function
+ * @public
+ *
+ * @description
+ * Try to get as most information as possible about the error
```
jviotti: Fixed!

lib/shared/errors.js
```diff
@@ -0,0 +1,366 @@
...
+ * @param {String} [description] - error description
+ * @param {Object} [options] - options
+ * @param {Boolean} [options.report] - report error
+ * @returns {Error} error
+ *
+ * @example
+ * const error = errors.createError('Foo', 'Bar');
+ * throw error;
+ */
+exports.createError = (title, description, options = {}) => {
```
jviotti: The description is an optional argument (see the JSDoc annotation above), and it feels that its overkill for some cases (e.g: where you just want to throw a real Etcher error).

tests/cli/errors.spec.js
```diff
@@ -1,132 +0,0 @@
...
-
-'use strict';
-
-const m = require('mochainon');
-const _ = require('lodash');
-const errors = require('../../lib/cli/errors');
-
-describe('CLI: Errors', function() {
-
-  describe('.HUMAN_FRIENDLY', function() {
```
jviotti: Oh, damn. I'll bring them back :)

lib/child-writer/writer-proxy.js
```diff
@@ -126,7 +127,7 @@ return isElevated().then((elevated) => {
       name: packageJSON.displayName
     }).then((stdout, stderr) => {
       if (!_.isEmpty(stderr)) {
-        throw new Error(stderr);
+        throw errors.createError(stderr);
```
jviotti: Not really sure. The purpose of this block is to bridge errors coming from the child script, which should already contain enough context.

lib/shared/robot/index.js
jviotti: Oops, fixed!

lib/gui/os/dialog/services/dialog.js
```diff
@@ -149,31 +150,9 @@ module.exports = function($q, SupportedFormatsModel) {
    * OSDialogService.showError('Foo Bar', 'An error happened!');
    */
   this.showError = (error, description) => {
```
jviotti: I think it makes a lot of sense. I'll update the PR.

lib/cli/options.js
```diff
@@ -69,7 +70,7 @@ module.exports = yargs
       robot.printError(error || message);
     } else {
       yargs.showHelp();
-      errors.print(error || message);
+      utils.printError(error || message);
```
jviotti: Makes a lot of sense as well. I updated the `utils` and the `robot` sides of this :)

jviotti:
> Only a minor niggle, but I wouldn't classify this as Fixes: #1098 as it only improves the error-reporting, it doesn't reduce the memory-usage of Etcher ;)

Not sure. I feel that the referenced ticket concluded that improving the error messages should be enough for now. Maybe we can create a new one to track reducing memory consumption, and have this PR close the referenced one?

> With all refactoring done here, I wonder what the 'overlap' now is between shared/errors.js and shared/messages.js? Should they be merged, or left separate? Are there any new strings that ought to be added to shared/messages.js?

Yeah, that needs to be unified as well, but I'll probably to do in another PR, once this is merged.

lurch:
> I feel that the referenced ticket concluded that improving the error messages should be enough for now.

I've edited the title of #1098 so I'm now happy for this PR to 'fix' that issue :-)

lib/shared/errors.js
```diff
@@ -0,0 +1,367 @@
...
+     */
+    title: _.constant('Your system ran out of memory'),
+
+    /**
+     * @property {Function} description
+     * @memberof ENOMEM
+     * @param {Error} error - error object
+     * @returns {String} message
+     */
+    description: _.constant('Make sure your system has enough available memory for this task')
```
lurch: Should this start with 'Please ...' to match the other `description`s?

tests/shared/robot.spec.js
```diff
@@ -131,16 +131,16 @@ describe('Shared: Robot', function() {
         data: {
           message: 'foo',
           code: 'MYERROR',
-          stacktrace: error.stack
+          stack: error.stack
         }
       });
     });
 
     it('should handle a string error', function() {
```
lurch: This test is causing the CI tests to fail (it should be removed, as `robot.buildErrorMessage` now only accepts Error objects, not Strings.

lib/gui/os/dialog/services/dialog.js
```diff
@@ -146,44 +147,15 @@ module.exports = function($q, SupportedFormatsModel) {
...
-    //   Underlying error:
-    //     Error processing argument at index 0, conversion failure
-    //
-    // This can be thrown if for some reason, either `title` or `message`
-    // are not strings.
-    electron.remote.dialog.showErrorBox(title.toString(), message.toString());
+  this.showError = (error) => {
+    const title = errors.getTitle(error);
+    const message = errors.getDescription(error);
+    electron.remote.dialog.showErrorBox(title, message.toString());
```
jviotti: If the description is an empty string, then no description is shown.

lib/shared/errors.js
jviotti: Makes sense, fixed!

lib/shared/errors.js
```diff
@@ -0,0 +1,367 @@
...
+  if (!_.isNil(options.report) && !options.report) {
+    error.report = false;
+  }
+
+  return error;
+};
+
+/**
+ * @summary Create a user error
+ * @function
```
jviotti: Done!

lib/shared/robot/index.js
```diff
@@ -119,19 +116,9 @@ exports.parseMessage = (string) => {
  *
  * console.log(error.data.message);
  * > 'foo'
- *
- * error.data.stacktrace === error.stack;
- * > true
  */
 exports.buildErrorMessage = (error) => {
```
jviotti: I suggest keeping it separate for testability purposes. `.printError` does a `console.log`, which will be hard to test.

tests/shared/robot.spec.js
jviotti: Fixed! I made the function handle strings anyway just to be on the safe side, since some node libraries throw strings (like UI Bootstrap). If this happens on the CLI writer, the current approach will swallow the error.

lib/shared/errors.js
jviotti: Done! I created a helper function instead of a full class though, given we were only using the constructor.

lib/shared/errors.js
```diff
@@ -0,0 +1,343 @@
...
+  EACCES: createErrorDetails({
+    title: 'You don\'t have access to this resource',
+    description: 'Please ensure you have necessary permissions to access this resource'
+  }),
+
+  /**
+   * @namespace ENOMEM
+   * @memberof HUMAN_FRIENDLY
+   */
+
```
jhermsmeier: Excess newline here

lib/shared/errors.js
jhermsmeier: Nice! Yeah, I was just thinking a class might have been optimal to re-use the hidden class in v8 (to avoid thrashing the heap with new objects), but that's probably over-optimising it ;)

lib/shared/errors.js
jviotti: Nice catch! Fixed!

tests/shared/robot.spec.js
lurch: ...except `robot.buildErrorMessage` is _only_ called from `robot.printError`, and you already changed `robot.printError` to only accepting `Error` objects?

Oh, you've changed `robot.printError` back to `{(Error|String)}` :-/
Looks like the only places `robot.printError` is used are https://github.com/resin-io/etcher/blob/master/lib/cli/options.js#L96 and https://github.com/resin-io/etcher/blob/master/lib/cli/etcher.js#L124 so I wonder if it'd be worth keeping `robot.printError` as only accepting `Error`, and then using `errors.createError` in https://github.com/resin-io/etcher/blob/master/lib/cli/etcher.js#L124 to create a proper Error object if we only get a string back?

Or am I just overcomplicating things? :-S

lurch:
Looks like the new [angular-mocks](https://www.npmjs.com/package/angular-mocks) is now breaking the CI tests! :-/
I dunno if that's the fault of `angular-mocks` though, or if our code is doing something it shouldn't be?

jviotti:
Hm, it looks like its a dependency indeed. The shrinkwrap file doesn't lock development dependencies, so either angular-mocks or electron-mocha is causing this.

jviotti:
Neither module was upgraded. `package.json` is already declaring the latest versions. On top of that, the issue only seems to happen in GNU/Linux and Windows, and I can't reproduce on OS X.

jviotti:
I eventually got to the bottom of it. See https://github.com/resin-io/etcher/pull/1168

tests/shared/robot.spec.js
jviotti: > Oh, you've changed robot.printError back to {(Error|String)} :-/

I don't I touched that one at all, but I edited similar signatures all over the place, so I don't remember very much anymore :P

>  I wonder if it'd be worth keeping robot.printError as only accepting Error, and then using errors.createError in https://github.com/resin-io/etcher/blob/master/lib/cli/etcher.js#L124 to create a proper Error object if we only get a string back?

I guess we could, however I'd like to avoid `_.isError(error)` checks all over the place by making the function a bit smarter.

I edited the signatures to say we only accept Errors (to kinda enforce that), but the function will still handle other values, just in case.

lurch: > I'd like to avoid ... checks all over the place

It's not "all over the place" - you'll see I pointed out above that `robot.printError` is only used in two places.

> I edited the signatures to say we only accept Errors (to kinda enforce that), but the function will still handle other values, just in case.

Fair enough. But if `robot.printError` accepts a string "just in case", then you ought to do the same for `utils.printError` because it appears just a couple of lines below `robot.printError` in `lib/cli/etcher.js` (with the same `error` variable being passed to both).

jviotti: > Fair enough. But if robot.printError accepts a string "just in case", then you ought to do the same for utils.printError because it appears just a couple of lines below robot.printError in lib/cli/etcher.js (with the same error variable being passed to both).

`utils.printError` does handle strings already, since it re-uses the same `lib/shared/errors.js` code than `robot.printError`.

lurch: Oh right, I've just double-checked and I can now see `.getTitle` and `.getDescription` do support being passed `String`s (with `.getDescription` simply returning `''` if passed a `String`).

lurch: (`robot.printError` by comparison doesn't use `.getTitle` and `.getDescription` - it just bubbles down to `errors.toJSON`)

PR #1161, opened 2017-03-07
feat(GUI): don't log absolute paths in Mixpanel

stefan-mihaila:
The image path contains user information that should not be logged
in Mixpanel. Instead, just log the image filename.

Change-Type: minor
Changelog-Entry: Don't include user paths in usage reports

### DIFF
--- a/lib/gui/modules/analytics.js
+++ b/lib/gui/modules/analytics.js
@@ -141,14 +141,15 @@ analytics.service('AnalyticsService', function($log, $window, $mixpanel, Setting
    * });
    */
   this.logEvent = (message, data) => {
-    const flatStartCaseData = utils.makeFlatStartCaseObject(data);
+    const debugData = utils.hideAbsolutePathsInObject(utils.makeFlatStartCaseObject(data));
+
     if (SettingsModel.get('errorReporting') && isRunningInAsar()) {
-      $mixpanel.track(message, flatStartCaseData);
+      $mixpanel.track(message, debugData);
     }
 
     const debugMessage = _.attempt(() => {
-      if (flatStartCaseData) {
-        return `${message} (${JSON.stringify(flatStartCaseData)})`;
+      if (debugData) {
+        return `${message} (${JSON.stringify(debugData)})`;
       }
 
       return message;
--- a/lib/shared/utils.js
+++ b/lib/shared/utils.js
@@ -17,8 +17,10 @@
 'use strict';
 
 const _ = require('lodash');
+_.mixin(require('lodash-deep'));
 const flatten = require('flat').flatten;
 const deepMapKeys = require('deep-map-keys');
+const path = require('path');
 
 /**
  * @summary Create a flattened copy of the object with all keys transformed in start case
@@ -81,3 +83,53 @@ exports.makeFlatStartCaseObject = (object) => {
   });
 
 };
+
+/**
+ * @summary Create an object clone with all absolute paths replaced with the path basename
+ * @function
+ * @public
+ *
+ * @param {Object} object - original object
+ * @returns {Object} transformed object
+ *
+ * @example
+ * const anonymized = utils.hideAbsolutePathsInObject({
+ *   path1: '/home/john/rpi.img',
+ *   simpleProperty: null,
+ *   nested: {
+ *     path2: '/home/john/another-image.img',
+ *     path3: 'yet-another-image.img',
+ *     otherProperty: false
+ *   }
+ * });
+ *
+ * console.log(anonymized);
+ * > {
+ * >   path1: 'rpi.img',
+ * >   simpleProperty: null,
+ * >   nested: {
+ * >     path2: 'another-image.img',
+ * >     path3: 'yet-another-image.img',
+ * >     otherProperty: false
+ * >   }
+ * > }
+ */
+exports.hideAbsolutePathsInObject = (object) => {
+  return _.deepMapValues(object, (value) => {
+    if (!_.isString(value)) {
+      return value;
+    }
+
+    // Don't alter disk devices, even though they appear as full paths
+    if (_.some([
+      _.startsWith(value, '/dev/'),
+      _.startsWith(value, '\\\\.\\')
+    ])) {
+      return value;
+    }
+
+    return path.isAbsolute(value) ? path.basename(value) : value;
+  });
+
+};
+
--- a/tests/shared/utils.spec.js
+++ b/tests/shared/utils.spec.js
@@ -18,6 +18,7 @@
 
 const m = require('mochainon');
 const utils = require('../../lib/shared/utils');
+const path = require('path');
 
 describe('Shared: Utils', function() {
 
@@ -120,6 +121,214 @@ describe('Shared: Utils', function() {
 
     it('should leave nested arrays nested', function() {
       m.chai.expect(utils.makeFlatStartCaseObject([ 1, 2, [ 3, 4 ] ])).to.deep.equal([ 1, 2, [ 3, 4 ] ]);
+    });
+
+  });
+
+  describe('hideAbsolutePathsInObject()', function() {
+
+    it('should return undefined if given undefined', function() {
+      m.chai.expect(utils.hideAbsolutePathsInObject(undefined)).to.be.undefined;
+    });
+
+    it('should return null if given null', function() {
+      m.chai.expect(utils.hideAbsolutePathsInObject(null)).to.be.null;
+    });
+
+    it('should return a clone of the object if there are no paths in the object', function() {
+      const object = {
+        numberProperty: 1,
+        nested: {
+          otherProperty: 'value'
+        }
+      };
+      m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.not.equal(object);
+      m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal(object);
+    });
+
+    describe('given UNIX paths', function() {
+
+      beforeEach(function() {
+        this.isAbsolute = path.isAbsolute;
+        this.basename = path.basename;
+        path.isAbsolute = path.posix.isAbsolute;
+        path.basename = path.posix.basename;
+      });
+
+      afterEach(function() {
+        path.isAbsolute = this.isAbsolute;
+        path.basename = this.basename;
+      });
+
+      it('should replace absolute paths with the basename', function() {
+        const object = {
+          prop1: 'some value',
+          prop2: '/home/john/rpi.img'
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          prop1: 'some value',
+          prop2: 'rpi.img'
+        });
+      });
+
+      it('should replace nested absolute paths with the basename', function() {
+        const object = {
+          nested: {
+            path: '/home/john/rpi.img'
+          }
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          nested: {
+            path: 'rpi.img'
+          }
+        });
+      });
+
+      it('should not alter /dev/sdb', function() {
+        const object = {
+          nested: {
+            path: '/dev/sdb'
+          }
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          nested: {
+            path: '/dev/sdb'
+          }
+        });
+      });
+
+      it('should not alter relative paths', function() {
+        const object = {
+          path: 'foo/bar'
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          path: 'foo/bar'
+        });
+      });
+
+      it('should handle arrays', function() {
+        m.chai.expect(utils.hideAbsolutePathsInObject({
+          foo: 'foo',
+          bar: [
+            {
+              path: '/foo/bar/baz'
+            },
+            {
+              path: '/foo/bar/baz'
+            },
+            {
+              path: '/foo/bar/baz'
+            }
+          ]
+        })).to.deep.equal({
+          foo: 'foo',
+          bar: [
+            {
+              path: 'baz'
+            },
+            {
+              path: 'baz'
+            },
+            {
+              path: 'baz'
+            }
+          ]
+        });
+      });
+
+    });
+
+    describe('given Windows paths', function() {
+
+      beforeEach(function() {
+        this.isAbsolute = path.isAbsolute;
+        this.basename = path.basename;
+        path.isAbsolute = path.win32.isAbsolute;
+        path.basename = path.win32.basename;
+      });
+
+      afterEach(function() {
+        path.isAbsolute = this.isAbsolute;
+        path.basename = this.basename;
+      });
+
+      it('should replace absolute paths with the basename', function() {
+        const object = {
+          prop1: 'some value',
+          prop2: 'C:\\Users\\John\\rpi.img'
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          prop1: 'some value',
+          prop2: 'rpi.img'
+        });
+
+      });
+
+      it('should replace nested absolute paths with the basename', function() {
+        const object = {
+          nested: {
+            path: 'C:\\Users\\John\\rpi.img'
+          }
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          nested: {
+            path: 'rpi.img'
+          }
+        });
+      });
+
+      it('should not alter \\\\.\\PHYSICALDRIVE1', function() {
+        const object = {
+          nested: {
+            path: '\\\\.\\PHYSICALDRIVE1'
+          }
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          nested: {
+            path: '\\\\.\\PHYSICALDRIVE1'
+          }
+        });
+      });
+
+      it('should not alter relative paths', function() {
+        const object = {
+          path: 'foo\\bar'
+        };
+
+        m.chai.expect(utils.hideAbsolutePathsInObject(object)).to.deep.equal({
+          path: 'foo\\bar'
+        });
+      });
+
+      it('should handle arrays', function() {
+        m.chai.expect(utils.hideAbsolutePathsInObject({
+          foo: 'foo',
+          bar: [ {
+            path: 'C:\\foo\\bar\\baz'
+          }, {
+            path: 'C:\\foo\\bar\\baz'
+          }, {
+            path: 'C:\\foo\\bar\\baz'
+          } ]
+        })).to.deep.equal({
+          foo: 'foo',
+          bar: [ {
+            path: 'baz'
+          }, {
+            path: 'baz'
+          }, {
+            path: 'baz'
+          } ]
+        });
+      });
+
     });
 
   });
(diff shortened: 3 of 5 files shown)


stefan-mihaila:
I opened this against the `improved-analytics` branch until we merge that one so the diff looks right.

lib/shared/utils.js
```diff
@@ -81,3 +82,24 @@ exports.makeFlatStartCaseObject = (object) => {
...
+ * @example
+ * hideImagePath({
+ *   size: 1000000000,
+ *   path: '/home/john/rpi.img'
+ * })
+ */
+exports.hideImagePath = (image) => {
+  const imageClone = _.clone(image);
+  Reflect.deleteProperty(imageClone, 'path');
+  imageClone.filename = path.basename(image.path);
```
jviotti: I think I'd prefer to keep the property name called `path`

lib/shared/utils.js
```diff
@@ -81,3 +82,24 @@ exports.makeFlatStartCaseObject = (object) => {
...
+/**
+ * @summary Create a clone of the image, removing .path and adding .filename
+ * @function
+ * @public
+ *
+ * @param {Object} image - image
+ * @returns {Object} image with filename instead of path
+ *
+ * @example
+ * hideImagePath({
```
jviotti: Can we add the `utils` prefix here?

lib/shared/utils.js
```diff
@@ -81,3 +82,24 @@ exports.makeFlatStartCaseObject = (object) => {
...
+ *   size: 1000000000,
+ *   path: '/home/john/rpi.img'
+ * })
```
jviotti: Let's add a semicolon here, so that we keep the examples code style consistent with the rest of the code.

jviotti: Maybe we can `console.log` the path property in the example to showcase that the directory name is removed?

jviotti:
@stefan-mihaila Can we add a WIP prefix to the PR title, until the dependent branch gets merged?

jviotti:
Keep in mind that the image path might still be printed in less obvious places, like in the child writer: https://github.com/resin-io/etcher/blob/master/lib/child-writer/writer-proxy.js#L93.

I guess you can go through the Etcher flow with the console opened and keep an eye on the output.

In any case, this PR already says "Mixpanel" in the title, so I guess a new one should be opened for that.

jviotti:
The main reason why we don't want to pass the full image path to Mixpanel is not because of the image itself, but because the path is absolute, and thus may reveal things like the username.

In order to protect ourselves from cases that contain absolute paths other than images, I suggest that we make the utility function more generic, and make it recursively traverse through every property of the passed object, check if the path is an absolute path with `path.isAbsolute`, and `path.basename` every occurrence.

Since the function is generic, it can now be moved to the analytics code, and in the near future, to your re-usable module.

stefan-mihaila:
> Keep in mind that the image path might still be printed in less obvious places, like in the child writer: https://github.com/resin-io/etcher/blob/master/lib/child-writer/writer-proxy.js#L93.

If I'm not missing something, information printed to the console don't go in Sentry / Mixpanel or other services at the moment, does it?

I assumed that the idea was to not collect sensitive information. Printing it to the console is an issue also?

stefan-mihaila:
updated PR; no need to use `improved-analytics` as base anymore

jviotti:
@stefan-mihaila The `console.log` output is piped to TrackJS atm.

jviotti:
Builds are failing due to an `angular-mocks` version incompatibility issue. See https://github.com/resin-io/etcher/pull/1168 for the fix.

lurch:
Would there be any way to refactor things so that the full-path gets displayed in `console.log` output, but only the truncated path gets logged to TrackJS?
I guess any "message objects" would need to store the paths separately, something like:
```js
{
    message: 'Couldn\'t copy "{srcpath}" to "{dstpath}"',
    paths: {
        srcpath: 'C:\Users\Andrew\Downloads\foo.img',
        dstpath: 'C:\Users\Andrew\test\bar.bin'
    }
}
```
and then console.log could display `Couldn't copy "C:\Users\Andrew\Downloads\foo.img" to "C:\Users\Andrew\test\bar.bin"` whereas TrackJS would only see `Couldn't copy "foo.img" to "bar.bin"`

That's possibly more trouble than it's worth though...

lib/gui/modules/analytics.js
```diff
@@ -39,6 +41,42 @@ const analytics = angular.module(MODULE_NAME, [
...
+ * >   path1: 'rpi.img',
+ * >   simpleProperty: null,
+ * >   nested: {
+ * >     path2: 'another-image.img',
+ * >     path3: 'yet-another-image.img',
+ * >     otherProperty: false
+ * >   }
+ * > }
+ */
+const hidePaths = (object) => {
```
lurch: Some unit-tests for this function would be great :-)

lib/gui/modules/analytics.js
```diff
@@ -39,6 +41,42 @@ const analytics = angular.module(MODULE_NAME, [
...
+ * @example
+ * const anonymized = hidePaths({
+ *   path1: '/home/john/rpi.img',
+ *   simpleProperty: null,
+ *   nested: {
+ *     path2: '/home/john/another-image.img',
+ *     path3: 'yet-another-image.img',
+ *     otherProperty: false
+ *   }
+ * });
```
lurch: With the code as it is currently, I guess it'd also 'anonymize' `/dev/sdb` -> `sdb` and `\\\\.\\PHYSICALDRIVE1` -> `PHYSICALDRIVE1` ?
(which of course isn't a problem, if we know to expect that when looking at the logs)

lib/gui/modules/analytics.js
lurch: (see https://github.com/resin-io/etcher/blob/master/tests/shared/drive-constraints.spec.js#L132 and https://github.com/resin-io/etcher/blob/master/tests/shared/drive-constraints.spec.js#L210 for examples of writing cross-platform path-manipulation tests)

jviotti:
We sadly have very little control over what TrackJS grabs. As soon as you import their module, it intercepts everything that is logged (stdout, stderr) and pipes it to its backend service.

Maybe there a way to do it in Sentry though. @stefan-mihaila ?

lib/gui/modules/analytics.js
jviotti: Ugh, good catch. I think we should provide a whitelist of base paths that are not anonymized. For example, we can whitelist everything that comes from `/dev`, etc. I'm not sure how `\\\\.\\PHYSICALDRIVE1` would react to this code, so better to unit test and make sure it doesn't get modified.

lib/gui/modules/analytics.js
```diff
@@ -39,6 +41,42 @@ const analytics = angular.module(MODULE_NAME, [
...
+ * >   simpleProperty: null,
+ * >   nested: {
+ * >     path2: 'another-image.img',
+ * >     path3: 'yet-another-image.img',
+ * >     otherProperty: false
+ * >   }
+ * > }
+ */
+const hidePaths = (object) => {
+  return _.deepMapValues(object, (value) => {
```
jviotti: Not sure if its worth or not, but you could implement this recursively to avoid the extra dependency. For example: if the value of a property is an object, you run `hidePath` on it again.

jviotti: BTW, can we also make sure this works on array of objects. For example:

```json
{
  "foo": "bar",
  "baz": [
    {
      "path": "/foo/bar/baz"
    },
    {
      "path": "/foo/bar/baz"
    },
    {
      "path": "/foo/bar/baz"
    }
  ]
}
```

stefan-mihaila:
I think there is a way to do it TrackJS too, by making use of the `onError` callback: http://docs.trackjs.com/tracker/configuration (search for `Custom onError() Callback` on the page).

The `payload` received by the callback has an array of 'console' objects, each object with its message, and they can be altered before sending them to TrackJS.

I'll also look for a way of doing it in Sentry.

jviotti:
The `onError` callback is called, well, on errors :P Looking at the same page, I found the following configuration options:

```
  console: {
    // By default TrackJS will watch all console activity and include that information in the Telemetry Timeline
    enabled: true,

    // Since TrackJS wraps the console, we can also prevent the messages from being output to the browser.
    // This is useful if you don't want your users seeing anything if they have the dev tools open.
    display: true,

    // By default any calls to console.error() will automatically trigger an error.
    // Set this to false if you don't want console.error() to trigger errors.
    error: true,

    // Whether to trigger an error on calls to `console.warn()`
    warn: false,

    // If you only want TrackJS to watch certain console functions, you may specify them here.
    watch: ["log","debug","info","warn","error"]

  },
```

I guess we can disable console support for certain console functions, like `console.info`, and use that to display sensitive information.

In any case, I think this is probably overkill. Let's normalize the absolute paths and see how we can improve it later on.

stefan-mihaila:
Aren't the `console.log`s being sent (along with AJAX GETs, clicks etc) to TrackJS, only when there is an error? Maybe it's not important for this task, but I'm looking at the TrackJS interface and can't seem to figure it out.

jviotti:
So TrackJS will send every single `console.log` to the service. When you as a developer want to inspect an error, you can see the "context" by checking what was printed before and after the error (you can go back and forth as much as you like).

stefan-mihaila:
Updated PR. I'll open a new one with a `hideObjectAbsolutePaths` function to hide paths in `console.log` strings.

lib/shared/utils.js
```diff
@@ -81,3 +84,53 @@ exports.makeFlatStartCaseObject = (object) => {
...
+ * >   }
+ * > }
+ */
+exports.hideObjectAbsolutePaths = (object) => {
+  return _.deepMapValues(object, (value) => {
+    if (!_.isString(value)) {
+      return value;
+    }
+
+    if (os.platform() !== 'win32' && _.startsWith(value, '/dev/')) {
```
jviotti: I think we should add some comments above these checks to describe why we're ignoring them.

tests/shared/utils.spec.js
```diff
@@ -124,4 +127,191 @@ describe('Shared: Utils', function() {
 
   });
 
+  describe('hideObjectAbsolutePaths', function() {
```
jviotti: This should be `.hideObjectAbsolutePaths()`, for consistency purposes.

tests/shared/utils.spec.js
```diff
@@ -124,4 +127,191 @@ describe('Shared: Utils', function() {
...
+        const object = {
+          prop1: 'some value',
+          prop2: '/home/john/rpi.img'
+        };
+
+        m.chai.expect(utils.hideObjectAbsolutePaths(object)).to.deep.equal({
+          prop1: 'some value',
+          prop2: 'rpi.img'
+        });
+
```
jviotti: This line should be removed.

tests/shared/utils.spec.js
```diff
@@ -124,4 +127,191 @@ describe('Shared: Utils', function() {
...
+          nested: {
+            path: '/dev/sdb'
+          }
+        });
+      });
+
+      it('should handle arrays', function() {
+        m.chai.expect(utils.hideObjectAbsolutePaths({
+          foo: 'foo',
+          bar: [ {
```
jviotti: Can we put the curly braces in separate lines, for consistency purposes? For example:

```js
foo: 'foo',
bar: [
  {
    path: '/foo/bar/baz'
  },
  {
    path: '/foo/bar/baz'
  },
  {
    path: '/foo/bar/baz'
  }
]
```

tests/shared/utils.spec.js
```diff
@@ -124,4 +127,191 @@ describe('Shared: Utils', function() {
...
+            path: '/foo/bar/baz'
+          }, {
+            path: '/foo/bar/baz'
+          }, {
+            path: '/foo/bar/baz'
+          } ]
+        })).to.deep.equal({
+          foo: 'foo',
+          bar: [ {
```
jviotti: Let's put curly braces in different lines here as well.

tests/shared/utils.spec.js
```diff
@@ -124,4 +127,191 @@ describe('Shared: Utils', function() {
...
+        numberProperty: 1,
+        nested: {
+          otherProperty: 'value'
+        }
+      };
+      m.chai.expect(utils.hideObjectAbsolutePaths(object)).to.not.equal(object);
+      m.chai.expect(utils.hideObjectAbsolutePaths(object)).to.deep.equal(object);
+    });
+
+    describe('given UNIX paths', function() {
```
jviotti: Let's test that relative paths are not modified.

lib/shared/utils.js
```diff
@@ -81,3 +84,53 @@ exports.makeFlatStartCaseObject = (object) => {
   });
 
 };
+
+/**
+ * @summary Create an object clone with all absolute paths replaced with the path basename
+ * @function
+ * @private
```
jviotti: This function is `@public` now.

lib/gui/modules/analytics.js
```diff
@@ -142,13 +142,15 @@ analytics.service('AnalyticsService', function($log, $window, $mixpanel, Setting
    */
   this.logEvent = (message, data) => {
     const flatStartCaseData = utils.makeFlatStartCaseObject(data);
+    const anonymizedFlatStartCaseData = utils.hideObjectAbsolutePaths(flatStartCaseData);
```
lurch: I wonder if `anonymizedFlatStartCaseData` could be given a nicer name? It's a bit of a mouthful! ;-)

lib/shared/utils.js
lurch: Do we even need to bother checking `os.platform()`? I can't imagine any scenarios where paths on Windows start with `/dev/` , or paths on Linux start with `\\.\` ?

lib/shared/utils.js
```diff
@@ -81,3 +84,53 @@ exports.makeFlatStartCaseObject = (object) => {
...
+ *
+ * @param {Object} object - original object
+ * @returns {Object} transformed object
+ *
+ * @example
+ * const anonymized = utils.hidePaths({
```
lurch: Should this be updated to `utils.hideObjectAbsolutePaths` ?

tests/shared/utils.spec.js
```diff
@@ -124,4 +127,191 @@ describe('Shared: Utils', function() {
...
+        };
+
+        m.chai.expect(utils.hideObjectAbsolutePaths(object)).to.deep.equal({
+          nested: {
+            path: 'rpi.img'
+          }
+        });
+      });
+
+      it('should not alter \\\\.\\PHYSICALDRIVE1', function() {
```
lurch: Should we also add tests for UNC-style filepaths? https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx

lib/shared/utils.js
```diff
@@ -81,3 +84,53 @@ exports.makeFlatStartCaseObject = (object) => {
...
+ * >   path1: 'rpi.img',
+ * >   simpleProperty: null,
+ * >   nested: {
+ * >     path2: 'another-image.img',
+ * >     path3: 'yet-another-image.img',
+ * >     otherProperty: false
+ * >   }
+ * > }
+ */
+exports.hideObjectAbsolutePaths = (object) => {
```
lurch: Only a minor niggle, but I'm not sure I like this function name. Could the word `Object` be removed? Alternatively, does `hideAbsolutePathsInObject` sound better?

lib/shared/utils.js
jviotti: Great catch!

lib/shared/utils.js
jviotti: Yeah, I think we are safe to drop them. This means we can keep a nice array of the paths we check, potentially re-using some logic. For example:

```js
_.each([ 
    '/dev/',
    '\\\\.\\'
], (whitelistedPath) => {
    if (_.startsWith(value, whitelistedPath)) {
        return;
    }
});
```

jviotti: Ah, wait. The `return` in there will exit from the `_.each` callback, but not from the outer flow :)

lib/gui/modules/analytics.js
stefan-mihaila: What about `const debugData = utils.hideAbsolutePathsInObject(utils.makeFlatStartCaseObject(data));`? It loses information fromthe naming, but is not a mouthful anymore.

tests/shared/utils.spec.js
stefan-mihaila: I'm not really familiar with UNC-style filepaths. Should we leave them as they are? Or they might actually contain sensitive information that we should hide? Or are they just a weird / rare case that we shouldn't care about?

lib/gui/modules/analytics.js
lurch: IMHO that's fine, given how short the enclosing function is.

tests/shared/utils.spec.js
lurch: Sorry, I meant maybe we should check that UNC paths also get **do** anonymized, and that the special-case code to not anonymize e.g. `\\.\PHYSICALDRIVE1` doesn't accidentally prevent e.g. `\\server\share\path\file` or `\\?\D:\very long path\file` from being anonymized?

> Or are they just a weird / rare case that we shouldn't care about?

Yeah, you might be right ;-)

lurch:
BTW, not sure if it falls within the remit of this PR, but looking at `lib/gui/modules/analytics.js` I've just noticed that we have `logDebug` which is documented as:
```
   * @summary Log a debug message
   * @description This function sends the debug message to TrackJS only.
   * @param {String} message - message
```
and `logEvent` which is documented as:
```
   * @summary Log an event
   * @description This function sends the debug message to TrackJS and Mixpanel.
   * @param {String} message - message
   * @param {Object} [data] - event data
```
and `logEvent` also (after doing some other stuff) calls `logDebug` (after joining `message` and `data` together into a single string).


Would it be more consistent to also change the signature of `logDebug` to:
```
   * @param {String} message - message
   * @param {Object} [data] - event data
```
?
This wouldn't break backwards-compatibility (as it is, `logDebug` is only used in two places in the entire codebase anyway), it would give `logDebug` and `logEvent` consistent signatures, and it'd allow us to simplify `logEvent` and move the creation of the `debugMessage` string entirely within `logDebug`.
?

lib/shared/utils.js
```diff
@@ -81,3 +83,53 @@ exports.makeFlatStartCaseObject = (object) => {
...
+  return _.deepMapValues(object, (value) => {
+    if (!_.isString(value)) {
+      return value;
+    }
+
+    // Don't alter disk devices, even though they appear as full paths
+    if (_.some([
+      _.startsWith(value, '/dev/'),
+      _.startsWith(value, '\\\\.\\')
+    ])) {
```
lurch: Is there any benefit to using `_.some` as opposed to using just a regular `||` ?

jviotti: Not really, but I don't have any strong opinions. Just to brainstorm, this could be also expressed as:

```js
_.some(_.map([
  '/dev/',
  '\\\\.\\'
], _.partial(_.startsWith, value))) {
    ...
}
```

Although its probably not so readable :P

tests/shared/utils.spec.js
jviotti: I think we can ignore them for now, given we're not using them anywhere.

jviotti:
@lurch Yeah, it looks like a great idea :+1: @stefan-mihaila  Should we do it in another PR?

lib/shared/utils.js
lurch: IMVHO for something as simple as this, readability i.e.
```
if (_.startsWith(value, '/dev/') || _.startsWith(value, '\\\\.\\')) {
```
should take priority over "lodash cleverness". (If there were three or more conditionals being tested, I might agree that a 'clever' solution might be more appropriate)

Entirely up to @stefan-mihaila though :)

jviotti: I merged the PR already, but @stefan-mihaila is welcomed to send a quick PR for this.

PR #1163, opened 2017-03-08
style(GUI): reduce `eslint-disable no-var` scope

stefan-mihaila:
Change-Type: patch


jviotti:
Amazing catch! I'll approve.

The tests are failing due to some angular mocks weirdness, which seems to be biting  https://github.com/resin-io/etcher/pull/1154 as well

Issue #1155, opened 2017-03-03
Publish etcher on chocolatey.org


pierrebeaucamp:
- **Etcher version:** 1.0.0-beta.19
- **Operating system and architecture:** Windows

Similar to Homebrew, [Chocolatey](https://chocolatey.org/) is the most used package manager for windows applications.

WasabiFan:
Related to https://github.com/resin-io/etcher/issues/829

jviotti:
Sounds interesting, and not hard to do, even though it'd be nice to get a hand from people used to creating Chocolatey packaged. Here's what's needed based on https://chocolatey.org/docs/create-packages. Please correct me if I'm missing something:

- Prepare the package. We distribute self-contained zips for Windows, so it looks like we can re-use those:

> If you have a self-contained package, you can remove the automation scripts
entirely and just include the runtime executables, they will automatically get shimmed,
which puts them on the path.

- Create a Nuspec configuration file
- Integrate the building routines into our build system (the Makefile + utility scripts). This includes:
  - Building the package with `cpack`
  - Uploading the package using Chocolatey

It'd be great if someone more familiar with Chocolatey can validate that the ZIPs that we create are good enough for publishing, and help us out with the Nuspec file :)

pierrebeaucamp:
I never created a package on chocolatey before, but since I opened this bug, I'd be happy to look into this some more this weekend and open a PR.

jviotti:
Sounds good, let us know how it goes!

pierrebeaucamp:
Turns out that building a package is pretty simple. Please take a look at https://github.com/pierrebeaucamp/chocolatey to see the necessary steps. I'd be happy to transfer the ownership of that repo to your account. 

I tested the scripts on a 64-bit Win10 machine, but I don't see why they would fail on a 32 bit machine.

jviotti:
That is amazing @pierrebeaucamp ! Thanks a lot for the awesome work. Don't bother with repository ownership transfer. I'll integrate the install script and the nuspec in the main repo as part of the build system at the beginning of the week :)

pierrebeaucamp:
Glad I could help. Two things to keep in mind:

* The nuspec I included is pretty basic. You can see a full list of supported fields here: https://docs.microsoft.com/en-ca/nuget/schema/nuspec. In particular, I didn't include the 'tags' field which lets people find the package more easily. 

* I don't know if your build system is running on Windows, but I think you will only be able to build & publish packages from a windows powershell.

[...]

PR #1167, opened 2017-03-08
WIP: chore: chocolatey nupkg support

jviotti:
This commit adds a Makefile target called `electron-installer-nupkg`,
which generates a `.nupkg` file that references Etcher installers
uploaded to S3.

This largely based on @pierrebeaucamp's work, originally published at
https://github.com/pierrebeaucamp/chocolatey.

Fixes: https://github.com/resin-io/etcher/issues/1155


jviotti:
@lurch What are your thoughts on this? This is the first release target that depends on externally hosted files, which means we have to hardcode URLs and checksums, which doesn't feel right.

jviotti:
Closing this for the time being, since a lot has changed in the build system since I created this PR.

PR #1168, opened 2017-03-09
upgrade: `angular` and `angular-mocks` to v1.6.3

jviotti:
We recently started hitting the following errors in all our CI builds:

```
TypeError: angular.module(...).info is not a function
    at /etcher/node_modules/angular-mocks/angular-mock
```

The `.info()` function was released in AngularJS 1.6.3. We declare
`angular-mocks@^1.6.1` as a development dependency, which means that
once 1.6.3 was out, npm started resolving `angular-mocks@1.6.3`, however
that version is incompatible with the main Angular version we're
shrinkwrapping (1.6.1), and thus the error.

As a solution, I've upgraded both `angular` and `angular-mocks` to
1.6.3, but also locked down `angular-mocks`, since similar errors will
happen again unless we manually keep these versions in sync.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3,9 +3,9 @@
   "version": "1.0.0-beta.19",
   "dependencies": {
     "angular": {
-      "version": "1.6.1",
-      "from": "angular@>=1.6.1 <2.0.0",
-      "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.1.tgz"
+      "version": "1.6.3",
+      "from": "angular@1.6.3",
+      "resolved": "https://registry.npmjs.org/angular/-/angular-1.6.3.tgz"
     },
     "angular-if-state": {
       "version": "1.0.0",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "removedrive": "^1.1.1"
   },
   "dependencies": {
-    "angular": "^1.6.1",
+    "angular": "^1.6.3",
     "angular-if-state": "^1.0.0",
     "angular-middle-ellipses": "^1.0.0",
     "angular-moment": "^1.0.1",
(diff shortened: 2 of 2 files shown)


lurch:
It's obviously a bit of an ugly hack, but if the versions of `angular` and `angular-mocks` need to be kept in sync, could your clean-shrinkwrap script update the version of `angular-mocks` in `package.json` to match `angular` in `npm-shrinkwrap.json` ?

Alternatively, would it make sense to lock down the version of `angular` in addition to locking down the version of `angular-mocks`? (to ensure that they stay synchronised)

lurch:
...but obviously with the large backlog of broken CI builds, the sooner this gets merged, the better!
We can always come back to make it 'nicer' later ;-)

jviotti:
> Alternatively, would it make sense to lock down the version of angular in addition to locking down the version of angular-mocks? (to ensure that they stay synchronised)

Yeah, that makes a lot of sense. I'll get this merged and will send a new PR for this.

> It's obviously a bit of an ugly hack, but if the versions of angular and angular-mocks need to be kept in sync, could your clean-shrinkwrap script update the version of angular-mocks in package.json to match angular in npm-shrinkwrap.json ?

Looks ugly indeed. I like the idea of automatically protecting ourselves from these kind of stuff though. Maybe worth creating a new bash script with the sole task of ensuring that the angular dependencies are compatible with each other? We could make it run on the CI serves like we do to avoid unstaged SCSS changes. What do you think?

lurch:
> Maybe worth creating a new bash script with the sole task of ensuring that the angular dependencies are compatible with each other? We could make it run on the CI serves like we do to avoid unstaged SCSS changes. What do you think?

Yeah, could do. But obviously that script would have very little to do if you lock down the version of `angular` ;-)
Are there any other script versions that need to be kept "in step" too, or is it (so far) just `angular` and `angular-mocks` ?

jviotti:
> Yeah, could do. But obviously that script would have very little to do if you lock down the version of angular ;-)

Sure, but this would protect us from upgrading angular without upgrading angular mocks, or vice versa.

> Are there any other script versions that need to be kept "in step" too, or is it (so far) just angular and angular-mocks ?

Just those two as far as I can see.

lurch:
Hah, coincidentally... https://github.com/resin-io/resin-ui/issues/508

PR #1172, opened 2017-03-09
chore: migrate from scss-lint to sass-lint

lurch:
https://github.com/brigade/scss-lint now says "NOTICE: Consider other tools
before adopting SCSS-Lint", and recommends sass-lint as an alternative.
sass-lint has the additional advantage of being written in NodeJS, which means
it slots nicely into our existing infrastructure, and we can completely
remove our build-time dependency on Ruby.


jviotti:
You can't imagine how much I love this PR :) Its great that we will not have to deal with ruby anymore!

PR #1174, opened 2017-03-09
chore: codespell improvements

lurch:
* fix the command so it actually works - see https://github.com/lucasdemarchi/codespell/issues/101
 * check the spelling of more files
 * ignore known-binary filetypes
 * fix a couple of spelling mistakes it caught

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "jslint": "eslint lib tests scripts bin versionist.conf.js",
     "sasslint": "sass-lint lib/gui/scss",
     "htmllint": "node scripts/html-lint.js",
-    "codespell": "codespell.py --dictionary=- --dictionary=dictionary lib tests",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "jslint": "eslint lib tests scripts bin versionist.conf.js",
     "scsslint": "scss-lint lib/gui/scss",
     "htmllint": "node scripts/html-lint.js",
-    "codespell": "codespell.py --dictionary=- --dictionary=dictionary lib tests",
+    "codespell": "codespell.py --skip *.gz,*.bz2,*.xz,*.zip,*.img lib tests docs scripts Makefile CHANGELOG.md README.md SUPPORT.md LICENSE",
```
jviotti: Should we use a wildcard for the top level markdown files? E.g: `*.md`

lurch: Good call, done :+1:

PR #1175, opened 2017-03-09
chore: automatically enforce compatible angular core module versions

jviotti:
We've recently had an incident were the `angular-mocks` version we were
running was incompatible with the `angular` version.

Each AngularJS core module is versioned equally, therefore we can write
a small scripts that automatically test that they match, so we're
guarded against forgetting to similarly upgrade different dependencies.

See: https://github.com/resin-io/etcher/pull/1168

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,11 +49,10 @@ install:
 
 script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      npm test;
-      ./scripts/ci/ensure-staged-sass.sh;
+      make sanity-checks && npm test;
     fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-      ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "xvfb-run --server-args=$XVFB_ARGS npm test && ./scripts/ci/ensure-staged-sass.sh";
+      ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "make sanity-checks && xvfb-run --server-args=$XVFB_ARGS npm test";
     fi
 
 notifications:
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "removedrive": "^1.1.1"
   },
   "dependencies": {
-    "angular": "^1.6.3",
+    "angular": "1.6.3",
     "angular-if-state": "^1.0.0",
     "angular-middle-ellipses": "^1.0.0",
     "angular-moment": "^1.0.1",
--- a/Makefile
+++ b/Makefile
@@ -329,6 +329,7 @@ endif
 TARGETS = \
 	help \
 	info \
+	sanity-checks \
 	clean \
 	distclean \
 	package \
@@ -422,6 +423,10 @@ info:
 	@echo "Host arch           : $(HOST_ARCH)"
 	@echo "Target platform     : $(TARGET_PLATFORM)"
 	@echo "Target arch         : $(TARGET_ARCH)"
+
+sanity-checks:
+	./scripts/ci/ensure-staged-sass.sh
+	./scripts/ci/ensure-npm-dependencies-compatibility.sh
 
 clean:
 	rm -rf $(BUILD_DIRECTORY)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,7 +41,7 @@ build: off
 test_script:
   - node --version
   - npm --version
-  - bash .\scripts\ci\ensure-staged-sass.sh
+  - make sanity-checks
   - cmd: npm test
 
 notifications:
--- a/scripts/ci/ensure-npm-dependencies-compatibility.sh
+++ b/scripts/ci/ensure-npm-dependencies-compatibility.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+###
+# Copyright 2017 resin.io
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
+set -u
+set -e
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+"$HERE/../build/check-dependency.sh" jq
+
+PACKAGE_JSON=package.json
+
+# Two pair-wise arrays, because associative arrays only work in Bash 4
+PRIMARY_VERSIONS=("dependencies[\"angular\"]")
+SECONDARY_VERSIONS=("devDependencies[\"angular-mocks\"]")
+
+function check_locked {
+    name=$1
+    version=$2
+    if [[ "$version" =~ "^\^" ]]; then
+        echo "Dependency: $name must be version-locked in $PACKAGE_JSON"
+        exit 1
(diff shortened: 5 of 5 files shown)


scripts/ci/ensure-angular-dependencies-compatibility.js
```diff
@@ -0,0 +1,41 @@
+/**
```
lurch: I think this would be simpler if it was just a little bash script using `jq` ?

package.json
```diff
@@ -62,7 +62,7 @@
     "removedrive": "^1.1.1"
   },
   "dependencies": {
-    "angular": "^1.6.3",
+    "angular": "1.6.3",
```
lurch: Oh, does this change not also result in a change to `npm-shrinkwrap.json`?

.travis.yml
```diff
@@ -55,9 +55,10 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       npm test;
       ./scripts/ci/ensure-staged-sass.sh;
+      node ./scripts/ci/ensure-angular-dependencies-compatibility.js;
```
lurch: As we now have two "post build consistency checks" maybe they should just both get run by another npm `scripts` target in `package.json`?
Then we wouldn't need to edit both `.travis.yml` and `appveyor.yml` in future when adding additional checks, only `package.json` :-)

lurch:
...and I think "a small scripts" in your commit-message should be "a small script" ?

.travis.yml
jviotti: Makes sense, I added a new `sanity-checks` npm script.

package.json
jviotti: The shrinkwrap is already referencing 1.6.3 (given I upgraded the dependency in the other PR), so there's no change there.

scripts/ci/ensure-angular-dependencies-compatibility.js
jviotti: You're more than welcomed :)

jviotti:
> ...and I think "a small scripts" in your commit-message should be "a small script" ? And then change "test" to "tests".

lol, a major mixup of singulars and plurals :P

scripts/ci/ensure-angular-dependencies-compatibility.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  {
+    name: 'angular-mocks',
+    value: packageJSON.devDependencies['angular-mocks']
+  }
+];
+
+_.each(dependencies, (dependency) => {
+  if (!/^\d+\.\d+\.\d+/.test(dependency.value)) {
+    console.error(`Dependency: ${dependency.name} should be locked`);
```
lurch: As long as we're checking that they're the same version, does it matter if they're locked or not?

jviotti: The package.json might be declaring the same versions, e.g: `~1.6.1`, however that doesn't necessarily mean that the exact version is going to be installed.

lurch: Ahhh. Then (since these `sanity-checks` get run *after* `make electron-develop`) would it be worth checking that the installed `node_modules` match up pair-wise too?  (i.e. if we had `angular` 1.6.3 and `angular-mocks` 1.6.3 in `package.json`, but `angular` 1.6.5 and `angular-mocks` 1.6.5 installed into `node_modules`, presumably that'd work fine?)

Hmmm, and should `npm run sanity-checks;` actually get run **before** `npm test;`? Because being told "angular is version 1.6.1 but angular-mocks is version 1.6.3" is probably more useful error-output than "TypeError: angular.module(...).info is not a function" ? ;-)

jviotti: > ) would it be worth checking that the installed node_modules match up pair-wise too?

That's probably overkill. If we're locking the dependencies to the same version but npm somehow manages to install different ones, then npm would be really flawed.

> Hmmm, and should npm run sanity-checks; actually get run before npm test;? 

Oh, good catch. It's running before for Appveyor for after for Travis.

lurch: > then npm would be really flawed

Biting my tongue... ;-)

scripts/ci/ensure-angular-dependencies-compatibility.js
lurch: Done :)
https://gist.github.com/lurch/43342a528a2187dd45a333eea78aaa90

jviotti: @lurch I get the following error on OS X:

```sh
./scripts/ci/ensure-angular-dependencies-compatibility.sh: line 28: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

Not sure what's the equivalent to this `-A` option.

jviotti: I pushed it here so we can see if the CI servers complain with another error. Feel free to push straight to this PR btw.

lurch: > Not sure what's the equivalent to this -A option.

Oh damn. `declare -A` creates an associative array and is a new feature in Bash 4, so I guess the bash version in OS X must be older than that. I'll have to rewrite it using two linear arrays instead of a single associative array. No biggie.

> I pushed it here so we can see if the CI servers complain with another error.

https://travis-ci.org/resin-io/etcher/jobs/209813569 fails with the same error, **but** TravisCI thinks the job passed! Eeek, do we need to tweak something somewhere so that if `npm run sanity-checks` fails then the whole CI job fails?

> Feel free to push straight to this PR btw.

Cool will do. I think we ought to fix the non-failing-job problem first though, before I update the PR ;-)

jviotti: > https://travis-ci.org/resin-io/etcher/jobs/209813569 fails with the same error, but TravisCI thinks the job passed! Eeek, do we need to tweak something somewhere so that if npm run sanity-checks fails then the whole CI job fails?

Oh, interesting. I wonder why we didn't catch this before. Looks like the commands inside conditionals should be chained with `&&` rather than terminated with a `;`.

lurch:
Cool, the OSX and Windows CI tests are now "failing properly", so now to update the bash script... :-)

lurch:
Hmmm, now works on both Linux and OSX, but Appveyor is still failing with `Dependency: dependencies["angular"] must be version-locked in package.json` - it'll have to wait until Monday ;-)

@jviotti Any ideas how to get rid of the separate `sanity-checks-unix` and `sanity-checks-windows` entries in `package.json`?

jviotti:
> @jviotti Any ideas how to get rid of the separate sanity-checks-unix and sanity-checks-windows entries in package.json?

Let's move the task to the Makefile (and make the npm script still call it if you like), and that should do it!

jviotti:
@lurch Windows seem to work now! \o/

lurch:
Yup, I've got the script moved to the `Makefile` too (in my local version), just need to test it on Windows and Docker and then I'll push a further commit to this PR. I'm thinking that it probably _doesn't_ make sense to add `make sanity-checks` to `package.json`, given that this is a new script (i.e. nobody is relying on it yet) and there's also no entry in `package.json` for `make electron-develop` either.

jviotti:
>  I'm thinking that it probably doesn't make sense to add make sanity-checks to package.json, given that this is a new script (i.e. nobody is relying on it yet) and there's also no entry in package.json for make electron-develop either.

Yeah, I agree. Let's only keep it in the Makefile.

jviotti:
@lurch I've moved the npm scripts to the Makefile and squashed the commits. Let me know how it looks now!

lurch:
> I've moved the npm scripts to the Makefile and squashed the commits.

LOL, you beat me to it. Sorry, I've been busy with other stuff.

> Let me know how it looks now!

It looks great!! :tada: 
Your changes to `Makefile`, `.travis.yml` and `appveyor.yml` are identical to my local versions :+1:

Issue #1170, opened 2017-03-09
Show friendly name for removable drive in main window


dlech:
- **Etcher version:**
beta19
- **Operating system and architecture:**
all of the above

It makes me really nervous flashing a drive when it just says `/dev/sdc`. I always double check by clicking "change", which lists slightly more drive information or I use another application to confirm that the drive is really what I think it is. It would be nice to do this with fewer clicks. I would rather see the "friendly" name instead of the path in the main window (and maybe show the path on hover).

lurch:
Pinging @konmouz , our UX designer.

I guess the trouble with showing _only_ the "friendly name" is that you might have multiple drives that appear with the same "friendly name", whereas the raw-drive-name is guaranteed to be unique. So IMHO we should always at least show the raw-drive-name, and maybe also show the friendly-name at the same time?

konmouz:
![screen shot 2017-03-09 at 13 23 38](https://cloud.githubusercontent.com/assets/15323961/23752062/c1355ea6-04cb-11e7-827f-1db03c8a0ffc.png)
![screen shot 2017-03-09 at 13 24 25](https://cloud.githubusercontent.com/assets/15323961/23752063/c159c098-04cb-11e7-92cd-ae2987004c76.png)

jviotti:
@konmouz Its easy to show the friendly name there, however notice that we need to deal with how long names get displayed/truncated. Any ideas?

konmouz:
We should used ellipsis "..." here as well. We can put the size (e.g. 15.8GB) in a new line, but this will be a bit confusing with the CTA just below (might look like another CTA to some users).

dlech:
What if you put the ellipsis in the middle, like `really lo...ng name`.

lurch:
> What if you put the ellipsis in the middle, like `really lo...ng name`?

That's exactly how the ellipses (in Etcher) already work, see the `resin-tes...afe02.img` in the above screenshot ;-)

jviotti:
As a fun fact, the ellipses module we're using was tested with JSVerify (http://jsverify.github.io) with great success. We should bring that to more Etcher components :)

Shou:
[...]

PR #1181, opened 2017-03-15
feat(GUI): show drive name instead of device name

Shou:
We show the friendly drive name on the main page with the drive's device name
or mount point now inside a tooltip. Drive names longer than 11 characters are
truncated with a middle ellipsis.

Changelog-Entry: Show friendly drive name instead of device name.
Closes: https://github.com/resin-io/etcher/issues/1170

### DIFF
--- a/lib/gui/pages/main/templates/main.tpl.html
+++ b/lib/gui/pages/main/templates/main.tpl.html
@@ -60,8 +60,12 @@
           <div class="step-selection-text"
             ng-class="{
               'text-disabled': main.shouldDriveStepBeDisabled()
-            }">
-            <span class="step-drive step-name">{{ main.selection.getDrive().name }}</span>
+            }"
+            uib-tooltip="{{ main.selection.getDrive().name }}">
+            <span class="step-drive step-name">
+              <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
+              {{ (main.selection.getDrive().description || "") | middleEllipses:11 }}
+            </span>
             <span class="step-drive step-size">{{ main.selection.getDrive().size | gigabyte | number:1 }} GB</span>
           </div>
           <button class="button button-link step-footer"


Shou:
As usual, here's a screenshot

image

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -61,7 +61,10 @@
             ng-class="{
               'text-disabled': main.shouldDriveStepBeDisabled()
             }">
-            <span class="step-drive step-name">{{ main.selection.getDrive().name }}</span>
+            <span class="step-drive step-name">
+            <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
```
jviotti: Can we align the comment with the code below? E.g:

```html
<span class="step-drive step-name">
  {{ (main.selection.getDrive().description || "") | middleEllipses:11 }}
</span>
```

jviotti:
I left a very small comment, but it all looks great. This PR will need to be rebased after https://github.com/resin-io/etcher/pull/1182 gets merged though.

Shou:
@jviotti fixed and rebased

lurch:
Just to be pedantic, we never showed the "mount point" previously, we were showing the "device name". The "mount point" is something entirely different ;-)
Could you update the PR title and commit-message?

lurch:
Well, actually it was "device name" on Linux and OSX, and "mount points" on Windows ;-)
https://github.com/resin-io/etcher/blob/master/lib/gui/modules/drive-scanner.js#L54

lurch:
Based on the discussion in #1170 can we get the `main.selection.getDrive().name` shown as a tooltip?

jviotti:
Sounds good. Let's do that first.

Shou:
Here we go. Maybe with #1089 we can also put the full (non-truncated) drive description in the popup?

jviotti:
Awesome, we'll merge once we sort this weird Windows build issue. 

> Maybe with #1089 we can also put the full (non-truncated) drive description in the popup?

@konmouz What do you think?

konmouz:
@jviotti sounds reasonable. Just to recap behavior: Only drive name and size are initially displayed. If name is too long, it will be truncated (ellipse in the middle of the name). On hover, a tooltip that includes the full drive name and mount point will appear.

jviotti:
Sounds good. Let's address that in another PR.

PR #1182, opened 2017-03-15
chore: get the CI tests working again

lurch:
* an updated eslint-plugin-lodash was creating extra linter errors where we
   were using built-in String methods instead of the lodash equivalents
 * an updated codespell package now installs with a different executable-name
   (no .py extension), and now supports multiple dictionaries
 * while I was at it I replaced the messy double-quoting-string logic with the
   command-join module

Changelog-type: patch


package.json
```diff
@@ -16,7 +16,7 @@
     "jslint": "eslint lib tests scripts bin versionist.conf.js",
     "sasslint": "sass-lint lib/gui/scss",
     "htmllint": "node scripts/html-lint.js",
-    "codespell": "codespell.py --skip *.gz,*.bz2,*.xz,*.zip,*.img lib tests docs scripts Makefile *.md LICENSE",
+    "codespell": "codespell --dictionary=- --dictionary=dictionary --skip *.gz,*.bz2,*.xz,*.zip,*.img lib tests docs scripts Makefile *.md LICENSE",
```
jviotti: How is this change related to this PR? Appveyor CI is throwing:

> 'codespell' is not recognized as an internal or external command,
operable program or batch file.

jviotti: > an updated codespell package now installs with a different executable-name
(no .py extension), and now supports multiple dictionaries

Oh, I see. Just `codespell` doesn't seem to be working anymore though.

lurch: Works on Travis... :-/

tests/gui/components/svg-icon.spec.js
```diff
@@ -33,9 +34,9 @@ describe('Browser: SVGIcon', function() {
 
     it('should inline the svg contents in the element', function() {
       const icon = '../../../../../lib/gui/assets/etcher.svg';
-      let iconContents = fs.readFileSync(path.join(__dirname, '../../../lib/gui/assets/etcher.svg'), {
+      let iconContents = _.split(fs.readFileSync(path.join(__dirname, '../../../lib/gui/assets/etcher.svg'), {
         encoding: 'utf8'
-      }).split(/\r?\n/);
+      }), /\r?\n/);
```
jhermsmeier: Since we're only replacing the first line, maybe a `replace()`, rather than a `split()` might be shorter & better to comprehend – and we could reuse `icon` and `iconContents`:

```js
let iconContents = fs.readFileSync( path.join( __dirname, icon ), 'utf8' )
iconContents = _.replace( iconContents, /^([^\n\r]+?)/, ( match, firstLine ) => {
  return `<!--${firstLine.slice(1)}-->`;
})
```

lurch: I have no idea what this code is doing (I can feel my brain starting to melt from all this lodash stuff and trying to fix the CI build!) but I'm happy to change that if @jviotti approves.

jviotti: lol, the change looks good to me, however notice that `path.join( __dirname, icon)` will not resolve correctly, given that `icon` has one extra `..`.

jhermsmeier: Yeah, you could forward resolve again with `path.join( __dirname, 'whichever/folder', icon )` though

jviotti: The missing piece seems to be the directory that you used to clone the repo, which might be anything.

jhermsmeier: True. Wait... according to the `icon` path `etcher.svg` resides 5 levels up, but we're only 3 levels deep here – why does it point to outside the repo root? Or am I completely misunderstanding this?

jviotti: AngularJS weirdness. The path is actually relative to the location of the `svg-icon` directive :P We'll hopefully move to React very soon :)

lurch:
Yay, glad we managed to get this merged :grinning:

PR #1186, opened 2017-03-17
doc: Add MSVC Build Tools option to Windows requisites

jhermsmeier:
Add Visual C++ 2015 Build Tools as a lighter option to a full blown Visual Studio installation for development requirements on Windows.

### DIFF
--- a/docs/RUNNING-LOCALLY.md
+++ b/docs/RUNNING-LOCALLY.md
@@ -21,9 +21,11 @@ Prerequisites
 
 - [Rimraf](https://github.com/isaacs/rimraf)
 - [NSIS v2.51](http://nsis.sourceforge.net/Main_Page) (v3.x won't work)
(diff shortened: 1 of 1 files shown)


jviotti:
Oh, very nice. I'll test this on my fresh Windows 8 installation to confirm it works :+1:

lurch:
Indeed, very nice. Would it be worth updating Appveyor to use this?

WasabiFan:
I have to ask... How'd you get a fresh Win8 installation?

jviotti:
> Indeed, very nice. Would it be worth updating Appveyor to use this?

Appveyor machines include the full Visual Studio by default, so we can use that instead of having to install something custom.

> I have to ask... How'd you get a fresh Win8 installation?

You can download an ISO here: https://www.microsoft.com/en-us/software-download/windows8ISO

PR #1187, opened 2017-03-17
minifix(GUI): show Windows warning for images starting with W(8|9|10)

jviotti:
Some Windows images start with `W8`, `W9`, or `W10`. This commit extends
the `.looksLikeWindowsImage()` regular expression to catch these cases.

Since a `W` followed by a number can occur in other parts of a file
name, we only attempt to detect this pattern at the beginning of the
file name.

### DIFF
--- a/lib/gui/models/supported-formats.js
+++ b/lib/gui/models/supported-formats.js
@@ -145,7 +145,7 @@ SupportedFormats.service('SupportedFormatsModel', function() {
    * }
    */
   this.looksLikeWindowsImage = (imagePath) => {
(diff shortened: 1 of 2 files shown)


jviotti:
I'm closing this, since we've decided to start checking the MBR instead of relying on file names.

lurch:
If you're referring to the recent conversation in gitter, I believe the `W10_july.img` filename only came about because the user manually renamed the file.
Or is there evidence of "official" Windows images using the "W10" naming convention?

Issue #1138, opened 2017-02-27
Slow start


PrzemekSkw:
- **Etcher version: Etcher-1.0.0-beta.19**
- **Operating system and architecture: Fedora 25 -Gnome, 64 bit**
- **Do you see any meaningful error information on DevTools? No**

Hi, it's possible to improve launch speed of Etcher because it's slowest app I have used. I have fast SSD drive and every app starting when I click on icon but for Etcher I must wait about 6 seconds :)

Regards,

lurch:
I believe this is a side-effect of AppImages and/or the executable-compression used by Etcher.
Changing either of those would make Etcher much harder to install (i.e. we wouldn't be able to use a single download to support all Linux distros), and make it take up lots more diskspace / be an even bigger download.
Unless @probonopd has some suggestions?

PrzemekSkw:
OK, thanks for an answer. Maybe flatpak will be better option than appimage ?

jviotti:
Its probably a combination of:

- AppImages
- Electron (which is not particularly a "fast" framework)
- UPX

I'll prepare some custom builds so we can try to find the current bottleneck.

lurch:
Coincidentally, I had a quick look at the flatpak website over the weekend, but unfortunately it doesn't support Ubuntu 14.04 (which I'm still running on my laptop), so I didn't look at it in much depth...

probonopd:
UPX? Why UPX? You could try squashfs-based AppImages for comparison, too. And the unpacked AppImage (then: AppDir).

jviotti:
UPX seems to slow down the binary considerably in some cases when we use the best possible compression. I'm uploading a set of builds as we talk, which should be enough to at least identity what to investigate further.

probonopd:
AppImages are compressed without the need for UPX anyway...

PrzemekSkw:
I try Kaku Music Player appimage and it start in 1 second :)

jviotti:
Yeah, I doubt AppImages are the cause. I'm still uploading btw, I should probably move to a country with better internet :P

jviotti:
@PrzemekSkw Here are the builds:

- AppImage without upx compression: https://drive.google.com/open?id=0B7tkbonGU-RyWEN1bGJqT0tWRDQ
- Raw build (without being packaged as an AppImage) without upx compression: https://drive.google.com/open?id=0B7tkbonGU-RyOGVTM1prX0wwVDA
[...]

PR #1188, opened 2017-03-17
chore: don't compress binaries/libraries with upx

jviotti:
upx allows to reduce the package size by some mega bytes, but also
considerably slows down the application start time.

This commit ditches upx, since we can re-gain those benefits (and even
increase them) by other means (e.g: minifying JavaScript).

See: https://github.com/resin-io/etcher/issues/1138

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,6 @@ install:
   - ps: Install-Product node $env:nodejs_version x64
   - npm install -g bower rimraf asar
   - choco install nsis -version 2.51
-  - choco install upx
   - choco install jq
   - set PATH=C:\Program Files (x86)\Windows Kits\8.1\bin\x86;%PATH%
   - set PATH=C:\Program Files (x86)\NSIS;%PATH%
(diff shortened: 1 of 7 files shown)


lurch:
Just to satisfy my curiosity, could you post the final output sizes both with and without UPX?

jviotti:
- GNU/Linux x64 without upx: 167 MB
- GNU/Linux x64 with upx: 120 MB
- Windows x64 without upx: 194 MB
- Windows x64 with upx: 178 MB

(note that we don't use upx on macOS, given it breaks the `*.app`)

lurch:
Out of curiosity, looking at what this PR is "undoing", how come on Windows we only UPX-ed the `.dll`s and not the `.exe` ? Something to do with code-signing?

jviotti:
> 
Out of curiosity, looking at what this PR is "undoing", how come on Windows we only UPX-ed the .dlls and not the .exe ? Something to do with code-signing?

Oh, looks like we missed that. I don't remember any upx problems on Windows.

PR #1191, opened 2017-03-18
chore: fix `lzma-native` build issues on Windows

jviotti:
We've been recently hitting a weird `lzma-native` build error on Windows
(both locally and on Appveyor CI):

```
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
  build
  The input line is too long.

C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with code 1. [C:\projects\etcher\node_modules\lzma-native\build\liblzma.vcxproj]
```

After a lot of experimentation, we realised the issue was gone if we
removed `node-sass` from the development dependencies.

The issue is that `node-gyp` was recently upgraded to v3.6.0, which was
picked up by `node-sass`, which declares `node-gyp` as a dependency. For
some reason, if `node-sass` causes `node-gyp` to be updated, then
`lzma-native` fails with the above cryptic error.

I was able to trace down the error to the following `node-gyp` commit:

https://github.com/nodejs/node-gyp/commit/ae141e19060ecb1c596a96acf9510d66b7376c18

As a solution, this commit starts to shrinkwrap development
dependencies, and locks `node-gyp` to v3.5.0 until the issue is fixed.

Fixes: https://github.com/addaleax/lzma-native/issues/30
See: https://github.com/nodejs/node-gyp/issues/1151

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -48,17 +48,6 @@ Use the following steps to ensure everything goes flawlessly:
   operating system specific dependencies that could get included in the
   previous step are removed from `npm-shrinkwrap.json`.
 
-Some npm versions seem to contain an issue were all development dependencies
-will be included in `npm-shrinkwrap.json` when attempting to modify it (e.g: by
-`npm install`, `npm uninstall`, etc). A bulletproof way to ensure only the
-necessary dependencies get added is to run the following commands:
-
-```sh
-make electron-develop
-npm prune --production
-npm shrinkwrap --dev
-```
-
 - Commit *both* `package.json` and `npm-shrinkwrap.json`.
 
 Testing
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -56,7 +56,7 @@ necessary dependencies get added is to run the following commands:
 ```sh
 make electron-develop
 npm prune --production
-npm shrinkwrap
+npm shrinkwrap --dev
 ```
 
 - Commit *both* `package.json` and `npm-shrinkwrap.json`.
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2,6 +2,65 @@
   "name": "etcher",
   "version": "1.0.0-beta.19",
   "dependencies": {
+    "abbrev": {
+      "version": "1.1.0",
+      "from": "abbrev@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz"
+    },
+    "acorn": {
+      "version": "4.0.11",
+      "from": "acorn@>=4.0.3 <5.0.0",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.11.tgz"
+    },
+    "acorn-jsx": {
+      "version": "3.0.1",
+      "from": "acorn-jsx@>=3.0.0 <4.0.0",
+      "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
+      "dependencies": {
+        "acorn": {
+          "version": "3.3.0",
+          "from": "acorn@>=3.0.4 <4.0.0",
+          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz"
+        }
+      }
+    },
+    "agent-base": {
+      "version": "1.0.2",
+      "from": "agent-base@>=1.0.1 <1.1.0",
+      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-1.0.2.tgz"
+    },
+    "ajv": {
+      "version": "4.11.5",
+      "from": "ajv@>=4.9.1 <5.0.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.5.tgz",
+      "dependencies": {
+        "json-stable-stringify": {
+          "version": "1.0.1",
+          "from": "json-stable-stringify@>=1.0.1 <2.0.0",
+          "resolved": "https://registry.npmjs.org/json-stable-stringify/-/json-stable-stringify-1.0.1.tgz"
+        }
+      }
+    },
+    "ajv-keywords": {
+      "version": "1.5.1",
+      "from": "ajv-keywords@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-1.5.1.tgz"
+    },
+    "align-text": {
+      "version": "0.1.4",
+      "from": "align-text@>=0.1.3 <0.2.0",
+      "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz"
+    },
+    "alter": {
+      "version": "0.2.0",
+      "from": "alter@>=0.2.0 <0.3.0",
+      "resolved": "https://registry.npmjs.org/alter/-/alter-0.2.0.tgz"
+    },
+    "amdefine": {
+      "version": "1.0.1",
+      "from": "amdefine@>=0.0.4",
+      "resolved": "https://registry.npmjs.org/amdefine/-/amdefine-1.0.1.tgz"
+    },
     "angular": {
       "version": "1.6.3",
       "from": "angular@1.6.3",
@@ -24,6 +83,11 @@
       "from": "angular-middle-ellipses@>=1.0.0 <2.0.0",
       "resolved": "https://registry.npmjs.org/angular-middle-ellipses/-/angular-middle-ellipses-1.0.0.tgz"
     },
+    "angular-mocks": {
+      "version": "1.6.3",
+      "from": "angular-mocks@1.6.3",
+      "resolved": "https://registry.npmjs.org/angular-mocks/-/angular-mocks-1.6.3.tgz"
+    },
     "angular-moment": {
       "version": "1.0.1",
       "from": "angular-moment@1.0.1",
@@ -43,6 +107,11 @@
       "version": "0.4.2",
       "from": "angular-ui-router@>=0.4.2 <0.5.0",
       "resolved": "https://registry.npmjs.org/angular-ui-router/-/angular-ui-router-0.4.2.tgz"
+    },
+    "ansi-escape-sequences": {
+      "version": "2.2.2",
+      "from": "ansi-escape-sequences@>=2.2.2 <3.0.0",
+      "resolved": "https://registry.npmjs.org/ansi-escape-sequences/-/ansi-escape-sequences-2.2.2.tgz"
     },
     "ansi-escapes": {
       "version": "1.4.0",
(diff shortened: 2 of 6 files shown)


lurch:
Have you created an issue against the `node-gyp` repo?
Should https://github.com/resin-io/etcher/blob/master/docs/CONTRIBUTING.md be updated to explain the `--dev` flag that you talk about [here](https://gitter.im/resin-io/etcher?at=58cc9b077b3f37e7540c435f) ?

Great to see the CI tests working again!

npm-shrinkwrap.json
```diff
@@ -1271,56 +4276,274 @@
...
+      "version": "0.1.0",
+      "from": "murmur-32@>=0.1.0 <0.2.0",
+      "resolved": "https://registry.npmjs.org/murmur-32/-/murmur-32-0.1.0.tgz"
+    },
     "mute-stream": {
       "version": "0.0.5",
       "from": "mute-stream@0.0.5",
       "resolved": "https://registry.npmjs.org/mute-stream/-/mute-stream-0.0.5.tgz"
     },
-    "nan": {
```
lurch: I was concerned that some modules have been removed from the shrinkwrap, but it's actually just github's diff-display getting confused ;-)

lurch:
...and when we next need to update the shrinkwrap file, how do we ensure that `node-gyp` gets locked to v3.5.0 ?

jviotti:
> Have you created an issue against the node-gyp repo?
Should https://github.com/resin-io/etcher/blob/master/docs/CONTRIBUTING.md be updated to explain the --dev flag that you talk about here ?

Good catch! Done!

> ...and when we next need to update the shrinkwrap file, how do we ensure that node-gyp gets locked to v3.5.0 ?

I just tried to upgrade `node-sass` to the latest version, and since `node-gyp` is being independently locked, it remains at v3.5.0 on the shrinkwrap file! We should still keep an eye during code reviews to ensure the dependency is not manually upgraded.

scripts/clean-shrinkwrap.js
```diff
@@ -25,7 +25,7 @@ const EXIT_CODES = require('../lib/shared/exit-codes');
 const SHRINKWRAP_PATH = path.join(__dirname, '..', 'npm-shrinkwrap.json');
 
 try {
-  console.log(childProcess.execSync('npm shrinkwrap', {
+  console.log(childProcess.execSync('npm shrinkwrap --dev', {
     cwd: path.dirname(SHRINKWRAP_PATH)
   }));
```
lurch: I think you need to change this to `  }).toString());` ? Otherwise I just get `<Buffer` and a load of hex-digits printed to the screen.

And I think this script might need to be using `npm prune` too? In fact, could the `CONTRIBUTING.md` instructions be updated to use _only_ this script (even if that means modifications are needed to this script also), as currently it's unclear in the instructions when to use `npm run clean-shrinkwrap` and when to use `npm shrinkwrap --dev`.

lurch:
> We should still keep an eye during code reviews to ensure the dependency is not manually upgraded.

Sounds like a job for another CI bash script. Give me 10 minutes... ;-)

scripts/clean-shrinkwrap.js
lurch: ...and when I run this on my local machine (which currently has npm 3.10.10) it adds a load of `"dev": true` entries into my `npm-shrinkwrap.json`, which aren't in the shrinkwrap file in this PR? :-/

e.g.
```patch
index 5556a28..f7916e9 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5,61 +5,72 @@
     "abbrev": {
       "version": "1.1.0",
       "from": "abbrev@>=1.0.0 <2.0.0",
-      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz"
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
+      "dev": true
     },
     "acorn": {
       "version": "4.0.11",
       "from": "acorn@>=4.0.3 <5.0.0",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.11.tgz"
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.11.tgz",
+      "dev": true
     },
     "acorn-jsx": {
```

lurch:
...I've added `scripts/ci/ensure-npm-shrinkwrap-versions.sh` to `make sanity-checks` :-)

scripts/build/dependencies-npm.sh
```diff
@@ -91,9 +91,19 @@ if [ "$ARGV_FORCE" == "true" ]; then
   INSTALL_OPTS="$INSTALL_OPTS --force"
 fi
 
-if [ "$ARGV_PRODUCTION" == "true" ]; then
-  INSTALL_OPTS="$INSTALL_OPTS --production"
-fi
+function run_install() {
+  npm install $INSTALL_OPTS
+
+  # Turns out that if `npm-shrinkwrap.json` contains development
```
jviotti: Done!

docs/CONTRIBUTING.md
```diff
@@ -56,7 +56,7 @@ necessary dependencies get added is to run the following commands:
 ```sh
 make electron-develop
 npm prune --production
```
jviotti: I"m removing this whole section, which described an approach to avoid development dependencies from being shrinkwrapped, which we are doing now.

scripts/clean-shrinkwrap.js
jviotti: Regarding this scripts, I think the complexity of this script lies in the fact that it call `npm shrinkwrap` itself, when its not really its responsibility. Now that we're shrinkwrapping development dependencies, we shouldn't need to call `npm shrinkwrap` ourselves (npm will do it automatically for us whenever we do `npm install`, `npm uninstall`, etc).

I edited the script to only scans the current shrinkwrap file and remove the optional dependencies.

jviotti: Regarding the `dev: true` properties, it seems that they are being added by a new npm version (mine is a bit older than yours). I'll upgrade and get those changes added here.

jviotti:
Amazing stuff! :D

jviotti:
@lurch Updated!

refack:
Hello from nodejs/node-gyp#1151,

> It's not a "length of command line" issue, it's a "not run from VS console" issue.
> The `lzma-native` build calls `lib.exe` which is not in the `path` unless `vsvarsall.bat` (or it's derivatives) is run.
> Up till `node-gyp:v3.5.x` the whole thing would fail. Now we can find & call `msbuild.exe` in the general case, but then we fail if the build is complicated ;(

I'm working on a better workflow, but till then you'll need to setup VC Vars...
Anyway I just found out that `lzma-native:2.0.0` uses pre-build binaries, maybe that will solve this?

jviotti:
Hi @refack , 

Thanks for reaching out. 

> I'm working on a better workflow, but till then you'll need to setup VC Vars...

Correct me if I'm wrong (I'm not a Windows expert), but this means that we should run `vsvarsall.bat` before compiling, right? If so, we're using the Visual Studio command prompt, which does that already AFAIK. In any case, I tried re-running the script before building the module, and it fails with the same error.

> Anyway I just found out that lzma-native:2.0.0 uses pre-build binaries, maybe that will solve this?

We were using pre-build binaries at some point, but decided to build ourselves after hitting an ABI issue after an electron upgrade (better to make sure they are always on sync by compiling ourselves)

refack:
Sorry, I was too hasty. it's not a `vcvarsall.bat` error.
It's a bug in `GYP`
https://github.com/nodejs/node-gyp/issues/1151#issuecomment-287839249

addaleax:
> > Anyway I just found out that lzma-native:2.0.0 uses pre-build binaries, maybe that will solve this?

> We were using pre-build binaries at some point, but decided to build ourselves after hitting an ABI issue after an electron upgrade (better to make sure they are always on sync by compiling ourselves)

Btw, earlier versions of the library also have pre-buit binaries available. But fwiw, if there’s any help I can offer by providing more pre-built binaries, I’m glad to look into that. (It would be really cool to know which electron versions to look out for in that case, though.)

jviotti:
Hi @addaleax

Thank you very much the good will. If you want to ensure `lzma-native` pre compiled binaries work out of the box for Electron project, I guess you could keep an eye on Electron releases (https://github.com/electron/electron/releases) and publish as soon as Electron publishes a new version that makes use of a new node ABI.

In any case, I don't think its reasonable for you to invest time into that. Compiling the module just takes a few seconds from our side, and allows you to not have to worry about being on top of nodejs releases all the time.

scripts/build/dependencies-npm.sh
```diff
@@ -94,15 +94,16 @@ fi
...
 
-  # Turns out that if `npm-shrinkwrap.json` contains development
-  # dependencies then `npm install --production` will also install
-  # those, despite knowing, based on `package.json`, that they are
-  # really development dependencies. As a workaround, we manually
-  # delete the development dependencies using `npm prune`.
   if [ "$ARGV_PRODUCTION" == "true" ]; then
+
+    # Turns out that if `npm-shrinkwrap.json` contains development
+    # dependencies then `npm install --production` will also install
```
lurch: Does this comment imply that when `"$ARGV_PRODUCTION" == "true"` we ought to be adding `--production` to `$INSTALL_OPTS`?

(I guess there's a slim chance that `npm install --production` might get fixed in future, and in that case `npm install --production; npm prune --production` will probably be quicker than `npm install; npm prune --production`?)

docs/CONTRIBUTING.md
```diff
@@ -48,17 +48,6 @@ Use the following steps to ensure everything goes flawlessly:
...
-will be included in `npm-shrinkwrap.json` when attempting to modify it (e.g: by
-`npm install`, `npm uninstall`, etc). A bulletproof way to ensure only the
-necessary dependencies get added is to run the following commands:
-
-```sh
-make electron-develop
-npm prune --production
-npm shrinkwrap --dev
-```
-
```
lurch: Ahh, much cleaner :+1:

lurch:
OTOH if pre-compiled versions of `lzma-native` _did_ work with all versions of Electron, would it allow us to remove the Visual Studio C++ requirement from the RUNNING-LOCALLY doc?

jviotti:
@lurch We will have various more native dependencies very soon:

- mountutils
- drivelist (at least for Windows due to the WMI size bug)
- Some module to do Windows lock/unlocking, etc

so I wouldn't try to remove the Visual Studio dependency.

scripts/build/dependencies-npm.sh
jviotti: I mean that when doing `npm install --production`, npm will also install the development dependencies (I know, npm is messed :P), which is the reason why we just do `npm install` + `npm prune --production`.

> (I guess there's a slim chance that npm install --production might get fixed in future, and in that case npm install --production; npm prune --production will probably be quicker than npm install; npm prune --production?)

Fair enough. Lets add it in case it gets fixed in the future :)

jviotti: Done!

refack:
[Found the bug in `node-gyp`](https://github.com/nodejs/node-gyp/issues/1151#issuecomment-287953524)

jviotti:
Amazing stuff, thanks a lot @refack ! Do you mind letting us know when you release a new version, so we can upgrade to it?

Issue #1144, opened 2017-02-28
Improve "Source Drive" drive selector badge wording 


jviotti:
From https://github.com/resin-io/etcher/issues/1136

"Source Drive" is hard to understand. Maybe something more obvious like "Drive contains image"?

EDIT: Orignally from https://github.com/resin-io/etcher/issues/1136#issuecomment-282983302

Issue #1143, opened 2017-02-28
Don't show System Drive and Source Drive badges at the same time


jviotti:
From https://github.com/resin-io/etcher/issues/1136

![7716ee38-fcde-11e6-8d6e-a8fa7cba9712](https://cloud.githubusercontent.com/assets/2192773/23408208/a3c80a38-fd9d-11e6-81a3-a529eada156a.png)

Which one should we prefer though? @konmouz ?

EDIT: Originally from https://github.com/resin-io/etcher/issues/1136#issuecomment-282776898

WasabiFan:
I think there's a case to be made for keeping both: Doing so reinforces the meaning of "source drive" (or whatever you rename it to) so that the user has a reasonable means of understanding that label.

jviotti:
That's a good point as well. Up to @konmouz I guess.

konmouz:
@jviotti I am not sure if I am the better to answer this, it is a bit technical. Is source drive a boot drive? We need to find a name that represents this drive and fits in one label.

jviotti:
@konmouz The source drive is the drive that contains the image that you previously selected (not necessarily a system drive)

konmouz:
@jviotti, ok I was completely out of topic. Now it makes sense why you might need two. So, since a 'source' drive is only one at a time, it should have a different color label (for instance orange). Also, may worth renaming to 'image source' and of course the drive would be greyed out. Maybe in that sense, we don't need the second label even if the drive is system or boot drive.

jviotti:
Perfect, I'll close this issue then!

lurch:
> Now it makes sense why you might need two. (...)
Maybe (...) we don't need the second label even if the drive is system or boot drive.

I'm not quite sure I follow. Which "second label" are you suggesting we omit @konmouz ?

lurch:
Re-opening this issue...

PR #1193, opened 2017-03-20
refactor(GUI): central error/warning messages function

Shou:
* Centralise drive error/warning messages with a function
  `.getDriveImageCompatibilityStatuses` in `lib/shared/drive-constraints.js` -- tests included
* Fix an error where several labels show at once
* Clarify the source drive label to 'Drive Contains Image'
* Remove label icons and make text bold to match Zeplin's design

Closes: #1143
Closes: #1144
Changelog-Entry: New central function as the source of drive error
messages.


jviotti:
Looks good, but I'm worried about how complex these conditionals are getting, and the fact that we can't unit test them if they live only in the template.

How do you feel about adding a function somewhere in `lib/shared` (maybe inside the drive constraints file) that takes the drive and the image, and returns the correct label to show. We can then unit test that a lot :)

Shou:
@jviotti that's a good idea. I'll retrofit this PR for that. How would we differentiate warnings from errors? 
Since "Not Recommended" still permits you to flash, maybe two functions—one for warnings and one for errors returning their respective strings if applicable and otherwise `null`?

jviotti:
We can make the function return an object, including a `name` and a `type` property, where `type` can be `warning`, `error`, etc.

lurch:
> How do you feel about adding a function somewhere in lib/shared (maybe inside the drive constraints file) that takes the drive and the image, and returns the correct label to show.

I'm fairly sure I suggested that (or at least something very similar) quite a while ago? ;-P

@Shou I restarted the failing CI builds and they're nearly all passing now (Travis seems to be suffering from network unreliability issues recently requiring manual build-restarts). Looks like if you rebase to the latest `master` then the `appveyor/branch` builds should pass too.

Shou:
How's this?

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -15,35 +15,14 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
-              ng-show="modal.constraints.isSourceDrive(drive, modal.state.getImage())">
-                <i class="glyphicon glyphicon-circle-arrow-down"></i>
-                Source Drive</span>
+            <span class="label"
+              ng-class="{
+                'label-warning': modal.constraints.getDriveFeedback(drive, modal.state.getImage()).type === 'warning',
+                'label-danger': modal.constraints.getDriveFeedback(drive, modal.state.getImage()).type === 'error'
+              }"
+              ng-show="modal.constraints.getDriveFeedback(drive, modal.state.getImage()).message">
+                <b>{{ modal.constraints.getDriveFeedback(drive, modal.state.getImage()).message }}</b>
```
lurch: If the same function is being called 4 times, should the `.getDriveFeedback` function be memoized?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,85 @@ exports.isDriveValid = (drive, image) => {
 exports.isDriveSizeRecommended = (drive, image) => {
   return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'recommendedDriveSize' ], UNKNOWN_SIZE);
 };
+
+const feedbackMessages = {
```
lurch: Doesn't seem to be any need for not specifying these directly "inline"?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,85 @@ exports.isDriveValid = (drive, image) => {
...
+
+/**
+ * @summary Get drive feedback in an object, if any
+ * @function
+ * @public
+ *
+ * @description
+ * If the drive-image combination results in a warning or error, this function
+ * returns an object containing the feedback type ('error', 'warning') and the
+ * feedback message.
```
lurch: Rather than returning `null`, would it be more consistent to return a feedback object with its `.type` set to something like `FEEDBACK_OKAY` ?  \*shrug\*

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,85 @@ exports.isDriveValid = (drive, image) => {
...
+  // Mind the order of the if-statements if you modify.
+  if (exports.isSourceDrive(drive, image)) {
+    return {
+      type: FEEDBACK_ERROR,
+      message: feedbackMessages.SOURCE
+    };
+
+  } else if (exports.isSystemDrive(drive)) {
+    return {
+      type: FEEDBACK_ERROR,
```
lurch: Hmm, what if "dangerous mode" is enabled?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,85 @@ exports.isDriveValid = (drive, image) => {
...
+ *
+ * const { type, message } = DriveConstraints.getDriveFeedback(drive, image);
+ *
+ * if (type === 'warning') {
+ *   // still allow flashing
+ * } else if (type === 'error') {
+ *   // disallow flashing
+ * }
+ */
+exports.getDriveFeedback = (drive, image) => {
```
lurch: I wonder if it should be renamed `.getDriveImageFeedback` (or something) to more accurately reflect its arguments? Or maybe that'd just make things more confusing?

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -15,35 +15,14 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
-                System Drive</span>
-
-            <span class="label label-danger"
-              ng-show="modal.constraints.isSourceDrive(drive, modal.state.getImage())">
-                <i class="glyphicon glyphicon-circle-arrow-down"></i>
-                Source Drive</span>
+            <span class="label"
+              ng-class="{
+                'label-warning': modal.constraints.getDriveFeedback(drive, modal.state.getImage()).type === 'warning',
+                'label-danger': modal.constraints.getDriveFeedback(drive, modal.state.getImage()).type === 'error'
```
lurch: Is it possible to compare these to e.g. the `FEEDBACK_ERROR` constants directly?
(and if that _is_ possible, then obviously the unit-tests ought to be updated to match)

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,133 @@ describe('Shared: DriveConstraints', function() {
...
+        });
+
+        m.chai.expect(result).to.deep.equal({
+          type: 'error',
+          message: 'Locked'
+        });
+      });
+    });
+
+    describe('given the drive is unrecommended size', () => {
```
lurch: I'd change "unrecommended size" to "smaller than the recommended size"

lurch:
Looking better now :)  :+1: 
I recall the last time that this was mentioned, there were actually two places where this logic was being duplicated. Is that still the case (in which case some other file probably needs to be updated to use `.getDriveFeedback` too), or has the code been refactored so much since then that the logic is now only used in one place?

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -15,35 +15,14 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
-                <i class="glyphicon glyphicon-hdd"></i>
-                System Drive</span>
-
-            <span class="label label-danger"
-              ng-show="modal.constraints.isSourceDrive(drive, modal.state.getImage())">
-                <i class="glyphicon glyphicon-circle-arrow-down"></i>
-                Source Drive</span>
+            <span class="label"
+              ng-class="{
+                'label-warning': modal.constraints.getDriveFeedback(drive, modal.state.getImage()).type === 'warning',
```
jviotti: `modal.constraints.getDriveFeedback(drive, modal.state.getImage())` is being repeated tons of times here. Maybe we should move it to the controller?

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
Shou: That's definitely possible, so I can do that.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
Shou: I had the same feeling but no clean idea of how to resolve it.

lib/shared/drive-constraints.js
Shou: I could potentially expose this along with the error types.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -15,35 +15,14 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
 
           <footer class="list-group-item-footer">
 
-            <span class="label label-warning"
-                ng-show="modal.constraints.isDriveLargeEnough(drive, modal.state.getImage()) && !modal.constraints.isDriveLocked(drive) && !modal.constraints.isDriveSizeRecommended(drive, modal.state.getImage())">
-                <i class="glyphicon glyphicon-warning-sign"></i>
```
jviotti: What happened with the icons? We should bring them back as well, unless they are not present in the new design?

lib/shared/drive-constraints.js
Shou: Hmm, it does feel more consistent but `null` might be simpler to deal with? I'm not sure here.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
Shou: Yeah, I thought I removed them a while back, but I noticed they were there. The icon removal reflects Zeplin.

lib/shared/drive-constraints.js
lurch: If you _do_ do that, then IMHO the constants should be given better (longer) names.

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,85 @@ exports.isDriveValid = (drive, image) => {
...
+
+  // Mind the order of the if-statements if you modify.
+  if (exports.isSourceDrive(drive, image)) {
```
jviotti: Maybe its just me, but given we're returning inside each conditional, something like this looks cleaner IMHO:

```js
if (exports.isSourceDrive(drive, image)) {
    return { ... };
}

if (exports.isSystemDrive(drive)) {
    return { ... };
}

if (!exports.isDriveLargeEnough(drive, image)) {
    return { ... };
}

...
```

lurch:
And of course as always, a screenshot would be really useful ;-)

lib/shared/drive-constraints.js
jviotti: Even better: what about exporting these, so the client code can check which message is which? This could be really handy to address https://github.com/resin-io/etcher/pull/1193/files#r107220266: this function can determine flashing to a system drive is an error, while the GUI can check if that's the case by comparing with one of these constants, and dismissing the error if the unsafe mode setting is enabled.

Shou:
That's true! Here you go.

image

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,85 @@ exports.isDriveValid = (drive, image) => {
...
 };
+
+const feedbackMessages = {
+  NOTREC: 'Not Recommended',
+  TOOSMALL: 'Too Small For Image',
+  LOCKED: 'Locked',
+  SYSTEM: 'System Drive',
+  SOURCE: 'Drive Contains Image'
+};
+const FEEDBACK_ERROR = 'error';
```
jviotti: Actually, can we move this constants inside the `getDriveFeedback()` function? They are very local to that piece of code, and probably don't make much sense as globals for the entire file.

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,133 @@ describe('Shared: DriveConstraints', function() {
...
+          name: 'USB Drive',
+          size: 100000000100
+        }, {
+          path: '/mnt/test/rpi.img',
+          size: 1000000000
+        });
+
+        m.chai.expect(result).to.deep.equal({
+          type: 'error',
+          message: 'Drive Contains Image'
```
jviotti: Lets export the messages constants, so we can make use of them here rather than hardcoding the messages.

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,133 @@ describe('Shared: DriveConstraints', function() {
 
   });
 
+  describe('.getDriveFeedback', function() {
+
+    describe('given there should be no errors or warnings', () => {
```
jviotti: I think this should be `given there are no errors or warnings`

jviotti:
I think we should add unit tests to determine what happens if two things go wrong. For example, what happens if the drive is a system drive, but its also too small? There are probably quite some combinations we can test.

lib/shared/drive-constraints.js
lurch: Yeah, I'm guessing "feedback" was chosen because Benedict was coming at this from a "visual feedback" point of view? I don't think `.getDriveImageFlashingMessage` works either, because it sounds like it might be the message displayed _during_ the flashing process.
What about something like `.getDriveImageCompatibilityStatus` ? (since it's effectively telling you whether a given image is 'allowed' to be flashed to a given drive, and the reason why it's not otherwise)

Also, perhaps we should check that this function works with a 'null' image supplied, so that we don't have to revisit this function if we swap around the GUI order to select the drive before the image?

lib/shared/drive-constraints.js
lurch: If unsafe mode is enabled, IMHO flashing a system drive should be downgraded from an error to a warning, and not simply treated as "not an error". But I have no preferences on the logic / mechanisms used to achieve that.

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,85 @@ exports.isDriveValid = (drive, image) => {
 exports.isDriveSizeRecommended = (drive, image) => {
   return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'recommendedDriveSize' ], UNKNOWN_SIZE);
 };
+
+const feedbackMessages = {
+  NOTREC: 'Not Recommended',
```
lurch: That's exactly the sort of thing I meant with "IMHO the constants should be given better (longer) names." :-)

lib/shared/drive-constraints.js
lurch: > But I have no preferences on the logic / mechanisms used to achieve that.

Although having said that, I guess it depends if we feel the unsafe-mode setting is part of the core application itself, or whether it's "just" a concern of the GUI sitting on top? (IYSWIM!)

lurch:
> I think we should add unit tests to determine what happens if two things go wrong. For example, what happens if the drive is a system drive, but its also too small? There are probably quite some combinations we can test.

Agreed. But given that we're just doing a "linear lookup", e.g.
```
if (conditionA) {
  return errorA;
} else if (conditionB) {
  return errorB;
} else if (conditionC) {
  return errorC;
}
```
then in addition to testing conditions A, B & C in isolation, I think it's probably just sufficient to check "conditionA + conditionB" and "conditionB + conditionC"; i.e. we don't need to explicitly test "conditionA + conditionC" or "conditionA + conditionB + conditionC" ?

lurch:
Hmmm, extremely odd that all the Appveyor builds are failing with the same unit-test failure, but all the Travis builds are passing. What's going on there then?? :-/

lib/shared/drive-constraints.js
Shou: Yup, you hit the nail on the head. :)

`.getDriveImageCompatibilityStatus` sounds good to me.

Nice catch, I should test for when the drive is `null` too, in case the behaviour ever changes.

jviotti: Yeah, I like that name!

> Nice catch, I should test for when the drive is null too, in case the behaviour ever changes.

Would it make sense if we make the function return the "OK" status if any of the arguments is null/undefined?

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,133 @@ describe('Shared: DriveConstraints', function() {
...
+        m.chai.expect(result).to.equal(null);
+      });
+    });
+
+    describe('given the drive contains the image', () => {
+
+      it('should return the source error', function() {
+        const result = constraints.getDriveFeedback({
+          device: '/dev/disk1',
+          mountpoints: [ {
```
jviotti: Can we put the curly braces in separate strings, for consistency purposes?

```
mountpoints: [
    {
        ...
    }
]
```

jviotti:
> Hmmm, extremely odd that all the Appveyor builds are failing with the same unit-test failure, but all the Travis builds are passing. What's going on there then?? :-/

@lurch @Shou The error happens when testing for the "source drive" case, which uses the `exports.isSourceDrive()` function. Since that function checks the paths, it will only work on Windows when using Windows-like paths. Can we edit the test cases to use Windows paths on Windows, and UNIX paths otherwise?

You can probably do something like this for the whole function tests:

```js
beforeEach(function() {
    this.mountpoints = 'E:\\';
  } else {
    this.mountpoints = '/mnt/foo';
  }
});
```

And re-use `this.mountpoints` everywhere.

lib/shared/drive-constraints.js
Shou: That's where I've ended up now. The test currently returns "Too Small For Image", and it's not completely wrong considering the drive is nonexistent. `OK` sounds weird here since it's really not okay, but making another message like `EMPTY` solely for this might not be ideal either? Maybe we can change from `OK` to sound less validating and more neutral?

lurch: Perhaps we need separate `DRIVE_OK`, `IMAGE_OK` and `DRIVE_AND_IMAGE_OK` statuses?

tests/shared/drive-constraints.spec.js
lurch: I think @jviotti  meant 'on separate lines' rather than 'in separate strings' ;-)

jviotti: Haha, my bad!

Shou:
Any thoughts on this commit?

jviotti:
Oh, I didn't realise this was updated. I'll review again.

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,115 @@ exports.isDriveValid = (drive, image) => {
...
+/**
+ * @summary Drive/image compatibility status messages.
+ * @public
+ *
+ * @description
+ * Status messages intended to be shown to the user.
+ *
+ * @type {Object}
+ */
+exports.COMPATIBILITY_STATUS_MESSAGES = {
```
jviotti: Maybe we can descriptions of what each status means, like in https://github.com/resin-io/etcher/blob/master/lib/shared/exit-codes.js#L24? For example, that "locked" means an SD Card locking (vs something else), etc?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,115 @@ exports.isDriveValid = (drive, image) => {
...
 };
+
+/**
+ * @summary Drive/image compatibility status messages.
+ * @public
+ *
+ * @description
+ * Status messages intended to be shown to the user.
+ *
+ * @type {Object}
```
jviotti: Can we move `@type` right below `@public`, for consistency purposes?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,115 @@ exports.isDriveValid = (drive, image) => {
...
+/**
+ * @summary Drive/image compatibility status types.
+ * @public
+ *
+ * @description
+ * Status types classifying what kind of message it is, i.e. ok, error, warning.
+ *
+ * @type {Object}
+ */
+exports.COMPATIBILITY_STATUS_TYPES = {
```
jviotti: I feel we're doing C programming :P

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,115 @@ exports.isDriveValid = (drive, image) => {
...
+  OK: 0,
+  ERROR: 1,
+  WARNING: 2
+};
+
+/**
+ * @summary Get drive feedback in an object, if any
```
jviotti: Should this summary be updated, now that the function has changed?

lib/shared/drive-constraints.js
jviotti: Looks like we should explain the "OK" message here.

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,115 @@ exports.isDriveValid = (drive, image) => {
...
+ * @public
+ *
+ * @description
+ * If the drive-image combination results in a warning or error, this function
+ * returns an object containing the feedback type ('error', 'warning') and the
+ * feedback message.
+ *
+ * @param {Object} drive - drive
+ * @param {Object} image - image
+ * @returns {Object} feedback object
```
jviotti: Maybe this should be changed to capability status object now?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,115 @@ exports.isDriveValid = (drive, image) => {
...
+  }
+
+  if (exports.isSystemDrive(drive)) {
+    return {
+      type: exports.COMPATIBILITY_STATUS_TYPES.ERROR,
+      message: exports.COMPATIBILITY_STATUS_MESSAGES.SYSTEM
+    };
+  }
+
+  if (Boolean(drive) && !exports.isDriveLargeEnough(drive, image)) {
```
jviotti: Is the intention here to catch whether drive is not defined? If so, I think the right way to do it would be by using `_.isNil()`.

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,229 @@ describe('Shared: DriveConstraints', function() {
...
+        });
+
+        m.chai.expect(result).to.deep.equal({
+          type: constraints.COMPATIBILITY_STATUS_TYPES.OK,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.OK
+        });
+      });
+    });
+
+    describe('given that everything is wrong', () => {
```
jviotti: lol, that sounds really pessimistic :P

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,229 @@ describe('Shared: DriveConstraints', function() {
...
+        const result = constraints.getDriveImageCompatibilityStatus(this.drive, {
+          path: '/mnt/disk2/rpi.img',
+          size: 1000000000
+        });
+
+        m.chai.expect(result).to.deep.equal({
+          type: constraints.COMPATIBILITY_STATUS_TYPES.OK,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.OK
+        });
+      });
```
jviotti: In all these test cases there is a blank between `describe` and `it`. Can we add a blank line here as well for symmetric purposes?

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,229 @@ describe('Shared: DriveConstraints', function() {
 
   });
 
+  describe('.getDriveImageCompatibilityStatus', function() {
+
+    beforeEach(function() {
+      if (process.platform === 'win32') {
+        this.mountpoint = 'E:\\foo';
```
lurch: Mountpoints on Windows are just short strings like 'E:' https://github.com/resin-io-modules/drivelist

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -15,35 +15,15 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
-              ng-show="modal.constraints.isSourceDrive(drive, modal.state.getImage())">
-                <i class="glyphicon glyphicon-circle-arrow-down"></i>
-                Source Drive</span>
+            <span class="label"
+              ng-class="{
+                'label-warning': modal.getDriveImageCompatibilityStatus(drive).type === modal.constraints.COMPATIBILITY_STATUS_TYPES.WARNING,
+                'label-warning': modal.constraints.isDriveValid(drive, modal.state.getImage()) && modal.getDriveImageCompatibilityStatus(drive).message === modal.constraints.COMPATIBILITY_STATUS_MESSAGES.SYSTEM,
+                'label-danger': modal.getDriveImageCompatibilityStatus(drive).type === modal.constraints.COMPATIBILITY_STATUS_TYPES.ERROR
+              }"
+              ng-show="modal.getDriveImageCompatibilityStatus(drive).message">
```
lurch: This should be changed to something like `modal.getDriveImageCompatibilityStatus(drive).type === modal.constraints.COMPATIBILITY_STATUS_TYPES.OK`

lib/shared/drive-constraints.js
lurch: Perhaps doing just:
```js
exports.COMPATIBILITY_STATUS_TYPE_OK = 0;
exports.COMPATIBILITY_STATUS_TYPE_ERROR = 1;
exports.COMPATIBILITY_STATUS_TYPE_WARNING = 2;
```
would feel more like Javascript and less like C? \*shrug\*

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,229 @@ describe('Shared: DriveConstraints', function() {
...
+        });
+
+        m.chai.expect(result).to.deep.equal({
+          type: constraints.COMPATIBILITY_STATUS_TYPES.WARNING,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.NOT_RECOMMENDED
+        });
+      });
+    });
+
+    describe('given the image is null', () => {
```
lurch: We should also check that it behaves as expected if e.g. the image is null, but the drive is locked

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,229 @@ describe('Shared: DriveConstraints', function() {
...
+        });
+
+        m.chai.expect(result).to.deep.equal({
+          type: constraints.COMPATIBILITY_STATUS_TYPES.ERROR,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.CONTAINS_IMAGE
+        });
+      });
+    });
+
+    describe('given everything but source drive is wrong', () => {
```
lurch: As I mentioned in an earlier comment, I think it's sufficient to just do pair-wise combination checks, to check that for each pair the expected error gets returned first. So something like:
conditionSourceDrive & conditionSystemDrive => errorSourceDrive
conditionSystemDrive & conditionTooSmall => errorSystemDrive
conditionTooSmall & conditionLocked => errorTooSmall
conditionLocked & conditionNotRecommended => errorLocked

A bit like by checking `1 < 2` and `2 < 3` and `3 < 4`, you automatically know that `1 < 3`, `1 < 4`, `2 < 4` without having to explicitly check all combinations ;-)

tests/shared/drive-constraints.spec.js
Shou: Does this mean [this](https://github.com/resin-io/etcher/blob/ae99b8b43506c667ff5282e27d2505b3bbdad3fe/tests/shared/drive-constraints.spec.js#L468) is wrong too? Should I fix it?

lurch: Theoretically yes. Up to @jviotti to confirm.

Shou:
Here's another update! I followed most of the advice here, additions include the new transitive-style tests that @lurch suggested and testing when images are `null` where applicable.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +170,26 @@ module.exports = function(
...
+   * containing the status type (ERROR, WARNING, and OK), and accompanying
+   * status message.
+   * Almost just a synonym of its namesake, the difference is that the image
+   * argument is implicit here for the sake of shortening.
+   *
+   * @param {Object} drive - drive
+   * @returns {Object} compatibility status object
+   *
+   * @example
+   * let { type, message } = DriveSelectorController.getDriveImageCompatibilityStatus(drive);
```
jviotti: Should we encourage the use of `const` here, for pure functional programming's sake?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,115 @@ exports.isDriveValid = (drive, image) => {
...
+ *   // do something
+ * } else if (type === constraints.COMPATIBILITY_STATUS_TYPES.ERROR) {
+ *   // do something else
+ * }
+ */
+exports.getDriveImageCompatibilityStatus = (drive, image) => {
+
+  // Mind the order of the if-statements if you modify.
+  if (exports.isSourceDrive(drive, image)) {
+    return {
```
jviotti: I think its fine as it is. Storing the object instances somewhere looks like to much hassle to me.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -15,35 +15,15 @@ <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gi
...
-                System Drive</span>
-
-            <span class="label label-danger"
-              ng-show="modal.constraints.isSourceDrive(drive, modal.state.getImage())">
-                <i class="glyphicon glyphicon-circle-arrow-down"></i>
-                Source Drive</span>
+            <span class="label"
+              ng-class="{
+                'label-warning': modal.getDriveImageCompatibilityStatus(drive).type === modal.constraints.COMPATIBILITY_STATUS_TYPES.WARNING,
+                'label-warning': modal.constraints.isDriveValid(drive, modal.state.getImage()) && modal.getDriveImageCompatibilityStatus(drive).message === modal.constraints.COMPATIBILITY_STATUS_MESSAGES.SYSTEM,
```
jviotti: @Shou I think this line is unnecessary if the fact that a drive is a system drive is made a warning rather than an error in the drive constraints file, which sounds logical IMHO.

lib/gui/components/drive-selector/controllers/drive-selector.js
Shou: I'm always open to smuggling functional programming in.

Shou:
Here's that small change from `let` → `const`.

lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
Shou: Oh I missed this. Sure, so we just wholly change it a warning instead?

jviotti: I think it makes a lot of sense as just a warning.

Shou: I pushed the change for this now

jviotti:
@Shou Approved, however the tests are failing after you made the system drive scenario a warning.

Shou:
Oh yeah, forgot to update the tests to compare against `WARNING`. Here we go.

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,162 @@ exports.isDriveValid = (drive, image) => {
...
+  }
+
+  if (exports.isSystemDrive(drive)) {
+    return {
+      type: exports.COMPATIBILITY_STATUS_TYPES.WARNING,
+      message: exports.COMPATIBILITY_STATUS_MESSAGES.SYSTEM
+    };
+  }
+
+  if (!_.isNil(drive) && !exports.isDriveLargeEnough(drive, image)) {
```
lurch: We should probably re-order these so that any ERROR conditions get tested before any WARNING conditions?
AIUI the code as it stands currently would mean that a drive that was both "system" and "locked" would only be labelled with "warning: system", whereas it probably makes sense to label it with "error: locked" ?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,162 @@ exports.isDriveValid = (drive, image) => {
...
+  }
+
+  if (exports.isDriveLocked(drive)) {
+    return {
+      type: exports.COMPATIBILITY_STATUS_TYPES.ERROR,
+      message: exports.COMPATIBILITY_STATUS_MESSAGES.LOCKED
+    };
+  }
+
+  if (!_.isNil(drive) && !exports.isDriveSizeRecommended(drive, image)) {
```
lurch: Hmmm, based on my comment above, this is a tricky one. If the drive is both a system drive _AND_ smaller than the image's recommended size, which warning should we display? From a logical point of view, the user really ought to be made aware of _both_ warnings, because they both signify different things.
If we only display the "system" warning, then the user isn't aware that the drive is also smaller than the image's recommended size.
If we display only the "not recommended" warning, then the user isn't aware that they'll potentially be overwriting one of their system drives.
:-S

(but given the choice of only displaying one warning, the "system" warning obviously has higher priority)

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,162 @@ exports.isDriveValid = (drive, image) => {
...
+ * @type {Object}
+ *
+ * @description
+ * Status types classifying what kind of message it is, i.e. ok, error, warning.
+ */
+exports.COMPATIBILITY_STATUS_TYPES = {
+  OK: 0,
+  ERROR: 1,
+  WARNING: 2
+};
```
lurch: Doesn't really make much difference, but I wonder if it might make sense to re-order these as:
```
  OK: 0,
  WARNING: 1,
  ERROR: 2
```
so that a higher number indicates higher severity?

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,266 @@ describe('Shared: DriveConstraints', function() {
...
+          type: constraints.COMPATIBILITY_STATUS_TYPES.ERROR,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.CONTAINS_IMAGE
+        });
+      });
+
+    });
+
+    describe('given the drive is a system drive', () => {
+
+      it('should return the system drive error', function() {
```
lurch: "system drive warning"

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,266 @@ describe('Shared: DriveConstraints', function() {
...
+          type: constraints.COMPATIBILITY_STATUS_TYPES.OK,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.OK
+        });
+      });
+
+    });
+
+    describe('given the drive contains the image', () => {
+
+      it('should return the source error', function() {
```
lurch: Given the names used in the test-code here, should this be referring to "contains-image error" rather than "source error"?

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,266 @@ describe('Shared: DriveConstraints', function() {
...
+
+    describe('given there are no errors or warnings', () => {
+
+      it('should return OK', function() {
+        const result = constraints.getDriveImageCompatibilityStatus(this.drive, {
+          path: '/mnt/disk2/rpi.img',
+          size: 1000000000
+        });
+
+        m.chai.expect(result).to.deep.equal({
```
lurch: Hmmm, I guess using `.to.deep.equal` means that all the unit-tests will instantly break if we want to add any extra attributes the the status-objects. Would a more flexible approach be to check:
```
m.chai.expect(result.type).to.equal(constraints.COMPATIBILITY_STATUS_TYPES.OK);
m.chai.expect(result.message).to.equal(constraints.COMPATIBILITY_STATUS_MESSAGES.OK);
```
?  \*shrug\*

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,162 @@ exports.isDriveValid = (drive, image) => {
...
+exports.COMPATIBILITY_STATUS_MESSAGES = {
+
+  /**
+   * @property {String} OK
+   * @memberof COMPATIBILITY_STATUS_MESSAGES
+   *
+   * @description
+   * The image and drive compatibility is OK.
+   */
+  OK: '',
```
lurch: It's slightly "symmetry-breaking", but rather than assigning an empty string to a not-actually-used-anywhere constant, I guess we could define the status objects returned by `getDriveImageCompatibilityStatus` to only have a `.message` attribute when the `.type` is `WARNING` or `ERROR` ?

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,266 @@ describe('Shared: DriveConstraints', function() {
...
+          type: constraints.COMPATIBILITY_STATUS_TYPES.ERROR,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.LOCKED
+        });
+      });
+
+    });
+
+    describe('given the drive is smaller than the recommended size', () => {
+
+      it('should return the smaller than recommended size warning', function() {
```
lurch: Again, maybe in the unit-tests it's better for the `it` string to more closely reflect the actual error name? (rather than the internal meaning of the error). IYSWIM ;)

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,266 @@ describe('Shared: DriveConstraints', function() {
...
+          type: constraints.COMPATIBILITY_STATUS_TYPES.OK,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.OK
+        });
+      });
+
+    });
+
+    describe('given a system drive and image is null', () => {
+
+      it('should return system drive error', function() {
```
lurch: warning ;)

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +556,266 @@ describe('Shared: DriveConstraints', function() {
...
+          type: constraints.COMPATIBILITY_STATUS_TYPES.WARNING,
+          message: constraints.COMPATIBILITY_STATUS_MESSAGES.SYSTEM
+        });
+      });
+
+    });
+
+    describe('given a too small and locked drive', () => {
+
+      it('should return the too small error by precedence', function() {
```
lurch: Maybe a "locked" error should be considered more important than a "too small" error?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,162 @@ exports.isDriveValid = (drive, image) => {
...
+
+/**
+ * @summary Drive/image compatibility status messages.
+ * @public
+ * @type {Object}
+ *
+ * @description
+ * Status messages intended to be shown to the user.
+ */
+exports.COMPATIBILITY_STATUS_MESSAGES = {
```
lurch: _Maybe_ it might make sense for these to be moved to `lib/shared/messages.js` ?

lib/shared/drive-constraints.js
jviotti: Sounds good to me!

lib/shared/drive-constraints.js
jviotti: I suggest to keep it as it is for now. The `lib/shared/messages.js` is a mess I'm looking forward to fix soon, given that their structure no longer matches what we're imposing on `lib/shared/errors.js`. I'll send a PR that refactors all these occurrences next week.

lib/shared/drive-constraints.js
Shou: Maybe we should turn it into a list of errors and warnings to be displayed, which comes with the benefit of 'OK' simply being represented by an empty list.

tests/shared/drive-constraints.spec.js
Shou: That's true.

lib/shared/drive-constraints.js
lurch: Doesn't returning a list defeat the whole original purpose of this PR? :confused: 

Although yeah from a logical point of view it probably makes sense for `getDriveImageCompatibilityStatus` to return a (possibly-empty) set of 'problems', and then there should be another method which decides which problem(s) are "most important" and should be displayed in the GUI / CLI ?

Shou: Originally it was just a bugfix but it's turned into a refactor meant to distance ourselves from Angular more so that the transition to React will be smoother. I will update the title to reflect that, because it's gone beyond the bugfix, and the point you raise here is an important one—we should show both warnings in this case.

jviotti: > If the drive is both a system drive AND smaller than the image's recommended size, which warning should we display? From a logical point of view, the user really ought to be made aware of both warnings, because they both signify different things.
If we only display the "system" warning, then the user isn't aware that the drive is also smaller than the image's recommended size.
If we display only the "not recommended" warning, then the user isn't aware that they'll potentially be overwriting one of their system drives.

I'd like to hear @konmouz opinion here. Which one should we show? Also, should we support more than one label on each drive in order to handle this case? If so, we can make this function returns a list containing one or two labels the UI should show.

jviotti:
@lurch How do you feel about this PR now?

tests/shared/drive-constraints.spec.js
```diff
@@ -556,4 +557,240 @@ describe('Shared: DriveConstraints', function() {
...
+
+      it('should return both warnings', function() {
+        this.drive.system = true;
+
+        const result = constraints.getDriveImageCompatibilityStatuses(this.drive, {
+          path: '/mnt/disk2/rpi.img',
+          size: 1000000000,
+          recommendedDriveSize: 5000000000
+        });
+        const expectedTuples = [ [ 'WARNING', 'SYSTEM' ], [ 'WARNING', 'SIZE_NOT_RECOMMENDED' ] ];
```
lurch: Given that this is checking against a list of tuples (rather than a set of tuples) does this mean this test would fail if `.getDriveImageCompatibilityStatuses` returned a list with 'SIZE_NOT_RECOMMENDED' before 'SYSTEM'? IMHO it might be nice if the test would still pass in that case?

lurch:
> @lurch How do you feel about this PR now?

Sorry for taking so long to get back to this - I've just added some more comments.

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,153 @@ exports.isDriveValid = (drive, image) => {
...
+  const list = [];
+
+  // Mind the order of the if-statements if you modify.
+  if (exports.isSourceDrive(drive, image)) {
+    list.push({
+      type: exports.COMPATIBILITY_STATUS_TYPES.ERROR,
+      message: exports.COMPATIBILITY_STATUS_MESSAGES.CONTAINS_IMAGE
+    });
+
+  } else if (exports.isDriveLocked(drive)) {
```
Shou: Maybe we could make the priority part of the error message object?

lurch: Yeah, after leaving my previous comments I started wondering if instead of returning a list of lists like `[ [ 'WARNING', 'SYSTEM' ], [ 'WARNING', 'SIZE_NOT_RECOMMENDED' ] ]`, it might be nicer to return something like `{ warnings: [ 'SYSTEM', 'SIZE_NOT_RECOMMENDED' ] }` ?

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,153 @@ exports.isDriveValid = (drive, image) => {
...
+ *     // do something
+ *   } else if (type === constraints.COMPATIBILITY_STATUS_TYPES.ERROR) {
+ *     // do something else
+ *   }
+ * }
+ */
+exports.getDriveImageCompatibilityStatuses = (drive, image) => {
+  const list = [];
+
+  // Mind the order of the if-statements if you modify.
```
jviotti: I can't avoid to hear "mind the gap between the train and the platform" in my head :P

lib/shared/drive-constraints.js
```diff
@@ -208,3 +208,153 @@ exports.isDriveValid = (drive, image) => {
...
+/**
+ * @summary Drive/image compatibility status types.
+ * @public
+ * @type {Object}
+ *
+ * @description
+ * Status types classifying what kind of message it is, i.e. ok, error, warning.
+ */
+exports.COMPATIBILITY_STATUS_TYPES = {
+  OK: 0,
```
jviotti: Yep, looks like we should remove `OK`.

lib/shared/drive-constraints.js
Shou: So apparently there are Sets in ECMA6 which is perfect! Although there is no set comparator operation in ECMA6 (hah) we can use Lodash `isEqual`.

tests/shared/drive-constraints.spec.js
Shou: So there are Sets in ECMA6, that's perfect here! Although there is no set comparator operation in ECMA6 (hah) we can use Lodash `isEqual`.

Shou:
Thoughts on this one now? The tests should allow unordered tuples.

tests/shared/drive-constraints.spec.js
```diff
@@ -771,4 +772,241 @@ describe('Shared: DriveConstraints', function() {
...
+
+      // Sort so that order doesn't matter
+      const expectedTypesSorted = _.sortBy(_.map(expectedTuples, (tuple) => {
+        return constraints.COMPATIBILITY_STATUS_TYPES[tuple[0]];
+      }));
+      const expectedMessagesSorted = _.sortBy(_.map(expectedTuples, (tuple) => {
+        return constraints.COMPATIBILITY_STATUS_MESSAGES[tuple[1]];
+      }));
+      const resultTypesSorted = _.sortBy(_.map(resultList, 'type'));
+      const resultMessagesSorted = _.sortBy(_.map(resultList, 'message'));
```
lurch: I'm not 100% sure what this does, but does it mean that e.g. `[ ['ERROR', 'B'], ['WARNING', 'A'] ]` would compare as identical to `[ ['ERROR', 'A'], ['WARNING', 'B'] ]` ?  (which IMHO would be wrong)

I dunno how sorting works in JavaScript / NodeJS / Lodash but can you use a 'nested' comparison function, so that it first sorts by `tuple[0]` and then sorts by `tuple[1]` ? So that e.g. `[ ['ERROR', 'Z'], ['WARNING', 'A'], ['ERROR', 'B'] ]` would sort as `[ ['ERROR', 'B'], ['ERROR', 'Z'], ['WARNING', 'A'] ]`

lib/shared/drive-constraints.js
```diff
@@ -232,3 +232,153 @@ exports.isDriveValid = (drive, image) => {
...
+ *
+ * for ({ type, message } of statuses) {
+ *   if (type === constraints.COMPATIBILITY_STATUS_TYPES.WARNING) {
+ *     // do something
+ *   } else if (type === constraints.COMPATIBILITY_STATUS_TYPES.ERROR) {
+ *     // do something else
+ *   }
+ * }
+ */
+exports.getDriveImageCompatibilityStatuses = (drive, image) => {
```
lurch: Apologies for jumping down the rabbit hole :wink: but is there some overlap between `isDriveValid` and `getDriveImageCompatibilityStatuses` ? Could they share code somehow?

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +170,85 @@ module.exports = function(
...
+  // The problem is that every time you call `myImmutableObject.toJS()`
+  // you will get a new object, whose reference is different from
+  // the one you previously got, even if the data is exactly the same.
+  /**
+   * @private
+   * @param {Function} func - function
+   * @param {Function} compare - function
+   * @returns {Function} function
+   */
+  const memoizeImmutableListReference = (func, compare = _.isEqual) => {
```
Shou: Definitely! I was wondering where to put it, but now we have `lib/shared/utils.js` so that's solved.

lib/shared/drive-constraints.js
Shou: I tried this, but the line ended up longer than before when using `.getDriveImageCompatibilityStatuses()` in `.isDriveValid()` and it'd be slightly less efficient because we'd also check for warnings. Maybe you had something else in mind though!

Shou:
Sorry for the mess, once #1253 is merged I'll rebase and all of this should go away. For now though, the changes are [here](https://github.com/resin-io/etcher/pull/1193/commits/2cc3269a78af193d591c787dc439bfb044548f23). Notable changes:

 * New utility function `.removeAngularHashes()` and tests
 * Removed unnecessary memoize usage
 * Remove 'OK' and usage (somehow this came back from the grave)

CHANGELOG.md
```diff
@@ -3,6 +3,45 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## v1.0.0-rc.2 - 2017-04-11
```
jviotti: Why is this PR including the changelog?

jviotti:
Something seems to have gone wrong with a rebase, since there are a lot of unrelated commits in here.

jviotti:
> Sorry for the mess, once #1253 is merged I'll rebase and all of this should go away. For now though, the changes are here. Notable changes:

Oh, I missed this comment, sorry about that :)

lib/gui/components/drive-selector/controllers/drive-selector.js
jviotti: We should definitely about duplicating this, however `lib/shared` it not the place, since this hack is *very* Angular.js specific.

In this comment (https://github.com/resin-io/etcher/pull/1265#discussion_r111381085), I realised the only other occurrence of this hack is also specific to this particular component, so this could be its only home for now.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -171,25 +172,86 @@ module.exports = function(
...
+   *   }
+   * }
    */
-  this.getDriveImageCompatibilityStatus = (drive) => {
-    return this.constraints.getDriveImageCompatibilityStatus(drive, this.state.getImage());
-  };
+  this.getDrivesAndStatuses = memoizeImmutableListReference(() => {
+    const statusList = [];
+
+    for (const drive of this.drives.getDrives()) {
```
jviotti: This function looks like a perfect case for a `_.map`.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -171,25 +172,86 @@ module.exports = function(
...
+      statusList.push({
+        drive,
+        statuses: this.constraints.getDriveImageCompatibilityStatuses(drive, this.state.getImage())
+      });
+    }
+
+    return statusList;
+  }, (newList, oldList) => {
+    if (newList.length !== oldList.length) {
```
jviotti: @Shou Can you give me some context for why this is needed? Is there a way we can simplify this whole thing?

Shou: We can remove it now, it's a relic of the block since I previously used `zipWith` which doesn't truncate in Lodash, however it is slightly more efficient when kept because it only checks the length at `O(1)` versus the structure at `O(n)`, but `n` is usually lower than the tens here so it might not be much of a problem...

lib/gui/components/drive-selector/controllers/drive-selector.js
Shou: I missed that, good call! I really like how functional traversions clarify intent.

Shou:
Fixed and rebased, but the test seems stuck.

jviotti:
@Shou You can amend the commit with no changes so the timestamp gets updated, and do a force push afterwards.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +171,81 @@ module.exports = function(
...
+      return {
+        drive,
+        statuses: this.constraints.getDriveImageCompatibilityStatuses(drive, this.state.getImage())
+      };
+    });
+
+    return statusList;
+  }, (newList, oldList) => {
+    return _.isEqual(
+      utils.removeAngularHashes(newList),
```
jviotti: I keep thinking there must be a better way to solve this angular hashes and immutable list references mess. I'll be doing some experimentation today.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +171,81 @@ module.exports = function(
...
+    };
+  };
+
+  /**
+   * @summary Get the list of drives and their compatibility status objects
+   * @function
+   * @public
+   *
+   * @description
+   * For each drive, return its compatibility status with the seleted image,
```
lurch: seleted -> selected

lib/shared/drive-constraints.js
lurch: Fair enough. I guess we'll just have to be careful that the logic in `isDriveValid` and `getDriveImageCompatibilityStatuses` doesn't get out of sync.

tests/shared/drive-constraints.spec.js
```diff
@@ -771,4 +772,233 @@ describe('Shared: DriveConstraints', function() {
...
+
+        m.chai.expect(result).to.deep.equal([]);
+      });
+
+    });
+
+    describe('given the drive contains the image', () => {
+
+      it('should return the contains-image error', function() {
+        this.image.path = `${this.mountpoint}${this.separator}rpi.img`;
```
lurch: Probably better to do something similar to what https://github.com/resin-io/etcher/blob/master/tests/shared/drive-constraints.spec.js#L134 does, rather than manually building paths by concatenating strings.

lib/gui/components/drive-selector/controllers/drive-selector.js
jviotti: OK, I had great success with this! Here's a diff with my local changes:

```diff
diff --git a/lib/gui/components/drive-selector/controllers/drive-selector.js b/lib/gui/components/drive-selector/controllers/drive-selector.js
index df8fc8d..1579ed0 100644
--- a/lib/gui/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/components/drive-selector/controllers/drive-selector.js
@@ -16,6 +16,7 @@
 
 'use strict';
 
+const angular = require('angular');
 const _ = require('lodash');
 const messages = require('../../../../shared/messages');
 const constraints = require('../../../../shared/drive-constraints');
@@ -197,13 +198,13 @@ module.exports = function(
    *
    * const memoizedFunction = memoizeImmutableListReference(getList);
    */
-  const memoizeImmutableListReference = (func, compare = _.isEqual) => {
+  const memoizeImmutableListReference = (func) => {
     let previous = [];
 
     return (...restArgs) => {
       const list = Reflect.apply(func, this, restArgs);
 
-      if (!_.isEqualWith(list, previous, compare)) {
+      if (!angular.equals(list, previous)) {
         previous = list;
       }
 
@@ -211,41 +212,12 @@ module.exports = function(
     };
   };
 
-  /**
-   * @summary Get the list of drives and their compatibility status objects
-   * @function
-   * @public
-   *
-   * @description
-   * For each drive, return its compatibility status with the seleted image,
-   * containing the status type (ERROR, WARNING), and accompanying
-   * status message.
-   *
-   * @returns {Object[]} list of objects containing drive and statuses
-   *
-   * @example
-   * const metadrives = DriveSelectorController.getDrivesAndStatuses();
-   *
-   * for ({ drive, statuses } of metadrives) {
-   *   for ({ type, message } of statuses) {
-   *     // do something
-   *   }
-   * }
-   */
-  this.getDrivesAndStatuses = memoizeImmutableListReference(() => {
-    const statusList = _.map(this.drives.getDrives(), (drive) => {
-      return {
-        drive,
-        statuses: this.constraints.getDriveImageCompatibilityStatuses(drive, this.state.getImage())
-      };
-    });
+  this.getDrives = memoizeImmutableListReference(() => {
+    return this.drives.getDrives();
+  });
 
-    return statusList;
-  }, (newList, oldList) => {
-    return _.isEqual(
-      utils.removeAngularHashes(newList),
-      utils.removeAngularHashes(oldList)
-    );
+  this.getDriveStatuses = memoizeImmutableListReference((drive) => {
+    return this.constraints.getDriveImageCompatibilityStatuses(drive, this.state.getImage());
   });
 
 };
diff --git a/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html b/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
index 690782a..3525ea6 100644
--- a/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
+++ b/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
@@ -5,17 +5,17 @@
 
 <div class="modal-body">
   <ul class="list-group">
-    <li class="list-group-item" ng-repeat="meta in modal.getDrivesAndStatuses() track by meta.drive.device"
-      ng-disabled="!modal.constraints.isDriveValid(meta.drive, modal.state.getImage())"
-      ng-dblclick="modal.selectDriveAndClose(meta.drive)"
-      ng-click="modal.toggleDrive(meta.drive)">
+    <li class="list-group-item" ng-repeat="drive in modal.getDrives() track by drive.device"
+      ng-disabled="!modal.constraints.isDriveValid(drive, modal.state.getImage())"
+      ng-dblclick="modal.selectDriveAndClose(drive)"
+      ng-click="modal.toggleDrive(drive)">
         <div>
-          <h4 class="list-group-item-heading">{{ meta.drive.description }} - {{ meta.drive.size | gigabyte | number:1 }} GB</h4>
-          <p class="list-group-item-text">{{ meta.drive.name }}</p>
+          <h4 class="list-group-item-heading">{{ drive.description }} - {{ drive.size | gigabyte | number:1 }} GB</h4>
+          <p class="list-group-item-text">{{ drive.name }}</p>
 
           <footer class="list-group-item-footer">
 
-            <span class="label" ng-repeat="status in meta.statuses"
+            <span class="label" ng-repeat="status in modal.getDriveStatuses(drive)"
               ng-class="{
                 'label-warning': status.type === modal.constraints.COMPATIBILITY_STATUS_TYPES.WARNING,
                 'label-danger': status.type === modal.constraints.COMPATIBILITY_STATUS_TYPES.ERROR
@@ -26,8 +26,8 @@
           </footer>
         </div>
         <span class="tick tick--success"
-          ng-show="modal.constraints.isDriveValid(meta.drive, modal.state.getImage())"
-          ng-disabled="!modal.state.isCurrentDrive(meta.drive.device)"></span>
+          ng-show="modal.constraints.isDriveValid(drive, modal.state.getImage())"
+          ng-disabled="!modal.state.isCurrentDrive(drive.device)"></span>
     </li>
     <li class="list-group-item"
       ng-show="!modal.drives.hasAvailableDrives()">
diff --git a/lib/shared/utils.js b/lib/shared/utils.js
index 337c7f9..c9e98e3 100644
--- a/lib/shared/utils.js
+++ b/lib/shared/utils.js
@@ -76,32 +76,3 @@ exports.percentageToFloat = (percentage) => {
 
   return percentage / PERCENTAGE_MAXIMUM;
 };
-
-/**
- * @summary Traverse an object and remove Angular hashes
- * @function
- * @public
- *
- * @param {Object} object - object
- * @returns {Object} object
- *
- * @example
- * const angularObjA = { $$hashKey: 'abc' };
- * const angularObjB = { $$hashKey: '123' };
- *
- * const areObjectsEqual = _.isEqual(
- *   utils.removeAngularHashes(angularObjA),
- *   utils.removeAngularHashes(angularObjB)
- * );
- */
-exports.removeAngularHashes = (object) => {
-  if (_.isArray(object)) {
-    return _.map(object, exports.removeAngularHashes);
-  }
-  if (_.isObject(object)) {
-    const omitted = _.omit(object, [ '$$hashKey' ]);
-    return _.mapValues(omitted, exports.removeAngularHashes);
-  }
-
-  return object;
-};
diff --git a/tests/shared/utils.spec.js b/tests/shared/utils.spec.js
index cea0b1d..fd604ec 100644
--- a/tests/shared/utils.spec.js
+++ b/tests/shared/utils.spec.js
@@ -131,74 +131,4 @@ describe('Shared: Utils', function() {
 
   });
 
-  describe('.removeAngularHashes()', function() {
-
-    it('should remove direct hashes', function() {
-      m.chai.expect(utils.removeAngularHashes({
-        attrA: 1,
-        $$hashKey: '123'
-      })).to.deep.equal({
-        attrA: 1
-      });
-    });
-
-    it('should remove hashes from objects in list', function() {
-      m.chai.expect(utils.removeAngularHashes([
-        {
-          attrA: 1,
-          $$hashKey: '123'
-        },
-        {
-          attrB: 2,
-          $$hashKey: 'abc'
-        }
-      ])).to.deep.equal([
-        {
-          attrA: 1
-        },
-        {
-          attrB: 2
-        }
-      ]);
-    });
-
-    it('should remove hashes from objects in objects', function() {
-      m.chai.expect(utils.removeAngularHashes({
-        attrA: {
-          nestedA: 1,
-          $$hashKey: '123'
-        },
-        attrB: {
-          nestedB: 2,
-          $$hashKey: 'abc'
-        },
-        $$hashKey: 'xyz'
-      })).to.deep.equal({
-        attrA: {
-          nestedA: 1
-        },
-        attrB: {
-          nestedB: 2
-        }
-      });
-    });
-
-    it('should return identity when no hashes', function() {
-      m.chai.expect(utils.removeAngularHashes({
-        attrA: 1
-      })).to.deep.equal({
-        attrA: 1
-      });
-    });
-
-    it('should return identity when not an object or list', function() {
-      m.chai.expect(utils.removeAngularHashes(1)).to.deep.equal(1);
-    });
-
-    it('should return identity when null', function() {
-      m.chai.expect(utils.removeAngularHashes(null)).to.deep.equal(null);
-    });
-
-  });
-
 });
```

This is what I did:

- Always use `angular.equals` to compare objects in the `memoizeImmutableListReference` function, since this function already deals with angular hashes correctly
- Thus, we can get rid of the whole `utils.removeAngularHashes` function
- Create two thin functions that simply wrap `drives.getDrives` and `constraints.getDriveImageCompatibilityStatuses` with `memoizeImmutableListReference`
- Use these functions on the template directly, without creating a temporary data structure to mix both

What do you think?

Shou: Awesome find of `angular.equals`! Unfortunately this doesn't wholly work because `memoizeImmutableListReference` would need to memoize on a per-argument basis, which is messy to do in a generic manner—when it's called with a different argument, the previous argument's returned list is used. So we can definitely get rid of the `removeAngularHashes` utility function, but not `getDrivesAndStatuses` (yet anyway).

jviotti: Oh, I see :( I suggest to push for a generic `memoizeImmutableListReference` using a `Map` or something like that. I didn't realize the limitation of such function, and it definitely makes little sense.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +172,81 @@ module.exports = function(
...
+   * @returns {Function} memoized function
+   *
+   * @example
+   * const getList = () => {
+   *   return Store.getState().toJS().myList;
+   * };
+   *
+   * const memoizedFunction = memoizeImmutableListReference(getList);
+   */
+  const memoizeImmutableListReference = (func) => {
```
lurch: Would there be any benefit to making this more generic by passing in `angular.equals` as a `compare_func` parameter?

lib/gui/components/drive-selector/controllers/drive-selector.js
lurch: Typo still there ;)

lib/gui/components/drive-selector/controllers/drive-selector.js
jviotti: I don't think we should bother about that. This "hack" is very focused, and will get removed once we start moving stuff to react.

Shou:
Is this good to go nows ince the memoization hack will be gone with React creeping into the codebase?

jviotti:
I vote to spend just a bit more time improving `memoizeImmutableListReference` so we can get rid of the temporary statuses data structure, since it shouldn't take that much time.

lurch:
Would it be worth testing this in a rc6 release of Etcher, before pushing out 1.0.0 ?

Shou:
So this should memoize on a per-argument basis, and it works in practice, but with all the mutability/state involved I'm very suspicious. I'm thinking maybe we should move it to `lib/shared/utils.js` and add tests? Either way, any feedback for this commit?

jviotti:
Its starting to look really good. I agree with the tests idea, but instead of moving it to `lib/shared`, let's expose this function as a private function on the controller and test it there (since its very specific)

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +171,93 @@ module.exports = function(
...
+   * @summary Get the list of drives and their compatibility status objects
+   * @function
+   * @public
+   *
+   * @description
+   * For each drive, return its compatibility status with the selected image,
+   * containing the status type (ERROR, WARNING), and accompanying
+   * status message.
+   *
+   * @returns {Object[]} list of objects containing drive and statuses
```
lurch: I guess you need to update the `@summary`, `@description` and `@returns` ?

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +171,93 @@ module.exports = function(
...
+   * caught in an infinite digest loop when using `ngRepeat`
+   * over a function that returns a mutable version of an
+   * ImmutableJS object.
+   *
+   * The problem is that every time you call `myImmutableObject.toJS()`
+   * you will get a new object, whose reference is different from
+   * the one you previously got, even if the data is exactly the same.
+   *
+   * @param {Function} func - function that returns an ImmutableJS list
+   * @param {Function} compare - function used to compare the old and new lists
```
lurch: This param doesn't exist?

Shou:
I've updated the JSDoc, fixed a bug, and added tests now. Let me know what you think.

lurch:
Failing the eslint checks though :-/

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -170,4 +171,92 @@ module.exports = function(
...
+        }
+
+        // Return the tuple unchanged
+        return [ oldArgs, oldState ];
+      });
+
+      // Add the state associated with these args to be memoized
+      if (!areArgsInTuple) {
+        previousTuples.push([ restArgs, state ]);
+        console.log(JSON.stringify(previousTuples, null));
```
lurch: Accidentally-left-in debug code?

tests/gui/components/drive-selector.spec.js
```diff
@@ -0,0 +1,102 @@
...
+        const angularObjectB = {
+          $$hashKey: 2,
+          keyA: true
+        };
+
+        m.chai.expect(memoizedStateA(angularObjectA)).to.equal(angularObjectA);
+        m.chai.expect(memoizedStateA(angularObjectB)).to.equal(angularObjectA);
+      });
+
+    });
```
lurch: Do any of these unit-tests check the tuple-memoization?
To check that things are _actually_ being memoized properly, do you need to test with objects and compare with `.to.deep.equal()`, rather than just testing with simple bools?

Shou: The last test (shown above here) tests the tuple-memoization by ensuring that the object is the same, we don't need deep equals here since the object is memoized. I've expanded the state test so that it's relevant here too, providing evidence that it doesn't unintentionally memoize.

Shou:
Fixed the linter error and removed the debug line. Also changed the state test as mentioned above.

tests/gui/components/drive-selector.spec.js
```diff
@@ -0,0 +1,103 @@
...
+
+        m.chai.expect(memoizedStateA()).to.be.false;
+
+        stateA = true;
+
+        m.chai.expect(memoizedStateA()).to.be.true;
+      });
+
+      it('should reflect different arguments', function() {
+        const memoizedStateA = controller.memoizeImmutableListReference(_.identity);
```
lurch: Minor nit, but can we not re-use the `memoizedStateA` name in this test and the next one? IMHO that's misleading.
What about `memoizedParameter`?

Shou: Good catch, I've amended that now.

lurch:
I've got nothing left to add, final approval is up to @jviotti :+1:

jviotti:
@Shou Fantastic work solving this whole refactoring story! :tada: The code looks great. I'll just give it some manual testing to ensure its all good before merging.

Issue #734, opened 2016-09-30
"rawr i'm a dinosaur" when writing bz2 file


imrehg:
- **Etcher version:** git head at e47a70606ce017a14a3ab29a1edeee5e6399b3b1
- **Operating system and architecture:** ArchLinux x64
- **Do you see any meaningful error information on DevTools?**

```
tracker.js:10 Error: rawr i'm a dinosaur
    at ChildProcess.child.on (/home/greg/resin/prog/etcher/lib/src/child-writer/index.js:92)
    at emitTwo (events.js:106)
    at ChildProcess.emit (events.js:191)
    at handleMessage (internal/child_process.js:718)
    at Pipe.channel.onread (internal/child_process.js:444)
a.(anonymous function)                       @ tracker.js:10
(anonymous function)                         @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:13708
logException                                 @ /home/greg/resin/prog/etcher/lib/gui/modules/analytics.js:157
reportException                              @ /home/greg/resin/prog/etcher/lib/gui/modules/error.js:45
ImageWriterService.flash.then.catch          @ /home/greg/resin/prog/etcher/lib/gui/pages/main/controllers/flash.js:82
processQueue                                 @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:16170
(anonymous function)                         @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:16186
$eval                                        @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:17444
$digest                                      @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:17257
(anonymous function)                         @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:17483
completeOutstandingRequest                   @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:5955
(anonymous function)                         @ /home/greg/resin/prog/etcher/node_modules/angular/angular.js:6234
d.util.hasFunction.f.(anonymous function)    @ tracker.js:3
```

Get this `rawr i'm a dinosaur`error while writing the `bz2` compressed version of the Raspbian Jessie image. Very slow writing (~3x slower than other compressed images), and breaks around ~95%.

![etcher_dinosaur](https://cloud.githubusercontent.com/assets/38863/18984840/780992b8-86ba-11e6-8a7e-0154cffbeb61.png)

Happens both with `1.0.0-beta15` AppImage, and git head `npm start`.

[...]

PR #1194, opened 2017-03-21
upgrade: `unbzip2-stream` to v1.0.11

jviotti:
We believe this issue fixes the "rawr i'm a dinosaur" error on certain
problematic images.

Gives this only happens on a minority of images, we couldn't confirm
that the fix really solves the issue, but we'll include the upgrade on
the next Etcher version and see how it goes in the real world.

Change-Type: patch
Changelog-Entry: Fix "rawr i'm a dinosaur" bzip2 error.
Fixes: https://github.com/resin-io/etcher/issues/734

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -6939,9 +6939,9 @@
       "dev": true
     },
     "unbzip2-stream": {
-      "version": "1.0.10",
-      "from": "unbzip2-stream@>=1.0.10 <2.0.0",
-      "resolved": "https://registry.npmjs.org/unbzip2-stream/-/unbzip2-stream-1.0.10.tgz"
+      "version": "1.0.11",
+      "from": "unbzip2-stream@1.0.11",
+      "resolved": "https://registry.npmjs.org/unbzip2-stream/-/unbzip2-stream-1.0.11.tgz"
     },
     "unc-path-regex": {
       "version": "0.1.2",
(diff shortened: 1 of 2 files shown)


npm-shrinkwrap.json
```diff
@@ -6939,9 +6939,9 @@
       "dev": true
     },
     "unbzip2-stream": {
-      "version": "1.0.10",
-      "from": "unbzip2-stream@>=1.0.10 <2.0.0",
-      "resolved": "https://registry.npmjs.org/unbzip2-stream/-/unbzip2-stream-1.0.10.tgz"
+      "version": "1.0.11",
+      "from": "unbzip2-stream@1.0.11",
```
lurch: Just another curiosity-question: Given the small change in `package.json`, why has the `"from":` line changed like this?

jviotti: I believe its caused by my npm upgrade, which triggered this change given we've updated this dependency from it.

lurch: Is it possible to "regenerate" `npm-shrinkwrap.json` using this upgraded npm format, to keep future commits "cleaner", or am I just worrying about nothing again?

jviotti: `from` only seems to change when you explicitly modify a dependency. Looks like we would have to uninstall/install every single dependency here to make such a change :(

lurch: Ah well, fair enough then ;-)

PR #1183, opened 2017-03-15
refactor(GUI): integrate `etcher-latest-version` into the main repo

jviotti:
`etcher-latest-version` was kept in a separate repository in order to
re-use it with the Etcher website, however the Etcher website is not
using it at all, and we're moving towards having the website in the main
repository.

Therefore, this commit brings back the logic from
`etcher-latest-version`, but introduces it as
`lib/shared/s3-packages.js`, in order to not tie ourselves to the
AngularJS framework, and as a step towards the Etcher SDK.

As a nice little bonus, this commit adds support for an
`ETCHER_FAKE_S3_LATEST_VERSION` environment variable that can be used to
trick Etcher that there is an available update, and therefore show the
update notifier modal.

Also, this commit adds support for snapshot builds update-checks, by
checking the `resin-nightly-downloads` S3 bucket if the current version
contains a git commit hash build number.

If the version is not a production release, then the update notifier
modal doesn't present the checkbox to disable update notifications for X
days.

We also add a property called `updates.semverRange` to `package.json`,
which can be used to fine control which versions are considered as
candidates for an update notification.

This commit adds a setting called `includeUnstableChannel`, which can be
used to tweak whether unstable (beta) releases are considered or not
when checking for the latest available version.

See: https://github.com/resin-io-modules/etcher-latest-version
Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

PR #1199, opened 2017-03-22
WIP chore: include timestamps in snapshot builds as pre-release tags

jviotti:
Current snapshot builds include a version that contains the shorter
current commit hash as a build tag, for example: v1.0.0-beta.19+fbfd6f0.

We want to support update notifications (and auto updates in the
future) on snapshot builds, however we realised that there is no way to
order snapshot build versions based on the information we have at the
moment. For example, how do we determine if v1.0.0-beta.19+fbfd6f0 came
before v1.0.0-beta.19+799ebc6?

Note that we can always roll our custom "smart" comparison logic
querying the git history, etc, however we want to rely on standard
semver comparison rules, so that we can leverage existing third party
auto updates server in the future.

As a solution, we decided to include the timestamp of the commit as part
of the pre release tag. For example: 1.0.0-beta.19.1490192871+fbfd6f0.
Semver implementations consider pre release tags for comparison
purposes, which means that any semver implementation can automatically
determine the ordering of our version.

See: https://github.com/resin-io/etcher/pull/1183

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,12 @@ APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)
 S3_BUCKET = resin-production-downloads
 endif
 ifeq ($(RELEASE_TYPE),snapshot)
-CURRENT_COMMIT_HASH = $(shell git log -1 --format="%h")
-APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)+$(CURRENT_COMMIT_HASH)
+CURRENT_COMMIT_INFO = $(shell git log -1 --format="%ct+%h")
+ifneq (,$(findstring -,$(PACKAGE_JSON_VERSION)))
+APPLICATION_VERSION = $(PACKAGE_JSON_VERSION).$(CURRENT_COMMIT_INFO)
+else
+APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)-$(CURRENT_COMMIT_INFO)
+endif
 S3_BUCKET = resin-nightly-downloads
 endif
 ifndef APPLICATION_VERSION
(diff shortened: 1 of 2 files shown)


jviotti:
We need to move the version from the footer to the settings before merging this.

lurch:
I think for such a UI change, it might be worth getting @alexandrosm 's opinion about moving the version out of the footer and only visible on the settings page?

scripts/publish/aws-s3.sh
```diff
@@ -31,23 +31,20 @@ function usage() {
...
-ARGV_PREFIX=""
 
-while getopts ":f:b:v:p:k:" option; do
+while getopts ":f:b:v:p:" option; do
   case $option in
     f) ARGV_FILE="$OPTARG" ;;
     b) ARGV_BUCKET="$OPTARG" ;;
     v) ARGV_VERSION="$OPTARG" ;;
     p) ARGV_PRODUCT_NAME="$OPTARG" ;;
-    k) ARGV_PREFIX="$OPTARG" ;;
```
lurch: The S3 snapshots repo could still be split into date-specific subfolders? Which would make it _much_ easier to 'manually browse' if we get some kind of UI on top of it.

Makefile
```diff
@@ -35,8 +35,12 @@ APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)
...
 endif
 ifeq ($(RELEASE_TYPE),snapshot)
-CURRENT_COMMIT_HASH = $(shell git log -1 --format="%h")
-APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)+$(CURRENT_COMMIT_HASH)
+CURRENT_COMMIT_INFO = $(shell git log -1 --format="%ct+%h")
+ifneq (,$(findstring -,$(PACKAGE_JSON_VERSION)))
+APPLICATION_VERSION = $(PACKAGE_JSON_VERSION).$(CURRENT_COMMIT_INFO)
+else
+APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)-$(CURRENT_COMMIT_INFO)
+endif
```
lurch: As this is now reasonably complex, perhaps it ought to be split out into a separate shell script?

jviotti:
I'm closing this since we're now going to to store files in GitHub Releases: https://github.com/resin-io/etcher/pull/1555

PR #1203, opened 2017-03-23
chore: Fix copying of variables into Docker

lurch:
Fixes `set -u` errors, and prevents variables being set to undefined values

Change-type: patch


jviotti:
Neat!

PR #1205, opened 2017-03-23
chore: check that all text files only contain expected ASCII chars

lurch:


petrosagg:
What is the problem with non-ascii characters?

lurch:
We had a problem recently (admittedly in another repo) where weird unicode characters had snuck in where they were unwanted: https://github.com/resin-io-modules/mountutils/pull/20

@jviotti suggested I knock up a batch script to catch them.

jviotti:
There is no technical problem, but they look weird in comments.

lurch:
P.S. I forgot to mention at the time... I started out using `grep` in this script, but when doing comprehensive tests I found that there's some non-ASCII byte ranges that `grep` can't even detect, and other byte-ranges that actually cause `grep` to crash with a segmentation fault. I guess it shouldn't be _too_ surprising that a tool designed for working on text files, isn't ideally suited to detecting binary (non-ASCII) byte values.
Hence why I switched to using `perl` instead...

@jviotti I'm beginning to wonder if the **Prerequisites** listed in `docs/RUNNING-LOCALLY.md` ought to be divided into separate lists of prerequisites needed to run Etcher, prerequisites needed to build Etcher, prerequisites needed to package Etcher, prerequisites needed to unit-test Etcher (including linting), prerequisites needed to sanity-check Etcher (primarily only done by the CI servers), prerequisites needed to publish Etcher, etc. etc.
Currently `docs/RUNNING-LOCALLY.md` lists `Asar` which I think is only needed to _package_ Etcher, and `Codespell` which I think is only needed to _lint_ Etcher? And it doesn't list `perl` or `git` needed to _sanity-check_ Etcher, or `AWS CLI` needed to _publish_ Etcher.

jviotti:
That's spot on. Lets update that document that with the latest additions, and separate them between CI, development, release, etc.

jviotti:
Also, what about having shell scripts that initialise your environment for Etcher development? They can detect the OS you're running, and whether you have `apt-get`, `pip`, `brew`, etc, and attempt to pull the necessary dependencies using the tools it finds on the system.

PR #1206, opened 2017-03-23
fix(CLI): don't print stack traces by default

jviotti:
Currently, the Etcher CLI will print scary stack traces for every single
error (e.g: if you forgot to pass an image to the tool), given that
`errors.getDescription()` will return a stack trace if no other
description could be found.

This commit introduces an `ETCHER_CLI_DEBUG` environment variable, which
when set, it will cause the Etcher CLI to output stack traces, plus a
boolean `userFriendlyDescriptionsOnly` option to
`errors.getDescription()`, so we can control whether
`errors.getDescription()` returns things like stack traces, or
stringified error objects.

Change-Type: minor
Changelog-Entry: Don't print stack traces by default in the CLI.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -43,3 +43,9 @@ Options
   --yes, -y      confirm non-interactively
   --unmount, -u  unmount on success
 ```
+
+Debug mode
+----------
+
+You can set the `ETCHER_CLI_DEBUG` environment variable to make the Etcher CLI
+print error stack traces.
--- a/lib/shared/errors.js
+++ b/lib/shared/errors.js
@@ -206,6 +206,8 @@ exports.getTitle = (error) => {
  * @public
  *
  * @param {Error} error - error
+ * @param {Object} options - options
+ * @param {Boolean} [options.userFriendlyDescriptionsOnly=false] - only return user friendly descriptions
  * @returns {String} error description
  *
  * @example
@@ -213,7 +215,11 @@ exports.getTitle = (error) => {
  * const description = errors.getDescription(error);
  * console.log(description);
  */
-exports.getDescription = (error) => {
+exports.getDescription = (error, options = {}) => {
+  _.defaults(options, {
+    userFriendlyDescriptionsOnly: false
+  });
+
   if (!_.isError(error) && !_.isPlainObject(error)) {
     return '';
   }
@@ -225,6 +231,10 @@ exports.getDescription = (error) => {
   const codeDescription = getUserFriendlyMessageProperty(error, 'description');
   if (!_.isNil(codeDescription)) {
     return codeDescription;
+  }
+
+  if (options.userFriendlyDescriptionsOnly) {
+    return '';
   }
 
   if (error.stack) {
--- a/tests/shared/errors.spec.js
+++ b/tests/shared/errors.spec.js
@@ -422,6 +422,60 @@ describe('Shared: Errors', function() {
       m.chai.expect(errors.getDescription(error)).to.equal('Memory error');
     });
 
+    describe('given userFriendlyDescriptionsOnly is false', function() {
+
+      it('should return the stack for a basic error', function() {
+        const error = new Error('Foo');
+        m.chai.expect(errors.getDescription(error, {
+          userFriendlyDescriptionsOnly: false
+        })).to.equal(error.stack);
+      });
+
+      it('should return the stack if the description is an empty string', function() {
+        const error = new Error('Foo');
+        error.description = '';
+        m.chai.expect(errors.getDescription(error, {
+          userFriendlyDescriptionsOnly: false
+        })).to.equal(error.stack);
+      });
+
+      it('should return the stack if the description is a blank string', function() {
+        const error = new Error('Foo');
+        error.description = '   ';
+        m.chai.expect(errors.getDescription(error, {
+          userFriendlyDescriptionsOnly: false
+        })).to.equal(error.stack);
+      });
+
+    });
+
+    describe('given userFriendlyDescriptionsOnly is true', function() {
+
+      it('should return an empty string for a basic error', function() {
+        const error = new Error('Foo');
+        m.chai.expect(errors.getDescription(error, {
+          userFriendlyDescriptionsOnly: true
+        })).to.equal('');
+      });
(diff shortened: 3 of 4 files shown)


lurch:
The CLI counterpart of #1154 ? ;-)
Is this another place where we can't use unit tests because we're writing directly to the `console`?
(would it be worth refactoring everywhere where we write to the console to a couple of global functions, which we _could_ then unit-test?)

jviotti:
> would it be worth refactoring everywhere where we write to the console to a couple of global functions, which we could then unit-test?

Sounds reasonable. Instead of splitting this function, what about adding a boolean "showStacktrace" option to `.getDescription()` so we avoid having to compare the description with the error stack directly, plus we move all this logic somewhere we can unit test it?

jviotti:
Hm, I think the `DEBUG` approach is flawed anyway. One would expect that even if the error has a description, setting `DEBUG` will still cause the stack trace to be shown, which is not the case now.

I'll think about it.

jviotti:
Done!

lib/shared/errors.js
```diff
@@ -228,7 +234,7 @@ exports.getDescription = (error) => {
   }
 
   if (error.stack) {
```
lurch: Should this be
```js
if (options.considerStacktrace && error.stack) {
    return error.stack;
}
```
?

tests/shared/errors.spec.js
```diff
@@ -371,6 +371,13 @@ describe('Shared: Errors', function() {
       m.chai.expect(errors.getDescription(error)).to.equal(error.stack);
     });
 
+    it('should return an empty string for a basic error when considerStacktrace is disabled', function() {
```
lurch: Should there be corresponding tests `when considerStacktrace is enabled` ?

docs/CLI.md
```diff
@@ -43,3 +43,9 @@ Options
   --yes, -y      confirm non-interactively
   --unmount, -u  unmount on success
 ```
+
+Debug mode
+----------
+
+You can set the `DEBUG` environment variable to make the Etcher CLI print error
```
lurch: I wonder if it should be more specific, like `ETCHER_CLI_DEBUG`? The user might have `DEBUG` set in their environment variables for other reasons, without wanting to debug the Etcher CLI? ;-)

jviotti: Done!

lib/shared/errors.js
```diff
@@ -206,14 +206,20 @@ exports.getTitle = (error) => {
  * @public
  *
  * @param {Error} error - error
+ * @param {Object} options - options
+ * @param {Boolean} [options.considerStacktrace=true] - consider stack trace
```
jviotti: Done!

lib/shared/errors.js
jviotti: Not quite. Note we explicitly return an empty string if there is a stack but we're not considering it.

tests/shared/errors.spec.js
jviotti: Done!

lib/shared/errors.js
lurch: Why should that be the case? Rather than let it carrying on with the rest of the code below?

docs/CLI.md
lurch: I just noticed that this "corresponds" nicely with the `ETCHER_CLI_ROBOT` environment variable used by `lib/shared/robot/index.js` - cool :sunglasses:

lib/shared/errors.js
jviotti: If we let it carry on, then the error object will get stringified, which is the worst possible case. I do get your point though, the `considerStacktrace` option is confusing if it also means an empty string is returned right away if there is no stack trace.

One solution would be to remove the string transformation fallback at the bottom of this function, however that would be really useful when dealing with strange errors coming from third party libraries that don't adhere to the NodeJS "guidelines" (e.g: a library throwing an object that is not a common `Error` instance).

Another solution would be to rephrase the option name to signify that we want a user friendly description, or nothing at all (thus ignoring non-user, but developer friendly things like stack traces and stringified versions of object). We could call this option `ignoreNonUserFriendlyForms()` or something like that.

What do you think?

lurch: Or you could just add a more detailed inline code-comment, explaining why we return an empty string if `error.stack` is set (which from what you say above, indicates it's a "standard" Error?) and `considerStacktrace` is false.  ;-)

Or perhaps `userFriendlyDescriptionsOnly` is a better name for `ignoreNonUserFriendlyForms` ?

jviotti: > Or perhaps userFriendlyDescriptionsOnly is a better name for ignoreNonUserFriendlyForms ?

That sounds great to me! Take a look now.

tests/shared/errors.spec.js
```diff
@@ -422,6 +422,60 @@ describe('Shared: Errors', function() {
...
+        })).to.equal('');
+      });
+
+      it('should return an empty string if the description is an empty string', function() {
+        const error = new Error('Foo');
+        error.description = '';
+        m.chai.expect(errors.getDescription(error, {
+          userFriendlyDescriptionsOnly: true
+        })).to.equal('');
+      });
```
lurch: Could we re-order these last two tests, so that we run the tests in the same order for both `userFriendlyDescriptionsOnly: false` and `userFriendlyDescriptionsOnly: true`? :)

lib/shared/errors.js
```diff
@@ -227,16 +233,20 @@ exports.getDescription = (error) => {
     return codeDescription;
   }
```
lurch: Might be neater if we just have a single:
```js
if (options.userFriendlyDescriptionsOnly) {
  return '';
}
```
here, as I think that'd then allow us to remove both the `!options.userFriendlyDescriptionsOnly` tests further below?

jviotti: That's a great observations. Done!

tests/shared/errors.spec.js
jviotti: :+1: Someday I'll build a static code analysis tool, and I'll call it "lurch" :P

Issue #985, opened 2016-12-23
Etcher freezing on "Starting..." dialog


Facel3ss1:
- **Etcher version: 1.0.0-beta.17**
- **Operating system and architecture: macOS 10.12.1 64-bit**
- **Do you see any meaningful error information on DevTools?** No

When trying to flash an image to an SD Card, Etcher freezes on the "Starting..." dialog on the flash button: 
![screen shot 2016-12-23 at 09 31 44](https://cloud.githubusercontent.com/assets/16763503/21450892/af985d34-c8f2-11e6-8ad7-353c656632af.png)
No errors come up in DevTools or the GUI...
![screen-shot-2016-12-22-at-19 14 24](https://cloud.githubusercontent.com/assets/16763503/21451014/b8d15aee-c8f3-11e6-80a5-3334119a9853.png)
...but if I try running the command from the DevTools in Terminal.app it comes up with this error message:
![screen shot 2016-12-23 at 08 47 36](https://cloud.githubusercontent.com/assets/16763503/21451040/e3f739a0-c8f3-11e6-8011-c65b891087fe.png)
Any help would be appreciated. Thanks!

jviotti:
Hm, very weird. So the command exits with an error code other than zero, but we don't know anything else about it.

BTW, did you run the command after closing Etcher? If so, the error might be due to the IPC server ID not longer being valid, and that might be obscuring the original issue.

If you can reliably reproduce (let me know if not), you should be able to execute the command while Etcher is frozen, but without closing it, and see if that reveals anything else.

Facel3ss1:
Okay, I have a lot to say, so bear with me:

Thank you for your reply (especially this close to Christmas! 🎄)! I did run the command in Etcher while the Terminal was closed. So, I ran it while Etcher was open as you said and that just freezes too (it freezes on 'diskutil' at the top'). No errors:
![screen shot 2016-12-24 at 10 07 29](https://cloud.githubusercontent.com/assets/16763503/21466362/e42e98e6-c9c0-11e6-9478-e564a628f2dd.png)
[...]

PR #1211, opened 2017-03-23
fix(image-stream): Destroy unused image streams

jhermsmeier:
When fetching metadata, a stream is opened, but never consumed. This destroys the stream when retrieving metadata to avoid dangling open file descriptors and such.

Change-Type: patch

### DIFF
--- a/lib/image-stream/index.js
+++ b/lib/image-stream/index.js
@@ -108,6 +108,11 @@ exports.getFromFilePath = (file) => {
  */
 exports.getImageMetadata = (file) => {
   return exports.getFromFilePath(file).then((image) => {
+
+    // Since we're not consuming this stream,
+    // destroy() it, to avoid dangling open file descriptors etc.
+    image.stream.destroy();
+
     return _.omitBy(image, isStream);
   });
 };


lib/image-stream/index.js
```diff
@@ -108,6 +108,9 @@ exports.getFromFilePath = (file) => {
  */
 exports.getImageMetadata = (file) => {
   return exports.getFromFilePath(file).then((image) => {
+    // Since we're not consuming this stream,
```
jviotti: Can we add a blank line above this comment for consistency purposes?

lib/image-stream/index.js
```diff
@@ -108,6 +108,9 @@ exports.getFromFilePath = (file) => {
...
+    // destroy() it, to avoid dangling open file descriptors etc.
+    image.stream.destroy();
```
jviotti: Can we also add a blank line below this? Otherwise I get the impression that the comment applies to this line, and to the one below.

jhermsmeier: Yup, mended

lurch:
This is obviously the "quick fix", but perhaps a better approach would be to modify the functions in `handlers.js` so that you can tell them that you _only_ want the metadata, so that the streams don't get created in the first place?

BTW @jviotti it looks strange that `getFromFilePath` is documented as returning **just** `{size, stream, transform}`, but `getImageMetadata` is documented as returning a whole load of (potential) other metadata properties, and yet the latter just calls the former. Should `getFromFilePath` also be documented as (potentially) returning these other metadata properties too?

And I've spotted a typo, which I'll submit a PR for ;-)

jviotti:
> This is obviously the "quick fix", but perhaps a better approach would be to modify the functions in handlers.js so that you can tell them that you only want the metadata, so that the streams don't get created in the first place?

Yep, the idea was to have this ASAP so we can move forward, and @jhermsmeier will to the refactoring tomorrow.

> Should getFromFilePath also be documented as (potentially) returning these other metadata properties too?

Yeah, we should do that! Are you sending a PR for that as well?

lurch:
> Yeah, we should do that! Are you sending a PR for that as well?

No, I think that would be better PRed by somebody who actually understands what the code is doing ;-P

jviotti:
Haha, I'll handle it tomorrow!

jviotti:
As a side note, this issue probably fixes https://github.com/resin-io/etcher/issues/985

PR #1212, opened 2017-03-23
feat(robot): Support logging arbitrary data to the console

jhermsmeier:
This adds a new command recognition for message type "log",
to enable `console.log()`ing arbitrary data to the parent process console.

Not sure if we'd really want this in, but it's very handy for debugging,
and would keep us from writing these 3 lines every time
we want to inspect something in between those processes.

Change-Type: minor

### DIFF
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -139,6 +139,8 @@ exports.write = (image, drive, options) => {
       // to provide better encapsulation.
       if (messageCommand === 'error') {
         emitError(robot.recomposeErrorMessage(parsedMessage));
+      } else if (messageCommand === 'log') {
+        console.log(messageData);
       } else {
         emitter.emit(messageCommand, messageData);
       }
--- a/lib/shared/robot/index.js
+++ b/lib/shared/robot/index.js
@@ -215,3 +215,17 @@ exports.printError = (error) => {
 exports.printMessage = (message, data) => {
   console.log(exports.buildMessage(message, data));
 };
+
+/**
+ * @summary Log a message to the host's console
+ * @function
+ * @public
+ *
+ * @param {Object} [data] - data
+ *
+ * @example
+ * robot.log({ example: 'data' });
+ */
+exports.log = (data) => {
+  exports.printMessage('log', data);
+};
--- a/lib/shared/robot/README.md
+++ b/lib/shared/robot/README.md
@@ -48,6 +48,16 @@ console.log(robot.getData(message));
 > }
 ```
 
+**Logging debug data to the console:**
+
+*Child process:*
+
+```js
+// This will log the passed data to parent's console,
+// as `console.log()`ing in the child will cause errors
+robot.log({ debugging: 'things' })
+```
+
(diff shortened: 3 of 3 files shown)


lurch:
Is there, or should there be, documentation of what methods / messages / etc. the 'robot' understands / emits?

jviotti:
I like the idea. Lets add some documentation for this like @lurch pointed out. Maybe inside the README at `lib/child-writer`? We could also have a small utility function that creates and prints a "log" robot message, so we don't have to manually construct it.

lurch:
IMHO we should wait until after #1221 has been merged, and then rebase this one.

jviotti:
Yeah, sounds like a good way to avoid some more complex conflicts.

jviotti:
The linked PR was merged! We have some conflicts here, but they are hopefully very simple to deal with.

lib/child-writer/index.js
```diff
@@ -139,6 +139,8 @@ exports.write = (image, drive, options) => {
       // to provide better encapsulation.
       if (messageCommand === 'error') {
         emitError(robot.recomposeErrorMessage(parsedMessage));
+      } else if (messageCommand === 'log') {
+        console.log(message.data);
```
lurch: Should this be `messageData`? Or `messageData.data`? Or is `message.data` indeed correct??
I think I need to go and lay down, I'm getting dizzy ;-)

jhermsmeier: Oh, missed that – fixed!

jviotti: Yeah, this should be `messageData`.

lurch:
\*bump\* https://github.com/resin-io/etcher/pull/1212#issuecomment-288897255  ;-)

lurch:
re-ping @jhermsmeier - do you want to address @jviotti 's comments in https://github.com/resin-io/etcher/pull/1212#issuecomment-288897255

jhermsmeier:
> We could also have a small utility function that creates and prints a "log" robot message, so we don't have to manually construct it.

Added `robot.log({ data: 'datadata' })` for even fewer chars to type (which just calls `robot.printMessage('log', data)`).

> I like the idea. Lets add some documentation for this like @lurch pointed out. Maybe inside the README at lib/child-writer?

Added an example to `lib/shared/robot/README.md`, since the message passing is already documented there.

lib/shared/robot/index.js
```diff
@@ -215,3 +215,17 @@ exports.printError = (error) => {
 exports.printMessage = (message, data) => {
   console.log(exports.buildMessage(message, data));
 };
+
+/**
+ * @summary Log a message to the host's console
+ * @function
+ * @public
+ *
+ * @param {Object} [data] - data
```
lurch: Does it always need to be an Object, or would passing in a String work too?

jhermsmeier: It does have to be an object, otherwise `buildMessage()` will yell (see line 58: https://github.com/resin-io/etcher/pull/1212/files#diff-0ccd387a7e7d9c117af87e3bdbd6a9ebR58)

PR #1214, opened 2017-03-24
Update .travis.yml to not clone entire nvm history

PeterDaveHello:
You don't need to have the whole commit history of `nvm`, let's speed up the clone process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ os:
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       rm -rf ~/.nvm;
-      git clone https://github.com/creationix/nvm.git ~/.nvm;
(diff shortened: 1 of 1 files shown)


jviotti:
Thanks a lot for the heads up @PeterDaveHello !

PeterDaveHello:
@jviotti no problem, btw, do you really want to use the "latest" code base - master branch? It may be not stable enough, though it'll come with the newest feature, but I wonder if you'll like to use a stable release here?

jviotti:
@PeterDaveHello I didn't consider that, but sounds like another great addition. Feel free to submit another PR for it :)

jviotti:
Oh wait, I think that won't be needed after all, since nvm is completely removed by https://github.com/resin-io/etcher/pull/1216

PeterDaveHello:
#1216 looks good, Travis CI will maintain its `nvm` :)

PR #1215, opened 2017-03-24
chore(scripts): Rebuild instead of removing `node_modules`

jhermsmeier:
As removing the `node_modules` completely and then re-installing is quite time consuming and SSD-thrashing, resort to pruning and rebuilding native add-ons when switching targets.

Change-Type: patch

### DIFF
--- a/scripts/build/dependencies-npm.sh
+++ b/scripts/build/dependencies-npm.sh
@@ -96,6 +96,18 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
 fi
 
 function run_install() {
+
+  # Since we use an `npm-shrinkwrap.json` file, if you pull changes
+  # that update a dependency and try to `npm install` directly, npm
+  # will complain that your `node_modules` tree is not equal to what
+  # is defined by the `npm-shrinkwrap.json` file, and will thus
+  # refuse to do anything but install from scratch.
+  npm prune
+
+  # When changing between target architectures, rebuild all dependencies,
+  # since compiled add-ons will not work otherwise.
+  npm rebuild
+
   npm install $INSTALL_OPTS
 
   if [ "$ARGV_PRODUCTION" == "true" ]; then
--- a/Makefile
+++ b/Makefile
@@ -471,18 +471,7 @@ endif
 
 .PHONY: $(TARGETS)
 
-# Note that we need to remove `node_modules` every time.
-# Since we use an `npm-shrinkwrap.json` file, if you pull changes
-# that update a dependency and try to `npm install` directly, npm
-# will complain that your `node_modules` tree is not equal to what
-# is defined by the `npm-shrinkwrap.json` file, and will thus
-# refuse to do anything but install from scratch.
-# The `node_modules` directory also needs to be wiped out if you're
-# changing between target architectures, since compiled add-ons
-# will not work otherwise.
-
 cli-develop:
-	rm -rf node_modules
 	./scripts/build/dependencies-npm.sh \
 		-r "$(TARGET_ARCH)" \
 		-v "$(ELECTRON_VERSION)" \
@@ -490,7 +479,6 @@ cli-develop:
 		-s "$(TARGET_PLATFORM)"
 
 electron-develop:
-	rm -rf node_modules
 	./scripts/build/dependencies-npm.sh \
 		-r "$(TARGET_ARCH)" \
 		-v "$(ELECTRON_VERSION)" \


scripts/build/dependencies-npm.sh
```diff
@@ -96,6 +96,18 @@ if [ "$ARGV_PRODUCTION" == "true" ]; then
...
+  # Since we use an `npm-shrinkwrap.json` file, if you pull changes
+  # that update a dependency and try to `npm install` directly, npm
+  # will complain that your `node_modules` tree is not equal to what
+  # is defined by the `npm-shrinkwrap.json` file, and will thus
+  # refuse to do anything but install from scratch.
+  npm prune
+
+  # When changing between target architectures, rebuild all dependencies,
+  # since compiled add-ons will not work otherwise.
+  npm rebuild
```
lurch: Could we optimise this further to only rebuild if the new target-arch is different to the old target-arch?

jhermsmeier: Yeah, that'd be nice, but I'm not sure we can reliably determine which target it has previously been built for. Maybe @jviotti knows something?

lurch: I guess we could write some kind of temporary marker file, but we'd have to be careful that it didn't get out of sync (i.e. that it always gets updated/deleted as appropriate).

jviotti: Yeah, I don't know if we can accomplish that. Lets test this in more detail first. Maybe npm is smart enough to only rebuild when needed?

jhermsmeier: Afaik it'll rebuild everything, which is also what the docs sound like:

> This  command runs the npm build command on the matched folders.  This is useful when you install a new version of node, and must recompile all your C++ addons with the new binary.

jviotti: I think its fine as it is now. We don't do that much compilation at the moment, so trying to reduce it is probably not worth the effort. We can re-consider it if compilation starts becoming more of a bottleneck.

lurch:
A question about this, if you don't mind...
When I run `make electron-develop` twice in a row, at the start of the second run it prints:
```
unbuild array-buffer-from-string@0.1.0
unbuild base32-encode@0.1.0
unbuild bplist-creator@0.0.7
unbuild cp-file@3.2.0
unbuild fmix@0.1.0
unbuild imul@1.0.1
unbuild murmur-32@0.1.0
unbuild nested-error-stacks@1.0.2
unbuild random-path@0.1.1
unbuild stream-buffers@2.2.0
unbuild tn1150@0.1.0
unbuild unorm@1.4.1
unbuild fs-temp@1.1.2
unbuild image-size@0.5.1
```
I'm guessing this is coming from `npm prune`? However these modules _are_ listed in `npm-shrinkwrap.json`, and do then get re-installed.
I noticed that they're "optional" dependencies, however it seems that there's also other optional dependencies in `npm-shrinkwrap.json` (below list created with `grep -B5 optional npm-shrinkwrap.json | grep '": {'|cut -d'"' -f2|sort|uniq` )
```
array-buffer-from-string
base32-encode
bcrypt-pbkdf
bplist-creator
camelcase
cliui
cp-file
ecc-jsbn
fmix
fs-temp
image-size
jodid25519
jsbn
murmur-32
nested-error-stacks
random-path
source-map
stream-buffers
tn1150
tweetnacl
uglify-js
uglify-to-browserify
unorm
window-size
wordwrap
yargs
```
and some of them _aren't_ 'unbuilt' by `npm prune`? Hmmm, and `imul` gets pruned, which **isn't** an optional dependency?
I also ran `npm run clean-shrinkwrap` as a test, but that didn't result in any changes to `npm-shrinkwrap.json`

I'm not complaining, as running `make electron-develop` a second time now seems much faster than before, I'm just trying to understand what's going on ;-)

jhermsmeier:
This isn't coming from `npm prune` – that's `npm rebuild`. And those `unbuild ...` messages are coming from packages that have build scripts.

lurch:
Ahhh, thanks Jonas.
Just out of curiosity, I also ran:
```bash
git clone --depth=1 https://github.com/resin-io/etcher etcher-fresh
cd etcher-fresh
rm npm-shrinkwrap.json
make electron-develop
npm --dev shrinkwrap
npm run clean-shrinkwrap
git diff
```
and ignoring the version-change differences (and using `grep` to double-check), it seems like these packages are in the repo-version of `npm-shrinkwrap.json` but _not_ in my freshly-generated version of `npm-shrinkwrap.json` :
```
array-buffer-from-string
base32-encode
bplist-creator
cp-file
fmix
fs-temp
image-size
imul
lodash.rest
murmur-32
nested-error-stacks
node-cmd
pkg-conf
random-path
spdx-exceptions
stream-buffers
symbol
tn1150
unorm
```

Obviously some of those will be due to the newer non-shrinkwrapped packages having different dependencies, but I wonder if maybe some of the packages in the list above _aren't_ actually needed anyway, and are merely being installed because they're still listed in `npm-shrinkwrap.json` when they should actually have been removed?

Again, I'm still an NPM newbie, so please feel free to point out where I'm going wrong :-)

PR #1216, opened 2017-03-24
chore(travis): Use `node` base, instead of `cpp`

jhermsmeier:
This changes the `.travis.yml` to use the `node_js` language as a base instead of `cpp` to avoid installing & setting up node manually.

Change-Type: patch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,24 @@
-language: cpp
+language: node_js
 sudo: false
+node_js:
+  - "6.1.0"
 
 services:
   - docker
 
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - libstdc++-6-dev
+
 env:
   global:
-    - NODE_VERSION="6.1.0"
+    - CCACHE_TEMPDIR=/tmp/.ccache-temp
+    - CCACHE_COMPRESS=1
+    - CC="clang"
+    - CXX="clang++"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,12 +1,25 @@
...
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - libstdc++-6-dev
+
 env:
   global:
     - NODE_VERSION="6.1.0"
```
jviotti: Should this be deleted now?

jviotti:
There is a small conflict here :)

.travis.yml
jhermsmeier: Yup, removed

jhermsmeier:
Oh yes, conflict resolved :)

lurch:
Just out of curiosity, what advantage does this give? (is installing `cpp` on top of a `node_js` base faster than installing `node_js` on top of a `cpp` base?)
Should something similar be done for Appveyor?

Might be worth still printing out `node --version` and `npm --version` for debugging purposes? \*shrug\*

jhermsmeier:
> Might be worth still printing out node --version and npm --version for debugging purposes?

Travis does that automatically with `language: node_js`, it's in the upper part of build output

> Just out of curiosity, what advantage does this give?

Less complexity. We don't have to install node etc. ourselves, for example

Issue #1210, opened 2017-03-23
Disallow reloading the renderer process


jviotti:
Users can reload the renderer process with Cmd+R and Ctrl+R. We should disallow this, given that it would result in undefined behavior, and weird errors.

Shou:
Is this platform-specific? I can't reload Etcher with Ctrl+R on Ubuntu, only from within the developer tools.

jviotti:
Hm, maybe F5 in that platform? I'll check here.

Shou:
On Ubuntu F5 only works from within the developer tools too.

jviotti:
You're right. It only happens on OS X. Before changing the issue tag to `platform:os x`, do you think there is a way to disable reloading from DevTools as well?

I think we should force reloading by re-running the app, given that refreshing, even from DevTools, can cause strange error that will get into TrackJS.

jviotti:
I'll fix it for the main Etcher window. We can't do much with DevTools since we have little control over it.

alexandrosm:
How about making the UI reload-tolerant? Is it a big job?

jviotti:
Each renderer process has a unique IPC identifier so that the child writer process can communicate with it. Making the UI reload-tolerant would involve storing that runtime calculated unique identifier somewhere where it could be matched with the reloaded renderer process but still not collide with other running instances of Etcher. On top of that, we would have to somehow restore the application state based on the fact that we detect that the child writer is running.

So in summary, it could be done, but its a lot of work. I don't think its worth the effort.

PR #1219, opened 2017-03-24
fix(GUI): disallow reloading the app in macOS using Cmd+R

jviotti:
Users can currently force the browser window to reload by using Cmd+R in
macOS. The `will-navigate` even handler in this same file is supposed to
protect us from this case, however its not fired at all just in macOS
for some reason.

As a solution, we attach dummy keyboard shortcut handlers for Cmd+R,
Ctrl+R, and F5, to override the normal browser window behaviour.

The reason behind disallowing reloads is that a reload during a critical
operation such as writing, validation, elevating, unmounting, drive
scanning, etc will cause undefined behaviour, resulting in non sense
errors that we can't solve coming to TrackJS.

See: https://github.com/electron/electron/issues/8841
Fixes: https://github.com/resin-io/etcher/issues/1210

### DIFF
--- a/lib/gui/etcher.js
+++ b/lib/gui/etcher.js
@@ -17,6 +17,7 @@
 'use strict';
 
 const electron = require('electron');
+const _ = require('lodash');
 const path = require('path');
 let mainWindow = null;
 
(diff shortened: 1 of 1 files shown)


lurch:
Couple of typos in the commit-message:
"even handler" -> "event handler"
"its not fired" -> "it's not fired"
"non sense" -> "nonsense"

The local pedant ;-)

jviotti:
Fixed! :)

mike-seekwell:
@jviotti It's worth noting this disables CMD+R across all applications the user has open (e.g. they won't be able to reload a page in Chrome using CMD+R)

Issue #1179, opened 2017-03-11
"Error: No polkit authentication agent found" when trying to flash an image as non-root


soulshake:
- **Etcher version:** 1.0.0-beta.19
- **Operating system and architecture:** 
```
$ uname -a
Linux patamushka 4.9.0-997-generic #201612270045 SMP Tue Dec 27 05:47:01 UTC 2016 x86_64 GNU/Linux

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.0 (stretch)
Release:	9.0
Codename:	stretch
```

## What I do

- open Etcher via CLI by running `etcher-electron` (as my normal system user)
- select an image I've just downloaded from resin.io
- (the disk autopopulates with my SD card)
- click "Flash"

## What I expect to happen

For the image to be written to the card

## What actually happens

I get the following popup:

![image](https://cloud.githubusercontent.com/assets/1585815/23824969/aa08901e-0646-11e7-846c-36279765f63b.png)

```
{ Error: No polkit authentication agent found.
    at /usr/share/etcher-electron/resources/app.asar/node_modules/sudo-prompt/index.js:132:23
    at ChildProcess.exithandler (child_process.js:225:5)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Socket.<anonymous> (internal/child_process.js:334:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at Pipe._handle.close [as _onclose] (net.js:493:12)
  cause: 
   Error: No polkit authentication agent found.
       at /usr/share/etcher-electron/resources/app.asar/node_modules/sudo-prompt/index.js:132:23
       at ChildProcess.exithandler (child_process.js:225:5)
       at emitTwo (events.js:106:13)
       at ChildProcess.emit (events.js:191:7)
       at maybeClose (internal/child_process.js:877:16)
       at Socket.<anonymous> (internal/child_process.js:334:11)
       at emitOne (events.js:96:13)
       at Socket.emit (events.js:188:7)
       at Pipe._handle.close [as _onclose] (net.js:493:12),
  isOperational: true }
```

```
    at Socket.child.stderr.on (/usr/share/etcher-electron/resources/app.asar/lib/child-writer/index.js:129:17)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
[...]

PR #1221, opened 2017-03-24
fix(GUI): provide a friendly message when no polkit agent is available

jviotti:
`sudo-prompt`, the module we use to provide elevation for GNU/Linux,
relies on polkit to show the elevation dialog. Polkit implements a
"backend" to control privileges, but relies on different packages to
provide a GUI frontend.

If no GUI frontend can be found on the system, then `sudo-prompt` will
fail with a `No polkit authentication agent found.` In this case, we
shouldn't present a scary stack trace to the user, and we should also
not send this error to TrackJS.

As a solution, we intercept such error in
`lib/child-writer/writer-proxy.js`, and convert it into a human readable
user error using `errors.createUserError()`, which is passed to the
Etcher GUI using the "robot" mechanism.

Fixes: https://github.com/resin-io/etcher/issues/1179
Change-Type: patch
Changelog-Entry: Provide a user friendly error message when no polkit authentication agent is available on the system.


lib/child-writer/index.js
```diff
@@ -113,25 +113,36 @@ exports.write = (image, drive, options) => {
...
-
+  /**
+   * @summary Bridge robot message to the child writer caller
+   * @function
+   * @private
+   *
+   * @param {String} message - robot message
+   *
+   * @example
+   * bridgeRobotMessage(robot.buildErrorMessage(new Error('foo')));
```
lurch: Presumably `message` commands will be much more common than `error` commands? To reflect that more common case, perhaps this example should be modified?

lib/child-writer/writer-proxy.js
```diff
@@ -157,6 +158,13 @@ return isElevated().then((elevated) => {
       message: 'User did not grant permission.'
     }, () => {
       process.exit(EXIT_CODES.CANCELLED);
+    }).catch({
+      message: 'No polkit authentication agent found.'
```
lurch: Is there any way that this message (and the other message a few lines further up) could be unit-tested, so that this code doesn't break if `sudoPrompt` should happen to change its error strings?

lib/child-writer/writer-proxy.js
```diff
@@ -157,6 +158,13 @@ return isElevated().then((elevated) => {
...
+    }, () => {
+      throw errors.createUserError(
+        'No polkit authentication agent found',
+        'Please install a polkit authentication agent for your desktop environment of choice to continue'
```
lurch: This error message is perhaps a bit obscure for the typical user that Etcher is aimed at? I wonder if we should maybe include a URL that the user could visit for more information?
(Or is it only likely to be 'hardcore' users that are running without a 'polkit authentication agent' anyway?)

lib/child-writer/index.js
```diff
@@ -113,25 +113,36 @@ exports.write = (image, drive, options) => {
...
-    } catch (error) {
-      return emitError(error);
-    }
+      const parsedMessage = robot.parseMessage(message);
+
+      // The error object is decomposed by the CLI for serialisation
+      // purposes. We compose it back to an `Error` here in order
+      // to provide better encapsulation.
+      if (robot.getCommand(parsedMessage) === 'error') {
+        emitError(robot.recomposeErrorMessage(parsedMessage));
```
lurch: Is doing:
```
const parsedMessage = robot.parseMessage(message);
if (robot.getCommand(parsedMessage) === 'error') {
  emitError(robot.recomposeErrorMessage(parsedMessage));
```
the same as doing:
```
const parsedMessage = robot.parseMessage(message);
if (robot.getCommand(parsedMessage) === 'error') {
  emitError(message);
```
?
(It's kinda hard to keep track of which parts are `String`s, and which parts are `Object`s ;)  )

lib/child-writer/index.js
jviotti: Makes sense, done!

lib/child-writer/index.js
jviotti: Not really. `message` is a string like `{"command":"foo","data":{"foo":"bar"}}`. `parsedMesage` is then an object like:

```js
{
  command: 'foo',
  data: {
    foo: 'bar'
  }
}
```

Finally, the result of `recomposeErrorMessage` is an instance of `Error`.

lib/child-writer/index.js
```diff
@@ -113,25 +113,36 @@ exports.write = (image, drive, options) => {
...
+      } else {
+        emitter.emit(robot.getCommand(parsedMessage), robot.getData(parsedMessage));
```
jviotti: This is basically the same as:

```js
emitter.emit(parsedMessage.command, parsedMessage.data);
```

with the exception that we use function accessors instead of referencing the property directly (so we can change the structure of a robot object at will without having to change code all over the place), so it doesn't look like a weird pattern IMHO.

I'm still happy to refactor if you think what would make things cleaner. What do you think?

lib/child-writer/writer-proxy.js
jviotti: I can't think of anything apart from checking `sudo-prompt` each time we upgrade it, to make sure we have the right error messages :( I agree that this means things can break out of the sudden if `sudo-prompt` changes the error messages, but looks like the best think we can do at the moment.

lib/child-writer/writer-proxy.js
jviotti: Virtually all user friendly GNU/Linux distributions will include a polkit agent, so if you don't have one then it means you either have a less known distribution, or you are an experienced user that removed it for some reason. In either case, it means that you're an advanced user, so I think its fine.

lib/child-writer/index.js
lurch: Aha, thanks!

lib/child-writer/index.js
lurch: Could we have a single function that returned both the `command` and `data` (e.g. as a tuple) at the same time? Or maybe to keep it more generic, it should be a function which accepts a list of properties, and returns an object with those properties set?

Or maybe `parsedMessage` itself should be an Object with accessor functions (this is where my JS knowledge runs out) so you could do `parsedMessage.getCommand()` and `parsedMessage.getData()` ?

If none of those make sense, or if they'd make the code less readable, then feel free to keep things as-is.

Alternatively, how likely is it that `robot.getCommand(parsedMessage)` will ever do anything _other_ than returning `parsedMessage.command` ?

\*shrug\*  You're obviously for more familiar with usual JS practices then I am.

lib/child-writer/writer-proxy.js
lurch: Fair enough. Would it be worth chucking some code-comments in here to reflect that fact?
(i.e. that the `sudo-prompt` error messages need to be checked if we ever upgrade it)

jviotti: Sounds good, will do!

lib/child-writer/index.js
jviotti: > Could we have a single function that returned both the command and data (e.g. as a tuple) at the same time?

Tuples are not very used in the JS world. We could return an object, but that'd be basically returning `parsedMessage` again :)

> Or maybe parsedMessage itself should be an Object with accessor functions (this is where my JS knowledge runs out) so you could do parsedMessage.getCommand() and parsedMessage.getData() ?

That would mean

```js
emitter.emit(robot.getCommand(parsedMessage), robot.getData(parsedMessage));
```

gets changed to:

```js
emitter.emit(parsedMessage.getCommand(), parsedMessage.getData());
```

Which is basically the same, plus I think we should go functional when possible.

> Alternatively, how likely is it that robot.getCommand(parsedMessage) will ever do anything other than returning parsedMessage.command ?

Probably nothing else. The main benefit of function getters is that JavaScript is a weakly typed language. We might accidentally do `message.comand`, which will silently be evaluated as `undefined`, and will not match any robot message handler, causing the data on that message to be ignored, etc. Having function getters protect us from that kind of things.

lurch: Fair enough. Perhaps I'd be happier if there was an inline code-comment mentioning that `robot.getCommand` and `robot.getData` are 'lightweight' accessor methods?

The reason I was concerned when I first saw
```
emitter.emit(robot.getCommand(parsedMessage), robot.getData(parsedMessage));
```
was because I was worried that `robot.getCommand` and `robot.getData` were possibly 'heavy' methods, which might be (internally) repeating the same code-actions to pull the required information out of `parsedMessage`. IYSWIM?

lib/child-writer/writer-proxy.js
```diff
@@ -153,10 +154,23 @@ return isElevated().then((elevated) => {
       if (!_.isEmpty(stderr)) {
         throw errors.createError(stderr);
       }
+
+    // We're hardcoding internal error messages declared by `sudo-prompt`.
+    // There doesn't seem to be a better way to handle these errors, so
+    // for now, we should make sure we double check if the error messages
+    // have changed everytime we upgrade `sudo-prompt`.
```
lurch: "everytime" -> "every time"

jviotti: Fixed!

lib/child-writer/index.js
jviotti: Sounds good, added!

lib/child-writer/index.js
```diff
@@ -113,25 +113,44 @@ exports.write = (image, drive, options) => {
...
+      // purposes. We compose it back to an `Error` here in order
+      // to provide better encapsulation.
+      if (robot.getCommand(parsedMessage) === 'error') {
+        emitError(robot.recomposeErrorMessage(parsedMessage));
+      } else {
+
+        // These are lighweight accessor methods for
+        // the properties of the parsed message
+        const messageCommand = robot.getCommand(parsedMessage);
+        const messageData = robot.getData(parsedMessage);
```
lurch: These could be moved a few lines higher up, to eliminate the duplicated `robot.getCommand(parsedMessage)` ?

lib/child-writer/index.js
```diff
@@ -113,25 +113,44 @@ exports.write = (image, drive, options) => {
...
+
+      // The error object is decomposed by the CLI for serialisation
+      // purposes. We compose it back to an `Error` here in order
+      // to provide better encapsulation.
+      if (robot.getCommand(parsedMessage) === 'error') {
+        emitError(robot.recomposeErrorMessage(parsedMessage));
+      } else {
+
+        // These are lighweight accessor methods for
+        // the properties of the parsed message
```
lurch: :+1:

lib/child-writer/index.js
jviotti: Done!

Issue #1125, opened 2017-02-21
Improve error when the child process starts, but the drive doesn't exist anymore


jviotti:
Steps to reproduce:

- Select image
- Select drive
- Click "Flash"
- Unplug the drive right afterwards

You might want to add a timeout before the CLI logic kicks in to increase the chances of reproducing.

This is the error I get:

<img width="433" alt="screenshot 2017-02-21 17 12 14" src="https://cloud.githubusercontent.com/assets/2192773/23187963/47e5d518-f863-11e6-9735-012f679ad4db.png">

Tasks:

- [ ] Add a human friendly error description (replacing the stack trace)
- [ ] Prevent this error from getting to TrackJS

jviotti:
Turns out this issue is already fixed. In any case, I'm sending a PR to improve the wording of the message.

PR #1222, opened 2017-03-24
style(CLI): improve error message when drive doesn't exist anymore

jviotti:
The current message is short and unfriendly.

Fixes: https://github.com/resin-io/etcher/issues/1125


jviotti:
We'll very soon decouple the full fledged CLI product from the writer process ran from the GUI, so we'll have a chance to tweak the messages in both contexts.

Issue #858, opened 2016-11-09
Display warning if image is too big when using CLI


lurch:
- **Etcher version:** 1.0.0-beta16
- **Operating system and architecture:** Linux x64

When using the Etcher CLI (see https://github.com/resin-io/etcher/blob/master/docs/CLI.md ) there's no check that the image is smaller than the drive size, which means it "allows" you to attempt to write e.g. a 3GB image onto a 2GB USB stick. It then fails with `ENOSPC: Not enough space on the drive`.

stefan-mihaila:
What's the desired behaviour? I see more options:

1. Allow the user to select a disk smaller than the image, and if he does, display a warning and prompt the user if he wishes to continue with the selected disk
2. Allow the user to select a disk smaller than the image, and if he does, bail out with an error without even trying to burn the image
3. Filter out (do not display) disks smaller than the image.

Option 3 is more work to implement, because from what I see, the resin-cli-visuals module is responsible for filtering out system drives: https://github.com/resin-io/resin-cli-visuals/blob/f3d643008be2cd6b5b0c51133080618a1d69d130/build/widgets/drive/index.js#L40. I don't think that the image size should get passed all the way to the visuals module.

lurch:
I guess it depends if we want the CLI to behave the same way as the GUI (which does option 3), or if we want the CLI to be a more 'hardcore' version, in which case I guess option 2 would be best?
I'm not sure why we'd ever want to allow option 1?

ping @konmouz @jviotti

jviotti:
I think it greatly depends on what we're doing with https://github.com/resin-io/etcher/issues/888, since both the CLI and the GUI should have a similar look 'n feel. 

At the time of this writing, the consensus on the issue linked before is that we will show every single drive on the drive selector modal (even if its non-removable, but making it very clear that that's the case), but we will present a warning message if the drive the user selected is not removable.

I honestly don't like how we're depending on `resin-cli-visuals` for this, and since the paradigm of drive selection is changing in the GUI, I feel we need to re-work this.

@stefan-mihaila How do you feel about re-working a CLI drive selection widget that mimics what the GUI will be in the future?

jviotti:
Yeah, I think we should try to mimic the GUI when possible. I propose showing all the drives, but not allowing selection of the ones that are small enough (and present a confirmation warning if selecting a system one).

lurch:
[...]

PR #1223, opened 2017-03-24
fix(CLI): prevent flashing an image that is larger than the drive

jviotti:
The CLI sill happily let you flash an image that doesn't fit on the
drive. This commit prevents such scenario right before flashing, but we
should still implement a smarter CLI drive detection widget for when the
user runs the CLI interactively.

Change-Type: patch
Changelog-Entry: Prevent flashing an image that is larger than the drive with the CLI.
Fixes: https://github.com/resin-io/etcher/issues/858

### DIFF
--- a/lib/cli/writer.js
+++ b/lib/cli/writer.js
@@ -22,6 +22,8 @@ const fs = Bluebird.promisifyAll(require('fs'));
 const os = require('os');
 const unmount = require('./unmount');
 const imageStream = require('../image-stream');
+const errors = require('../shared/errors');
+const constraints = require('../shared/constraints');
 
 /**
  * @summary Write an image to a disk drive
@@ -67,6 +69,15 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
     return fs.openAsync(drive.raw, 'rs+');
   }).then((driveFileDescriptor) => {
     return imageStream.getFromFilePath(imagePath).then((image) => {
+      if (!image.size.final.estimation && !constraints.isDriveLargeEnough(drive, {
+        size: image.size.final.value
+      })) {
+        throw errors.createUserError(
+          'The image you selected is too big for this drive',
+          'Plug in a bigger drive and try again'
+        );
+      }
+
       return imageWrite.write({
         fd: driveFileDescriptor,
         device: drive.raw,


lib/cli/writer.js
```diff
@@ -67,6 +68,13 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
     return fs.openAsync(drive.raw, 'rs+');
   }).then((driveFileDescriptor) => {
     return imageStream.getFromFilePath(imagePath).then((image) => {
+      if (!image.size.final.estimation && image.size.final.value > drive.size) {
+        throw errors.createUserError(
+          'The image you selected is too big for this drive',
+          'Plug a bigger drive and try again'
```
lurch: Should this be using `notEnoughSpaceInDrive` from `lib/shared/messages.js` ?

lurch:
> but we should still implement a smarter CLI drive detection widget for when the
user runs the CLI interactively.

Do you want to create a separate issue for that? ;)  I'll let you create it, because it sounds like you have a better idea of what's needed.

lib/cli/writer.js
jviotti: We should refactor the whole messages thing (I'll making a note to myself), given that the messages in `lib/shared/messages.js` contain the title and description mixed in a single string, where `errors.createUserError` requires them as separate arguments. 

I'll refactor in another PR.

jviotti:
Done: https://github.com/resin-io/etcher/issues/1227

lib/cli/writer.js
```diff
@@ -67,6 +68,13 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
     return fs.openAsync(drive.raw, 'rs+');
   }).then((driveFileDescriptor) => {
     return imageStream.getFromFilePath(imagePath).then((image) => {
+      if (!image.size.final.estimation && image.size.final.value > drive.size) {
```
lurch: I guess this should also be refactored to use the `isDriveLargeEnough` method that the GUI drive-selector code uses?
At the very least the `>` in the line above should be changed to `>=`

lurch:
The "sill" typo in the commit comment should be changed to "still"

lib/cli/writer.js
jviotti: Oh, I completely forgot about that function, good catch! :)

jviotti: Hm, but that function takes an image object, where here we just have a size :( I'll create a dummy image object and pass that as an argument.

jviotti:
Fixed!

lib/cli/writer.js
lurch: Hmmm, does that mean that the `image` object expected by `isDriveLargeEnough` is different to the `image` object returned by `getFromFilePath` ?
Is that deliberate? (and if so, why?) Could/Should more refactoring be done here?

jviotti: It took quite a while, but I was able to normalise these. The resulting refactoring PR is far from trivial, so can we approve and merge this one, and I'll send the refactor PR right afterwards?

lurch: Yeah, go for it :)

lib/cli/writer.js
```diff
@@ -67,6 +69,15 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
     return fs.openAsync(drive.raw, 'rs+');
   }).then((driveFileDescriptor) => {
     return imageStream.getFromFilePath(imagePath).then((image) => {
+      if (!image.size.final.estimation && !constraints.isDriveLargeEnough(drive, {
+        size: image.size.final.value
+      })) {
+        throw errors.createUserError(
+          'The image you selected is too big for this drive',
+          'Plug a bigger drive and try again'
```
jhermsmeier: Should this be "Plug in [...]"? Also, do we want to refer to "drives" as "devices", as most of them these days aren't technically drives?

jviotti: I fixed the "plug in" thing, but I"m not sure about drive/device. I guess it depends on which one users feel more comfortable with. In any case, we're using drive all across the code base, to probably not something we should handle here; we can send a PR that replaces drives with devices across the code base later.

lurch: I also thought the "Plug..." wording was strange, but I didn't bother commenting because Juan mentioned he'd be refactoring all the error messages soon anyway ;-)

jviotti: Oh, I meant I'll refactor how/where messages are stored. I'm definitely open to improve the wording :)

lurch: Ah! Well in that case I think "Please connect a ..." sounds nicer than "Plug a ..." :)

jviotti: See https://github.com/resin-io/etcher/pull/1239

PR #1224, opened 2017-03-24
chore(travis): move deploy command to a separate script

jviotti:
We're passing a shell conditional to `deploy.script`, however Travis CI
seems to get confused about this in the deploy section, causing
GNU/Linux to run `make publish-aws-s3` directly on the Travis CI build
instead of in the Docker container, which causes the deployment to
eventually fail because of missing dependencies.

According to Travis CI:

> `deploy.script` must be a scalar pointing to an executable file or
> command.

See https://travis-ci.org/resin-io/etcher/jobs/214723725 for a failure
example.

We also move the Appveyor deploy script to a separate file for symmetry
purposes.

### DIFF
--- a/scripts/ci/deploy-travis.sh
+++ b/scripts/ci/deploy-travis.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+###
+# Copyright 2017 resin.io
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
+set -e
+set -u
+
+./scripts/build/check-dependency.sh make
+
+if [ -z "$TRAVIS_OS_NAME" ]; then
+  echo "This script is only meant to run in Travis CI" 1>&2
+  exit 1
+fi
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  ./scripts/build/docker/run-command.sh -r "$TARGET_ARCH" -s "$(pwd)" -c "make publish-aws-s3"
+else
+  make publish-aws-s3
+fi
--- a/.gitattributes
+++ b/.gitattributes
@@ -16,6 +16,7 @@ LICENSE text
 Makefile text
 *.md text
 *.sh text
+*.bat text
 *.svg text
 *.yml text
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_install:
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-      ./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "make info && make electron-develop";
+      ./scripts/build/docker/run-command.sh -r "${TARGET_ARCH}" -s "${PWD}" -c "make info && make electron-develop";
     else
       pip install codespell==1.9.2 awscli;
       npm install -g bower asar;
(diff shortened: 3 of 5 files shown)


scripts/ci/deploy-travis.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
lurch: This should probably be `#!/bin/bash` for consistency with all our other scripts

scripts/ci/deploy-travis.sh
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
+set -e
+set -u
+
```
lurch: Would it be worth including a dependency-check for `make` ?

scripts/ci/deploy-travis.sh
```diff
@@ -0,0 +1,26 @@
...
+if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+  ./scripts/build/docker/run-command.sh -r $TARGET_ARCH -s $PWD -c "make publish-aws-s3"
```
lurch: This should probably use the same "$HERE/...etc..." that other scripts use?

(Hmmm, would it be worth submitting a separate PR to update _all_ our scripts to use the "$HERE/...etc..." style of referencing paths?)

lurch:
Ahhh, if Travis is indeed running `make publish-aws-s3` instead of `./scripts/build/docker/run-command.sh -r ${TARGET_ARCH} -s ${PWD} -c "make publish-aws-s3"`, that would explain the weird `Dependency missing: bower` error. Good job @jviotti !

Doesn't really make much difference, but for the sake of symmetry would it be worth adding a `scripts/ci/deploy-appveyor.sh` too?

scripts/ci/deploy-travis.sh
jviotti: Done!

scripts/ci/deploy-travis.sh
jviotti: Makes sense, done!

scripts/ci/deploy-travis.sh
jviotti: > (Hmmm, would it be worth submitting a separate PR to update all our scripts to use the "$HERE/...etc..." style of referencing paths?)

Yeah, lets do that instead!

jviotti:
:) Done!

scripts/ci/deploy-travis.sh
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+###
+
+set -e
+set -u
+
+./scripts/build/check-dependency.sh make
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  ./scripts/build/docker/run-command.sh -r $TARGET_ARCH -s $PWD -c "make publish-aws-s3"
```
lurch: We should quote these variables, and `$PWD` might be better as `"$(pwd)"`

(and the variables in `.travis.yml` ought to be quoted too)

lurch:
Should we make these `deploy...` scripts fail if `$CI` isn't set to 1?
And/or we could make the deploy-travis script fail if it doesn't detect the Travis-specific envvars, and make the deploy-appveyor script fail if it doesn't detect the Appveyor-specific envvars?

scripts/ci/deploy-travis.sh
jviotti: Fixed!

jviotti:
> Should we make these deploy... scripts fail if $CI isn't set to 1?
And/or we could make the deploy-travis script fail if it doesn't detect the Travis-specific envvars, 

It will fail out of the box given that we use `set -u`. Do you think we should still add specific conditionals in order to fail with a better error message?

> and make the deploy-appveyor script fail if it doesn't detect the Appveyor-specific envvars?

I'll add a conditional there.

lurch:
IMHO it'd be better to explicitly check (as you're now doing in the appveyor script), rather than "hoping" that `set -u` always remains set and using that to catch any errors.

jviotti:
Sounds good. Fixed!

PR #1226, opened 2017-03-26
chore: don't bother sending any build context to Docker

lurch:
We don't actually use any build context, as we just mount the local
directory as a volume anyway. Omitting it speeds up the container
start-up time. This is an enhanced version of #1173


jviotti:
Cool stuff. Looks like you're getting to a "Docker expert" level already :)

PR #1229, opened 2017-03-28
refactor: unify image objects across GUI and image-stream

jviotti:
`image-stream` returns image objects that look like this:

```js
{
  stream: <readable stream>,
  transform: <transform stream>,
  size: {
    original: <number>,
    final: {
      value: <number>,
      estimation: <boolean>
    }
  },
  ...
}
```

While the GUI handles image objects that look like this:

```sh
{
  path: <string>,
  size: <number>,
  ...
}
```

It looks like we should share a common structure between both, so we can
use `image-stream` images in `drive-constraints`, for example.

Turns out that we actually transform `image-stream` image objects to GUI
image objects when the user selects an image using the image selector
dialog, which is another indicator that we should normalise this
situation.

As a solution, this commit does the following:

- Add `path` to `image-stream` image object
- Reuse `image-stream` image objects in the GUI, given they are a
  superset of GUI image objects

See: https://github.com/resin-io/etcher/pull/1223#discussion_r108165110
Fixes: https://github.com/resin-io/etcher/issues/1232

### DIFF
--- a/lib/cli/writer.js
+++ b/lib/cli/writer.js
@@ -69,9 +69,7 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
     return fs.openAsync(drive.raw, 'rs+');
   }).then((driveFileDescriptor) => {
     return imageStream.getFromFilePath(imagePath).then((image) => {
-      if (!image.size.final.estimation && !constraints.isDriveLargeEnough(drive, {
-        size: image.size.final.value
-      })) {
+      if (!constraints.isDriveLargeEnough(drive, image)) {
         throw errors.createUserError(
           'The image you selected is too big for this drive',
           'Please connect a bigger drive and try again'
--- a/lib/gui/models/store.js
+++ b/lib/gui/models/store.js
@@ -287,8 +287,22 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         throw errors.createError('Missing image size');
       }
 
-      if (!_.isNumber(action.data.size)) {
+      if (!_.isPlainObject(action.data.size)) {
         throw errors.createError(`Invalid image size: ${action.data.size}`);
+      }
+
+      const MINIMUM_IMAGE_SIZE = 0;
+
+      if (!_.isInteger(action.data.size.original) || action.data.size.original < MINIMUM_IMAGE_SIZE) {
+        throw errors.createError(`Invalid original image size: ${action.data.size.original}`);
+      }
+
+      if (!_.isInteger(action.data.size.final.value) || action.data.size.final.value < MINIMUM_IMAGE_SIZE) {
+        throw errors.createError(`Invalid final image size: ${action.data.size.final.value}`);
+      }
+
+      if (!_.isBoolean(action.data.size.final.estimation)) {
+        throw errors.createError(`Invalid final image size estimation flag: ${action.data.size.final.estimation}`);
       }
 
       if (action.data.url && !_.isString(action.data.url)) {
--- a/lib/gui/pages/main/templates/main.tpl.html
+++ b/lib/gui/pages/main/templates/main.tpl.html
@@ -1,6 +1,6 @@
 <div class="page-main row around-xs">
   <div class="col-xs" ng-controller="ImageSelectionController as image">
-    <div class="box text-center relative" os-dropzone="image.selectImage($file)">
+    <div class="box text-center relative" os-dropzone="image.selectImageByPath($file)">
 
       <svg-icon
         class="center-block"
--- a/lib/image-stream/archive.js
+++ b/lib/image-stream/archive.js
@@ -195,6 +195,7 @@ exports.extractImage = (archive, hooks) => {
     }).then((results) => {
       results.metadata.stream = results.imageStream;
       results.metadata.transform = new PassThroughStream();
+      results.metadata.path = archive;
 
       results.metadata.size = {
         original: imageEntry.size,
--- a/lib/image-stream/handlers.js
+++ b/lib/image-stream/handlers.js
@@ -48,6 +48,7 @@ module.exports = {
    */
   'application/x-bzip2': (file, options) => {
     return Bluebird.props({
+      path: file,
       stream: fs.createReadStream(file),
       size: {
         original: options.size,
@@ -76,6 +77,7 @@ module.exports = {
   'application/gzip': (file, options) => {
     return gzip.getUncompressedSize(file).then((uncompressedSize) => {
       return Bluebird.props({
+        path: file,
         stream: fs.createReadStream(file),
         size: {
           original: options.size,
@@ -109,6 +111,7 @@ module.exports = {
       });
     }).then((metadata) => {
       return {
+        path: file,
         stream: fs.createReadStream(file),
         size: {
           original: options.size,
@@ -151,6 +154,7 @@ module.exports = {
    */
   'application/octet-stream': (file, options) => {
     return Bluebird.props({
+      path: file,
       stream: fs.createReadStream(file),
       size: {
         original: options.size,
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -142,7 +142,31 @@ exports.isSourceDrive = (drive, image) => {
  * }
  */
 exports.isDriveLargeEnough = (drive, image) => {
-  return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'size' ], UNKNOWN_SIZE);
+  const driveSize = _.get(drive, [ 'size' ], UNKNOWN_SIZE);
+
+  if (_.get(image, [ 'size', 'final', 'estimation' ])) {
+
+    // If the drive size is smaller than the original image size, and
+    // the final image size is just an estimation, then we stop right
+    // here, based on the assumption that the final size will never
+    // be less than the original size.
+    if (driveSize < _.get(image, [ 'size', 'original' ], UNKNOWN_SIZE)) {
+      return false;
+    }
+
+    // If the final image size is just an estimation then consider it
+    // large enough. In the worst case, the user gets an error saying
+    // the drive has ran out of space, instead of prohibiting the flash
+    // at all, when the estimation may be wrong.
+    return true;
+
+  }
+
+  return driveSize >= _.get(image, [
+    'size',
+    'final',
+    'value'
+  ], UNKNOWN_SIZE);
 };
 
 /**
--- a/tests/gui/models/selection-state.spec.js
+++ b/tests/gui/models/selection-state.spec.js
@@ -221,7 +221,13 @@ describe('Browser: SelectionState', function() {
       beforeEach(function() {
         this.image = {
           path: 'foo.img',
-          size: 999999999,
+          size: {
+            original: 999999999,
+            final: {
+              estimation: false,
+              value: 999999999
+            }
+          },
           recommendedDriveSize: 1000000000,
           url: 'https://www.raspbian.org',
           supportUrl: 'https://www.raspbian.org/forums/',
@@ -336,7 +342,13 @@ describe('Browser: SelectionState', function() {
         it('should override the image', function() {
           SelectionStateModel.setImage({
             path: 'bar.img',
-            size: 999999999
+            size: {
+              original: 999999999,
+              final: {
+                estimation: false,
+                value: 999999999
+              }
+            }
           });
 
           const imagePath = SelectionStateModel.getImagePath();
@@ -369,7 +381,13 @@ describe('Browser: SelectionState', function() {
         it('should be able to set an image', function() {
           SelectionStateModel.setImage({
             path: 'foo.img',
-            size: 999999999
+            size: {
+              original: 999999999,
+              final: {
+                estimation: false,
+                value: 999999999
+              }
+            }
           });
 
           const imagePath = SelectionStateModel.getImagePath();
@@ -381,7 +399,13 @@ describe('Browser: SelectionState', function() {
         it('should throw if no path', function() {
           m.chai.expect(function() {
             SelectionStateModel.setImage({
-              size: 999999999
+              size: {
+                original: 999999999,
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
             });
           }).to.throw('Missing image path');
         });
@@ -390,7 +414,13 @@ describe('Browser: SelectionState', function() {
           m.chai.expect(function() {
             SelectionStateModel.setImage({
               path: 123,
-              size: 999999999
+              size: {
+                original: 999999999,
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
             });
           }).to.throw('Invalid image path: 123');
         });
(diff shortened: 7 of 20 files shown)


lib/image-stream/archive.js
```diff
@@ -195,6 +195,7 @@ exports.extractImage = (archive, hooks) => {
     }).then((results) => {
       results.metadata.stream = results.imageStream;
       results.metadata.transform = new PassThroughStream();
+      results.metadata.path = archive;
```
lurch: Could we add some unit-tests to check that when given e.g. `/path/to/rpi.img.xz` or `/path/to/rpi.zip` then the path stays as `/path/to/rpi.img.xz` or `/path/to/rpi.zip` and doesn't get transformed to `/path/to/rpi.img` or `rpi.img` ?

lib/image-stream/handlers.js
```diff
@@ -76,6 +77,7 @@ module.exports = {
   'application/gzip': (file, options) => {
     return gzipUncompressedSize.fromFileAsync(file).then((uncompressedSize) => {
       return Bluebird.props({
+        path: file,
```
lurch: What would the `path` be set to in the case where we're doing a streaming download?

lib/cli/writer.js
```diff
@@ -23,7 +23,7 @@ const os = require('os');
 const unmount = require('./unmount');
 const imageStream = require('../image-stream');
 const errors = require('../shared/errors');
-const constraints = require('../shared/constraints');
+const constraints = require('../shared/drive-constraints');
```
lurch: Ooops, nasty typo in #1223 ? :-(
Does the fact that this wasn't caught mean that it wasn't being unit-tested properly? Could this PR add appropriate unit-tests?

lurch:
Great work @jviotti  :-D

jviotti:
Drag and drop is broken in this PR.

tests/shared/drive-constraints.spec.js
```diff
@@ -303,52 +303,219 @@ describe('Shared: DriveConstraints', function() {
...
+      };
+    });
+
+    describe('given the final image size estimation flag is false', function() {
+
+      describe('given the original size is less than the drive', function() {
+
+        beforeEach(function() {
+          this.image = {
+            path: 'rpi.img',
```
jviotti: Done!

lib/image-stream/handlers.js
jviotti: The path would be the URL.

lib/cli/writer.js
jviotti: This writer file ties together things like unmounting, and actual writing, so it'd be hard to automatically test without some NUCs and a custom test runner :)

lib/image-stream/archive.js
jviotti: This particular test is covering that: https://github.com/resin-io/etcher/blob/f242443d29864346ef9860879ea447bded283220/tests/image-stream/zip.spec.js#L74

lib/shared/drive-constraints.js
```diff
@@ -142,7 +142,22 @@ exports.isSourceDrive = (drive, image) => {
  * }
  */
 exports.isDriveLargeEnough = (drive, image) => {
-  return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'size' ], UNKNOWN_SIZE);
+
+  // If the final image size is just an estimation then consider it
+  // large enough. In the worst case, the user gets an error saying
+  // the drive has ran out of space, instead of prohibiting the flash
+  // at all, when the estimation may be wrong.
```
jviotti: > surely the 'real' final size will always be >= the original image size?

Sounds like a reasonable assumption. I'll update the code and test suites.

lib/cli/writer.js
lurch: You mean something like an AutoHAT? :wink: 

This typo inspired #1237 BTW ;)

jviotti: I wish we could re-use auto hat. I spoke with the team at some point, but they would only support GNU/Linux :(

jviotti:
Drag and drop is up and running again!

lib/cli/writer.js
lurch: Even "only" running hardware tests on Linux, would still be better than no hardware tests at all? ;)

jviotti:
We have plans to build such a testing infrastructure soon, and even got the green line to order any equipment we need. We just need to put some time into it :) 


—
Juan Cruz Viotti

--- original message ---
On Wed, Mar 29, 2017 at 12:36 pm, <notifications@github.com> Andrew Scheller wrote:


@lurch commented on this pull request.

In lib/cli/writer.js:

> @@ -23,7 +23,7 @@ const os = require('os');
const unmount = require('./unmount');
const imageStream = require('../image-stream');
const errors = require('../shared/errors');
-const constraints = require('../shared/constraints');
+const constraints = require('../shared/drive-constraints');

Even "only" running hardware tests on Linux, would still be better than no hardware tests at all? ;)

—

You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub, or mute the thread.
--- end of original message ---

lib/cli/writer.js
jviotti: We have plans to build such a testing infrastructure soon, and even got the green sign to order any equipment we need. We just need to put some time into it :)

lib/shared/drive-constraints.js
```diff
@@ -142,7 +142,31 @@ exports.isSourceDrive = (drive, image) => {
  * }
  */
 exports.isDriveLargeEnough = (drive, image) => {
-  return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'size' ], UNKNOWN_SIZE);
+  const driveSize = _.get(drive, [ 'size' ], UNKNOWN_SIZE);
+
+  if (_.get(image, [ 'size', 'final', 'estimation' ])) {
```
lurch: Does there need to be some kind of fallback value here in case the image is null?

tests/shared/drive-constraints.spec.js
```diff
@@ -303,52 +303,219 @@ describe('Shared: DriveConstraints', function() {
...
+      };
+    });
+
+    describe('given the final image size estimation flag is false', function() {
+
+      describe('given the original size is less than the drive size', function() {
+
+        beforeEach(function() {
+          this.image = {
+            path: path.join(__dirname, 'rpi.img'),
```
lurch: There's lots of other image relative-only `path` properties scattered throughout the test scripts. If you like I could make the change you've done here to all of them, in a separate PR after you've merged this one?

tests/shared/drive-constraints.spec.js
```diff
@@ -303,52 +303,219 @@ describe('Shared: DriveConstraints', function() {
...
+        it('should return false if the final size is greater than the drive size', function() {
+          this.image.size.final.value = this.drive.size + 1;
+          m.chai.expect(constraints.isDriveLargeEnough(this.drive, this.image)).to.be.false;
+        });
+
       });
 
-      m.chai.expect(result).to.be.true;
-    });
+      describe('given the original size is equal to the drive', function() {
```
lurch: "the drive size"

tests/shared/drive-constraints.spec.js
```diff
@@ -303,52 +303,219 @@ describe('Shared: DriveConstraints', function() {
...
+            }
+          };
+        });
+
+        it('should return true if the final size is less than the drive size', function() {
+          this.image.size.final.value = this.drive.size - 1;
+          m.chai.expect(constraints.isDriveLargeEnough(this.drive, this.image)).to.be.true;
+        });
+
+        it('should return true if the final size is equal to the drive', function() {
```
lurch: "the drive size" ;)

tests/shared/drive-constraints.spec.js
```diff
@@ -303,52 +303,219 @@ describe('Shared: DriveConstraints', function() {
...
```
lurch: ...  ;-)

tests/shared/drive-constraints.spec.js
```diff
@@ -303,52 +303,219 @@ describe('Shared: DriveConstraints', function() {
...
-    it('should return false if the drive size is less than the image size', function() {
-      const result = constraints.isDriveLargeEnough({
-        device: '/dev/disk1',
-        name: 'USB Drive',
-        size: 1000000000,
-        protected: false
-      }, {
-        path: 'rpi.img',
-        size: 1000000001
+    describe('given the final image size estimation flag is true', function() {
```
lurch: Would 'given the final image size is an estimation' be more readable? Or is it better to have the test descriptions be brutally accurate?

lib/shared/drive-constraints.js
jviotti: If the image is `null`, then this line will resolve to `undefined`, which is casted to `false`, and eventually results in `true` in the last line of this function.

tests/shared/drive-constraints.spec.js
jviotti: Yeah, that'd be great!

tests/shared/drive-constraints.spec.js
jviotti: Oh, looks like `s/than the drive/than the drive size` was not enough :P Fixed.

tests/shared/drive-constraints.spec.js
jviotti: I kinda like the way it is at the moment, otherwise you have no idea what "is an estimation" means. What do you think?

tests/shared/drive-constraints.spec.js
lurch: At least you can rest safe in the knowledge that I'll probably catch any omissions :laughing:

tests/shared/drive-constraints.spec.js
lurch: Fair enough, leave it as it is then :+1:

tests/gui/models/selection-state.spec.js
```diff
@@ -403,20 +433,62 @@ describe('Browser: SelectionState', function() {
...
+              size: {
+                original: 999999999,
+                final: {
+                  estimation: 'false',
+                  value: 999999999
+                }
+              }
             });
-          }).to.throw('Invalid image size: 999999999');
+          }).to.throw('Invalid final image size estimation flag: false');
```
lurch: Same comment as my previous one ;-)

tests/shared/drive-constraints.spec.js
```diff
@@ -490,29 +657,53 @@ describe('Shared: DriveConstraints', function() {
...
+            }
+          }
         })).to.be.false;
       });
 
       it('should return false if the drive is large enough and is a source drive', function() {
         m.chai.expect(constraints.isDriveValid(this.drive, {
-          path: path.join(this.mountpoint, 'rpi.img'),
-          size: 2000000000
+          path: path.join(this.mountpoint, path.join(__dirname, 'rpi.img')),
```
lurch: ??
We don't really need two `path.join`s here do we?

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -1,6 +1,6 @@
 <div class="page-main row around-xs">
   <div class="col-xs" ng-controller="ImageSelectionController as image">
-    <div class="box text-center relative" os-dropzone="image.selectImage($file)">
+    <div class="box text-center relative" os-dropzone="image.selectImageByPath($file.path)">
```
Shou: `$file` seems to be a path so it needs to be called as-is, so `$file.path` is `undefined` and results in an exception raised by some function nested in the call stack of `selectImagePath`.

jviotti: Amazing catch! Looks like I forgot to update this after refactoring `dropzone`. Fixed!

tests/shared/drive-constraints.spec.js
jviotti: Oh, a global replace glitch :D

tests/gui/models/selection-state.spec.js
```diff
@@ -403,20 +433,62 @@ describe('Browser: SelectionState', function() {
...
+              path: 'foo.img',
+              size: {
+                original: '999999999',
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
+            });
+          }).to.throw('Invalid original image size: 999999999');
```
jviotti: That doesn't look trivial to implement, unless we check if the invalid thing is a string, and then manually quote ourselves, otherwise we'd be unnecessarily quoting things like objects, or booleans. This errors are things being caught by our test suite, so I think its fine. Seeing that `999999999` is invalid instantly leads to the fact that is really a string after taking a glance at the code.

lurch: Does Javascript / NodeJS have anything like the `repr` function from Python?
But this is obviously only a small point so feel free to ignore.

jviotti: I don't think so. `JSON.stringify()` is the closest thing I can think of, but not sure if we should abuse that function for this purpose.

lurch: I did a quick search and found https://gist.github.com/JacobOscarson/734620 but I agree this probably isn't worth changing.

tests/shared/drive-constraints.spec.js
```diff
@@ -490,29 +657,53 @@ describe('Shared: DriveConstraints', function() {
 
       it('should return false if the drive is not large enough and is a source drive', function() {
         m.chai.expect(constraints.isDriveValid(this.drive, {
-          path: path.join(this.mountpoint, 'rpi.img'),
-          size: 5000000000
+          path: path.join(this.mountpoint, path.join(__dirname, 'rpi.img')),
```
lurch: Another not-wanted double use of `path.join`?  ;-P

lib/image-stream/handlers.js
jviotti: It doesn't sound crazy, but I think we should wait for the streaming functionality to come in before doing such change.

lib/shared/drive-constraints.js
```diff
@@ -142,7 +142,31 @@ exports.isSourceDrive = (drive, image) => {
...
+
+    // If the final image size is just an estimation then consider it
+    // large enough. In the worst case, the user gets an error saying
+    // the drive has ran out of space, instead of prohibiting the flash
+    // at all, when the estimation may be wrong.
+    return true;
+
+  }
+
+  return driveSize >= _.get(image, [
```
jviotti: Sounds good. Done!

tests/shared/drive-constraints.spec.js
jviotti: Good catch, fixed!

lib/gui/models/store.js
```diff
@@ -287,10 +287,22 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         throw errors.createError('Missing image size');
       }
 
-      if (!_.isNumber(action.data.size)) {
+      if (!_.isPlainObject(action.data.size)) {
         throw errors.createError(`Invalid image size: ${action.data.size}`);
       }
 
+      if (!_.isNumber(action.data.size.original)) {
```
jviotti: Done! BTW, I'm making a note to myself to use something like JSON schema to decoratively validate these objects, so we don't have to write dozens of conditionals.

lurch: I'm assuming you meant declaratively rather than decoratively ;-)

Sounds good, as long as parsing and validating the schema doesn't slow things down.
Hmm, I wonder if we could get the best of both worlds and use some module which (at build time - like we do with the sass) converts a schema into a list of conditionals to test? :)

lib/gui/models/store.js
```diff
@@ -287,10 +287,24 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
         throw errors.createError(`Invalid image size: ${action.data.size}`);
       }
 
+      const MINIMUM_DRIVE_SIZE = 0;
+
+      if (!_.isInteger(action.data.size.original) || action.data.size.original < MINIMUM_DRIVE_SIZE) {
+        throw errors.createError(`Invalid original image size: ${action.data.size.original}`);
+      }
+
+      if (!_.isInteger(action.data.size.final.value) || action.data.size.final.value < MINIMUM_DRIVE_SIZE) {
```
lurch: Probably not worth it for just two if-tests, but I guess if there were more you could abstract this into an `isPositiveInteger` function.

lib/shared/drive-constraints.js
```diff
@@ -142,7 +142,31 @@ exports.isSourceDrive = (drive, image) => {
  * }
  */
 exports.isDriveLargeEnough = (drive, image) => {
-  return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'size' ], UNKNOWN_SIZE);
+  const driveSize = _.get(drive, [ 'size' ], UNKNOWN_SIZE);
+
+  if (_.get(image, [ 'size', 'final', 'estimation' ])) {
+
+    // If the original image size is larger than the drive size, and
```
lurch: Another minor nit, but it _might_ be worth updating the comment to match the code? i.e. "If the drive size is smaller than the original image size, ..."

tests/shared/drive-constraints.spec.js
lurch: Are you sure it's fixed? Or is GH displaying an out-of-date diff for some reason? :-/

lurch:
All the `pr` CI builds are failing with an error in the `.getImageBasename()` test.

lib/gui/models/store.js
jviotti: I don't know where I wrote it, but I made a note to myself to abstract all these error checks in some declarative way (maybe JSON schema?), which should address this issue as well.

lib/shared/drive-constraints.js
jviotti: Definitely, fixed!

tests/shared/drive-constraints.spec.js
jviotti: I think its a GH issue. The chance has been made locally, and I have no un-pushed commits. Can you try cloning the PR?

jviotti:
Fixed!

lurch:
@Shou Can you re-test this please and make sure the DnD is all working properly now?

tests/shared/drive-constraints.spec.js
lurch: Still there in a local checkout (commit a7341086a2e825f6b1ee016f51a0707c67c4c3ea) of the `unify-image-object` branch:
```bash
$ grep -n path\.join tests/shared/drive-constraints.spec.js 
321:            path: path.join(__dirname, 'rpi.img'),
352:            path: path.join(__dirname, 'rpi.img'),
383:            path: path.join(__dirname, 'rpi.img'),
418:            path: path.join(__dirname, 'rpi.img'),
449:            path: path.join(__dirname, 'rpi.img'),
480:            path: path.join(__dirname, 'rpi.img'),
511:        path: path.join(__dirname, 'rpi.img'),
551:        path: path.join(__dirname, 'rpi.img'),
566:        path: path.join(__dirname, 'rpi.img'),
581:        path: path.join(__dirname, 'rpi.img'),
596:        path: path.join(__dirname, 'rpi.img'),
605:        path: path.join(__dirname, 'rpi.img'),
660:          path: path.join(this.mountpoint, 'rpi.img'),
686:          path: path.join(this.mountpoint, 'rpi.img'),
720:          path: path.join(this.mountpoint, path.join(__dirname, 'rpi.img')),
746:          path: path.join(this.mountpoint, path.join(__dirname, 'rpi.img')),
```

Shou:
Drag and drop does indeed work well now

lib/gui/models/store.js
lurch: It was in another comment on this PR ;-)

jviotti:
Ah, there were more duplicated `path.join` in other parts of the file (I was looking at just a certain line number). I fixed all of them!

lib/gui/models/store.js
```diff
@@ -287,10 +287,24 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         throw errors.createError('Missing image size');
       }
 
-      if (!_.isNumber(action.data.size)) {
+      if (!_.isPlainObject(action.data.size)) {
         throw errors.createError(`Invalid image size: ${action.data.size}`);
       }
 
+      const MINIMUM_DRIVE_SIZE = 0;
```
lurch: This should probably be `MINIMUM_IMAGE_SIZE` because it's actually the sizes of the images that we're checking here.

tests/gui/models/selection-state.spec.js
```diff
@@ -403,20 +433,122 @@ describe('Browser: SelectionState', function() {
...
+                  estimation: 'false',
+                  value: 999999999
+                }
+              }
             });
-          }).to.throw('Invalid image size: 999999999');
+          }).to.throw('Invalid final image size estimation flag: false');
         });
 
         it('should throw if url is defined but its not a string', function() {
```
lurch: minor nit: the `its` here should be `it's`

lib/gui/models/store.js
```diff
@@ -287,10 +287,24 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         throw errors.createError('Missing image size');
       }
 
-      if (!_.isNumber(action.data.size)) {
+      if (!_.isPlainObject(action.data.size)) {
         throw errors.createError(`Invalid image size: ${action.data.size}`);
```
lurch: Looks like there's not actually a unit-test for this now?

tests/shared/drive-constraints.spec.js
```diff
@@ -526,29 +717,53 @@ describe('Shared: DriveConstraints', function() {
 
       it('should return false if the drive is not large enough and is a source drive', function() {
         m.chai.expect(constraints.isDriveValid(this.drive, {
-          path: path.join(this.mountpoint, 'rpi.img'),
-          size: 5000000000
+          path: path.join(this.mountpoint, __dirname, 'rpi.img'),
```
lurch: Ummm, huh? Shouldn't this have been left as `path: path.join(this.mountpoint, 'rpi.img'),` ?

lib/gui/models/store.js
jviotti: True, I'm glad you have my back :D

tests/gui/models/selection-state.spec.js
jviotti: Fixed!

tests/shared/drive-constraints.spec.js
jviotti: lol, these `path.join` issues keep biting me :P Fixed!

lib/gui/models/store.js

Issue #1225, opened 2017-03-25
Flickering On The Screen


Eliteguardians:
- **Etcher version:**
- **Operating system and architecture:**
- **Do you see any meaningful error information on DevTools?**

     Hi I was wondering how do i remove this key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 379CE192D401AB61 and the update from the linux mint 17.3. rosa?

jviotti:
Hi @Eliteguardians ,

I'm afraid I don't understand the issue. The title describes a flickering on the screen, but the description mentions some PGP issue and a Linux Mint 17.3 update. Can you clarify?

lurch:
I wonder if @Eliteguardians is asking how to "undo" all the steps in https://github.com/resin-io/etcher/blob/master/README.md#debian-and-ubuntu-based-package-repository-gnulinux-x86x64 ?

It should simply be a case of:
```bash
sudo apt-get remove etcher-electron
sudo apt-key del D401AB61
sudo rm /etc/apt/sources.list.d/etcher.list
sudo apt-get update
```

Although AFAIK Etcher shouldn't make your screen flicker, so that might be being caused by some other update you installed.

jviotti:
I see. I sent a PR to add uninstall instructions to the README.

lurch:
@Eliteguardians Feel free to comment and re-open this issue if it doesn't solve your problem.

PR #1231, opened 2017-03-28
docs(README): uninstall instructions for debian/brew

jviotti:
We have inline instructions to install Etcher from package managers.
This commit extends them to quickly mention how to uninstall as well.

Fixes: https://github.com/resin-io/etcher/issues/1225

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,14 @@ installers for all supported operating systems.
     sudo apt-get install etcher-electron
     ```
 
+##### Uninstall
+
+```sh
+sudo apt-get remove etcher-electron
+sudo rm /etc/apt/sources.list.d/etcher.list
+sudo apt-get update
+```
+
 #### Brew Cask (macOS)
 
 Note that the Etcher Cask has to be updated manually to point to new versions,
@@ -66,6 +74,12 @@ release.
 
 ```sh
 brew cask install etcher
+```
+
+##### Uninstall
+
+```sh
+brew cask uninstall etcher
 ```
 
 Support


README.md
```diff
@@ -58,6 +58,15 @@ installers for all supported operating systems.
     sudo apt-get install etcher-electron
     ```
 
+##### Uninstall
+
+```sh
+sudo apt-get remove etcher-electron
+sudo apt-key del D401AB61
```
lurch: In hindsight, I'm now not sure if the key-removal should be in the uninstall instructions. AFAIK it doesn't hurt to leave it in, and there _might_ be other packages on the user's system also installed from bintray?

jviotti: So do you think we should close this PR? Uninstall instructions for these kinds of installations are probably easy to find in the internet, and we can probably handle in case by case basis if really needed.

lurch: Nah, I meant we should keep the PR but we should probably just remove the `sudo apt-key del D401AB61` line?

jhermsmeier: I agree with @lurch, and suggest removing the key removal step, but keeping the uninstall instructions. Can't hurt to have them – at least people looking for it will find it quickly, without having to search the interwebs.

jviotti: Oh, got it. Done!

lurch:
Unrelated to this PR, but something I've just spotted: looks strange that the Debian package is `etcher-electron` but the Brew package is `etcher` ?

jviotti:
The brew package was created by the community, so we didn't have a chance to discuss naming :(

jviotti:
I think we could do it, however what about waiting until the CLI is properly released, so we have something concrete to show them that we might have even more Etcher frontends in the future?

PR #1233, opened 2017-03-29
feat(GUI): display drive name and description in tooltip

Shou:
In the main page we show the drive description, and on hover a tooltip
shows the drive name. This behaviour is changed to show both in the
tooltip, un-truncatetd.

Closes: https://github.com/resin-io/etcher/issues/1170
Changelog-Entry: Display both drive name and description in the drive tooltip.


Shou:
image

PR #1235, opened 2017-03-29
chore: make `thread-sleep` and `spawn-sync` production dependencies

jviotti:
These are dynamic dependencies we've recently added to the list of top
level optional dependencies in order to be able to resolve them when
using browserify to package the Etcher CLI.

The issue is that `thread-sleep` is using a very old `node-pre-gyp`
version (0.6.9) that doesn't support Electron at all, thus resulting in
the following error when setting `npm_config_runtime=electron`:

```
node-pre-gyp ERR! UNCAUGHT EXCEPTION
node-pre-gyp ERR! stack Error: Unknown Runtime: 'electron'
```

This commit makes these dependencies production dependencies, so we can
tweak the `node-pre-gyp` version to make it work.

See: https://github.com/resin-io/etcher/pull/1228
See: https://github.com/mapbox/node-pre-gyp/issues/281

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1019,19 +1019,16 @@
       "version": "1.5.2",
       "from": "concat-stream@>=1.5.1 <1.6.0",
       "resolved": "https://registry.npmjs.org/concat-stream/-/concat-stream-1.5.2.tgz",
-      "dev": true,
       "dependencies": {
         "isarray": {
           "version": "1.0.0",
           "from": "isarray@~1.0.0",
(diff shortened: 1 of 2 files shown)


lurch:
Sheesh, it seems bonkers how such a small change in `package.json` can result in such a huge change to `npm-shrinkwrap.json` :confused:

jviotti:
Yeah, that's why I'm confident that getting rid of trivial dependencies will result in a much smaller shrinkwrap file.

lurch:
Based on our discussions in Gitter, I guess this PR would be better off being reverted?

jviotti:
Done!

PR #1236, opened 2017-03-29
chore(package): Make clean-shrinkwrap remove optional dependencies

jhermsmeier:
Previously dependencies weren't actually removed from `node_modules`,
this runs `npm rm` on the optional dependencies, effectively excluding
them, and their dependencies from the shrinkwrap file.

Also the script has been hooked to the `preshrinkwrap` hook,
to remove the need of having to run it manually.

**NOTE:** This depends on #1234 being merged first, to avoid mangling the `npm-shrinkwrap.json` several times.

Connects-To: #1234
Change-Type: patch

### DIFF
--- a/scripts/clean-shrinkwrap.js
+++ b/scripts/clean-shrinkwrap.js
@@ -17,13 +17,28 @@
 
 const _ = require('lodash');
 const path = require('path');
-const jsonfile = require('jsonfile');
 const packageJSON = require('../package.json');
+const spawn = require('child_process').spawn;
 const shrinkwrapIgnore = _.union(packageJSON.shrinkwrapIgnore, _.keys(packageJSON.optionalDependencies));
-const SHRINKWRAP_PATH = path.join(__dirname, '..', 'npm-shrinkwrap.json');
 
-const shrinkwrapContents = jsonfile.readFileSync(SHRINKWRAP_PATH);
-shrinkwrapContents.dependencies = _.omit(shrinkwrapContents.dependencies, shrinkwrapIgnore);
-jsonfile.writeFileSync(SHRINKWRAP_PATH, shrinkwrapContents, {
-  spaces: 2
-});
+console.log('Removing:', shrinkwrapIgnore.join(', '));
+
+/**
+ * Run an npm command
+ * @param {Array} command - list of arguments
+ * @returns {ChildProcess}
+ */
+const npm = (command) => {
+  return spawn('npm', command, {
+    cwd: path.join(__dirname, '..'),
+    env: process.env,
+    stdio: [ process.stdin, process.stdout, process.stderr ]
+  });
+};
+
+npm([ 'rm', '--ignore-scripts' ].concat(shrinkwrapIgnore))
+  .once('close', () => {
+    npm([ 'prune' ]).once('close', () => {
+      console.log('Done.');
+    });
+  });
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -44,10 +44,6 @@ Use the following steps to ensure everything goes flawlessly:
 - Install the new version of the dependency. For example: `npm install --save
   <package>@<version>`. This will update the `npm-shrinkwrap.json` file.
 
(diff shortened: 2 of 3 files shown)


lurch:
You probably want to update `docs/CONTRIBUTING.md` too ;-)

scripts/clean-shrinkwrap.js
```diff
@@ -17,13 +17,28 @@
 
 const _ = require('lodash');
 const path = require('path');
-const jsonfile = require('jsonfile');
-const packageJSON = require('../package.json');
-const shrinkwrapIgnore = _.union(packageJSON.shrinkwrapIgnore, _.keys(packageJSON.optionalDependencies));
-const SHRINKWRAP_PATH = path.join(__dirname, '..', 'npm-shrinkwrap.json');
+const pkg = require('../package.json');
```
lurch: Could we put this back to `const packageJSON` ? It's what all the rest of the code uses...

scripts/clean-shrinkwrap.js
```diff
@@ -17,13 +17,28 @@
...
+const spawn = require('child_process').spawn;
+const ignore = _.union(pkg.shrinkwrapIgnore, _.keys(pkg.optionalDependencies));
```
lurch: And for diff-viewing purposes, would it hurt to leave this as `shrinkwrapIgnore` ?

scripts/clean-shrinkwrap.js
```diff
@@ -17,13 +17,28 @@
...
+    cwd: path.join(__dirname, '..'),
+    env: process.env,
+    stdio: [ process.stdin, process.stdout, process.stderr ]
+  });
+};
+
+npm([ 'rm', '--ignore-scripts' ].concat(ignore))
+  .once('close', () => {
+    npm([ 'prune' ]).once('close', () => {
+      console.log('Done.');
```
lurch: If all this is doing is executing external commands, my gut instinct is saying that it might be easier done in a bash script? ;)

EDIT: But then again, perhaps we ought to restrict Bash scripts from being run from the `Makefile`, and (for cross-platform / self-containment reasons) any jobs run from `package.json` probably do work best as Javascript.

jhermsmeier:
Updated `CONTRIBUTING.md`

scripts/clean-shrinkwrap.js
jhermsmeier: Done

scripts/clean-shrinkwrap.js
jhermsmeier: Done

lurch:
LGTM, but I think @jviotti ought to review as he's more familiar with the intricacies of NPM than I am.

Although I still wonder if it'd be worth running this script automatically on the CI servers and erorr-ing if it results in a change to `npm-shrinkwrap.json`, just to try and stop anything slipping through the net? (see also [here](https://github.com/resin-io/etcher/pull/1234#issuecomment-290139862) - I dunno if these two steps do the same thing or if they're in fact different?)

scripts/clean-shrinkwrap.js
```diff
@@ -17,13 +17,28 @@
...
-const shrinkwrapContents = jsonfile.readFileSync(SHRINKWRAP_PATH);
-shrinkwrapContents.dependencies = _.omit(shrinkwrapContents.dependencies, shrinkwrapIgnore);
-jsonfile.writeFileSync(SHRINKWRAP_PATH, shrinkwrapContents, {
-  spaces: 2
-});
+console.log('Removing:', shrinkwrapIgnore.join(', '));
+
+/**
+ * Run an npm command
+ * @param {Array} command - list of arguments
```
jviotti: We can probably remove doc annotations on these kind of scripts.

jhermsmeier: Well, it doesn't hurt, and the linters will yell about it otherwise (guess we could turn that rule off for the scripts folder?)

jviotti: Yeah, not really a priority :+1:

PR #1237, opened 2017-03-29
chore: Add a require() checking script

lurch:
This adds another sanity-check bash script to scan the Etcher source files
and check all required node modules are available.

More bash magic ;-)


jviotti:
Very cool stuff! Do you think we can also do the opposite? e.g: complaining if there are unused dependencies?

lurch:
> Do you think we can also do the opposite? e.g: complaining if there are unused dependencies?

Yup, can do, although that should be a separate script rather than adding extra functionality to this script.
Expect another PR later today... ;)

lurch:
Obviously this is failing because of the typo that #1229 fixes.
@jviotti Do you want me to fix that typo in this PR, and then you can rebase #1229 after merging this PR, or would you rather merge #1229 first and then have me rebase this PR afterwards? ;-)

jviotti:
Go ahead and create a PR. My PR will take a while before its ready to merge, given we've found out it breaks the drag n drop functionality.

lurch:
See #1238

lurch:
BTW...

> Do you think we can also do the opposite? e.g: complaining if there are unused dependencies?

Made a script to do that last night. It's still a bit rough around the edges (got some tidying up to do before PRing it, as there's a bunch of modules I need to whitelist) but it identified that the dependencies `file-tail` and `tail` are unused (as you've already fixed in #1242), and that the devDependency `electron-packager` is unused.

jviotti:
> and that the devDependency electron-packager is unused.

This dependency is used by the build scripts, so that's probably the reason your script is not catching it.

lurch:
Can you show me where? https://github.com/resin-io/etcher/search?utf8=✓&q=electron-packager

See also #897

jviotti:
Oh, I got confused with `electron-builder`. Looks like `electron-packager` is indeed unused :+1:

PR #1238, opened 2017-03-29
style: fix require() typo in lib/cli/writer.js

lurch:


jviotti:
Approved. Can we edit the title to match the commit guidelines, like `style: fix require() typo` or something like that?

PR #1241, opened 2017-03-29
chore: make the CI servers fail-fast if one of their builds fails

jviotti:
This means that if any architecture build fails, then we stop the other
ones right away, for performance reasons.

See: atom/atom@ce51074
See: https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ env:
     - TARGET_ARCH=x86
 
 matrix:
+  fast_finish: true
   exclude:
     - os: osx
       env: TARGET_ARCH=x86
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,6 +19,9 @@ environment:
     - TARGET_ARCH: x64
     - TARGET_ARCH: x86
 
+matrix:
+  fast_finish: true
+
 install:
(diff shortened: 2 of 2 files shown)


lurch:
Nice. Does Travis offer anything similar?  It's a shame that https://www.appveyor.com/docs/appveyor-yml/ doesn't allow hyperlinking to a particular section, otherwise I'd have suggested including it in the commit message.

I guess the only downside to this PR that if the first build fails due to a network-timeout, you don't get a chance to see if the second build would have completed successfully. But given that on Appveyor you have to rebuild both arches anyway, rather than only rebuilding the one that had the network failure, I guess that's less of a problem.

jviotti:
Yeah, I found the Travis CI equivalent, and updated this PR with it!

lurch:
Nice! IMHO the commit title is now a bit confusing though. Perhaps "make the CI servers fail-fast if one of their builds fails" ? Or you could bracket it as "stop builds in the matrix if one fails (for both Travis and Appveyor)".
And you ought to use "other ones" instead of "other one" as Travis has 3 arches ;)

jviotti:
Sounds good. Fixed!

PR #1243, opened 2017-03-29
refactor(image-stream): get rid of read-chunk

jviotti:
We also took the opportunity to make `utils.getArchiveMimeType()`
asynchronous.

### DIFF
--- a/lib/image-stream/utils.js
+++ b/lib/image-stream/utils.js
@@ -17,7 +17,8 @@
 'use strict';
 
 const _ = require('lodash');
-const readChunk = require('read-chunk');
+const Bluebird = require('bluebird');
+const fs = Bluebird.promisifyAll(require('fs'));
 const archiveType = require('archive-type');
 
 /**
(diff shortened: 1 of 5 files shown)


lib/image-stream/utils.js
```diff
@@ -36,8 +37,22 @@ exports.getArchiveMimeType = (file) => {
...
+  }), (fileDescriptor) => {
+    const BUFFER_START = 0;
+    const chunk = new Buffer(MAGIC_NUMBER_BUFFER_LENGTH);
+
+    return fs.readAsync(
+      fileDescriptor,
+      chunk,
+      BUFFER_START,
+      MAGIC_NUMBER_BUFFER_LENGTH,
+      MAGIC_NUMBER_BUFFER_START
```
lurch: Looking at https://nodejs.org/api/fs.html#fs_fs_read_fd_buffer_offset_length_position_callback I guess we could use `null` here, and get rid of `MAGIC_NUMBER_BUFFER_START`.

Also, IMHO `MAGIC_NUMBER` usually indicates a complete number that's being read and matched against exactly. Which isn't actually what we're doing here, it's more of a heuristic-match than an exact-match. So maybe something like `ARCHIVE_TYPE_IDENTIFICATION_BYTES_LENGTH` would be better than `MAGIC_NUMBER_BUFFER_LENGTH` ?

tests/image-stream/utils.spec.js
```diff
@@ -25,29 +25,44 @@ describe('ImageStream: Utils', function() {
 
   describe('.getArchiveMimeType()', function() {
 
-    it('should return application/x-bzip2 for a bz2 archive', function() {
+    it('should resolve application/x-bzip2 for a bz2 archive', function(done) {
       const file = path.join(DATA_PATH, 'bz2', 'raspberrypi.img.bz2');
-      m.chai.expect(utils.getArchiveMimeType(file)).to.equal('application/x-bzip2');
+      utils.getArchiveMimeType(file).then((type) => {
```
lurch: I can't quite tell, but IMHO it looks like this change is 'extra' to just 'getting rid of read-chunk'. (i.e. you could have got rid of `read-chunk` without also switching this from `return` to `resolve`?)
If that's the case, could you perhaps mention it in the commit-message too?

Does the `getArchiveMimeType` docstring need updating too?

npm-shrinkwrap.json
```diff
@@ -5570,11 +5570,6 @@
       "resolved": "https://registry.npmjs.org/read/-/read-1.0.7.tgz",
       "dev": true
     },
-    "read-chunk": {
-      "version": "2.0.0",
-      "from": "read-chunk@>=2.0.0 <3.0.0",
-      "resolved": "https://registry.npmjs.org/read-chunk/-/read-chunk-2.0.0.tgz"
-    },
```
lurch: Given that this only results in a single package being removed from `npm-shrinkwrap.json`, does `read-chunk` "need" to be removed?

lib/image-stream/utils.js
jviotti: Cool stuff. I addressed both comments!

tests/image-stream/utils.spec.js
jviotti: I took the chance to make this function async. I'll update the commit.

> Does the getArchiveMimeType docstring need updating too?

Oops, good catch :)

npm-shrinkwrap.json
jviotti: There is also `pify`, but there are other modules still depending on it, which is the reason why it still doesn't go away. Luckily more stuff will start getting removed once we get rid of more packages.

PR #1244, opened 2017-03-29
refactor(image-stream): get rid of `isstream`

jviotti:
The dependency is not yet removed from `npm-shrinkwrap.json` given that
there another one that keeps requiring it.


lurch:
As with #1243 , I wonder if this "needs" to be removed, given the minimal impact it has on dependencies?

jviotti:
Yeah, the module has no dependencies at all, but its also so silly that makes little sense to have it (where we can just do an `instanceof`).

PR #1246, opened 2017-03-29
refactor(image-stream): get rid of rindle

jviotti:

### DIFF
--- a/lib/image-stream/archive.js
+++ b/lib/image-stream/archive.js
@@ -18,10 +18,10 @@
 
 const path = require('path');
 const Bluebird = require('bluebird');
-const rindle = require('rindle');
 const _ = require('lodash');
 const PassThroughStream = require('stream').PassThrough;
 const supportedFileTypes = require('./supported');
+const utils = require('./utils');
 const errors = require('../shared/errors');
 
 /**
@@ -82,7 +82,7 @@ const extractEntryByPath = (archive, filePath, options) => {
   }
 
   return options.hooks.extractFile(archive, options.entries, fileEntry.name)
-    .then(rindle.extract);
+    .then(utils.extractStream);
 };
 
 /**
@@ -137,9 +137,9 @@ const extractArchiveMetadata = (archive, basePath, options) => {
       checksum: results.manifest.checksum,
       bytesToZeroOutFromTheBeginning: results.manifest.bytesToZeroOutFromTheBeginning,
       recommendedDriveSize: results.manifest.recommendedDriveSize,
-      logo: results.logo,
-      bmap: results.bmap,
-      instructions: results.instructions
+      logo: _.invoke(results.logo, [ 'toString' ]),
+      bmap: _.invoke(results.bmap, [ 'toString' ]),
+      instructions: _.invoke(results.instructions, [ 'toString' ])
     };
   });
 };
--- a/package.json
+++ b/package.json
@@ -94,7 +94,6 @@
     "redux-localstorage": "^0.4.1",
     "resin-cli-form": "^1.4.1",
     "resin-cli-visuals": "^1.2.8",
-    "rindle": "^1.3.0",
     "rx": "^4.1.0",
     "semver": "^5.1.0",
     "sudo-prompt": "^6.1.0",
--- a/lib/image-stream/utils.js
+++ b/lib/image-stream/utils.js
@@ -58,3 +58,38 @@ exports.getArchiveMimeType = (file) => {
     });
   });
 };
+
+/**
+ * @summary Extract the data of a readable stream
+ * @function
+ * @public
+ *
+ * @description
+ * You should be careful when using this function, since you can only
+ * extract files that are not bigger than the available computer memory.
+ *
+ * @param {StreamReadable} stream - stream
+ * @fulfil {Buffer} - data
+ * @returns {Promise}
+ *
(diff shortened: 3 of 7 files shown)


lib/image-stream/archive.js
```diff
@@ -81,8 +80,20 @@ const extractEntryByPath = (archive, filePath, options) => {
     return Bluebird.resolve(options.default);
   }
 
-  return options.hooks.extractFile(archive, options.entries, fileEntry.name)
-    .then(rindle.extract);
```
lurch: Is the functionality that you're replacing here (and in the unit tests), generic enough to be moved to e.g. the `utils` module?

package.json
```diff
@@ -98,7 +98,6 @@
     "redux-localstorage": "^0.4.1",
     "resin-cli-form": "^1.4.1",
     "resin-cli-visuals": "^1.2.8",
-    "rindle": "^1.3.0",
```
lurch: Haha, that @jviotti writes such [wasteful code](https://www.npmjs.com/package/rindle) :rofl:

lib/image-stream/archive.js
```diff
@@ -81,8 +80,20 @@ const extractEntryByPath = (archive, filePath, options) => {
...
+    return new Bluebird((resolve, reject) => {
+      let data = '';
+
+      stream.on('data', (chunk) => {
+        data += chunk;
+      });
+
+      stream.on('error', reject);
+      stream.on('end', () => {
+        return resolve(data);
```
jhermsmeier: No need to `return` here, I believe

lib/image-stream/archive.js
```diff
@@ -81,8 +80,20 @@ const extractEntryByPath = (archive, filePath, options) => {
     return Bluebird.resolve(options.default);
   }
 
-  return options.hooks.extractFile(archive, options.entries, fileEntry.name)
-    .then(rindle.extract);
+  return options.hooks.extractFile(archive, options.entries, fileEntry.name).then((stream) => {
+    return new Bluebird((resolve, reject) => {
+      let data = '';
```
jhermsmeier: Considering that the file to be extracted might not always be a text file, what do you think about returning a buffer instead?

```js
let chunks = [];

stream.on('data', (chunk) => {
  chunks.push(chunk)
});

stream.on('end', () => {
  resolve(Buffer.concat(chunks))
});
```

Or, in case we do know the uncompressed file size beforehand, use a preallocated buffer and copy into it:

```js
let data = Buffer.alloc(uncompressedSize, 0);
let offset = 0;

stream.on('data', (chunk) => {
  chunk.copy(data, offset);
  offset += chunk.length;
});

stream.on('end', () => {
  resolve(data);
});
```

lurch: Would the pre-allocated buffer work if the uncompressed file size is larger than the free memory?

In fact, does the current code work if the file being extracted is larger than the computer's free memory?

jhermsmeier: Afaik, this is for extracting metadata files from archives, not the images – which means that memory exhaustion is not a concern (and before it would be, you'll get a `RangeError: Array buffer allocation failed`). The images themselves are streamed.

lurch: I wonder if that should be mentioned in a docstring? ;-)

package.json
jviotti: lol, I was also ranting about a dependency called `angular-seconds-to-date`, when I discovered it was something I published long ago :P (https://www.npmjs.com/package/angular-seconds-to-date)

lib/image-stream/archive.js
jviotti: True. Fixed!

lib/image-stream/archive.js
jviotti: Makes a lot of sense. Done!

lib/image-stream/archive.js
jviotti: Done! Not sure what I should mention in the docstring. Do you mind elaborating?

lurch: Something along the lines of only being able to use this method for "small" files, as it won't work for trying to extract files bigger than the computer's free memory.

jviotti: Done!

lurch:
Dunno what's happened with the "missing" Appveyor status here? :-/

jviotti:
Weird. I triggered a new build. Lets see if that fixes the issue.

lib/image-stream/archive.js
```diff
@@ -137,9 +137,9 @@ const extractArchiveMetadata = (archive, basePath, options) => {
       checksum: results.manifest.checksum,
       bytesToZeroOutFromTheBeginning: results.manifest.bytesToZeroOutFromTheBeginning,
       recommendedDriveSize: results.manifest.recommendedDriveSize,
-      logo: results.logo,
-      bmap: results.bmap,
-      instructions: results.instructions
+      logo: _.invoke(results.logo, [ 'toString' ]),
```
lurch: Huh? Why is the `logo` now being converted to a string?

jviotti: The `logo` is on SVG format (we enforce that), so it needs to be a string :)

lurch: Ahhh :)

PR #1247, opened 2017-03-29
refactor(image-stream): get rid of `gzip-uncompressed-size`

jviotti:

### DIFF
--- a/lib/image-stream/gzip.js
+++ b/lib/image-stream/gzip.js
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2016 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


lurch:
:-1: 
IMHO for something as low-level / technically-tricky as this, it makes _far_ more sense to use a 3rd-party module, rather than rolling our own code. Unless there's very compelling reasons otherwise...

jhermsmeier:
For reading a uint32 from the end of a file? I not sure I see how that's technically tricky ;)
Also, the 3rd party module we've been using was written by @jviotti and this is just an adjusted copy of https://github.com/jviotti/node-gzip-uncompressed-size/pull/1, as discussed there

lurch:
> Also, the 3rd party module we've been using was written by @jviotti

Ahh, I'd totally missed that ;-)
I guess in that case we have total control over its dependencies, so why does it need "de-modularising" ?

jviotti:
I rather have it inside Etcher, since its something we will want to keep tweaking and improving specifically for Etcher. Once we start publishing the SDK, this functionality will be re-usable through the image stream module, so we won't affect distributability at all.

Issue #1189, opened 2017-03-17
Check if an image has a MBR or GPT, and complain if not


jviotti:
Images without a MBR nor GPR will probably not boot. This simple check covers most edge cases:

- Hybrid ISOs still have a MBR
- Windows images don't have a MBR

jhermsmeier:
We should probably still let the user decide if they want to proceed with flashing, to avoid false positives from getting in the way.

lurch:
Is this a duplicate of #553 ? ;-)

jviotti:
Good catch. I'll close the other one in flavour of this, which sets a clearer way forward.

jhermsmeier:
Closing, as has been done in https://github.com/resin-io/etcher/pull/1248

Issue #1116, opened 2017-02-19
Identity partition table type in analytics & error reporting


jviotti:
I'd be great if we can know in advance if an image has a MBR or a GPT.

lurch:
If we're gathering this info for analytics purposes, it might also be useful to collect which filesystems are being used in the images being written? (FAT16/FAT32/exFAT/ext3/ext4/btrfs/etc.)

I'll assign this to @jhermsmeier because he already has modules to do some of this, but I dunno what priority this task has.

jhermsmeier:
Closed in https://github.com/resin-io/etcher/pull/1248

lurch:
Apologies if I'm being dense, but what part of #1248 closes this? All I can see is
`analytics.logEvent('Missing partition table', image);`

jhermsmeier:
The image metadata gets logged, which is what contains the MBR/GPT info

lurch:
Ah, thanks. Does this also include the partition / filesystem types, as I mentioned in my comment above?

jhermsmeier:
Yup it does

Issue #1115, opened 2017-02-19
Attempt to detect incomplete images


jviotti:
Some images might not be complete (e.g: the download stops halfway through). We could implement the following heuristics to determine if the image is complete or not:

- Inspect the partition table and ensure every position referenced there exists
- If the image contains a GPT, the GPT is also duplicated at the end of the file. Check that it exists at the end as well

 If we detect the image is incomplete, we should present an error/warning to the user.

jviotti:
Maybe @jhermsmeier has other ideas as well?

lurch:
As mentioned elsewhere, until I see evidence to the contrary, I still suspect all images will be an exact multiple of 512 bytes... ;-)

jviotti:
Fair enough, but if the image download stopped in the middle, then the changes are suddenly way higher.

lurch:
OTOH, I suspect _most_ people will download images as a compressed file (`.zip`, `.xz`, etc.) in which case I believe the archive won't even open if "the download stopped in the middle" ?

jviotti:
Maybe. The module we found to unzip while streaming ignores the zip integrity warnings altogether in order to be able to work. Who knows what we will encounter :)

lurch:
I guess streaming images will mean that we won't be able to apply these heuristics until after we've already streamed the whole file.

jviotti:
Closing stale issues

PR #1248, opened 2017-03-30
feat(image-stream): Read MBR & GPT in .getImageMetadata()

jhermsmeier:
**TODO:**
- [x] Correct `GPT_PROTECTIVE`/ `GPT_HYBRID` constant values (not sure I got them right)
- [x] Update the tests
- [x] Warn on missing MBR (see #1189)
- [x] Expose MBR/GPT data in a way that isn't going to be mangled by the logging
- [ ] Send off basic info about image format & partitioning scheme (see #1116)
- [x] ~~Potentially warn if partition sizes are OOB of file size (see #1115)~~~

**EDIT:** Warning if partitions don't span enough of the image turns out to be immensely inaccurate for several reasons, from unknown block size to images padded with gigabytes of zeros.

Note that having access to the partition tables will also aid us in blockmapping an image (without a `.bmap`) on the fly.

**Warning when attempting to flash an image without a partition table:**

screen shot 2017-03-31 at 18 49 21

Depends on: #1283 
Connects to: #1189, #1116, #1115 
Change-Type: minor


lib/image-stream/index.js
```diff
@@ -113,14 +113,12 @@ exports.getFromFilePath = (file) => {
...
 exports.getImageMetadata = (file) => {
-  return exports.getFromFilePath(file).then((image) => {
-
-    // Since we're not consuming this stream,
-    // destroy() it, to avoid dangling open file descriptors etc.
-    image.stream.destroy();
-
-    return _.omitBy(image, isStream);
-  });
+  const parsePartitions = require('./parse-partitions');
```
jviotti: Is there any reason why this require is dynamic?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Get the GPT/EFI partition from a MBR
+ * @param {MBR} mbr - Master Boot Record
+ * @returns {MBR.Partition|undefined} partition
+ */
+const getEfiPart = (mbr) => {
+  const GPT_HYBRID = 0xEE;
+  const GPT_PROTECTIVE = 0xEF;
+  return _.filter(mbr.partitions, (partition) => {
```
jviotti: This sounds like it should be a `_.find`, which allows you to get rid of the `[ZERO]`

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,92 @@
...
+'use strict';
+
+const _ = require('lodash');
+const Bluebird = require('bluebird');
+const MBR = require('mbr');
+const GPT = require('gpt');
+
+const MAX_BYTES = 65536;
+const BLOCK_SIZE = 512;
+const ZERO = 0;
```
jviotti: lol. The intention of disallowing magic numbers is to be able to pick a constant name that better explains what the value means in a particular context. In this case, the number 0 means two different things.

- Its used on `getEfiPart` to get the first value of the filtered array. You can use `_.first` to better showcase the meaning, or you can replace `_.filter` + `_.first` with `_.find`, which is the best way to go IMO

- Its used on the exported function as the initial value of a variable that is increased through the function. In this case, a good variable name would be `LENGTH_INITIAL_VALUE` or something like that

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Get the GPT/EFI partition from a MBR
+ * @param {MBR} mbr - Master Boot Record
+ * @returns {MBR.Partition|undefined} partition
+ */
+const getEfiPart = (mbr) => {
+  const GPT_HYBRID = 0xEE;
```
jviotti: I think we could clarify that these are partition types. Maybe `PARTITION_TYPE_GPT_HYBRID` would do it?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        console.log(image);
+        resolve(image);
+        return;
+      }
+
+      // Read chunks
+      let chunk = this.read();
+      while (chunk) {
```
jviotti: Could we implement this using a reducer?

lib/image-stream/parse-partitions.js
jhermsmeier: Updated to `mbr@1.1.0`, which now has the `.getEFIPart()` method, and renamed that ZERO :)

lib/image-stream/index.js
jhermsmeier: Nope, slip up – fixed!

lib/image-stream/parse-partitions.js
jhermsmeier: What would the benefits be? This seems like the fastest & simplest way to me

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+    stream.on('error', reject);
+
+    // We need to use the "old" flowing mode here,
+    // as some dependencies don't implement the "readable"
+    // mode properly (i.e. bzip2)
+    stream.on('data', (chunk) => {
+      chunks.push(chunk);
+      length += chunk.length;
+
+      // Once we've read enough bytes, terminate the stream
```
lurch: Possibly a micro-optimisation that wouldn't be worth it, but would it be worth reading the first 512 bytes, seeing if we have a valid MBR, and then seeing if the MBR indicates we also have a GPT, and only _then_ reading up to the `MAX_BYTES` boundary?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+ * @param {Buffer} buffer - Buffer
+ * @returns {undefined}
+ */
+const parsePartitions = (image, buffer) => {
+
+  let mbr = null;
+  let gpt = null;
+
+  try {
+    mbr = MBR.parse(buffer);
```
lurch: I realise it goes somewhat against @jviotti 's earlier suggestion, but IMHO this should probably be `mbr = MBR.parse(buffer.slice(0, 512));` (with the numbers replaced with suitable constants). Perhaps you could even import `MBR_OFFSET` and `MBR_LENGTH` constants from your `mbr` module?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
+/*
+ * Copyright 2016 resin.io
```
lurch: 2017 ;-)

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+ */
+
+'use strict';
+
+const _ = require('lodash');
+const Bluebird = require('bluebird');
+const MBR = require('mbr');
+const GPT = require('gpt');
+
+const MAX_BYTES = 8192;
```
lurch: Just out of curiosity, why this has reduced from 64K in an earlier version of this PR, to just 8K?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+  while (blockSize <= MAX_BLOCK_SIZE) {
+    try {
+      gpt = GPT.parse(buffer.slice(blockSize));
+    } catch (error) {
+      // Ignore error
+    }
+    if (gpt) {
+      return gpt;
+    }
+    blockSize += blockSize;
```
lurch: I guess this is effectively the same as `blockSize *= 2` ? That might be more readable?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Attempt to parse the GPT from various block sizes
+ * @private
+ * @param {Buffer} buffer - Buffer
+ * @returns {GPT|null}
+ */
+const parseGPT = (buffer) => {
+
+  let blockSize = INITIAL_BLOCK_SIZE;
```
lurch: Is this definitely `blockSize`, or is it `gptStartOffset`?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+  let gpt = null;
+
+  // Attempt to parse the GPT from several offsets,
+  // as the block size of the image may vary (512,1024,2048,4096);
+  // For example, ISOs will usually have a block size of 4096,
+  // but raw images a block size of 512 bytes
+  while (blockSize <= MAX_BLOCK_SIZE) {
+    try {
+      gpt = GPT.parse(buffer.slice(blockSize));
```
lurch: Is the GPT always a fixed size? If so, perhaps you could fill in the second `slice` parameter?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+    } catch (error) {
+      // Ignore error
+    }
+    if (gpt) {
+      return gpt;
+    }
```
lurch: I guess this could theoretically go inside the `try` block? (which would allow you to localise the `gpt` variable)

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+    blockSize += blockSize;
+  }
+
+  return gpt;
```
lurch: Will this always effectively be doing `return null;`? Would it be better to be explicit?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+
+  let mbr = null;
+  let gpt = null;
+
+  try {
+    mbr = MBR.parse(buffer);
+    image.hasMBR = true;
+    if (mbr.getEFIPart()) {
+      image.hasGPT = true;
+      gpt = parseGPT(buffer);
```
lurch: minor nit: you could possibly swap these previous two lines, to match the `hasMBR` part ;)

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+    });
+  } else if (image.hasMBR) {
+    image.partitions = _.map(mbr.partitions, (partition) => {
+      return {
+        type: partition.type,
+        id: null,
+        name: null,
+        start: partition.firstLBA,
+        end: partition.lastLBA,
+        extended: partition.extended
```
lurch: If you're tracking `extended`, would it make more sense to track `primary` and `logical`?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+  // we normalize them here to make them easier to deal with and
+  // avoid clutter in what's sent to analytics
+  if (image.hasGPT) {
+    image.partitions = _.map(gpt.partitions, (partition) => {
+      return {
+        type: partition.type.toString(),
+        id: partition.guid.toString(),
+        name: partition.name,
+        start: partition.firstLBA,
+        end: partition.lastLBA,
```
lurch: IMHO you should either rename these properties `startLBA` and `endLBA`, or you should leave them as `start` and `end` and convert the values they store to actual byte offsets.

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+    // mode properly (i.e. bzip2)
+    stream.on('data', (chunk) => {
+      chunks.push(chunk);
+      length += chunk.length;
+
+      // Once we've read enough bytes, terminate the stream
+      if (length >= MAX_BYTES && !destroyed) {
+
+        // Prefer close() over destroy(), as some streams
+        // from dependencies exhibit quirky behavior when destroyed
```
lurch: Hmmm, was this also a problem when the stream was destroyed in `getImageMetadata`?
(just out of curiosity: quirky how? ;)  )

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+};
+
+/**
+ * Attempt to read the MBR and GPT from an imagestream
+ * NOTE: This operation will consume the first `MAX_BYTES`
+ * of the stream and then destroy the stream.
+ * @param {Object} image - image metadata
+ * @returns {Promise}
+ */
+module.exports = (image) => {
```
lurch: Is this common NodeJS syntax? Feels a bit odd now that `parse-partitions.js` also has a private method named `parsePartitions` ?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+        // from dependencies exhibit quirky behavior when destroyed
+        if (image.stream.close) {
+          image.stream.close();
+        } else {
+          image.stream.destroy();
+        }
+
+        // Remove references to stream to allow them being GCed
+        image.stream = null;
+        image.transform = null;
```
lurch: Should the properties be removed from the object too?

package.json
```diff
@@ -78,9 +78,11 @@
     "etcher-latest-version": "^1.0.0",
     "file-type": "^4.1.0",
     "flexboxgrid": "^6.3.0",
+    "gpt": "^1.0.0",
     "immutable": "^3.8.1",
     "lodash": "^4.5.1",
     "lzma-native": "^1.5.2",
+    "mbr": "^1.1.1",
```
lurch: Minor point, but would it be worth creating a higher-level module which requires `mbr` and `gpt` (and then does much of the work that you're doing here), and then have Etcher depend on that module?

tests/gui/models/supported-formats.spec.js
```diff
@@ -62,62 +62,80 @@ describe('Browser: SupportedFormats', function() {
     describe('.isSupportedImage()', function() {
 
       it('should return false if the file has no extension', function() {
-        const isSupported = SupportedFormatsModel.isSupportedImage('/path/to/foo');
+        const isSupported = SupportedFormatsModel.isSupportedImage({
+          path: '/path/to/foo'
+        });
```
lurch: Rather than creating lots of objects with just a `path` property, perhaps it'd be better to have a 'proper' image object, which all the unit-tests would use, just changing the `path` property each time?

tests/image-stream/img.spec.js
```diff
@@ -53,7 +53,43 @@ describe('ImageStream: IMG', function() {
               estimation: false,
               value: expectedSize
             }
-          }
+          },
+          hasMBR: true,
+          hasGPT: false,
```
lurch: We'd obviously want one or more unit-tests with `hasGPT: true` as well ;-)

lurch:
I guess it'd mess up the streaming aspect, but for those images where we're not streaming, and we have a GPT, would it be "within Etcher's remit" to warn if the backup GPT doesn't match the primary GPT?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,110 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const _ = require('lodash');
+const Bluebird = require('bluebird');
+const MBR = require('mbr');
+const GPT = require('gpt');
```
jhermsmeier: MBR & GPT are constructors, thus pascal cased, and also acronyms, hence all caps – could change it to `Mbr` and `Gpt` respectively (which would be mildly irritating to my eyes, but yeah)

lib/image-stream/parse-partitions.js
jhermsmeier: Because the first time around I wasn't sure how many bytes it'd need from ISOs, the change to 8K is based on the max block size we're checking for, which is 4096. But now that I'm thinking about it, this would need changing to `MAX_BLOCK_SIZE * 34` because the GPT partition entries go from LBA 2 -> LBA34 (the first 4 at LBA2, 5-128 in LBA 3-34)

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,110 @@
...
+
+const MAX_BYTES = 65536;
+const BLOCK_SIZE = 512;
+const INITIAL_LENGTH = 0;
+
+/**
+ * Attempt to read the MBR and GPT from an imagestream
+ * NOTE: This operation will consume the first `MAX_BYTES`
+ * of the stream and then destroy the stream.
```
jhermsmeier: Nope, because the stream would be partially consumed before the method returns (or resolves), so that stream would be unusable even if left alive. It's safer to just open a new stream for a full read in most cases anyways.

lib/image-stream/parse-partitions.js
jhermsmeier: Definitely blockSize, yes – because the GPT should always be at LBA 1 (the MBR at LBA 2) – and doesn't have a GPT specific offset, it's always connected to the block size

lib/image-stream/index.js
```diff
@@ -115,16 +116,13 @@ exports.getFromFilePath = (file) => {
  * });
  */
 exports.getImageMetadata = (file) => {
-  return exports.getFromFilePath(file).then((image) => {
-
-    // Since we're not consuming this stream,
-    // destroy() it, to avoid dangling open file descriptors etc.
-    image.stream.destroy();
```
jhermsmeier: I see you've discovered the moved destroy() below :)

lib/image-stream/parse-partitions.js
jhermsmeier: Theoretically yes, but since this is a loop (variable declaration within loops is strongly discouraged) and it's generally best to keep as much as possible out of try catch blocks, it's not in there

lib/image-stream/parse-partitions.js
jhermsmeier: Then we'd have to add another constant (because of no-magic-numbers), which doesn't make any sense to me at all. And I think adding the block size to itself is pretty clear in terms of what it does.

lib/image-stream/parse-partitions.js
jhermsmeier: Slicing buffers is a rather expensive op, and unnecessary, as it will only "parse" the first 512 bytes anyways

lib/image-stream/parse-partitions.js
jhermsmeier: Possibly, yes – but the reasoning here is that if the GPT parsing fails (i.e. because it's corrupted, malformed, a bug in the GPT module), we'll at least still have an indicator that there's supposed to be one there.

lib/image-stream/parse-partitions.js
jhermsmeier: We can't convert them to byte offsets, as we don't know the block size – happy to change them to `firstLBA` and `lastLBA`

lib/image-stream/parse-partitions.js
jhermsmeier: Nope, because we aren't reading the full image we can't read logical partitions (and thus logical would always be `false`), and primary is ambiguous as the primary partition could also denote the partition with its status marked "active"

lib/image-stream/parse-partitions.js
jhermsmeier: Considering that fs read streams will usually read in at least 4K at a time (and the default high water mark in `fs.ReadStream` is 64K), we'd usually already have more than `MAX_BYTES` on the first or second `data` event (because listening on `data` puts it into flowing mode) – so I don't think this would yield anything apart from more complexity

lib/image-stream/parse-partitions.js
jhermsmeier: They are in `image-stream/index.js`, note the `_.omitBy()`

package.json
jhermsmeier: I guess that'd be https://github.com/jhermsmeier/node-disk, but since we're doing very specific stream slicing here, I don't think that'd be a good idea (just another minor abstraction to maintain), especially when considering that most of this will be done quite differently soon-ish.

lib/gui/models/supported-formats.js
```diff
@@ -128,25 +128,43 @@ SupportedFormats.service('SupportedFormatsModel', function() {
...
+   * @public
+   *
+   * @param {Object} image - image metadata
+   * @returns {Boolean} whether the image is missing a partition table
+   * @example
+   * if (SupportedFormatsModel.missingPartitionTable(image)) {
+   *   console.log('Looks like it might not boot');
+   * }
+   */
+  this.missingPartitionTable = (image) => {
```
jviotti: Yeah, I agree. This function name style reminds me to Objective-C :P

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -78,19 +78,26 @@ module.exports = function(
     }
 
     Bluebird.try(() => {
-      if (!SupportedFormatsModel.looksLikeWindowsImage(image.path)) {
-        return false;
+      let message = null;
```
jviotti: Can we refactor this to the following expression, so we get rid of `let` (its always nice to be able to make things inmmutable)?

```js
const message = _.attempt(() => {
    if (SupportedFormatsModel.looksLikeWindowsImage(image)) {
        return messages.warning.looksLikeWindowsImage();
    } else if (SupportedFormatsModel.missingPartitionTable(image)) {
        return messages.warning.missingPartitionTable();
    }
});
```

lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -78,19 +78,26 @@ module.exports = function(
...
+        // TODO: `Continue` should be on a red background (dangerous action) instead of `Change`.
+        // We want `X` to act as `Continue`, that's why `Continue` is the `rejectionLabel`
+        return WarningModalService.display({
+          confirmationLabel: 'Change',
+          rejectionLabel: 'Continue',
+          description: message
+        });
       }
 
-      AnalyticsService.logEvent('Possibly Windows image', image);
```
jviotti: Yeah, we shouldn't delete this, plus we should add another event to track images without partitions.

lib/image-stream/handlers.js
```diff
@@ -58,7 +58,7 @@ module.exports = {
           value: options.size
         }
       },
-      transform: Bluebird.resolve(unbzip2Stream())
+      transform: unbzip2Stream()
```
jviotti: Oh, for some reason I had the impression that we needed to resolve a value here.

lib/image-stream/parse-partitions.js
jviotti: Yeah, if they are constructors (and they are acronyms), then all uppercase its a standard convention :+1:

lib/image-stream/parse-partitions.js
jviotti: Would it be worth adding JSDoc annotations to these constants explaining what they are for?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,166 @@
...
+const MBR = require('mbr');
+const GPT = require('gpt');
+
+const MAX_BYTES = 8192;
+const INITIAL_LENGTH = 0;
+const INITIAL_BLOCK_SIZE = 512;
+const MAX_BLOCK_SIZE = 4096;
+
+/**
+ * Attempt to parse the GPT from various block sizes
```
jviotti: The title should be preceded by `@summary`.

tests/image-stream/dmg.spec.js
```diff
@@ -94,7 +96,9 @@ describe('ImageStream: DMG', function() {
                 estimation: false,
                 value: uncompressedSize
               }
-            }
+            },
+            hasMBR: false,
+            hasGPT: false
```
jviotti: Should we return an empty `partitions` array in these cases?

lib/image-stream/parse-partitions.js
lurch: Does that indicate that there ought to be separate `.hasGPT` and `.hasValidGPT` flags?

lib/image-stream/parse-partitions.js
lurch: Ahh, I see. Does your MBR module check that there's at most one extended partition?

lib/image-stream/parse-partitions.js
lurch: I wondered if something like that might be the case ;-)  Thanks for clarifying :+1:

lib/image-stream/parse-partitions.js
lurch: Ahhh, I didn't realise there was slightly different conventions between importing a module, and importing a constructor directly.

lib/image-stream/parse-partitions.js
lurch: Interesting - does that mean GPT always stores 4 partitions per block, regardless of whether the blocksize is 512 or 4096?

lib/image-stream/parse-partitions.js
lurch: Fair enough. Possibly a stupid question, but is there any kind of buffer-backed stream-like object that _can_ be rewound back to zero, assuming that the current filepos < sizeof(buffer) ?

lib/image-stream/parse-partitions.js
lurch: Did you mean MBR at LBA 1 and GPT at LBA 2? (or is it GPT at LBA 1 and MBR at LBA 0 - can't remember if the LBAs are 0-indexed or 1-indexed.)
IIRC the MBR always starts at offset 0, and has size 512 bytes, regardless of the blocksize?

lib/image-stream/parse-partitions.js
lurch: Ah, more NodeJS conventions I was unaware of - thanks :)

lib/image-stream/parse-partitions.js
lurch: > Slicing buffers is a rather expensive op

Ahh, didn't know that ;-)

lib/image-stream/parse-partitions.js
lurch: > as we don't know the block size

Oh? Based on an earlier comment, does that mean you can't read more than the first 4 GPT partitions? How is one "supposed" to determine the blocksize, theoretically?

lib/image-stream/parse-partitions.js
lurch: Yeah, I just thought it might be more explicit to keep everything localised - if they're being set to null anyway, might make sense to delete them here and then remove the `.omitBy` from the other place?

lib/gui/models/supported-formats.js
```diff
@@ -128,25 +128,43 @@ SupportedFormats.service('SupportedFormatsModel', function() {
...
+    return !image.hasMBR;
```
lurch: For clarity, I wonder if it'd be better to do something 'more explicit' like https://github.com/resin-io/etcher/pull/1183/files#diff-dec8509d5c0e865bccdce35d2dc3c9ccR47 ?
````js
_.get(options, [ 'ignoreSleepUpdateCheck' ], false)
```

lurch:
Conflicts and failing CI tests, oh my ;-)

lib/image-stream/parse-partitions.js
jhermsmeier: Oh, yes (what have I written there?!) – MBR is at LBA 0, GPT at LBA 1. Not sure what I meant there, the GPT's partition table starts at LBA 2.

lib/image-stream/parse-partitions.js
jhermsmeier: Yeah, sure you can do seekable streams, but that doesn't make things simpler – and becomes very tricky when dealing with streaming from the network.

lib/image-stream/parse-partitions.js
jhermsmeier: No, that's not checked. Why would only one extended partition be allowed?

lurch: I thought that's what the MBR spec said? Up to 4 primary partitions, one of which can optionally be an extended partition, containing an unlimited number of logical partitions. But perhaps I'm mis-remembering... (or perhaps it's just convention rather than spec?)

lurch:
> EDIT: Warning if partitions don't span enough of the image turns out to be immensely inaccurate for several reasons, from unknown block size to images padded with gigabytes of zeros.

Agree that it doesn't make sense to warn if the partitions don't "fill enough" of the disk, but it's probably still worth warning (or erroring) if the partition tables point to data outside of the size of the image?

lib/image-stream/parse-partitions.js
jhermsmeier: Not really, the GPT has a header in one block, then the first 4 partitions in another block, and then partitions 5-128 in yet another block (see [Wikipedia/GUID_Partition_Table_Scheme.svg](https://en.wikipedia.org/wiki/GUID_Partition_Table#/media/File:GUID_Partition_Table_Scheme.svg) for a good illustration)

lib/image-stream/parse-partitions.js
jhermsmeier: Not necessarily Javascript specific; shouldn't be done in most languages afaik

lib/image-stream/parse-partitions.js
jhermsmeier: The primary GPT with all its parts is 31 blocks in size (the backup another 31), so it's not particularly fixed, but dependant on the block size.

lib/image-stream/parse-partitions.js
jhermsmeier: Mended

lib/image-stream/parse-partitions.js
jhermsmeier: Added

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+const INITIAL_BLOCK_SIZE = 512;
+
+/** @const {Number} Maximum block size to check for */
+const MAX_BLOCK_SIZE = 4096;
+
+/**
+ * @summary Attempt to parse the GPT from various block sizes
+ * @private
+ * @param {Buffer} buffer - Buffer
+ * @returns {GPT|null}
```
jviotti: Can we show an `@example` here?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+  let gpt = null;
+
+  // Attempt to parse the GPT from several offsets,
+  // as the block size of the image may vary (512,1024,2048,4096);
+  // For example, ISOs will usually have a block size of 4096,
+  // but raw images a block size of 512 bytes
+  while (blockSize <= MAX_BLOCK_SIZE) {
+    try {
+      gpt = GPT.parse(buffer.slice(blockSize));
+    } catch (error) {
```
jviotti: Maybe we worth using `_.attempt` if we want to swallow the error?

```js
while (blockSize <= MAX_BLOCK_SIZE) {
  const gpt = _.attempt(GPT.parse, buffer.slice(blockSize));

  if (!_.isError(gpt)) {
    return gpt;
  }

  blockSize += blockSize;
}
```

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+  return gpt;
+
+};
+
+/**
+ * @summary Attempt to parse the MBR & GPT from a given buffer
+ * @private
+ * @param {Object} image - Image metadata
+ * @param {Buffer} buffer - Buffer
+ * @returns {undefined}
```
jviotti: Can we put an example here?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+    blockSize += blockSize;
+  }
+
+  return gpt;
+
+};
+
+/**
+ * @summary Attempt to parse the MBR & GPT from a given buffer
+ * @private
```
jviotti: Can we do `@summary {...}`, `@function`, `@private` for consistency purposes?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+  let gpt = null;
+
+  try {
+    mbr = MBR.parse(buffer);
+    image.hasMBR = true;
+    if (mbr.getEFIPart()) {
+      image.hasGPT = true;
+      gpt = parseGPT(buffer);
+    }
+  } catch (error) {
```
jviotti: Should we use `_.attempt`?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+      };
+    });
+  }
+
+};
+
+/**
+ * @summary Attempt to read the MBR and GPT from an imagestream
+ * @description
+ *   This operation will consume the first `MAX_BYTES`
```
jviotti: Why is the description indented?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+        lastLBA: partition.lastLBA,
+        extended: partition.extended
+      };
+    });
+  }
+
+};
+
+/**
+ * @summary Attempt to read the MBR and GPT from an imagestream
```
jviotti: Should use set `@function` and `@public` here?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+ * @description
+ *   This operation will consume the first `MAX_BYTES`
+ *   of the stream and then destroy the stream.
+ * @param {Object} image - image metadata
+ * @returns {Promise}
```
jviotti: Can we write an example?

jviotti: Can we add `@fulfil` as well?

tests/image-stream/img.spec.js
jviotti: Yeah, we don't have any tests where `hasGPT` is `true`

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+'use strict';
+
+const _ = require('lodash');
+const Bluebird = require('bluebird');
+const MBR = require('mbr');
+const GPT = require('gpt');
+
+/** @const {Number} Maximum number of bytes to read from the stream */
```
jviotti: Can we use:

```
@summary ...
@type {Number}
@constant
```

For consistency purposes?

lib/image-stream/parse-partitions.js
jviotti: A function that doesn't return anything should usually not include a `@returns` annotation tag.

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+ * @param {Object} image - Image metadata
+ * @param {Buffer} buffer - Buffer
+ * @returns {undefined}
+ */
+const parsePartitions = (image, buffer) => {
+
+  let mbr = null;
+  let gpt = null;
+
+  try {
```
jviotti: Which is the function that may throw? Can we only add the exception handler in a specific point?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+    // Ignore error
+  }
+
+  // As MBR and GPT partition entries have a different structure,
+  // we normalize them here to make them easier to deal with and
+  // avoid clutter in what's sent to analytics
+  if (image.hasGPT) {
+    image.partitions = _.map(gpt.partitions, (partition) => {
+      return {
+        type: partition.type.toString(),
```
jviotti: Why do we use `.toString()` here, but not in the type declaration below?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+  return new Bluebird((resolve, reject) => {
+
+    const chunks = [];
+    let length = INITIAL_LENGTH;
+    let destroyed = false;
+
+    image.hasMBR = false;
+    image.hasGPT = false;
+
+    let stream = image.stream.pipe(image.transform);
```
jviotti: Can we use `const` here?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+
+    // We need to use the "old" flowing mode here,
+    // as some dependencies don't implement the "readable"
+    // mode properly (i.e. bzip2)
+    stream.on('data', (chunk) => {
+      chunks.push(chunk);
+      length += chunk.length;
+
+      // Once we've read enough bytes, terminate the stream
+      if (length >= MAX_BYTES && !destroyed) {
```
jviotti: Can we clarify what `MAX_BYTES` means here? Maximum bytes of what? Maybe we can call this constant `PARTITION_TABLE_MAX_BYTES`?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+
+        // Prefer close() over destroy(), as some streams
+        // from dependencies exhibit quirky behavior when destroyed
+        if (image.stream.close) {
```
jviotti: Can this ever be true?

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,174 @@
...
+    stream.on('error', reject);
+
+    // We need to use the "old" flowing mode here,
+    // as some dependencies don't implement the "readable"
+    // mode properly (i.e. bzip2)
+    stream.on('data', (chunk) => {
+      chunks.push(chunk);
+      length += chunk.length;
+
+      // Once we've read enough bytes, terminate the stream
```
jviotti: Looks like what you're doing here is slicing a stream, for which there are utility modules that can abstract the gory details for you like https://www.npmjs.com/package/slice-stream. We already use this in the writer, so can we use it here as well?

lib/image-stream/parse-partitions.js
jhermsmeier: Sure, done

lib/image-stream/parse-partitions.js
jhermsmeier: Done

lib/image-stream/parse-partitions.js
jhermsmeier: Sure, added

lib/image-stream/parse-partitions.js
jhermsmeier: Yup, added

jhermsmeier: Aye, removed

lib/image-stream/parse-partitions.js
jhermsmeier: Removed the try-catch in favor of `_.attempt()` now

lib/image-stream/parse-partitions.js
jhermsmeier: Updated to use `_.attempt()`, even though not having type consistency for a value bugs me a bit :)

lib/image-stream/parse-partitions.js
jhermsmeier: Partition types on GPTs are GUIDs that are structs in [`gpt/lib/guid.js`](https://github.com/jhermsmeier/node-gpt/blob/master/lib/guid.js), and for simplicity's sake we use the string representation here (easier to match etc.)

lib/image-stream/parse-partitions.js
jhermsmeier: Yup, added

jhermsmeier: Added, but that doesn't seem to be a valid jsdoc tag?

lib/image-stream/parse-partitions.js
jhermsmeier: Done

lib/image-stream/parse-partitions.js
jhermsmeier: Renamed to `MAX_STREAM_BYTES`

lib/image-stream/parse-partitions.js
jhermsmeier: Sure, with streams that have a file handle, or otherwise implement this method (fs, net, etc. streams for example)

lib/image-stream/parse-partitions.js
jhermsmeier: I must've thought that was supposed to be that way... unindented now.

lib/image-stream/parse-partitions.js
jhermsmeier: Added

lib/image-stream/parse-partitions.js
jhermsmeier: Sure, added

lib/image-stream/parse-partitions.js
jhermsmeier: I'd prefer not to use another dep for this, especially since it's a rather simple operation to do. Also `slice-stream` doesn't end / close / destroy the stream once the required slice is read.

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,205 @@
...
+ * @type {Number}
+ * @constant
+ */
+const MAX_BLOCK_SIZE = 4096;
+
+/**
+ * @summary Attempt to parse the GPT from various block sizes
+ * @function
+ * @private
+ * @param {Buffer} buffer - Buffer
```
jviotti: Can we add blank lines above the first `@param` and below `@returns` for consistency purposes? (it also makes it easier to the eye to spot the signature of the function from the headers, examples, etc)

lib/image-stream/parse-partitions.js
jviotti: I think you missed `@function` here.

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,205 @@
...
+
+};
+
+/**
+ * @summary Attempt to parse the MBR & GPT from a given buffer
+ * @private
+ * @param {Object} image - Image metadata
+ * @param {Buffer} buffer - Buffer
+ * @example
+ * parsePartitionTables(image, buffer)
```
jviotti: Let's use semi-colons in the examples, for consistency with the linting guidelines of the rest of the code.

lib/image-stream/parse-partitions.js
```diff
@@ -0,0 +1,205 @@
...
+        extended: partition.extended
+      };
+    });
+  }
+
+};
+
+/**
+ * @summary Attempt to read the MBR and GPT from an imagestream
+ * @description
```
jviotti: Can we move `@description` below `@public` and `@function` for consistency purposes?

jhermsmeier:
Cool, I'll add some GPT test cases soon – just need to do the whole disk image creation dance again :)

jhermsmeier:
@jviotti Added a GPT test image & case now

lib/image-stream/parse-partitions.js
jviotti: Ping @jhermsmeier

lib/image-stream/parse-partitions.js
jviotti: Ping @jhermsmeier

lib/image-stream/parse-partitions.js
jviotti: Ping @jhermsmeier

lib/image-stream/parse-partitions.js
jviotti: Ping @jhermsmeier

jhermsmeier:
@jviotti oh, I don't recall seeing them before – mended now!

PR #1252, opened 2017-03-31
refactor(GUI): get rid of `basename` Angular filter

jviotti:


tests/gui/pages/main.spec.js
```diff
@@ -147,6 +150,33 @@ describe('Browser: MainPage', function() {
...
+
+      it('should return the basename of the selected image', function() {
+        const controller = $controller('ImageSelectionController', {
+          $scope: {}
+        });
+
+        SelectionStateModel.setImage({
+          path: path.join(__dirname, 'foo', 'bar.img'),
+          size: 999999999
+        });
```
lurch: Seems like lots of different unit tests now all end up creating their own temporary `image` objects. Would it be worth having a standard / shared test-image-object that all the unit-tests could re-use, or at least use as a common starting point? Or is it better for the unit tests to remain verbose and "independent" of each other?

jviotti: True. Maybe we should add some "factory" objects to create common things like images, drives, etc for testing purposes. What do you think?

lurch: Sounds good to me :-)

PR #1253, opened 2017-03-31
refactor(GUI): make OSWindowProgress a simple CommonJS module

jviotti:
There is no need to tie this module to Angular by using an Angular
service.


lib/gui/os/window-progress.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {Number} percentage - percentage
+ *
+ * @example
+ * windowProgress.set(85);
+ */
+exports.set = (percentage) => {
+  const PERCENTAGE_MINIMUM = 0;
+  const PERCENTAGE_MAXIMUM = 100;
+
+  if (percentage > PERCENTAGE_MAXIMUM || percentage < PERCENTAGE_MINIMUM) {
```
lurch: There now seems to be several places throughout Etcher where we're checking that a percentage lies within the 0 -> 100 range, which leads to many re-definitions of `ZERO` and `HUNDRED`. Maybe it's time to add a shared `isValidPercentage` function?

lib/gui/os/window-progress.js
```diff
@@ -0,0 +1,70 @@
...
+    throw errors.createError(`Invalid window progress percentage: ${percentage}`);
+  }
+
+  exports.currentWindow.setProgressBar(percentage / PERCENTAGE_MAXIMUM);
+};
+
+/**
+ * @summary Clear the operating system window progress bar
```
lurch: "operating system window progress bar" feels like a bit of a mouthful (and a bit vague). Is there a more succinct name for this?

lurch:
Requested a review from @Shou because I guess it ties up with some of the Angular-removal work he's also been doing :-)

lib/gui/os/window-progress.js
jviotti: Done!

lib/gui/os/window-progress.js
jviotti: I reduced it to "Clear the window progress bar."

tests/shared/utils.spec.js
```diff
@@ -333,4 +333,32 @@ describe('Shared: Utils', function() {
 
   });
 
+  describe('.isValidPercentage()', function() {
```
lurch: Could also test it with floating-point numbers? E.g. -0.1, 0.1, 12.3, 99.9, 100.1

jviotti: Done!

lib/gui/os/window-progress.js
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @example
+ * windowProgress.set(85);
+ */
+exports.set = (percentage) => {
+  if (!utils.isValidPercentage(percentage)) {
+    throw errors.createError(`Invalid window progress percentage: ${percentage}`);
+  }
+
+  const PERCENTAGE_MAXIMUM = 100;
```
lurch: I wonder if this constant could / should be exposed from `utils`?

Or perhaps there could be a `utils.percentageToFloat` function? \*shrug\*

lib/shared/utils.js
```diff
@@ -133,3 +133,26 @@ exports.hideAbsolutePathsInObject = (object) => {
...
+ *
+ * @param {Number} percentage - percentage
+ * @returns {Boolean} whether the percentage is valid
+ *
+ * @example
+ * if (utils.isValidPercentage(85)) {
+ *   console.log('The percentage is valid');
+ * }
+ */
+exports.isValidPercentage = (percentage) => {
```
lurch: :tada:

lib/gui/os/window-progress.js
jviotti: That's a great idea. Implemented!

tests/shared/utils.spec.js
```diff
@@ -333,4 +333,88 @@ describe('Shared: Utils', function() {
...
+      m.chai.expect(utils.isValidPercentage(101)).to.be.false;
+    });
+
+    it('should return false if percentage is a float greater than 100', function() {
+      m.chai.expect(utils.isValidPercentage(100.001)).to.be.false;
+    });
+
+  });
+
+  describe('.percentageToFloat()', function() {
```
lurch: No test for passing in strings? ;)

jviotti: Done!

jviotti:
Oh, this PR has now tons of conflicts :(

lurch:
Re-pinging @Shou for a review.

tests/shared/utils.spec.js
Shou: Maybe `null` would be good to test too considering `null <= 100` is `true`, to make it explicit we avoid this pitfall?

lib/shared/utils.js
```diff
@@ -0,0 +1,78 @@
...
+ * @returns {Number} float percentage
+ *
+ * @example
+ * const value = utils.percentageToFloat(50);
+ * console.log(value);
+ * > 0.5
+ */
+exports.percentageToFloat = (percentage) => {
+  if (!exports.isValidPercentage(percentage)) {
+    throw errors.createError(`Invalid window progress percentage: ${percentage}`);
```
Shou: Also, we should make this error generic instead of referencing the window progress.

Shou: Also, we should make this error generic instead of referencing the window progress.

lurch: Good spot!

tests/shared/utils.spec.js
jviotti: Oh, that's some hardcore JavaScript weirdness :P Fixed!

lib/shared/utils.js
jviotti: Oh, right, good catch!

tests/shared/utils.spec.js
```diff
@@ -0,0 +1,126 @@
...
+
+    it('should throw an error if given a string percentage', function() {
+      m.chai.expect(function() {
+        utils.percentageToFloat('50');
+      }).to.throw('Invalid percentage: 50');
+    });
+
+    it('should throw an error if given a null percentage', function() {
```
lurch: IMHO you should add this test to the `.isValidPercentage()` section too.

jviotti: Done!

PR #1257, opened 2017-04-05
feat(image-stream): Support Apple Disk Images

jhermsmeier:
This adds support for flashing Apple Disk Images (.dmg)

**To Do:**
- [x] Add tests and test images
- [x] Implement ADC decompression in UDIF

Change-Type: minor


lib/image-stream/utils.js
```diff
@@ -37,6 +38,10 @@ const archiveType = require('archive-type');
  */
 exports.getArchiveMimeType = (file) => {
 
+  if (mime.lookup(file)) {
```
jviotti: Could `mime-types` replace `archive-type`?

lib/image-stream/handlers.js
```diff
@@ -123,6 +124,46 @@ module.exports = {
...
+   * @param {String} file - file path
+   * @param {Object} options - options
+   * @param {Number} [options.size] - file size
+   *
+   * @fulfil {Object} - image metadata
+   * @returns {Promise}
+   */
+  'application/x-apple-diskimage': (file, options) => {
+    return new Bluebird((resolve, reject) => {
+      const dmg = new UDIF.Image(file);
```
jviotti: I guess we can promisify `new UDIF.Image(file)` to simplify the whole flow:

```js
const dmg = Bluebird.promisifyAll(new UDIF.Image(file));

const info = {
  stream: UDIF.createReadStream(file),
  size: {
    original: options.size,
    final: {
      estimation: false,
      value: options.size
    }
  },
  transform: new PassThroughStream()
};

return dmg.openAsync().then(() => {
  info.size.final.value = dmg.getUncompressedSize();
  return dmg.closeAsync().return(info);
});
```

jviotti: BTW, `info.size.final.value` always gets replaced, and `info` is only used at the end, so I guess this could be:

```js
const dmg = Bluebird.promisifyAll(new UDIF.Image(file));

return dmg.openAsync().then(() => {
  return dmg.closeAsync().return({
    stream: UDIF.createReadStream(file),
    size: {
      original: options.size,
      final: {
        estimation: false,
        value: dmg.getUncompressedSize()
      }
    },
    transform: new PassThroughStream()
  });
});
```

jviotti: Oh, the issue is that you need to get the uncompressed size before closing...

jviotti: Maybe:

```js
const dmg = Bluebird.promisifyAll(new UDIF.Image(file));

return dmg.openAsync().then(() => {
  const info = {
    stream: UDIF.createReadStream(file),
    size: {
      original: options.size,
      final: {
        estimation: false,
        value: dmg.getUncompressedSize()
      }
    },
    transform: new PassThroughStream()
  };

  return dmg.closeAsync().return(info);
});
```

?

jviotti: Or even better:

```js
const dmg = Bluebird.promisifyAll(new UDIF.Image(file));

return dmg.openAsync().return({
  stream: UDIF.createReadStream(file),
  size: {
    original: options.size,
    final: {
      estimation: false,
      value: dmg.getUncompressedSize()
    }
  },
  transform: new PassThroughStream()
}).tap(dmg.closeAsync);
```

And now I'm going to stop over-optimizing this promise chain :P

jhermsmeier: Hahaha 😁  I'm thinking I could just stick the open -> getUncompressedSize -> close into a method on UDIF, i.e. `UDIF.getUncompressedSize( filename, callback )` and then we can avoid doing this dance altogether :)

lib/image-stream/utils.js
jhermsmeier: Yeah, I guess – but probably should do that in a separate, preflight PR – what do you think?

jviotti: Depends on how much work it is. If leaving just `mime-types` makes the tests green, lets go for it here.

lib/image-stream/handlers.js
jhermsmeier: Updated to use `UDIF.getUncompressedSize( filename, callback )` now

lib/image-stream/utils.js
jhermsmeier: Removed `archive-type` now

lib/image-stream/index.js
```diff
@@ -67,15 +67,15 @@ const errors = require('../shared/errors');
  * });
  */
 exports.getFromFilePath = (file) => {
-  return utils.getArchiveMimeType(file).then((type) => {
-    if (!_.has(handlers, type)) {
-      throw errors.createUserError('Invalid image', `The ${type} format is not supported`);
-    }
+  const type = utils.getArchiveMimeType(file);
```
lurch: As you know I'm not a Promise-expert :wink: , but I'm wondering why this code was changed?

lib/image-stream/utils.js
```diff
@@ -16,47 +16,30 @@
...
 
 /**
  * @summary Get archive mime type
  * @function
  * @public
  *
- * @param {String} file - file path
+ * @param {String} filename - file path
  * @fulfil {String} - mime type
  * @returns {Promise}
```
lurch: Does this need updating? If this no longer returns a Promise then that'd answer my earlier question :)

lib/image-stream/utils.js
```diff
@@ -16,47 +16,30 @@
...
-      fileDescriptor,
-      chunk,
-      BUFFER_START,
-      ARCHIVE_TYPE_IDENTIFICATION_BYTES_LENGTH,
-      null
-    ).then(() => {
-      return _.get(archiveType(chunk), [ 'mime' ], 'application/octet-stream');
-    });
-  });
+  return mime.lookup(filename) || 'application/octet-stream';
```
lurch: Hmmm, so AFAICT previously we were using `archive-type` and examining the first 261 bytes of the image to determine the image's file-type, and instead we're now using `mime-types` to determine the file-type purely from the filename?

Would there be any benefit into combining both approaches? E.g. if `mime-types` doesn't recognise the filename, then fall back to using `archive-type` ?

lib/image-stream/index.js
jhermsmeier: No need to use a Promise here anymore as `. getArchiveMimeType()` isn't async anymore

lib/image-stream/utils.js
jhermsmeier: Ah yes, this needs mending

lib/image-stream/utils.js
jhermsmeier: Yes, `mime-types` determines the mime type by file extension, which is very reliable as most of them are specced in RFCs by IANA / IETF / W3C / etc., and people usually don't just change file extensions around. It's also plenty faster than doing I/O to figure it out.

> Would there be any benefit into combining both approaches?

I don't think so, as `archive-type` is limited to certain file formats, and detection by magic number can be mighty unreliable when we get into the realms of other file formats (`.dmg`s for example are known to be notorious detection busters as they have no header and begin with the data-fork which can contain anything).
If we'd really want highest-grade format detection then we'll have to implement specific format tests (i.e. by having something like `handler.test( filename, callback )`) – but that's material for another time.

jviotti: Oh, I didn't know `mime-types` would just look at the file name :( The I/O overhead is probably minimal, so I guess its a question of how much we trust file names.

From one side, I agree with @jhermsmeier that users are not going to change the extension, however my instinct tells me that if we can do it the reliable way (by inspecting the file itself), we should go for that.

Did `archive-type` fail to recognize DMGs?

jviotti: > .dmgs for example are known to be notorious detection busters as they have no header and begin with the data-fork which can contain anything).

I see. Lets go with @lurch suggestion and combine both then, by using `mime-types` only if `archive-type` returns `application/octet-stream`. If this is the case but `mime-types` detects another type, then use that instead.

jhermsmeier: > Did archive-type fail to recognize DMGs?

I most likely will (note how none of the following are correctly identified):

```
file test/data/*.dmg
test/data/encrypted.dmg:           VAX COFF executable not stripped
test/data/gpt.dmg:                 VAX COFF executable - version 10374
test/data/Patchwork-3.2.2-mac.dmg: bzip2 compressed data, block size = 100k
test/data/raw.dmg:                 Macintosh HFS Extended version 4 data last mounted by: 'HFSJ', created: Thu Apr  6 14:17:01 2017, last modified: Thu Apr  6 11:37:25 2017, last checked: Thu Apr  6 10:17:01 2017, block size: 4096, number of blocks: 572, free blocks: 403
test/data/zlib-compressed.dmg:     VAX COFF executable not stripped - version 8502
```

jhermsmeier: I also just noticed that `archive-type` is just a tiny wrapper around [file-type](https://github.com/sindresorhus/file-type) – we should probably use that instead.

And to mitigate those false detections, I think it'd be better if we combine them the other way around: Checking with `mime-types` first, and then with `file-type` second, if the first doesn't figure it out.

jhermsmeier: I also just checked `file-type`, and it would detect a few of those samples correctly, but only by the rather slim chance that the `.dmg` starts with `0x78 0x01`.

jviotti: Sounds good to me! :+1:

lurch: I initially agreed with @jhermsmeier that we ought to check `mime-types` first and `file-type` second, but I decided to have a look at this in a bit more detail...

As @jhermsmeier points out, `archive-type` is just a slim wrapper around `file-type` - with the difference being that `archive-type` only needs 261 bytes (although see https://github.com/kevva/archive-type/issues/7), and can only identify "archive" file types (7z, bz2, gz, rar, tar, zip, xz); whereas `file-type` needs the first 4100 bytes, but can identify a much larger range of file types.
Up until now, using just `archive-type` has sufficed for Etcher, since it's only "archive" types that we need to do any special processing on. However as @jhermsmeier points out, even `file-type` is unlikely to be able to identify DMGs, and we _do_ need to do special processing on DMGs, which is why he's switched to using `mime-types`.

So IMHO we should stick with using `archive-type` rather than `file-type`, and given that restriction (i.e. that archive-type doesn't even attempt to identify DMGs) I think it probably makes sense to go with @jviotti 's suggestion of checking with `archive-type` first, and only if that returns `false` proceeding with checking via `mime-types`. And I believe if `mime-types` returns anything other than 'application/x-apple-diskimage' then we should force `getArchiveMimeType` to return 'application/octet-stream'. (@jviotti does every mime-type returned by `getArchiveMimeType` need a corresponding handler in `lib/image-stream/handlers.js`?)

Alternatively, if all `mime-types` is doing is checking the file-extension, maybe it'd be easier to just check the file extension for `.dmg` ourselves, and not add the extra `mime-types` dependency?

tests/image-stream/utils.spec.js
```diff
@@ -26,44 +26,34 @@ describe('ImageStream: Utils', function() {
 
   describe('.getArchiveMimeType()', function() {
 
-    it('should resolve application/x-bzip2 for a bz2 archive', function(done) {
+    it('should resolve application/x-bzip2 for a bz2 archive', function() {
```
lurch: If `.getArchiveMimeType` isn't returning a Promise, should the unit-tests be updated to say 'return' instead of 'resolve'?

lib/image-stream/utils.js
jhermsmeier: I don't see the benefit in using `archive-type` as it's just an intermediate dependency, and the same can be done in 3 LOC:

```js
var type = fileType(buf)
var exts = [ '7z', 'zip', ... ]
return exts.contains( type ) ? type : false
```

Plus, we don't even need to narrow it down like `archive-type` does, because image-stream checks for a handlers existence based on the mime type.

> whereas file-type needs the first 4100 bytes, but can identify a much larger range of file types.

Where does it say that? Looking at the code, the highest offset into the buffer I could find is 261.

> and given that restriction (i.e. that archive-type doesn't even attempt to identify DMGs)

It doesn't, but as outlined earlier, it will detect other file types incorrectly *instead of* a DMG, because DMGs can contain anything in the first couple of bytes as they don't have any data structures or magic numbers there.
So if we use `file-type` before `mime-types`, then it will incorrectly detect a DMG as something else.

tests/image-stream/utils.spec.js
jhermsmeier: Changed it back to Promises now, so not necessary anymore

lib/image-stream/utils.js
```diff
@@ -16,47 +16,30 @@
...
+const mime = require('mime-types');
 
 /**
  * @summary Get archive mime type
  * @function
  * @public
  *
- * @param {String} file - file path
+ * @param {String} filename - file path
  * @fulfil {String} - mime type
```
jhermsmeier: It returns a Promise again now, so not necessary

lib/image-stream/utils.js
jhermsmeier: Mended

jhermsmeier:
So, I have now added ADC block decompression to UDIF, added tests and test images to image-stream, and replaced `archive-type` with `file-type` under addition of `mime-types`. Should be ready for another round of eyeballing @jviotti  @lurch :)

lib/image-stream/handlers.js
```diff
@@ -23,6 +23,7 @@ const lzma = Bluebird.promisifyAll(require('lzma-native'));
 const zlib = require('zlib');
 const unbzip2Stream = require('unbzip2-stream');
 const gzip = require('./gzip');
+const UDIF = require('udif');
```
jhermsmeier: Yeah, changed that to lowercase

lib/image-stream/utils.js
lurch: > Where does it say that? Looking at the code, the highest offset into the buffer I could find is 261.

Yeah, I was wondering that too! Eventually figured out (by looking through git history) that it's due to https://github.com/sindresorhus/file-type/blob/master/index.js#L175

> So if we use file-type before mime-types, then it will incorrectly detect a DMG as something else.

Not if we filter out anything not in the `[ '7z', 'zip', ... ]` list first ;)
Although I've just spotted `test/data/Patchwork-3.2.2-mac.dmg: bzip2 compressed data, block size = 100k` in your test results above, which I guess means that yeah it does need to check the file extension first, to prevent it mis-detecting a DMG as a BZIP2. Sorry.
But as I mentioned, do we really need to pull in all of `mime-types` just to determine whether a file ends with `.dmg` or not??

jhermsmeier: > Yeah, I was wondering that too! Eventually figured out (by looking through git history) that it's due to https://github.com/sindresorhus/file-type/blob/master/index.js#L175

Oh, I missed that one! Well, that's not going to fail if there are fewer than 4100 bytes, and archive-type would hit that case as well, because it only filters for archive extensions after calling `fileType()`.

> to prevent it mis-detecting a DMG as a BZIP2.

Yeah, exactly – and DMGs are also not the only format being an elusive bugger with magic number detection – so this way around we should also avoid those automagically, should they arise.

> But as I mentioned, do we really need to pull in all of mime-types just to determine whether a file ends with .dmg or not??

I'd recommend it, rather than hand-rolling it, especially since it has pretty good reusability and we (hopefully) won't have to think about this again when adding new formats.

tests/image-stream/utils.spec.js
lurch: If .getArchiveMimeType is now back to Promises, does this need the `done` that the original version had?

lib/image-stream/supported.js
```diff
@@ -56,5 +56,9 @@ module.exports = [
   {
     extension: 'raw',
     type: 'image'
+  },
+  {
+    extension: 'dmg',
+    type: 'image'
```
lurch: @jhermsmeier From what you've been saying, I guess this is actually a 'compressed image'? Do we need to make that distinction? @jviotti What are these `type` labels used for?

npm-shrinkwrap.json
```diff
@@ -5982,6 +6017,11 @@
       "from": "sax@>=0.6.0",
       "resolved": "https://registry.npmjs.org/sax/-/sax-1.2.2.tgz"
     },
+    "seek-bzip": {
```
lurch: Is it worth re-using this elsewhere in Etcher? E.g. instead of `unbzip2-stream`

tests/image-stream/utils.spec.js
jhermsmeier: Nope, quite the opposite – mocha knows how to handle promises and the `done()` in this case would stall the test on failure as unhandled exceptions from promises don't throw and don't terminate the process – which is why the promise is *returned* and no `done()` is being used.

lib/image-stream/utils.js
```diff
@@ -18,45 +18,40 @@
...
 
-  // `archive-type` only needs the first 261 bytes
-  // See https://github.com/kevva/archive-type
-  const ARCHIVE_TYPE_IDENTIFICATION_BYTES_LENGTH = 261;
+  const FILE_TYPE_ID_BYTES = 261;
 
-  return Bluebird.using(fs.openAsync(file, 'r').disposer((fileDescriptor) => {
-    return fs.closeAsync(fileDescriptor);
-  }), (fileDescriptor) => {
+  return Bluebird.using(fs.openAsync(filename, 'r').disposer((fd) => {
```
jhermsmeier: I'd rather keep it like it is, as I find that `file` would indicate the value being an object of sorts, whereas `filename` makes it quite clear that it's a string.

jhermsmeier: Also, I'm not sure that the smallest diff is something we should be optimising for

lib/image-stream/utils.js
```diff
@@ -18,45 +18,40 @@
...
-    return fs.readAsync(
-      fileDescriptor,
-      chunk,
-      BUFFER_START,
-      ARCHIVE_TYPE_IDENTIFICATION_BYTES_LENGTH,
-      null
-    ).then(() => {
-      return _.get(archiveType(chunk), [ 'mime' ], 'application/octet-stream');
+    return fs.readAsync(fd, buffer, BUFFER_START, FILE_TYPE_ID_BYTES, null).then(() => {
+      return _.get(fileType(buffer), [ 'mime' ], 'application/octet-stream');
```
jhermsmeier: That happens in `lib/image-stream/index.js`, line 71 – nothing changed there, so no need to do it twice here

lib/image-stream/supported.js
jhermsmeier: It can be raw, or compressed – that depends on how it was created. Wasn't sure what to go for... but I guess most of the time the contents would be compressed.

npm-shrinkwrap.json
jhermsmeier: Hmm, not sure – looks like streaming isn't implemented in `seek-bzip`, so we're probably better off waiting for an alternative for bzipping things.

tests/image-stream/utils.spec.js
lurch: Oh. I'm afraid I don't know what any of that means, but does it mean that @jviotti 's original code was 'wrong'? Could this same problem also be affecting other unit-tests?

lib/image-stream/utils.js
lurch: \*shrug\*  Fair enough, I just find smaller diffs easier to review ;)

npm-shrinkwrap.json
lurch: Does that mean that your udif-reading code won't be able to stream images either?
(apologies if that's a dumb question)

jhermsmeier: As you can see from `udif.createReadStream()`, it is perfectly capable of streaming an image. The UDIF compresses each block range separately, so there's no need for a streaming bzip decompressor to stream an image.

tests/image-stream/utils.spec.js
jhermsmeier: Not necessarily wrong, the failures would just take longer, and manifest in timeouts rather than having a little more helpful stack trace and error message.
Yes, I think there are still a few tests where this is the case, can do a check & follow up PR on those.

lib/image-stream/utils.js
lurch: Thanks for the clarification :+1: 
But given how `.getArchiveMimeType` worked previously, where it only returned a known archive (or compression) format, or `application/octet-stream` otherwise, it meant that (if somebody really wanted to) they could write a file named `blah.png` to their SD card, because it'd just get recognised as `application/octet-stream`. I guess that wouldn't work with the new code? We had an issue previously where someone was trying to flash a disk-image-format we didn't officially recognise, but it still worked due to Etcher identifying it as `application/octet-stream`.

Probably best if @jviotti determines what the "expected behaviour" here is?
(I'm not trying to be argumentative here, I'm just trying to check that by only adding support for `.dmg`, we're not also "turning off" functionality that used to work)

lib/image-stream/utils.js
```diff
@@ -18,45 +18,40 @@
...
  * @example
- * utils.getArchiveMimeType('path/to/raspberrypi.img.gz').then((mimeType) => {
- *   console.log(mimeType);
- * });
+ * const mimeType = utils.getArchiveMimeType('path/to/raspberrypi.img.gz');
  */
-exports.getArchiveMimeType = (file) => {
+exports.getArchiveMimeType = (filename) => {
+  if (mime.lookup(filename)) {
+    return Bluebird.resolve(mime.lookup(filename));
```
lurch: Could this be tweaked to not make two calls to `mime.lookup` ?

lib/image-stream/utils.js
jviotti: Just a nit, but we can we say `fileDescriptor` instead of `fd` for consistency purposes (sorry, I really hate these abbreviations :P)?

lib/image-stream/utils.js
jhermsmeier: Hmm, yeah – this will cause image-stream to bail out with `The ${type} format is not supported` now if there's no handler for i.e. `image/png`, but I guess people could just rename whatever they want to flash to `.img` or `.bin` if it's not "officially" supported by Etcher.

lib/image-stream/handlers.js
```diff
@@ -123,6 +124,41 @@ module.exports = {
...
+   * @param {String} file - file path
+   * @param {Object} options - options
+   * @param {Number} [options.size] - file size
+   *
+   * @fulfil {Object} - image metadata
+   * @returns {Promise}
+   */
+  'application/x-apple-diskimage': (file, options) => {
+    return new Bluebird((resolve, reject) => {
+      udif.getUncompressedSize(file, (error, size) => {
```
jviotti: Can we promisify `udif` to simplify this promise chain? For example:

```js
const udif = Bluebird.promisifyAll(require('udif'));

return udif.getUncompressedSizeAsync(file).then((size) => {
  return {
    stream: udif.createReadStream(file),
    size: {
      original: options.size,
      final: {
        estimation: false,
        value: size
      }
    },
    transform: new PassThroughStream()
  };
});
```

lib/image-stream/supported.js
jviotti: That's tricky. I guess `compressed` is the way to go. In any case, these type labels are only used to show an ordered tooltip of the supported file types.

tests/image-stream/utils.spec.js
jviotti: > mocha knows how to handle promises

Wow, I had no idea this was the case, and I've been using `done` all this time. If you can correct those, that'd be awesome :+1:

tests/image-stream/dmg.spec.js
```diff
@@ -0,0 +1,64 @@
...
+        path.join(DMG_PATH, 'zlib-compressed.dmg'),
+        path.join(IMAGES_PATH, 'zlib-compressed.img'));
+    });
+
+  });
+
+  describe('.getImageMetadata()', function() {
+
+    it('should return the correct metadata', function() {
+      const image = path.join(DMG_PATH, 'zlib-compressed.dmg');
```
jviotti: Should we also add uncompressed DMGs to the test suite?

lib/image-stream/utils.js
jviotti: Lets either:

- Bring back the old behavior (only return supported MIME types or `application/octet-stream`), or
- Allow a `The ${type} format is not supported` error, but make it a user error, so that it doesn't get to TrackJS/Sentry

tests/image-stream/dmg.spec.js
jhermsmeier: I've got a range of test images in the `udif` module here: https://github.com/jhermsmeier/node-udif/tree/master/test/data Not sure we'd need to test the same stuff in here, as that's internal stuff Etcher doesn't interface with. WDYT?

jviotti: I suggest to only test compressed/uncompressed, so we cover the final image size thing, and the rest can remain in the UDIF repo.

lib/image-stream/utils.js
```diff
@@ -18,45 +18,40 @@
...
- * @param {String} file - file path
+ * @param {String} filename - file path
  * @fulfil {String} - mime type
  * @returns {Promise}
  *
  * @example
- * utils.getArchiveMimeType('path/to/raspberrypi.img.gz').then((mimeType) => {
- *   console.log(mimeType);
- * });
+ * const mimeType = utils.getArchiveMimeType('path/to/raspberrypi.img.gz');
```
jhermsmeier: Yup, fixed!

lib/image-stream/utils.js
```diff
@@ -18,45 +18,40 @@
...
-exports.getArchiveMimeType = (file) => {
+exports.getArchiveMimeType = (filename) => {
+  if (mime.lookup(filename)) {
+    return Bluebird.resolve(mime.lookup(filename));
+  }
 
-  // `archive-type` only needs the first 261 bytes
-  // See https://github.com/kevva/archive-type
-  const ARCHIVE_TYPE_IDENTIFICATION_BYTES_LENGTH = 261;
+  const FILE_TYPE_ID_BYTES = 261;
```
jhermsmeier: Oh yes (even though I don't think it matters for the file types we're after)

lib/image-stream/utils.js
jhermsmeier: Changed it back :)

lib/image-stream/supported.js
jhermsmeier: Alright, changed it to `compressed`

lib/image-stream/utils.js
jhermsmeier: Yup, done!

lib/image-stream/handlers.js
jhermsmeier: Sure, done!

lib/image-stream/utils.js
jhermsmeier: > Allow a The ${type} format is not supported error, but make it a user error, so that it doesn't get to TrackJS/Sentry

It currently does the following:
```js
throw errors.createUserError('Invalid image', `The ${type} format is not supported`);
```
Does that result in a user error, or does that bubble up to the analytics?

jviotti: Ah, excellent, its an user error already :)

tests/image-stream/utils.spec.js
```diff
@@ -26,44 +26,46 @@ describe('ImageStream: Utils', function() {
...
-      utils.getArchiveMimeType(file).then((type) => {
+      return utils.getArchiveMimeType(file).then((type) => {
         m.chai.expect(type).to.equal('application/octet-stream');
-        done();
-      }).catch(done);
+      });
+    });
+
+    it('should resolve application/x-apple-diskimage for an Apple disk image', function() {
+      const file = path.join(DATA_PATH, 'dmg', 'zlib-compressed.dmg');
```
jviotti: I guess we should add a case for the uncompressed DMG part as well.

tests/image-stream/dmg.spec.js
jhermsmeier: Added

tests/image-stream/utils.spec.js
jhermsmeier: Added

Issue #1171, opened 2017-03-09
Flash Process slows to a crawl around 95% on big gzip images


petealbertson:
- **Etcher version:**
1.0.0-beta.19
- **Operating system and architecture:**
MacOS Sierra 10.12.3
MacBook Air
- **Do you see any meaningful error information on DevTools?**
No errors in the console. Only lots of progress logs. 

I'm attempting to flash an image of Android 6 to a 32BG Sandisk microSD card and for the second consecutive attempt, it slows to a crawl around 95%. It's now at a write speed of 0.64MB/s and slowing. Is there an abort protocol?

jhermsmeier:
Hey @petealbertson, can you tell us which exact image you're using, and (if not obvious) where to get it, so that we can have an in depth look at this?

lurch:
Sounds similar to #638 ?

jhermsmeier:
Not sure, Android images usually aren't that big, and come in `.zip`s

jviotti:
Yeah, the gz issue was the first thing that came to my mind as well. Let's see what image the user was flashing.

petealbertson:
Yes, it is an Android image from https://www.pine64.org/?page_id=1986

jhermsmeier:
Which one of those exactly did you use?

petealbertson:
The 32gb version direct from Pine (the first link under 32gb).

jhermsmeier:
Okay, so those images are gzipped, not `.zip`s, plus their contents are all over 4GB, which means #638 is most likely the culprit.

A workaround would be to extract the `.img` before flashing it, if you have enough disk space (the pine image will decompress to ~32GB).

jviotti:
> the pine image will decompress to ~32GB

Wow, that's the record so far :)

@jhermsmeier Did you have a chance to investigate the gzip issue in more detail? I'd love to hear your thoughts about what can we do to fix it.

lurch:
That's insane, the Pine developers have obviously never heard of partition resizing and filesystem expansion ;-)  (note that each of the different images compresses down to a similar size, indicating that they all have almost identical contents)

lurch:
....oh, except of course these are **Android** images, and Android probably doesn't have the tools to resize partitions like Linux does. Whoops :confused:

lurch:
[...]

Issue #638, opened 2016-08-19
.gz doesn't return correct file size for content above 2^32 B = 4 GB


imrehg:
- 1.0.0-beta13
- Linux 64bit

Taking it out from #629, apparently the gzip file format cannot accurately return the size of
 files above 4GB (2^32 bytes), but returns the modulo.

Looks like on the command line people recommended something like `zcat file.gz | wc -c` or `gzip -dc file | wc -c` which give the correct value - though then decompresses the the file twice. Might have to do that for gzip in the end, though, since likely >4GB files are common for Etcher's use case.

This might let images to start to be burned onto cards that are too small (in worst case), or affects the progress bar.

From testing with a 4100MiB > 4096MiB image, indeed `.gz` version lets to select a 512MB SD card, while the same file's `.xz` archive does not.
For the progress bar, the MB/s reading seems to be affected (shows very low speed, eg. 0.01MB/s)  but the progress percentage does not (shows correctly for the burning process), so it's not too bad.

jviotti:
So as far as I understand, the user will eventually hit `ENOSPC`, without any further weird behaviour, right?

I would love to be able to reliably get the drive size (or at least closely estimate), however decompressing the whole thing twice sounds like a very bad solution.

I'll research on this and see if there is a way we can do it.

imrehg:
@jviotti I think your description is still mixing up two uncovered issues, broken out to two parts to here and to #629. In that other issue (as described the reproducible `ENOSPC`) that happens regardless of compression.

In this issue:

* `ENOSPC` would happen if there's a `gz` image with `SIZE > 2^32` bytes,  and the user is trying to burn onto a card which has `CAPACITY < SIZE` but `CAPACITY > SIZE mod 2^32`. Then it would cause an issue, because the initial capacity check in etcher couldn't figure out the correct size

* If using a card that `CAPACITY > SIZE`, then the only effect is that the "speed" bar is wrong, but everything else works properly (including the progress bar), and the user won't run into `ENOSPC`.

Decompressing things twice might be a bad solution, but judging by the comments, it's just `gz` not designed for these big files (nor to return correct size estimate either), so curious to see if there's any other solution than run through the file twice. Should not be too bad, especially if that has some UI display such as "checking archive contents", so people know it not all just hung. I think doing things correctly for `gz` is more important than taking a bit longer time. Decompression itself with no data storage just byte counting seems to be pretty fast.

jviotti:
[...]

PR #1259, opened 2017-04-05
fix(writer): Use uncompressed or estimated size for progress

jhermsmeier:
This makes the writer use the image's uncompressed size, instead of the original size to avoid the progress percentage going above 100%.

Change-Type: patch

### DIFF
--- a/lib/cli/writer.js
+++ b/lib/cli/writer.js
@@ -84,7 +84,7 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
         size: drive.size
       }, {
         stream: image.stream,
-        size: image.size.original
+        size: image.size.final.value
       }, {
         check: options.validateWriteOnSuccess,
         transform: image.transform,


jviotti:
@jhermsmeier Can you link me to an image that reproduces the issue, so I can play around with it?

lurch:
This might connect to #1171 and #638 ?

@jhermsmeier Surely if the final size is only an estimate, then the percentage might still go above 100% if the actual size is bigger than the estimated size?

jhermsmeier:
@jviotti You should be able to see it quite significantly on the support-dmg branch when flashing the Etcher .dmg, as there are a couple hundred MB of compressed zeros in there (and that'll rocket up to 400%+). Otherwise those [Pine64](https://www.pine64.org/?page_id=1986) images should also do the trick.

jhermsmeier:
@lurch The issue with gzip is that the original size is stored as a uint32 as the modulo of 2^32 – it can't encode sizes larger than 2^32 - 1, as it doesn't store the factor – meaning, sizes for files over 4GB will always be horribly wrong by design.

Anyways, I don't think this relates to anything gzip specific.

lurch:
Okay, what about for bz2 images then? Will they end up with progress percentages > 100%

jhermsmeier:
@lurch this has no relation to any file format whatsoever, if the estimates are off because the file format's weird then there's nothing we can do about it. This just mends the logic behind the progress values, as previously the percentage would be calculated based on the file size (i.e. `bytesWritten / fileSize`), but of course the written bytes would exceed that value, since the uncompressed image will always be larger. Now it's `bytesWritten / uncompressedSize`, which should not exceed the 100% unless the estimate is too rough or inaccurate.

lurch:
Agreed. But the commit message was implying that this fixes the progress percentage to _never_ go above 100%, and I was just clarifying (as you've also done) that this isn't always the case ;-)
Might be worth clarifying that in the commit message?

And it does relate to #638 a little bit, see [this comment](https://github.com/resin-io/etcher/issues/638#issuecomment-284044760) and [this comment](https://github.com/resin-io/etcher/issues/638#issuecomment-285464517). (although I'm not familiar with that part of the code, so I dunno if those comments still apply - @jviotti ?)
And if my suggestion in [this comment](https://github.com/resin-io/etcher/issues/638#issuecomment-285410734) got implemented, then we'd indeed never end up with a progress greater than 100% ;-)

jviotti:
I'm still struggling to understand this fix. If you check the writer code (https://github.com/resin-io-modules/etcher-image-write/blob/master/lib/write.js#L113), you'll see that the pipe starts with the compressed image stream (https://github.com/resin-io-modules/etcher-image-write/blob/master/lib/write.js#L121), and applies the transform stream **after** the progress stream (https://github.com/resin-io-modules/etcher-image-write/blob/master/lib/write.js#L133).

This was done because we should still be able to calculate a correct progress bar even for images that we don't have a correct uncompressed size for (like bzip2).

With this change, we'd be calculating the percentage from the compressed stream using the uncompressed size, which wouldn't be correct.

Also, what happens if the final value is just an estimate?

jhermsmeier:
Hmm, I see – I was on a completely wrong path with this then.
Let's close this and I'll get to this issue while integrating the `blockmap` and `pipage` anyways (which is what I'm working on now).

PR #1261, opened 2017-04-06
refactor(GUI): remove angular from SettingsModel

Shou:
We remove the usage of Angular in SettingsModel in our efforts to make
Etcher as agnostic as possible as we make our move to React.

Changelog-Entry: Remove Angular usage from SettingsModel

### DIFF
--- a/lib/gui/components/update-notifier/controllers/update-notifier.js
+++ b/lib/gui/components/update-notifier/controllers/update-notifier.js
@@ -16,14 +16,16 @@
 
 'use strict';
 
-module.exports = function($uibModalInstance, SettingsModel, AnalyticsService, UPDATE_NOTIFIER_SLEEP_DAYS, options) {
+const settings = require('../../../models/settings');
+
+module.exports = function($uibModalInstance, AnalyticsService, UPDATE_NOTIFIER_SLEEP_DAYS, options) {
 
   // We update this value in this controller since its the only place
   // where we can be sure the modal was really presented to the user.
   // If the controller is instantiated, means the modal was shown.
   // Compare that to `UpdateNotifierService.notify()`, which could
   // have been called, but the modal could have failed to be shown.
-  SettingsModel.set('lastUpdateNotify', Date.now());
+  settings.set('lastUpdateNotify', Date.now());
 
   /**
    * @summary The number of days the update notified can be put to sleep
@@ -38,7 +40,7 @@ module.exports = function($uibModalInstance, SettingsModel, AnalyticsService, UP
    * @type {Object}
    * @public
    */
-  this.settings = SettingsModel;
+  this.settings = settings;
 
   /**
    * @summary Modal options
--- a/lib/gui/models/settings.js
+++ b/lib/gui/models/settings.js
@@ -20,64 +20,55 @@
  * @module Etcher.Models.Settings
  */
 
-const angular = require('angular');
 const Store = require('./store');
-const MODULE_NAME = 'Etcher.Models.Settings';
-const SettingsModel = angular.module(MODULE_NAME, []);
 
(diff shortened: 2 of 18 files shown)


Shou:
@jviotti thoughts on this one? If good, I could treat other modules in `lib/gui/models` to the same.

jviotti:
@Shou Amazing stuff! I'll leave a couple of comments but this is really in good track, and we should do it with the rest of the models. There is another challenge that we need to think before, though: the Redux store (and thus all the models which are just a front-end to the store) relies on localStorage to persist settings, so that won't run on the main process or the CLI, and therefore it can't be moved to `lib/shared`. I don't have a good solution yet, and its something to keep in mind, but lets apply similar changes to the other models as a first step, and we'll figure it out afterwards.

lib/gui/components/update-notifier/controllers/update-notifier.js
```diff
@@ -16,7 +16,9 @@
 
 'use strict';
 
-module.exports = function($uibModalInstance, SettingsModel, AnalyticsService, UPDATE_NOTIFIER_SLEEP_DAYS, options) {
+const SettingsModel = require('../../../models/settings');
```
jviotti: Can we name the object just `settings` now? The uppercase first letter and the `Model` suffix are Angular-ish conventions, so we should use node naming conventions now.

lib/gui/models/settings.js
```diff
@@ -20,12 +20,15 @@
...
-const SettingsModel = angular.module(MODULE_NAME, []);
 
-SettingsModel.service('SettingsModel', function() {
+/**
+ * @summary Setting/getting values
+ * @class
+ * @public
+ *
+ */
+class SettingsModel {
```
jviotti: This class is only exposing static functions, so I think we should simply attach these functions to `exports`, like in https://github.com/resin-io/etcher/pull/1251/files

lib/gui/modules/drive-scanner.js
```diff
@@ -26,13 +26,13 @@ const _ = require('lodash');
 const angular = require('angular');
 const EventEmitter = require('events').EventEmitter;
 const drivelist = require('drivelist');
+const SettingsModel = require('../models/settings');
 
 const MODULE_NAME = 'Etcher.Modules.DriveScanner';
 const driveScanner = angular.module(MODULE_NAME, [
```
jviotti: We should put the brackets on the same line now that it's empty:

```js
const driveScanner = angular.module(MODULE_NAME, []);
```

Shou:
This commit addresses the feedback

tests/gui/components/update-notifier.spec.js
```diff
@@ -16,19 +17,17 @@ describe('Browser: UpdateNotifier', function() {
...
         UpdateNotifierService = _UpdateNotifierService_;
-        SettingsModel = _SettingsModel_;
         UPDATE_NOTIFIER_SLEEP_DAYS = _UPDATE_NOTIFIER_SLEEP_DAYS_;
       }));
 
       describe('given the `sleepUpdateCheck` is disabled', function() {
 
         beforeEach(function() {
-          SettingsModel.set('sleepUpdateCheck', false);
+          settings.set('sleepUpdateCheck', false);
```
lurch: Does this use some kind of 'temporary' settings object, or does running the unit tests actually change your 'global' Etcher settings on your PC?

Shou: Good question, unless the unit tests mock localStorage it would modify the actual Etcher settings I believe. I'm unsure myself, any suggestions on a way forward here?

jviotti: It modifies the Electron Mocha settings, so it will not affect normal Etcher installations.

PR #1262, opened 2017-04-06
refactor: get rid of `is-elevated`

jviotti:
This commit implements our own version of the `is-elevated` module. The
reason behind this is that `is-elevated` relies on lots of unnecessary
modules, that eventually require `spawn-sync` and `try-thread-sleep`,
which conditionally require dependencies that are not even declared in
their `package.json`, causing issues when concatenating the Etcher CLI.

See: https://github.com/resin-io/etcher/pull/1235
See: https://github.com/resin-io/etcher/pull/1228

### DIFF
--- a/lib/shared/permissions.js
+++ b/lib/shared/permissions.js
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2016 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const os = require('os');
+const Bluebird = require('bluebird');
+const childProcess = Bluebird.promisifyAll(require('child_process'));
+const _ = require('lodash');
+
+/**
+ * @summary The user id of the UNIX "superuser"
+ * @constant
+ * @type {Number}
+ */
+const UNIX_SUPERUSER_USER_ID = 0;
+
+/**
+ * @summary Check if the current process is running with elevated permissions
+ * @function
+ * @public
+ *
+ * @description
+ * This function has been adapted from https://github.com/sindresorhus/is-elevated,
+ * which was originally licensed under MIT.
+ *
+ * We're not using such module directly given that is
+ * contains dependencies with dynamic undeclared dependencies,
+ * causing a mess when trying to concatenate the code.
+ *
+ * @fulfil {Boolean} - whether the current process has elevated permissions
+ * @returns {Promise}
+ *
+ * @example
+ * permissions.isElevated().then((isElevated) => {
+ *   if (isElevated) {
+ *     console.log('This process has elevated permissions');
+ *   }
+ * });
+ */
+exports.isElevated = () => {
+  if (process.platform === 'win32') {
+
+    // `fltmc` is available on WinPE, XP, Vista, 7, 8, and 10
+    // Works even when the "Server" service is disabled
+    // See http://stackoverflow.com/a/28268802
+    return childProcess.execAsync('fltmc')
+      .then(_.constant(true))
+      .catch({
+        code: os.constants.errno.EPERM
+      }, _.constant(false));
+
+  }
+
+  return Bluebird.resolve(process.geteuid() === UNIX_SUPERUSER_USER_ID);
+};
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2704,11 +2704,6 @@
       "resolved": "https://registry.npmjs.org/get-stdin/-/get-stdin-4.0.1.tgz",
       "dev": true
     },
-    "get-stream": {
-      "version": "3.0.0",
-      "from": "get-stream@>=3.0.0 <4.0.0",
-      "resolved": "https://registry.npmjs.org/get-stream/-/get-stream-3.0.0.tgz"
-    },
     "get-uri": {
       "version": "0.1.4",
       "from": "get-uri@>=0.1.0 <0.2.0",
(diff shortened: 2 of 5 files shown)


lib/shared/permissions.js
```diff
@@ -0,0 +1,64 @@
...
+ * @example
+ * permissions.isElevated().then((isElevated) => {
+ *   if (isElevated) {
+ *     console.log('This process has elevated permissions');
+ *   }
+ * });
+ */
+exports.isElevated = () => {
+  if (process.platform !== 'win32') {
+    return Bluebird.resolve(process.getuid() === UNIX_SUPERUSER_USER_ID);
```
lurch: Should this be checking `geteuid` rather than `getuid`?

lib/shared/permissions.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+const UNIX_SUPERUSER_USER_ID = 0;
+
+/**
+ * @summary Check if the current process is running with elevated permissions
+ * @function
+ * @public
+ *
+ * @description
+ * This function has been adapted from https://github.com/sindresorhus/is-elevated.
```
lurch: Might be worth noting that that code is MIT-licensed?

lib/shared/permissions.js
```diff
@@ -0,0 +1,64 @@
...
+    return Bluebird.resolve(process.getuid() === UNIX_SUPERUSER_USER_ID);
+  }
+
+  // `fltmc` is available on WinPE, XP, Vista, 7, 8, and 10
+  // Works even when the "Server" service is disabled
+  // See http://stackoverflow.com/a/28268802
+  return childProcess.execAsync('fltmc')
+    .then(_.constant(true))
+    .catch({
+      code: 1
```
lurch: Out of curiosity, how come the `1` here isn't tripping up the linter?

lib/shared/permissions.js
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @summary Check if the current process is running with elevated permissions
+ * @function
+ * @public
+ *
+ * @description
+ * This function has been adapted from https://github.com/sindresorhus/is-elevated.
+ * We're not using such module directly given that is
+ * contains dependencies with dynamic undeclared dependencies,
+ * causing a mess when trying to concatenate the code.
```
lurch: Might also be worth mentioning that our code is much smaller, and gets away with not needing these dependencies, due to the fact we're using a fixed NodeJS version?

lib/shared/permissions.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  // `fltmc` is available on WinPE, XP, Vista, 7, 8, and 10
+  // Works even when the "Server" service is disabled
+  // See http://stackoverflow.com/a/28268802
+  return childProcess.execAsync('fltmc')
+    .then(_.constant(true))
+    .catch({
+      code: 1
+    }, _.constant(false));
```
lurch: Given that Windows is the special-case here, it might be neater to do:
```js
exports.isElevated = () => {
    return childProcess.execAsync('fltmc') etc....
  }

  return Bluebird.resolve(process.getuid() === UNIX_SUPERUSER_USER_ID);
}
```

npm-shrinkwrap.json
```diff
@@ -6350,11 +6293,6 @@
       "from": "strip-bom@>=2.0.0 <3.0.0",
       "resolved": "https://registry.npmjs.org/strip-bom/-/strip-bom-2.0.0.tgz"
     },
-    "strip-eof": {
-      "version": "1.0.0",
-      "from": "strip-eof@>=1.0.0 <2.0.0",
-      "resolved": "https://registry.npmjs.org/strip-eof/-/strip-eof-1.0.0.tgz"
-    },
```
lurch: Nice to say goodbye to so many modules! :tada: :grinning:

lib/shared/permissions.js
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2016 resin.io
```
lurch: Something else to _maybe_ update, when we decide what we're doing with the Copyright years ;)

jviotti: Haha, I'll send a PR, I promise.

lib/shared/permissions.js
jviotti: Makes sense!

lib/shared/permissions.js
jviotti: Weird. I moved it to a constant anyway.

lib/shared/permissions.js
jviotti: Indeed, good catch. Fixed!

lib/shared/permissions.js
jviotti: I don't think a modern version is the main reason we can get rid of this modules. Most of the stuff introduced here is non-sense stuff, like modules that promisify other modules, etc.

lib/shared/permissions.js
jviotti: Yeah, I added a `@license` tag to the function.

lib/shared/permissions.js
lurch: Nah, I wasn't suggesting applying a specific license to this function, I was just suggesting noting that the "function we've adapted" was originally licensed under MIT, which gives us freedom to take it. modify it, and re-license it under Apache2 ;)

lib/shared/permissions.js
lurch: Or you could just edit this file here, so that it doesn't need to be changed in future? ;)

lib/shared/permissions.js
```diff
@@ -0,0 +1,68 @@
...
+ * @example
+ * permissions.isElevated().then((isElevated) => {
+ *   if (isElevated) {
+ *     console.log('This process has elevated permissions');
+ *   }
+ * });
+ */
+exports.isElevated = () => {
+  if (process.platform === 'win32') {
+    const FLTMC_PERMISSIONS_EXIT_CODE = 1;
```
jviotti: My intention was to signify that an exit code 1 represents that we don't have enough permissions to do the task. Other types of failures may set different exit codes. What do you think?

lib/shared/permissions.js
jviotti: Oh, I see. Done!

lib/shared/permissions.js
jhermsmeier: What do you think about using `os.constants.errno` instead? That should use the proper error codes on the respective operating systems.
```js
const errno = require('os').constants.errno;
// ...
.catch({ code: errno.EPERM }, ... );
```

jviotti: Cool, `errno.EPERM` seems to correctly be 1.

lib/shared/permissions.js
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @summary Check if the current process is running with elevated permissions
+ * @function
+ * @public
+ *
+ * @description
+ * This function has been adapted from https://github.com/sindresorhus/is-elevated,
+ * which was originally licensed under MIT.
+ *
+ * We're not using such module directly given that is
```
lurch: "given that is" -> "given that it" ;)

jviotti: Fixed by https://github.com/resin-io/etcher/pull/1272 :D

PR #1263, opened 2017-04-06
fix(GUI): wrap Math.min on ProgressButton controller

Shou:
We replace the erroneous direct Math.min usage in ProgressButton's
template with a wrapped version on the controller.

Changelog-Entry: Wrap Math.min on ProgressButton's controller.

### DIFF
--- a/lib/gui/components/progress-button/templates/progress-button.tpl.html
+++ b/lib/gui/components/progress-button/templates/progress-button.tpl.html
@@ -3,5 +3,5 @@
     'progress-button--striped': striped && striped != 'false'
   }">
     <span class="progress-button__content" ng-transclude></span>
-    <span class="progress-button__bar" ng-style="{ width: Math.min(100, percentage) + '%' }"></span>
+    <span class="progress-button__bar" ng-style="{ width: (percentage > 100 ? 100 : percentage) + '%' }"></span>
 </button>


lib/gui/components/progress-button/controllers/progress-button.js
```diff
@@ -0,0 +1,23 @@
+
```
lurch: This is throwing a linting error, killing all the CI builds.

Shou:
All should be well here now

lib/gui/components/progress-button/templates/progress-button.tpl.html
```diff
@@ -3,5 +3,5 @@
     'progress-button--striped': striped && striped != 'false'
   }">
     <span class="progress-button__content" ng-transclude></span>
-    <span class="progress-button__bar" ng-style="{ width: Math.min(100, percentage) + '%' }"></span>
+    <span class="progress-button__bar" ng-style="{ width: progress.min(100, percentage) + '%' }"></span>
```
jviotti: Maybe worth just replicating `Math.min` with a ternary conditional operator, and avoid the controller altogether?

```js
percentage > 100 ? 100 : percentage
```

jviotti: I think it might need to be enclosed in parenthesis, otherwise `%` will be added to the else condition.

Shou:
Using ternary operator instead now, removed the controller.

lurch:
Does the commit message need updating?

jviotti:
I updated the commit message myself right before merging :)

Shou:
Awesome, thanks

PR #1264, opened 2017-04-07
refactor(GUI): remove Angular from FlashStateModel and rename

Shou:
We remove the usage of Angular from 'FlashStateModel' and its usage
throughout the project comes under the new 'flashState' moniker.

Changelog-Entry: Remove the usage of Angular from 'FlashStateModel' and
rename its usage to 'flashState'.
Depends: #1261

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -31,6 +31,7 @@ const Bluebird = require('bluebird');
 const EXIT_CODES = require('../shared/exit-codes');
 const messages = require('../shared/messages');
 const packageJSON = require('../../package.json');
+const flashState = require('./models/flash-state');
 
 const Store = require('./models/store');
 
@@ -46,7 +47,6 @@ const app = angular.module('Etcher', [
 
   // Models
   require('./models/selection-state'),
-  require('./models/flash-state'),
   require('./models/drives'),
 
   // Components
@@ -129,9 +129,9 @@ app.run((AnalyticsService, ErrorService, UpdateNotifierService, SelectionStateMo
 
 });
 
-app.run((AnalyticsService, OSWindowProgressService, FlashStateModel) => {
+app.run((AnalyticsService, OSWindowProgressService) => {
   Store.subscribe(() => {
-    const flashState = FlashStateModel.getFlashState();
+    const currentFlashState = flashState.getFlashState();
 
     // There is usually a short time period between the `isFlashing()`
     // property being set, and the flashing actually starting, which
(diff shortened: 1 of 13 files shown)


jviotti:
Amazing stuff :D Can we rebase master before reviewing though (so I don't see the settings diff as well)?

lib/gui/models/flash-state.js
```diff
@@ -20,212 +20,203 @@
...
+ */
+exports.setProgressState = (state) => {
+  Store.dispatch({
+    type: Store.Actions.SET_FLASH_STATE,
+    data: {
+      type: state.type,
+      percentage: state.percentage,
+      eta: state.eta,
+
+      speed: _.attempt(() => {
```
lurch: I know I've mentioned this before, so apologies for sounding like a broken record, but my gut feeling is still that this conversion from bytes to Megabytes 'belongs' in the GUI itself, rather than in the internal model.
(but there may well be other aspects to this that I'm not aware of)

Shou:
@jviotti rebase done!

lib/gui/models/flash-state.js
jviotti: I agree. Maybe @Shou wants to handle it?

lib/gui/app.js
```diff
@@ -140,17 +140,17 @@ app.run((AnalyticsService, OSWindowProgressService, FlashStateModel) => {
     //
     // We use the presence of `.eta` to determine that the actual
     // writing started.
-    if (!FlashStateModel.isFlashing() || !flashState.eta) {
+    if (!flashState.isFlashing() || !flashState.eta) {
```
jviotti: Shoudn't `flashState.eta` be `currentFlashState.eta`?

lib/gui/models/flash-state.js
jviotti: (as another PR of course)

lib/gui/app.js
Shou: Good catch

lurch: I'm confused - you spotted a mistake but then merged anyway? :-/

jviotti: Oh damn, I incorrectly thought it was fixed already :/ I'll send a quick fix.

jviotti: See https://github.com/resin-io/etcher/pull/1273. Sorry about that, I got confused by the `flashState`/`currentFlashState` naming and assumed that `flashState` was the desired state.

PR #1268, opened 2017-04-07
test: Use mocha with promises instead of callbacks

jhermsmeier:
This changes the test cases to just return the Promises, to avoid timing out on failures and to provide better error messages and stack traces.

Change-Type: patch


jviotti:
@jhermsmeier Looks like there are some conflicts here :(

PR #1269, opened 2017-04-07
chore(ci): Limit git clone depth

jhermsmeier:
This may speed up cloning on CI services by reducing the clone depth.

Change-Type: patch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ sudo: false
 node_js:
   - "6.1.0"
 
+git:
+  depth: 3
+
 services:
   - docker
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,6 +19,8 @@ environment:
     - TARGET_ARCH: x64
     - TARGET_ARCH: x86
 
+clone_depth: 3
+
 matrix:
   fast_finish: true
 


jviotti:
Any reason why not doing depth 1?

jhermsmeier:
Yeah, from https://docs.travis-ci.com/user/customizing-the-build/#Git-Clone-Depth:

> Please note that if you use a depth of 1 and have a queue of jobs, Travis CI won’t build commits that are in the queue when you push a new commit.

And I figured we probably want to avoid that.

lurch:
I wonder if that means that (effectively) there can only be as many jobs in the Travis build queue as the selected clone depth?
Given that Travis CI is often quite slow, and we often have many jobs queued up, I guess time will tell if a depth of 3 is 'enough' ;-)

jviotti:
> I wonder if that means that (effectively) there can only be as many jobs in the Travis build queue as the selected clone depth?

Oh, is that indeed the case @jhermsmeier? As far as I can see, Travis CI clones the main branch of the repo (which after this change means the last 3 commits), but then calls `git fetch` on the branch like this:

```sh
git fetch origin +refs/pull/1272/merge:
```

Which means that the branch will always be fetched from the internet, and therefore the depth here doesn't apply.

Am I missing something?

jhermsmeier:
Honestly, I don't really know, but from what I understood `depth: 1` is a special case, so build behavior shouldn't change.

jviotti:
@lurch was right. This is causing failures when the Travis CI backlog is greater than 3. See:

- https://travis-ci.org/resin-io/etcher/jobs/220811455
- https://travis-ci.org/resin-io/etcher/jobs/220811463
- https://travis-ci.org/resin-io/etcher/jobs/220802663

I reverted this commit for now.

lurch:
> @lurch was right.

Hah, lucky guess ;-)

jhermsmeier:
Oh boy, sorry about that :/ Lesson learned.

PR #1273, opened 2017-04-10
fix(GUI): correctly reference the `eta` property in `lib/gui/app.js`

jviotti:
See: https://github.com/resin-io/etcher/pull/1264/files#r110662965


jviotti:
Travis OS X builds seem to be hanging without any apparent reason. Restarting various times eventually makes it work, but I can manage to get one latest OS X build done. I'll merge this anyway since its blocking the rc1 release, and we can investigate afterwards.

PR #1275, opened 2017-04-10
chore: remove duplicated curl dependency in Dockerfiles

jviotti:


lurch:
Nice catch! Just for reference, this "double curl" was introduced by #1078

PR #1276, opened 2017-04-11
chore: fix jq crashing on Windows when using relative paths

jviotti:
`jq` crashes when setting a value on a relative file. We workaround the
issue by using `jq` in a pipe instead of passing the file to it as an
argument.

See: https://github.com/stedolan/jq/issues/1155


lurch:
Yep, see also https://github.com/resin-io/etcher/blob/master/scripts/ci/ensure-all-node-requirements-available.sh#L31

PR #1277, opened 2017-04-11
fix(image-stream): interpret iso9660 as application/octet-stream

jviotti:
This is a regression caused by
https://github.com/resin-io/etcher/pull/1257, which introduced a new way
to detect mime types by using the `mime-types` module.

This module, contrary to `file-type`, will detect certain ISO files as
`application/x-iso9660-image`, which Etcher doesn't know how to handle,
and will therefore should at the user that the format is unsupported.

Change-Type: patch
Changelog-Entry: Don't interpret certain ISO images as unsupported.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -27,6 +27,7 @@ Makefile text
 *.icns binary
 *.ico binary
 *.img binary
+*.iso binary
 *.png binary
 *.xz binary
 *.zip binary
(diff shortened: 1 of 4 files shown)


lurch:
Whoops! Nice catch @jviotti 

Are there now unit-tests for _all_ the file types that Etcher claims to support?

jviotti:
Yeah, looks like we're covered now!

lurch:
Great :+1:

PR #1279, opened 2017-04-11
fix(image-stream): throw user error if image is a directory

jviotti:
If the user tries to drag and drop a directory to the application, then
he'll get a scary `EISDIR` error message. This commit catches this
error, and display a nice user friendly message instead.

Change-Type: patch
Changelog-Entry: Prevent uncaught `EISDIR` when dropping a directory to the application.

### DIFF
--- a/tests/image-stream/directory.spec.js
+++ b/tests/image-stream/directory.spec.js
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2016 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


tests/image-stream/directory.spec.js
```diff
@@ -0,0 +1,62 @@
...
+const m = require('mochainon');
+const path = require('path');
+const DATA_PATH = path.join(__dirname, 'data');
+const IMAGES_PATH = path.join(DATA_PATH, 'images');
+const errors = require('../../lib/shared/errors');
+const imageStream = require('../../lib/image-stream/index');
+
+describe('ImageStream: Directory', function() {
+
+  this.timeout(20000);
```
lurch: Probably won't take 20 seconds to determine if a path is a directory or not ;-)

jviotti: Good catch. See https://github.com/resin-io/etcher/pull/1285

PR #1283, opened 2017-04-11
fix(image-stream): Fix Apple disk image detection & reading

jhermsmeier:
This fixes two things: The format detection, and a bug in `udif`.

First, by categorizing the `.dmg` extension as compressed image,
`.isSupportedImage()` would attempt to detect the format after stripping
the extension, causing it to be misdetected.

Second, `udif`'s ReadStream didn't add the `dataForkOffset` to its
position when reading blocks, causing the wrong data to be read for some images,
in turn causing zlib to error on invalid headers.

Changes:
- Classify `.dmg` as `type: 'image'`
- Update `udif` to 0.8.0

Change-Type: patch
Changelog-Entry: Fix Apple disk image detection & streaming

### DIFF
--- a/lib/image-stream/README.md
+++ b/lib/image-stream/README.md
@@ -58,4 +58,19 @@ These are the rules for handling archive images:
 
 The module throws an error if the above rules are not met.
 
+Supported Formats
+-----------------
+
+There are currently three image types in supported formats: `image`, `compressed` and `archive`.
+
+An extension tagged `image` describes a format which can be directly written to a device by its handler,
+and an extension tagged `archive` denotes an archive containing an image, and will cause an archive handler
+to open the archive and search for an image file.
+
+Note that when marking an extension as `compressed`, the filename will be stripped of that extension,
+and the leftover extension examined to determine the uncompressed image format (i.e. `.img.gz -> .img`).
+
+As an archive (such as `.tar`) might be additionally compressed, this will allow for constructs such as
+`.tar.gz` (a compressed archive, containing a file with an extension tagged as `image`) to be handled correctly.
+
 [etcher-image-write]: https://github.com/resin-io-modules/etcher-image-write
--- a/lib/image-stream/supported.js
+++ b/lib/image-stream/supported.js
@@ -16,6 +16,14 @@
 
 'use strict';
 
+/**
+ * @summary Supported filename extensions
+ * @description
+ * NOTE: Extensions with type: 'compressed' will be stripped
+ * from filenames to determine the format of the uncompressed image.
+ * For details, see lib/image-stream/README.md
+ * @const {Array}
+ */
 module.exports = [
   {
     extension: 'zip',
@@ -38,10 +46,6 @@ module.exports = [
     type: 'compressed'
   },
   {
-    extension: 'dmg',
-    type: 'compressed'
-  },
-  {
     extension: 'img',
     type: 'image'
   },
@@ -59,6 +63,10 @@ module.exports = [
   },
   {
     extension: 'raw',
+    type: 'image'
+  },
+  {
+    extension: 'dmg',
     type: 'image'
   }
 ];
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -6772,9 +6772,9 @@
       "dev": true
     },
     "udif": {
-      "version": "0.7.0",
+      "version": "0.8.0",
       "from": "udif@latest",
-      "resolved": "https://registry.npmjs.org/udif/-/udif-0.7.0.tgz",
+      "resolved": "https://registry.npmjs.org/udif/-/udif-0.8.0.tgz",
       "dependencies": {
         "base64-js": {
           "version": "1.1.2",
--- a/tests/shared/supported-formats.spec.js
+++ b/tests/shared/supported-formats.spec.js
@@ -26,7 +26,7 @@ describe('Shared: SupportedFormats', function() {
 
     it('should return the supported compressed extensions', function() {
       const extensions = supportedFormats.getCompressedExtensions();
-      m.chai.expect(extensions).to.deep.equal([ 'gz', 'bz2', 'xz', 'dmg' ]);
+      m.chai.expect(extensions).to.deep.equal([ 'gz', 'bz2', 'xz' ]);
     });
 
   });
@@ -35,7 +35,7 @@ describe('Shared: SupportedFormats', function() {
 
     it('should return the supported non compressed extensions', function() {
       const extensions = supportedFormats.getNonCompressedExtensions();
-      m.chai.expect(extensions).to.deep.equal([ 'img', 'iso', 'dsk', 'hddimg', 'raw' ]);
+      m.chai.expect(extensions).to.deep.equal([ 'img', 'iso', 'dsk', 'hddimg', 'raw', 'dmg' ]);
     });
 
   });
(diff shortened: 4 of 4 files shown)


jhermsmeier:
Tested it with a MacOS Sierra image, which features a non-zero `dataForkOffset`:

screen shot 2017-04-11 at 16 12 37

lurch:
> First, by categorizing the .dmg extension as compressed image,
.isSupportedImage() would attempt to detect the format after stripping
the extension, causing it to be misdetected.

Hmmm, should that have been picked up by a unit-test? If not, could you please add a unit-test to prevent it again in future? :)

Maybe you could also add some notes into `lib/image-stream/supported.js` about the impact of the different `type` attributes?

jhermsmeier:
> Maybe you could also add some notes into lib/image-stream/supported.js about the impact of the different type attributes?

Good idea! I've added notes to both the `README` and `supported.js`

lib/image-stream/README.md
```diff
@@ -58,4 +58,11 @@ These are the rules for handling archive images:
 
 The module throws an error if the above rules are not met.
 
+Supported Formats
+-----------------
+
+There are currently three image types in supported formats: `image`, `compressed` and `archive`.
+Note that when marking an extension as `compressed`, the filename will be stripped of that extension,
+and the leftover extension examined to determine the compressed image format (i.e. `.img.gz -> .img`).
```
lurch: I guess this is a potential pitfall of identifying mime-types purely by file extension ;-)

lurch: Would be nice to also explain the differences between `image` and `archive`.

lurch:
I still think an extra unit-test would be nice too.

lib/image-stream/supported.js
```diff
@@ -16,6 +16,13 @@
 
 'use strict';
 
+/**
+ * Supported filename extensions
```
jviotti: Can we use `@summary` for the title, and `@description` for the long comment? For example:

```js
/**
 * @summary Supported filename extensions
 * @description
 * NOTE: Extensions with type: 'compressed' will be stripped
 * ...
 */
```

lib/image-stream/README.md
jviotti: Haha, yeah, that'd be a compressed archive containing an image.

lib/image-stream/supported.js
jhermsmeier: Yup, mended!

lib/image-stream/README.md
jhermsmeier: Updated with added notes on all three types

jhermsmeier:
> I still think an extra unit-test would be nice too.

I'm not sure how we could test against this – or what exactly we should test here. Any ideas?

lurch:
> I'm not sure how we could test against this – or what exactly we should test here. Any ideas?

Okay, let me rephrase the question then...

_Why_ wasn't
> First, by categorizing the .dmg extension as compressed image,
.isSupportedImage() would attempt to detect the format after stripping
the extension, causing it to be misdetected.

picked up by any unit-tests?

jhermsmeier:
Ah yes, because there are two places we check for image support – one in image-stream, and another in the GUI (Etcher.Models.SupportedFormats).

Added tests for all supported formats in `gui/models/supported-formats.spec` now, but that still won't save us from making that logical error in this setup.

lib/image-stream/README.md
```diff
@@ -58,4 +58,19 @@ These are the rules for handling archive images:
...
+
+An extension tagged `image` describes a format which can be directly written to a device by its handler,
+and an extension tagged `archive` denotes an archive containing an image, and will cause an archive handler
+to open the archive and search for an image file.
+
+As an archive (such as `.tar`) might be additionally compressed, this will allow for constructs such as
+`.tar.gz` (a compressed archive, containing a file with an extension tagged as `image`) to be handled correctly.
+
+Note that when marking an extension as `compressed`, the filename will be stripped of that extension,
+and the leftover extension examined to determine the compressed image format (i.e. `.img.gz -> .img`).
```
lurch: Might read better with this sentence moved before the previous one?

lib/image-stream/supported.js
```diff
@@ -16,6 +16,14 @@
 
 'use strict';
 
+/**
+ * @summary Supported filename extensions
+ * @description
+ * NOTE: Extensions with type: 'compressed' will be stripped
+ * from filenames to determine the format of the compressed image.
```
lurch: Perhaps change "of the compressed image." to "of the uncompressed image." ?

tests/gui/models/supported-formats.spec.js
```diff
@@ -61,6 +62,17 @@ describe('Browser: SupportedFormats', function() {
 
     describe('.isSupportedImage()', function() {
 
+      _.forEach(imageStream.supportedFileTypes, (fileType) => {
```
lurch: To match up with the style of the other unit tests, I think this should be explicitly spelled out, rather than just imported from `imageStream` ?  @jviotti to confirm :)

(and would that also prevent the "but that still won't save us from making that logical error in this setup" you refer to?)

lib/image-stream/README.md
jhermsmeier: Done

lib/image-stream/supported.js
jhermsmeier: Fixed

tests/gui/models/supported-formats.spec.js
jhermsmeier: Done. I guess it'll at least be another barrier at which one could notice.

lib/image-stream/README.md
```diff
@@ -58,4 +58,19 @@ These are the rules for handling archive images:
...
 The module throws an error if the above rules are not met.
 
+Supported Formats
+-----------------
+
+There are currently three image types in supported formats: `image`, `compressed` and `archive`.
+
+An extension tagged `image` describes a format which can be directly written to a device by its handler,
+and an extension tagged `archive` denotes an archive containing an image, and will cause an archive handler
+to open the archive and search for an image file.
```
lurch: Ridiculous-hypothetical-question: If a compressed image was inside a zip, which was inside another zip, would that transparently "just work"? ;-)

lib/image-stream/README.md
```diff
@@ -58,4 +58,19 @@ These are the rules for handling archive images:
...
+
+Note that when marking an extension as `compressed`, the filename will be stripped of that extension,
+and the leftover extension examined to determine the compressed image format (i.e. `.img.gz -> .img`).
```
lurch: As with my earlier comment, I guess this could be "to determine the uncompressed image format" ?

tests/gui/models/supported-formats.spec.js
```diff
@@ -61,6 +61,32 @@ describe('Browser: SupportedFormats', function() {
...
+        'path/to/filename.img.bz2',
+        'path/to/filename.img.xz',
+
+        // Type: 'image'
+        'path/to/filename.img',
+        'path/to/filename.iso',
+        'path/to/filename.dsk',
+        'path/to/filename.hddimg',
+        'path/to/filename.raw',
+        'path/to/filename.dmg'
```
lurch: Would it be worth it / possible to check that all these filenames resolve to the correct expected mime-type?
(or is that already covered elsewhere?)

P.S. Looking at https://github.com/resin-io/etcher/pull/1283/files#diff-dd8adcb7744d5236c18272b3e227071eR144 I can now see why the problem that this PR fixes wasn't picked up by the unit-tests - they only check the _first_ of each supported extension type, so of course `.dmg` didn't get tested because it was on the end of the list.

tests/gui/models/supported-formats.spec.js
```diff
@@ -61,6 +61,32 @@ describe('Browser: SupportedFormats', function() {
...
+
+      ], (filename) => {
+        it(`should return true for ${filename}`, function() {
+          const isSupported = SupportedFormatsModel.isSupportedImage(filename);
```
lurch: :+1:

tests/gui/models/supported-formats.spec.js
jhermsmeier: That's already covered in the image-stream specs

lib/image-stream/README.md
jhermsmeier: Nope, because Etcher only checks one level deep in archives, and yells if it can't find a suitable image

lib/image-stream/README.md
jhermsmeier: Mended

jviotti:
I approved, but should we add a `Changelog-Entry` footer tag describing the fix?

jhermsmeier:
@jviotti Added

PR #1286, opened 2017-04-11
upgrade: `etcher-image-write` to v9.0.2

jviotti:
This release fixes an uncaught error when there is a validation error.

See: https://github.com/resin-io-modules/etcher-image-write/pull/94
Change-Type: patch
Changelog-Entry: Fix "function createError(opts) {}" error on validation failure.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2193,9 +2193,9 @@
       "dev": true
     },
     "etcher-image-write": {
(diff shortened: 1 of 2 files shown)


jviotti:
@lurch The user literally sees a "function createError(opts) {}" error, so I guess it will make a lot of sense to the people who encountered. What do you think?

lurch:
Fair enough, didn't realise that.

PR #1289, opened 2017-04-12
chore(package): Fix codespell binary file warnings

jhermsmeier:
This avoids codespell printing warnings for binary files
which have been added recently (`.iso`, `.dmg`), as well as dotfiles (i.e. `.DS_Store`).

Change-Type: patch


lurch:
Good catch - if was feeling adventurous we could maybe get this info from `.gitattributes` ;-)
I'm :-1: on ignoring _all_ dotfiles, as there's a number of those in the repo. Could you add specific exclusions for `.DS_Store`?

jhermsmeier:
Updated with specific `.DS_Store` selector

PR #1290, opened 2017-04-12
Fix debian package run script

dlech:
The rc2 debian package is not working because the path where the package files
is installed has changed. The change in paths comes from an upstream change
in the electron-installer-debian node package. So, updating the path here to
match the actuall install location.

### DIFF
--- a/scripts/build/debian/etcher-electron.sh
+++ b/scripts/build/debian/etcher-electron.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 export ETCHER_DISABLE_UPDATES=1
-/usr/share/etcher-electron/etcher "$@"
+/usr/lib/etcher-electron/etcher "$@"


lurch:
Hmm, curious.
@jviotti Maybe this is somewhere else where we need to lock down the package version? :-(   https://github.com/resin-io/etcher/search?utf8=%E2%9C%93&q=electron-installer-debian&type=

@dlech Out of curiosity, do you know why the path was changed?

dlech:
It's all my fault.

https://github.com/unindented/electron-installer-debian/pull/47

jviotti:
Oh, great catch @dlech! I'm on mobile, but I'll merge as soon as I get home, and will re-publish the debian packages!

jviotti:
I re-uploaded v1.0.0-rc.1 and v1.0.0-rc.2 with the fix.

jviotti:
@dlech Looks like this change is actually breaking v1.0.0-rc.3. `/usr/share/etcher-electron` seems to be the correct path, at least on my Ubuntu 16.04 system.

jviotti:
Never mind, I was running an old version of `electron-installer-debian` on my linux box.

lurch:
> I was running an old version of electron-installer-debian on my linux box.

@jviotti IMHO we should either make `electron-installer-debian` a locally-installed NPM package ( https://github.com/unindented/electron-installer-debian#programmatically ) so that we can lock down the version, rather than installing it globally; or we should modify `scripts/build/electron-installer-debian-linux.sh` to check the output of `electron-installer-debian --version`.

jviotti:
Yeah, makes sense. I'll bring it in as a development dependency.

PR #1295, opened 2017-04-12
feat(child-writer): log non-robot messages for debugging purposes

jviotti:
Currently, if the child writer receives a message from the writer
process that is not a valid robot object, then it will throw an error.

Now, we check if the message is a robot message, and if so, we try to
parse it (throwing errors if its indeed a malformed/incomplete robot
message), however we log it if not.

The main motivation behind this feature is that it will allows us to
print debugging information on the mountutils module, and have it
redirected to DevTools.

Change-Type: minor

### DIFF
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -126,8 +126,28 @@ exports.write = (image, drive, options) => {
    * }));
    */
   const bridgeRobotMessage = (message) => {
+    const ROBOT_COMMAND_LOG = 'log';
+
+    const parsedMessage = _.attempt(() => {
+      if (robot.isMessage(message)) {
+        return robot.parseMessage(message);
+      }
+
+      // Don't be so strict. If a message doesn't look like
+      // a robot message, then make the child writer log it
+      // for debugging purposes.
+      return robot.parseMessage(robot.buildMessage(ROBOT_COMMAND_LOG, {
+        message
+      }));
+
+    });
+
+    if (_.isError(parsedMessage)) {
+      emitError(parsedMessage);
+      return;
+    }
+
     try {
-      const parsedMessage = robot.parseMessage(message);
 
       // These are lighweight accessor methods for
       // the properties of the parsed message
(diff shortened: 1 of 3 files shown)


lib/child-writer/index.js
```diff
@@ -126,8 +126,22 @@ exports.write = (image, drive, options) => {
    * }));
    */
   const bridgeRobotMessage = (message) => {
+    const ROBOT_COMMAND_LOG = 'log';
+
     try {
-      const parsedMessage = robot.parseMessage(message);
+      const parsedMessage = _.attempt(() => {
```
jhermsmeier: Since this is wrapped in an `_.attempt()`, I think this can go outside of the try-catch-block

tests/shared/robot.spec.js
```diff
@@ -91,6 +91,97 @@ describe('Shared: Robot', function() {
...
+      m.chai.expect(robot.isMessage(robot.buildMessage('foo', {
+        message: 'bar'
+      }))).to.be.true;
+    });
+
+    it('should return true for a message built with .buildErrorMessage()', function() {
+      const error = new Error('foo');
+      m.chai.expect(robot.isMessage(robot.buildErrorMessage(error))).to.be.true;
+    });
+
```
jhermsmeier: As strings can be JSON encoded as well (`JSON.stringify('asd')` ->` '"asd"'`), it might be good to add that case as well

lib/child-writer/index.js
jviotti: The point of this whole try/catch block is to emit the error as an event instead of just throwing. Even though we used `_.attempt`, we still want to emit any error that could come from it.

tests/shared/robot.spec.js
jviotti: Good catch, fixed!

lib/child-writer/index.js
jhermsmeier: I see, but doesn't it *return* the error instead of throwing? If so, wouldn't that mean that we'd have to check `if(_.isError(parsedMessage))` and then robot out the error, or am I missing something?

jviotti: Oh, true. I completely misunderstood the comment. Fixed!

lib/child-writer/index.js
```diff
@@ -126,8 +126,28 @@ exports.write = (image, drive, options) => {
    * }));
    */
   const bridgeRobotMessage = (message) => {
+    const ROBOT_COMMAND_LOG = 'log';
```
lurch: Hmmm, why add a constant for `ROBOT_COMMAND_LOG` but not also a constant for e.g. `ROBOT_COMMAND_ERROR` ?

jviotti: Only the log string is used multiple times, which was the reason it was extracted as a constant. I'll add the error one just for symmetry purposes.

PR #1299, opened 2017-04-14
upgrade: resin-corvus to 1.0.0-beta.25

stefan-mihaila:
This version handles better non-error objects sent to Sentry.

Change-Type: patch
Changelog-Entry: Upgrade `resin-corvus` to 1.0.0-beta.25.


jviotti:
Awesome work @stefan-mihaila, thanks for the quick PR!

PR #1303, opened 2017-04-14
fix(GUI): correctly log debugging messages on Windows

jviotti:
Windows will not react to dynamically set environment variables (e.g:
`process.env.MOUNTUTILS_DEBUG = '1'`), so in order to workaround this
limitation, we set `MOUNTUTILS_DEBUG` in
`lib/child-writer/writer-proxy.js` when spawning the CLI, and we also
set `DEBUG='*'` in `lib/start.js`, so it gets picked up by the child
writer module, which spawns the writer proxy by inheriting the
environment.

### DIFF
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -200,7 +200,12 @@ return permissions.isElevated().then((elevated) => {
             PATH: process.env.PATH,
 
             ELECTRON_RUN_AS_NODE: 1,
-            ETCHER_CLI_ROBOT: 1
+            ETCHER_CLI_ROBOT: 1,
+
+            // Enable extra logging from mountutils
+            // See https://github.com/resin-io-modules/mountutils
+            MOUNTUTILS_DEBUG: 1
+
           }
         });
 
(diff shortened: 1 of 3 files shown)


lurch:
Apologies if this is a dumb question, but if `process.env.MOUNTUTILS_DEBUG = true;` doesn't work, then how come `process.env.DEBUG = '*';` does work?
Or is this to do with processes vs sub-processes?

jviotti:
The environment variable trick doesn't work on Windows native modules, but works everywhere else. The `DEBUG` variable is not used by native modules, so we don't have to worry about that one.

lurch:
Apologies for another dumb question, but what's the difference between a "Windows native module" and other kinds of modules?

Ohh, I see now (I think) - it's actually the C++ mountutils code that wasn't picking up the env-var, but the `DEBUG` env-var is only used by pure JS code?

jviotti:
The issue is that dynamically setting an environment variable using `process.env.VARIABLE = 'value'` is detected on the fly by JS code and native modules on GNU/Linux and macOS, but not on Windows.

So if you have a C++ Windows Node add-on that relies on variable `FOO` and we call `process.env.FOO = 'value'` right before calling the native function, then the Windows native module will not pick it up for some reason (Windows apparently requires env vars to be set at startup).

PR #1304, opened 2017-04-14
upgrade: `elevator` to v2.2.2

jviotti:
This version prints debug messages to `stdout`.

See: https://github.com/resin-io-modules/elevator/pull/11

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     }
   },
   "optionalDependencies": {
-    "elevator": "^2.2.1"
+    "elevator": "^2.2.2"
   },
   "dependencies": {
     "angular": "1.6.3",


lurch:
Shouldn't this have bumped the shrinkwrap too?

jviotti:
This is an optional dependency, which is not tracked by shrinkwrap.

lurch:
Ahhh. Does that mean we'll always get the latest version installed anyway?

jviotti:
Yeah, that's right, but there is not much we can do about that, given that NPM shrinkwrap has some known limitations with regarding to optional dependencies. Maybe we should at least lock the version in package.json?

jviotti:
@lurch Maybe this is something to add to the integrity checks (e.g: optional dependencies can't have ranges)?

PR #1318, opened 2017-04-18
fix(GUI): catch EIO errors and display a user friendly message

jviotti:
If we get EIO at this point, then it means that the writer did
everything it could to recover (like multiple retries), and the error
is truly an input/output error coming from the operating system.

In this commit, we show a nice user friendly message explaining what
happened, and advising users to try again with another drive, reader, or
port instead of showing an uncaught EIO error.

Change-Type: patch
Changelog-Entry: Show a friendly user message on EIO after many retries.

### DIFF
--- a/lib/gui/pages/main/controllers/flash.js
+++ b/lib/gui/pages/main/controllers/flash.js
@@ -93,6 +93,12 @@ module.exports = function(
           image,
           drive
         });
+      } else if (error.code === 'EIO') {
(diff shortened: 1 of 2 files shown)


lurch:
advicing -> advising ;-)

jviotti:
Fixed!

lurch:
Fixed in the PR comment, but not in the commit comment ;-)

lurch:
LOL, sneaky :)

PR #1319, opened 2017-04-18
chore: perform GNU/Linux builds on Debian Jessie docker containers

jviotti:
We recently hit an issue where Etcher builds produced on Ubuntu relied
on a too new glibc version, making them incompatible with GNU/Linux
distributions such as Debian Jessie.

As a solution, we will start producing builds on Debian Jessie, which
ensures that the builds will be compatible with the majority of
GNU/Linux versions out there.

### DIFF
--- a/scripts/build/docker/run-command.sh
+++ b/scripts/build/docker/run-command.sh
@@ -89,6 +89,7 @@ done
 # The `-t` and TERM setup is needed to display coloured output.
 docker run -t \
   --env "TERM=xterm-256color" \
+  --env "TARGET_ARCH=$ARGV_ARCHITECTURE" \
   ${DOCKER_ENVVARS[@]+"${DOCKER_ENVVARS[@]}"} \
   --cap-add SYS_ADMIN \
   --device /dev/fuse:/dev/fuse:mrw \
--- a/scripts/build/docker/Dockerfile-i686
+++ b/scripts/build/docker/Dockerfile-i686
@@ -1,4 +1,4 @@
-FROM toopher/ubuntu-i386:14.04
+FROM 32bit/debian:jessie
 
 # Install dependencies
 RUN apt-get update && apt-get install -y \
--- a/scripts/build/docker/Dockerfile-x86_64
+++ b/scripts/build/docker/Dockerfile-x86_64
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM debian:jessie
 
 # Install dependencies
 RUN apt-get update && apt-get install -y \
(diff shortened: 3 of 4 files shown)


scripts/build/docker/run-command.sh
```diff
@@ -89,6 +89,7 @@ done
 # The `-t` and TERM setup is needed to display coloured output.
 docker run -t \
   --env "TERM=xterm-256color" \
+  --env "TARGET_ARCH=$ARGV_ARCHITECTURE" \
```
lurch: Just out of curiosity - where does this modification fit in with other things?

jviotti: It is because the Jessie 32-bit container still reports itself as 64-bit.

lurch: Ahhh, I remember reading something about that on the Docker Ubuntu images we were using.
Did a bit of searching and found https://github.com/docker-32bit/ubuntu/issues/18 which says "I think that it makes no sense to run a 32bit container in a 64bit host" :-/
https://hub.docker.com/r/32bit/ubuntu/ and https://github.com/docker-32bit/ubuntu/issues/12 seem to contain more info.

I guess setting the TARGET_ARCH env-var is just as effective as setting the Docker entrypoint though. Have you checked that the 32-bit files that this generates do indeed run in a pure 32-bit VM? (not just a 32-bit container using a 64-bit kernel).

Although thinking about it a bit more, if we're supplying the $TARGET_ARCH anyway, is there any benefit to using separate containers, rather than just running everything in a single 64-bit container?  This mixing of arches and containers and builds gets a bit confusing... ;-)

...

Or is this only an issue for local-development, and things actually _would_ work as expected on Travis without this "TARGET_ARCH=$ARGV_ARCHITECTURE" since it does run the x86 build in an x86 VM? (In which case, maybe we could check for the CI variable, and only manually override the TARGET_ARCH if CI isn't set?)

jviotti: > I guess setting the TARGET_ARCH env-var is just as effective as setting the Docker entrypoint though.

What do you mean by entry point?

> Have you checked that the 32-bit files that this generates do indeed run in a pure 32-bit VM? (not just a 32-bit container using a 64-bit kernel).

Yeah, I have a Debian Jessie 32-bit installation that runs it fine.

> Although thinking about it a bit more, if we're supplying the $TARGET_ARCH anyway, is there any benefit to using separate containers, rather than just running everything in a single 64-bit container? This mixing of arches and containers and builds gets a bit confusing... ;-)

The Jessie x86 container is indeed an x64, but its customized to contain all 32-bit packages in user land, so its not quite the same.

lurch:
@jviotti Just for future reference, maybe you could leave a comment detailing which version of glibc we were linking against previously which was "too new", and which version of glibc we're linking against now? Thanks.

jviotti:
Debian Jessie ships with 2.19-18, while we were linking to 2.23-0 before.

PR #1320, opened 2017-04-18
refactor(errors): add isUserError utility function

jviotti:
This utility function is useful to avoid duplicating the logic of
checking whether an error is a user error across the code base.

### DIFF
--- a/lib/shared/errors.js
+++ b/lib/shared/errors.js
@@ -293,6 +293,25 @@ exports.createUserError = (options) => {
 };
 
 /**
+ * @summary Check if an error is an user error
+ * @function
+ * @public
+ *
+ * @param {Error} error - error
+ * @returns {Boolean} whether the error is a user error
+ *
+ * @example
+ * const error = errors.createUserError('Foo', 'Bar');
+ *
+ * if (errors.isUserError(error)) {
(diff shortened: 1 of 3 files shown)


lurch:
Huh? You've added the function, and tests for it, but not actually used it anywhere?
Given how closely coupled this is to resin-corvus, are we sure that the separation between Etcher-functionality and resin-corvus-functionality is correct?  (I'm sure it probably is, but it doesn't hurt to ask :wink: )
Ping @stefan-mihaila

jviotti:
I'm using it in the test suite. Is just a convenient function to we don't check the property directly in the tests. Regarding corvus, would it make sense for the module to accept a function to determine whether the error should be reported or not, so that each application can provide its own custom logic if so?

lurch:
> so that each application can provide its own custom logic if so?

Sounds reasonable to me (but I've not even looked at any of resin-corvus yet).

jviotti:
I created a ticket for it: https://github.com/resin-io-modules/resin-corvus/issues/11

lurch:
Needs to be rebased ;)

jviotti:
Fixed!

jviotti:
I'll merge this so we can keep going.

PR #1322, opened 2017-04-18
refactor(errors): createError and createUserError accept an object

jviotti:
Currently, both of these functions accept two arguments: the error
title, and the error description. This function signature makes is hard
to keep adding options to these error creation functions, like an error
code, so this commit refactors them to take a single argument: an
options object containing `title` and `description` properties.

### DIFF
--- a/tests/shared/errors.spec.js
+++ b/tests/shared/errors.spec.js
@@ -390,12 +390,18 @@ describe('Shared: Errors', function() {
   describe('.createError()', function() {
 
     it('should not set `error.report` by default', function() {
-      const error = errors.createError('Foo', 'Something happened');
+      const error = errors.createError({
+        title: 'Foo',
+        description: 'Something happened'
+      });
+
       m.chai.expect(error.report).to.be.undefined;
     });
 
     it('should set `error.report` to false if `options.report` is false', function() {
-      const error = errors.createError('Foo', 'Something happened', {
+      const error = errors.createError({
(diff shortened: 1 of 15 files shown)


tests/shared/errors.spec.js
```diff
@@ -390,68 +390,99 @@ describe('Shared: Errors', function() {
...
+      });
+
       m.chai.expect(errors.getDescription(error)).to.equal(error.stack);
     });
 
-    it('should throw if no message', function() {
+    it('should throw if no title', function() {
       m.chai.expect(() => {
-        errors.createError();
+        errors.createError({});
```
lurch: Could you also add a test where you pass in e.g. `{description: 'foo'}` ?

tests/shared/errors.spec.js
```diff
@@ -390,68 +390,99 @@ describe('Shared: Errors', function() {
...
       }).to.throw('Invalid error title: undefined');
     });
 
-    it('should throw if message is empty', function() {
+    it('should throw if title is empty', function() {
```
lurch: Nice to see these tightened up from `message` to `title` :+1:

lurch:
Giving @stefan-mihaila a quick ping too, given how much error-handling stuff he's been doing recently :)

tests/shared/errors.spec.js
jviotti: Done!

PR #1326, opened 2017-04-19
chore: make electron-installer-debian a development dependency

jviotti:
We currently require this module to be available in the environment in
order to be able to build debian packages.

This commit adds it as a development dependency, so we can have control
over the version that we use, to prevent any potential incompatibility.

We also update the `udif` package.json version to match what its on the
shrinkwrap file to avoid npm complaining about version mismatches.

See: https://github.com/resin-io/etcher/pull/1290

### DIFF
--- a/scripts/build/electron-installer-debian-linux.sh
+++ b/scripts/build/electron-installer-debian-linux.sh
@@ -25,7 +25,9 @@ if [[ "$OS" != "Linux" ]]; then
   exit 1
 fi
 
-./scripts/build/check-dependency.sh electron-installer-debian
+ELECTRON_INSTALLER_DEBIAN="./node_modules/.bin/electron-installer-debian"
+
+./scripts/build/check-dependency.sh "$ELECTRON_INSTALLER_DEBIAN"
 
 function usage() {
   echo "Usage: $0"
@@ -65,7 +67,7 @@ fi
 DEBIAN_ARCHITECTURE=$(./scripts/build/architecture-convert.sh -r "$ARGV_ARCHITECTURE" -t debian)
 
 cp scripts/build/debian/etcher-electron.sh "$ARGV_DIRECTORY"
-electron-installer-debian \
+"$ELECTRON_INSTALLER_DEBIAN" \
   --src "$ARGV_DIRECTORY" \
   --dest "$ARGV_OUTPUT" \
   --config "$ARGV_DEBIAN_CONFIGURATION" \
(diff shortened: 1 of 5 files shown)


lurch:
Failing on Appveyor, since `electron-installer-debian` isn't supported on Windows :-/

jviotti:
I see. I'll make it an optional dependency then.

jviotti:
Done!

scripts/build/electron-installer-debian-linux.sh
```diff
@@ -62,10 +62,15 @@ then
   usage
 fi
 
+if [ ! -d "$PWD/node_modules" ]; then
```
lurch: This should probably be `"./node_modules"` to match the usage on line 73? (and line 28)

lurch:
> We also update the udif package.json version to match what its on the
shrinkwrap file to avoid npm complaining about version mismatches.

Out of curiosity, how did they get out of sync in the first place?

jviotti:
Looks like it was accidentally introduced by https://github.com/resin-io/etcher/pull/1283. The PR updates the shrinkwrap file, but not the package.json

scripts/build/electron-installer-debian-linux.sh
jviotti: Good catch! I actually made the `check-dependency.sh` script check for the dependency in `./node_modules/.bin` and removed this check, since if the check dependency script succeeded, then there is no way this would fail.

lurch:
Oh yeah, naughty @jhermsmeier ! ;-)

To prevent this in future, what particular NPM command should we have the sanity-checks run?

scripts/build/electron-installer-debian-linux.sh
lurch: Aha, doubly good catch! :+1: :+1:

lurch:
...along similar lines, would it be worth doing the same thing for `asar` too? Or is that much more "stable" and so the exact version doesn't matter?

jviotti:
> To prevent this in future, what particular NPM command should we have the sanity-checks run?

We could try generating a shrinkwrap file with `npm shrinkwrap --dev`. It will fail with mismatches errors in those cases.

> ...along similar lines, would it be worth doing the same thing for asar too? Or is that much more "stable" and so the exact version doesn't matter?

Yeah, let's do it!

jhermsmeier:
> Oh yeah, naughty @jhermsmeier ! ;-)

So, naughty me screwed it up again :D
This one should fix the cause of that mishap: https://github.com/resin-io/etcher/pull/1338

PR #1330, opened 2017-04-20
chore: make `asar` a development dependency

jviotti:
To prevent any potential version incompatibility caused by simply installing
the latest version of the packages in the CI services.

See: https://github.com/resin-io/etcher/pull/1326

### DIFF
--- a/docs/RUNNING-LOCALLY.md
+++ b/docs/RUNNING-LOCALLY.md
@@ -12,13 +12,11 @@ Prerequisites
 - [NodeJS](https://nodejs.org) (at least v6)
 - [Python](https://www.python.org)
 - [jq](https://stedolan.github.io/jq/)
-- [Asar](https://github.com/electron/asar)
 - [Codespell](https://github.com/lucasdemarchi/codespell)
 - [curl](https://curl.haxx.se/)
 
 ### Windows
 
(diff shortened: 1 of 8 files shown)


lurch:
I think we're trying to _prevent_ version incompatibility, rather than ensure it?? ;-D
(and I'd put "caused" between "incompatibility" and "by")

docs/RUNNING-LOCALLY.md
```diff
@@ -12,13 +12,11 @@ Prerequisites
 - [NodeJS](https://nodejs.org) (at least v6)
 - [Python](https://www.python.org)
 - [jq](https://stedolan.github.io/jq/)
-- [Asar](https://github.com/electron/asar)
 - [Codespell](https://github.com/lucasdemarchi/codespell)
 - [curl](https://curl.haxx.se/)
 
 ### Windows
 
-- [Rimraf](https://github.com/isaacs/rimraf)
```
lurch: Aha, I'd been wondering for ages if rimraf was actually unused, but never got round to actually testing it :)

jviotti:
> I think we're trying to prevent version incompatibility, rather than ensure it?? ;-D

lol, hopefully that's not what the commit addresses :P 

I'll fix it while merging, so I don't have to push again.

PR #1332, opened 2017-04-20
refactor(child-writer): add robot commands constant

jviotti:
See: https://github.com/resin-io/etcher/pull/1295#discussion_r112463686

### DIFF
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -126,8 +126,6 @@ exports.write = (image, drive, options) => {
    * }));
    */
   const bridgeRobotMessage = (message) => {
-    const ROBOT_COMMAND_LOG = 'log';
-
     const parsedMessage = _.attempt(() => {
       if (robot.isMessage(message)) {
         return robot.parseMessage(message);
@@ -136,7 +134,7 @@ exports.write = (image, drive, options) => {
       // Don't be so strict. If a message doesn't look like
       // a robot message, then make the child writer log it
       // for debugging purposes.
-      return robot.parseMessage(robot.buildMessage(ROBOT_COMMAND_LOG, {
+      return robot.parseMessage(robot.buildMessage(robot.COMMAND.LOG, {
         message
       }));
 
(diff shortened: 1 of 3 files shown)


lib/child-writer/index.js
```diff
@@ -127,6 +127,7 @@ exports.write = (image, drive, options) => {
    */
   const bridgeRobotMessage = (message) => {
     const ROBOT_COMMAND_LOG = 'log';
+    const ROBOT_COMMAND_ERROR = 'error';
```
jhermsmeier: Agree; `robot.COMMAND_LOG` would follow the usual pattern of how constants come from places

jviotti: Yeah, lets do it!

jviotti: Done!

lib/child-writer/index.js
```diff
@@ -136,7 +134,7 @@ exports.write = (image, drive, options) => {
       // Don't be so strict. If a message doesn't look like
       // a robot message, then make the child writer log it
       // for debugging purposes.
-      return robot.parseMessage(robot.buildMessage(ROBOT_COMMAND_LOG, {
+      return robot.parseMessage(robot.buildMessage(robot.COMMANDS.LOG, {
```
jhermsmeier: To stick with the conventional naming of enums, should this be singular, rather than plural, i.e. `robot.COMMAND.LOG`?

jviotti: Fixed!

lib/shared/robot/index.js
```diff
@@ -255,5 +282,5 @@ exports.printMessage = (message, data) => {
  * robot.log({ example: 'data' });
  */
 exports.log = (data) => {
-  exports.printMessage('log', data);
+  exports.printMessage(exports.COMMAND.LOG, data);
```
lurch: :tada:

tests/shared/robot.spec.js
```diff
@@ -22,6 +22,23 @@ const robot = require('../../lib/shared/robot');
...
+      m.chai.expect(_.every(_.values(robot.COMMAND), _.isString)).to.be.true;
+    });
+
+    it('should contain only unique values', function() {
+      const numberOfKeys = _.size(_.keys(robot.COMMAND));
+      m.chai.expect(_.size(_.uniq(_.values(robot.COMMAND)))).to.equal(numberOfKeys);
+    });
+
+  });
+
```
lurch: Given that later tests rely on `command: 'error'`, perhaps it'd be worth checking here that `robot.COMMAND.ERROR === 'error'` ?

(and it also looks like there are no tests for `robot.log`?)

jviotti: Fixed! `robot.log` only prints to `stdout`, which is quite tricky to test. It internally uses `.buildMessage`, which is already well tested.

PR #1336, opened 2017-04-21
[WIP] resin-corvus node analytics

genesis-block:


tests/gui/components/update-notifier.spec.js
```diff
@@ -4,10 +4,19 @@ const m = require('mochainon');
...
 require('angular-mocks');
 
 describe('Browser: UpdateNotifier', function() {
 
+  before(function() {
+    analytics.disableConsoleOutput();
+  });
+
+  after(function() {
+    analytics.enableConsoleOutput();
```
lurch: Given that these are only unit-tests, is it worth re-enabling analytics console output?

lurch: ...would it make more sense to just disable analytics altogether during the unit-tests?

lib/gui/app.js
```diff
@@ -26,13 +26,16 @@ var angular = require('angular');
...
+const _ = require('lodash');
 const electron = require('electron');
 const Bluebird = require('bluebird');
 const EXIT_CODES = require('../shared/exit-codes');
 const messages = require('../shared/messages');
 const packageJSON = require('../../package.json');
 const flashState = require('./models/flash-state');
 const windowProgress = require('./os/window-progress');
+const nodeAnalytics = require('../shared/analytics');
+
```
lurch: The extra blank line here is tripping up the CI builds

tests/gui/components/update-notifier.spec.js
jviotti: I believe test frameworks set some global variables so that packages can know if they are running on a test suite. We should check what variables mocha expose and disable this altogether on corvus itself.

lib/gui/app.js
```diff
@@ -162,7 +175,9 @@ app.run(($timeout, DriveScannerService, DrivesModel, ErrorService) => {
     // available drives list has changed, and incorrectly
     // keeps asking the user to "Connect a drive".
     $timeout(() => {
-      DrivesModel.setDrives(drives);
+      if (!_.isEqual(DrivesModel.getDrives(), drives)) {
+        DrivesModel.setDrives(drives);
```
Shou: The interface is the same so this should work as usual.

lurch: Even with the custom comparison functions you needed to add? Or am I getting my wires crossed? (more than likely!)

jviotti: The `setDrives()` function should already be preventing the same drives from getting set again. Is this not the case? If not, can we fix it on `DrivesModel`?

lib/gui/models/store.js
```diff
@@ -394,10 +405,17 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
 
     case ACTIONS.REMOVE_DRIVE: {
+      analytics.logEvent('Unselect drive', {
+        device: state.getIn([ 'selection', 'drive' ])
+      });
       return state.deleteIn([ 'selection', 'drive' ]);
     }
 
     case ACTIONS.REMOVE_IMAGE: {
+      const image = state.getIn([ 'selection', 'image' ]) && state.getIn([ 'selection', 'image' ]).toJS();
```
jviotti: I *think* this can be simplified to:

```js
const image = _.invoke(state.getIn([ 'selection', 'image' ]), 'toJS');
```

The `_.invoke` function will make sure `toJS` is not called if the image is undefined.

jviotti:
I'm closing this one since there are a lot of conflicts, and we're moving to a unified analytics library.

PR #1338, opened 2017-04-21
chore(scripts): Remove prune from shrinkwrap cleanup

jhermsmeier:
Newer npm versions (4.4+ I believe) even remove dependencies
not shrinkwrapped yet while pruning causing newly installed dependencies
to not always be shrinkwrapped / updated in the shrinkwrap.
Removing the prune allows for this to work properly again and in the future,
with the drawback that care must be taken to not have extraneous dependencies
in the module tree.

Change-Type: patch

### DIFF
--- a/scripts/clean-shrinkwrap.js
+++ b/scripts/clean-shrinkwrap.js
@@ -38,7 +38,5 @@ const npm = (command) => {
 
 npm([ 'rm', '--ignore-scripts' ].concat(shrinkwrapIgnore))
   .once('close', () => {
-    npm([ 'prune' ]).once('close', () => {
-      console.log('Done.');
-    });
+    console.log('Done.');
   });


lurch:
For people who are still NPM-newbies :wink: could you clarify what you mean with "with the drawback that care must be taken to not have extraneous dependencies in the module tree."?
I'm happy to attempt writing another magical Bash script to help check things :-)

jhermsmeier:
Sure, `npm` should actually yell at you when you do an `npm ls` and exit with a non-zero status code, like this:

```
... omitted dependency tree ...
└─┬ yauzl@2.6.0
  ├── buffer-crc32@0.2.5
  └─┬ fd-slicer@1.0.1
    └── pend@1.2.0

npm ERR! missing: drivelist@^5.0.19, required by etcher@1.0.0-rc.3
npm ERR! extraneous: blockmap@1.1.0 /Users/Jonas/Work/etcher/node_modules/blockmap
npm ERR! extraneous: pipage@1.0.0 /Users/Jonas/Work/etcher/node_modules/pipage
```
Since shrinkwrapping seems to change a bit with every minor npm version, I can't reliably tell whether `npm shrinkwrap` will yell & exit with a non-zero status (I think it should, and did once, but not anymore).

So checking for that manually (and automatically) shouldn't be much of a problem, just something to keep in mind before shrinkwrapping.

lurch:
So, is it worth checking the exit codes of both `npm ls` _and_ `npm shrinkwrap` in case one of them errors and the other doesn't ;-)

PR #1343, opened 2017-04-21
feat(GUI): collect archive and image extension in analytics

stefan-mihaila:
Change-Type: patch

### DIFF
--- a/lib/image-stream/handlers.js
+++ b/lib/image-stream/handlers.js
@@ -22,10 +22,12 @@ const PassThroughStream = require('stream').PassThrough;
 const lzma = Bluebird.promisifyAll(require('lzma-native'));
 const zlib = require('zlib');
 const unbzip2Stream = require('unbzip2-stream');
+const path = require('path');
 const gzip = require('./gzip');
 const udif = Bluebird.promisifyAll(require('udif'));
 const archive = require('./archive');
 const zipArchiveHooks = require('./archive-hooks/zip');
+const utils = require('./utils');
 
 /**
  * @summary Image handlers
@@ -50,6 +52,8 @@ module.exports = {
   'application/x-bzip2': (file, options) => {
     return Bluebird.props({
       path: file,
+      archiveExtension: path.extname(file),
+      extension: utils.getImageExtension(file),
       stream: fs.createReadStream(file),
       size: {
         original: options.size,
@@ -79,6 +83,8 @@ module.exports = {
     return gzip.getUncompressedSize(file).then((uncompressedSize) => {
       return Bluebird.props({
         path: file,
+        archiveExtension: path.extname(file),
+        extension: utils.getImageExtension(file),
         stream: fs.createReadStream(file),
         size: {
           original: options.size,
@@ -113,6 +119,8 @@ module.exports = {
     }).then((metadata) => {
       return {
         path: file,
+        archiveExtension: path.extname(file),
+        extension: utils.getImageExtension(file),
         stream: fs.createReadStream(file),
         size: {
           original: options.size,
@@ -143,6 +151,7 @@ module.exports = {
     return udif.getUncompressedSizeAsync(file).then((size) => {
       return {
         path: file,
+        extension: path.extname(file),
         stream: udif.createReadStream(file),
         size: {
           original: options.size,
@@ -186,6 +195,7 @@ module.exports = {
   'application/octet-stream': (file, options) => {
     return Bluebird.props({
       path: file,
+      extension: path.extname(file),
       stream: fs.createReadStream(file),
       size: {
         original: options.size,
--- a/lib/image-stream/handlers.js
+++ b/lib/image-stream/handlers.js
@@ -22,12 +22,11 @@ const PassThroughStream = require('stream').PassThrough;
 const lzma = Bluebird.promisifyAll(require('lzma-native'));
 const zlib = require('zlib');
 const unbzip2Stream = require('unbzip2-stream');
-const path = require('path');
 const gzip = require('./gzip');
 const udif = Bluebird.promisifyAll(require('udif'));
 const archive = require('./archive');
 const zipArchiveHooks = require('./archive-hooks/zip');
-const utils = require('./utils');
+const fileExtensions = require('../shared/file-extensions');
 
 /**
  * @summary Image handlers
@@ -52,8 +51,8 @@ module.exports = {
   'application/x-bzip2': (file, options) => {
     return Bluebird.props({
       path: file,
-      archiveExtension: path.extname(file),
-      extension: utils.getImageExtension(file),
+      archiveExtension: fileExtensions.getLastFileExtension(file),
+      extension: fileExtensions.getPenultimateFileExtension(file),
       stream: fs.createReadStream(file),
       size: {
         original: options.size,
@@ -83,8 +82,8 @@ module.exports = {
     return gzip.getUncompressedSize(file).then((uncompressedSize) => {
       return Bluebird.props({
         path: file,
-        archiveExtension: path.extname(file),
-        extension: utils.getImageExtension(file),
+        archiveExtension: fileExtensions.getLastFileExtension(file),
+        extension: fileExtensions.getPenultimateFileExtension(file),
         stream: fs.createReadStream(file),
         size: {
           original: options.size,
@@ -119,8 +118,8 @@ module.exports = {
     }).then((metadata) => {
       return {
         path: file,
-        archiveExtension: path.extname(file),
-        extension: utils.getImageExtension(file),
+        archiveExtension: fileExtensions.getLastFileExtension(file),
+        extension: fileExtensions.getPenultimateFileExtension(file),
         stream: fs.createReadStream(file),
         size: {
           original: options.size,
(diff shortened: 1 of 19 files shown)


lib/image-stream/handlers.js
```diff
@@ -50,6 +52,8 @@ module.exports = {
   'application/x-bzip2': (file, options) => {
     return Bluebird.props({
       path: file,
+      archiveExtension: path.extname(file),
```
jviotti: I think that the archive extension is redundant, given that it can be obtained from `path`. Can we only add `extension`?

tests/image-stream/bz2.spec.js
```diff
@@ -48,6 +48,8 @@ describe('ImageStream: BZ2', function() {
       return imageStream.getImageMetadata(image).then((metadata) => {
         m.chai.expect(metadata).to.deep.equal({
           path: image,
+          extension: '.img',
```
jviotti: Can we save the extensions without leading periods?

lib/image-stream/utils.js
```diff
@@ -97,3 +97,28 @@ exports.extractStream = (stream) => {
...
+ * @param {String} archiveName - archive name
+ * @returns {String} - image extension
+ *
+ * @example
+ * const extension = utils.getImageExtension('quirky-pi2-sd-8gb-xerus-8.1.img.xz');
+ * console.log(extension);
+ *
+ * > '.img'
+ */
+exports.getImageExtension = (archiveName) => {
```
jviotti: Can we somehow re-use the logic from https://github.com/resin-io/etcher/blob/master/lib/shared/supported-formats.js#L109? E.g: maybe this function can use the extension retrieval logic from `supported-formats.js`, and then `isSupportedImage()` can re-use this function.

lurch:
I'm not sure where it would "fit", but there should probably also be some code-check and/or unit-test that every `image` object has an `extension` property?

lib/image-stream/handlers.js
lurch: I guess it depends if it makes statistics / results easier to generate / sort / filter, if the `archiveExtension` is 'automatically' there, without having to manually extract it from the filename?

jviotti:
> I'm not sure where it would "fit", but there should probably also be some code-check and/or unit-test that every image object has an extension property?

Done!

jviotti:
I'll merge this so we can release today. I'm happy to address any comments afterwards!

tests/gui/models/selection-state.spec.js
```diff
@@ -425,10 +448,75 @@ describe('Browser: SelectionState', function() {
...
+              size: {
+                original: 999999999,
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
+            });
+          }).to.throw('Invalid image archive extension: 1');
+        });
```
lurch: Should there also be tests that e.g. `{ path: 'foo.img.xz', extension: 'img', archiveExtension: 'gz' }` and `{ path: 'foo.img.xz', extension: 'ifg', archiveExtension: 'xz' }` throw the appropriate errors?

tests/shared/file-extensions.spec.js
```diff
@@ -0,0 +1,141 @@
+/*
+ * Copyright 2016 resin.io
```
lurch: Should have been 2017 for a new file ;-P

tests/shared/file-extensions.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        'img',
+        'gz'
+      ]);
+    });
+
+  });
+
+  describe('.getLastFileExtension()', function() {
+
+    it('should return undefined in the file path has no extension', function() {
```
lurch: in -> if

tests/shared/file-extensions.spec.js
```diff
@@ -0,0 +1,141 @@
...
+      m.chai.expect(fileExtensions.getPenultimateFileExtension('foo')).to.be.undefined;
+    });
+
+    it('should return undefined if there is only one extension', function() {
+      m.chai.expect(fileExtensions.getPenultimateFileExtension('foo.img')).to.be.undefined;
+    });
+
+    it('should return the first extension if there are two extensions', function() {
```
lurch: Doesn't make much difference, but I think for consistency you should use "penultimate" here instead of "first".

tests/shared/file-extensions.spec.js
```diff
@@ -0,0 +1,141 @@
...
+describe('Shared: fileExtensions', function() {
+
+  describe('.getFileExtensions()', function() {
+
+    _.forEach([
+
+      // No extension
+      {
+        file: 'path/to/filename',
+        extension: []
```
lurch: Only a minor nit, but these test-case `extension` properties could be renamed to `extensions`, given that they sometimes contain more than one element.
Also, perhaps you should add a `foo.bar.img.gz` testcase?

tests/shared/file-extensions.spec.js
```diff
@@ -0,0 +1,141 @@
...
+
+    it('should return undefined in the file path has no extension', function() {
+      m.chai.expect(fileExtensions.getLastFileExtension('foo')).to.be.undefined;
+    });
+
+    it('should return the extension if there is only one extension', function() {
+      m.chai.expect(fileExtensions.getLastFileExtension('foo.img')).to.equal('img');
+    });
+
+    it('should return the last extension if there two extensions', function() {
```
lurch: if there **are** two

lib/shared/file-extensions.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @param {String} filePath - file path
+ * @returns {String[]} extensions
+ *
+ * @example
+ * const extensions = fileExtensions.getFileExtensions('path/to/foo.img.gz');
+ * console.log(extensions);
+ * > [ 'img', 'gz' ]
+ */
+exports.getFileExtensions = (filePath) => {
```
jviotti: Oh, yeah, great catch :+1:

lib/shared/file-extensions.js
```diff
@@ -0,0 +1,74 @@
...
+  return _.last(exports.getFileExtensions(filePath));
+};
+
+/**
+ * @summary Get the penultimate file extension
+ * @function
+ * @public
+ *
+ * @param {String} filePath - file path
+ * @returns {(String|Undefined)} penultimate extension
```
jviotti: Do you mean another example?

lib/shared/file-extensions.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @example
+ * const extension = fileExtensions.getLastFileExtension('path/to/foo.img.gz');
```
jviotti: Fixed!

lib/shared/file-extensions.js
```diff
@@ -0,0 +1,74 @@
...
+ * console.log(extension);
+ * > [ 'img' ]
+ */
+exports.getPenultimateFileExtension = (filePath) => {
+  return _.last(_.initial(exports.getFileExtensions(filePath)));
```
jviotti: Haha, this is a pattern I've seen in some Haskell code :P

tests/shared/file-extensions.spec.js
jviotti: God, I'll never get a chance to do that. Mind sending a massive search and replace PR for me? :P

tests/shared/file-extensions.spec.js
jviotti: I wonder why I always end up missing those verbs.

tests/shared/file-extensions.spec.js
jviotti: Done!

tests/shared/file-extensions.spec.js
jviotti: Done!

tests/gui/models/selection-state.spec.js
jviotti: I added a test case for the first one. The other case is a bit tricky to test. If a file is `foo.zip` and contains a file called `bar.img` inside, then the file path doesn't reflect the internal file's extension.

lib/shared/file-extensions.js
lurch: Yeah, I guess an extra example might be the easiest way to show that. Just so that it's obvious that `.getPenultimateFileExtension('path/to/foo.zip') !== 'foo'`

tests/shared/file-extensions.spec.js
lurch: As I've said before (I think!), IMHO the year should only reflect the date that the file was first added. No massive search&replace needed ;P

tests/shared/file-extensions.spec.js
lurch: Maybe there's a gremlin living in your keyboard ;-)

tests/shared/file-extensions.spec.js
lurch: To clarify, I was suggesting testing a filename with 3 extensions :)

tests/shared/file-extensions.spec.js
jviotti: Oh, I see!

tests/shared/file-extensions.spec.js
jviotti: Ah, fixed!

tests/shared/file-extensions.spec.js
jviotti: That's probably it! :)

tests/shared/file-extensions.spec.js
jviotti: Fixed!

PR #1344, opened 2017-04-21
feat(GUI): add uuid to flash events

stefan-mihaila:
Change-Type: patch


lib/gui/models/flash-state.js
```diff
@@ -220,3 +220,34 @@ exports.getLastFlashSourceChecksum = () => {
...
+
+/**
+ * @summary Generate flash uuid
+ * @function
+ * @public
+ *
+ * @example
+ * flashState.generateFlashUuid();
+ */
+exports.generateFlashUuid = () => {
```
jviotti: I don't think we should expose this as something the user can call, since it means it can easily get out of sync. I think the UUID should be generated when the `SET_FLASHING_FLAG` is called. What do you think?

lib/gui/models/store.js
```diff
@@ -236,6 +238,11 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         .set('flashResults', DEFAULT_STATE.get('flashResults'));
     }
 
+    // TODO: Remove this action and generate the uuid on 'SET_FLASHING_FLAG'
```
jviotti: Oh, you already added a TODO for it. Any reason to not do this now?

jviotti:
Let me take over this PR. `ImageWriterService` is on the angular world, and that point we can emit the flash event with access to the flash uuid, so I'll move the flash analytics over there (which is the correct place to have those anyway), and move the uuid generation to the unset flashing flag step.

jviotti:
Done!

jviotti:
@stefan-mihaila @Shou @jhermsmeier @lurch Can anybody review my changes here?

jhermsmeier:
@jviotti what did you change? There aren't any changes showing up here for me after you reviewed this... or am I missing something?

lib/gui/models/flash-state.js
```diff
@@ -220,3 +220,20 @@ exports.getLastFlashSourceChecksum = () => {
...
+ * @function
+ * @public
+ *
+ * @description
+ * This function returns undefined if there was no last flash.
+ *
+ * @returns {String} the last flash uuid
+ *
+ * @example
+ * const uuid = flashState.getLastFlashUuid();
```
lurch: Huh? Function-name here doesn't match that below

lib/gui/modules/image-writer.js
```diff
@@ -102,12 +113,32 @@ imageWriter.service('ImageWriterService', function($q, $rootScope) {
         flashState.setProgressState(state);
       });
 
-    }).then(flashState.unsetFlashingFlag).catch((error) => {
+    }).then(flashState.unsetFlashingFlag).then(() => {
+      if (flashState.wasLastFlashCancelled()) {
+        AnalyticsService.logEvent('Elevation cancelled', analyticsData);
```
lurch: I guess theoretically, all these logEvents only need to report the uuid, since we can get all the other details from the 'Flash' event with the matching uuid?

lib/gui/pages/main/controllers/flash.js
```diff
@@ -60,69 +58,33 @@ module.exports = function(
...
-        drive
-      });
-      $state.go('success');
     })
     .catch((error) => {
       OSNotificationService.send('Oops!', messages.error.flashFailure());
 
+      // TODO: All these error codes to messages translations
+      // should go away if the writer emitted user friendly
+      // messages on the first place.
```
lurch: But OTOH if/when we add localisation, it might be better to do that within Etcher rather than within the writer?

tests/gui/models/flash-state.spec.js
```diff
@@ -441,6 +470,52 @@ describe('Browser: flashState', function() {
...
+          sourceChecksum: '1234',
+          cancelled: false
+        });
+
+        m.chai.expect(UUID_REGEX.test(uuid1)).to.be.true;
+        m.chai.expect(UUID_REGEX.test(uuid2)).to.be.true;
+        m.chai.expect(UUID_REGEX.test(uuid3)).to.be.true;
+
+        m.chai.expect(uuid1).to.not.equal(uuid2);
+        m.chai.expect(uuid2).to.not.equal(uuid3);
```
lurch: I guess for completion, you could also do `m.chai.expect(uuid3).to.not.equal(uuid1);` ;)

tests/gui/models/flash-state.spec.js
```diff
@@ -441,6 +470,52 @@ describe('Browser: flashState', function() {
 
     });
 
+    describe('.getFlashUuid()', function() {
+
+      const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/;
```
lurch: Minor nit, but perhaps this should be `UUID_V4_REGEX`

EDIT: forget that, I've just seen they match the same regex ;-)  https://github.com/kelektiv/node-uuid

tests/gui/models/flash-state.spec.js
```diff
@@ -441,6 +470,52 @@ describe('Browser: flashState', function() {
...
+        flashState.unsetFlashingFlag({
+          sourceChecksum: '1234',
+          cancelled: false
+        });
+
+        flashState.setFlashingFlag();
+        const uuid2 = flashState.getFlashUuid();
+        flashState.unsetFlashingFlag({
+          sourceChecksum: '1234',
+          cancelled: false
```
lurch: Would it make sense for at least one of these to set `cancelled: true` ?

tests/gui/models/flash-state.spec.js
```diff
@@ -441,6 +470,52 @@ describe('Browser: flashState', function() {
 
     });
 
+    describe('.getFlashUuid()', function() {
+
+      const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/;
+
+      it('should be initially undefined', function() {
```
lurch: Presumably this is the _only_ time it'll ever be undefined?

EDIT: Nope, I've just seen that `ACTIONS.RESET_FLASH_STATE` does `.delete('flashUuid')`, but I've also just spotted that you _do_ check that with `it('should unset the flash uuid',` ;)

lib/gui/models/flash-state.js
```diff
@@ -220,3 +220,20 @@ exports.getLastFlashSourceChecksum = () => {
...
   return exports.getFlashResults().errorCode;
 };
+
+/**
+ * @summary Get last flash uuid
+ * @function
+ * @public
+ *
+ * @description
+ * This function returns undefined if there was no last flash.
```
lurch: "if no flash has been performed yet" or ""if no flash has been started yet" might be more readable than "if there was no last flash"?

lib/gui/models/flash-state.js
```diff
@@ -220,3 +220,20 @@ exports.getLastFlashSourceChecksum = () => {
 exports.getLastFlashErrorCode = () => {
   return exports.getFlashResults().errorCode;
 };
+
+/**
+ * @summary Get last flash uuid
```
lurch: IMHO this description might be a bit misleading - perhaps "Get current (or last) flash uuid" instead?

lib/gui/models/flash-state.js
lurch: Should this mention that this will also be undefined after `resetState` ?
(apologies for the confusing questions, but I initially misunderstood how this worked)

lib/gui/modules/image-writer.js
```diff
@@ -102,12 +113,32 @@ imageWriter.service('ImageWriterService', function($q, $rootScope) {
...
+      if (error.code === 'EVALIDATION') {
+        AnalyticsService.logEvent('Validation error', analyticsData);
+      } else if (error.code === 'EUNPLUGGED') {
+        AnalyticsService.logEvent('Drive unplugged', analyticsData);
+      } else if (error.code === 'EIO') {
+        AnalyticsService.logEvent('Input/output error', analyticsData);
+      } else if (error.code === 'ENOSPC') {
+        AnalyticsService.logEvent('Out of space', analyticsData);
+      } else {
+        AnalyticsService.logEvent('Flash error', analyticsData);
```
lurch: If this is catching cases where we haven't checked a matching `error.code`, maybe it'd be worth this logEvent reporting the `error.code` that triggered it?

EDIT: Ah, looks like the previous code in `lib/gui/pages/main/controllers/flash.js` actually used to log the entire `error` object :)  Probably worth doing the same still?

lurch:
> @jviotti what did you change?

Looks like @jviotti made his changes between his "Let me take over this PR." and "Done!" comments, and then squashed everything down to a single commit.

lib/gui/modules/image-writer.js
alexandrosm: @lurch you're assuming that the flash event makes it to mixpanel, but it's not certain. Also, it will make it very difficult to query (basically we won't be able to use mixpanel). As such, whoever is managing this pr, please don't implement the optimisation proposed by Andrew.

lib/gui/models/flash-state.js
lurch: Hmmm, looks like `resetState` is _only_ used in `lib/gui/components/flash-error-modal/services/flash-error-modal.js`, which (AFAICT) means the UUID will be cleared if a flash error occurs, but will remain set if the flash completes? Would it be more consistent for a successful flash to delete the UUID too, and then this `@description` could become something like `This function returns undefined if there is no flash operation currently taking place.` which _might_ make more sense?

lib/gui/modules/image-writer.js
lurch: Fair enough :-)  Can I get away with my 'theoretically' disclaimer? :wink:

lurch:
@jviotti If you've taken over this PR, do you want to remove your approval? ;)

lib/gui/modules/image-writer.js
alexandrosm: So long as we don't practically change the code :P

lib/gui/models/flash-state.js
jviotti: Sounds good, done!

lib/gui/models/flash-state.js
jviotti: Fixed!

lib/gui/modules/image-writer.js
jviotti: Good catch, fixed!

lib/gui/pages/main/controllers/flash.js
jviotti: `etcher-image-write` will be part of the main Etcher repo (I'm looking forward to do that very soon), so that won't be a problem.

tests/gui/models/flash-state.spec.js
jviotti: Fixed!

tests/gui/models/flash-state.spec.js
jviotti: Done!

jviotti:
Yeah, sorry about the confusion. I'm too used to amend and force push :(

lib/gui/models/flash-state.js
jviotti: `.resetState()` is only called when you click "retry" (or when you dismiss the flash error modal, which has the same effect as "retry"), so doesn't necessarily happen after an error.

jviotti: Obviously the "retry" button is the only path after an error (which I guess makes the "the UUID will be cleared if a flash error occurs" statement valid), but its not part of the flash UUID core logic.

jviotti:
I'll merge this so we can release today. I'm happy to address any comments afterwards!

lib/gui/pages/main/controllers/flash.js
lurch: Oh, isn't it also used by the resin-cli? Or am I getting muddled up with something else?

lib/gui/modules/image-writer.js
lurch: What impact (if any) does this have on https://github.com/resin-io/etcher/pull/1344/files#diff-1e0579ca641fcc9114c426954e51435aR210 ?

lib/gui/pages/main/controllers/flash.js
jviotti: By the resin CLI? Do you mean the Etcher CLI?

lurch: Turns out I was right ;-)
https://github.com/resin-io/resin-cli/blob/master/package.json#L48

jviotti: I see what you mean now. I think we can provide way better errors messages from `etcher-image-write` that don't include the word "Etcher", and thus would fit nicely in the CLI.

PR #1347, opened 2017-04-24
refactor: address file-extensions review comments

jviotti:
See: https://github.com/resin-io/etcher/pull/1343#pullrequestreview-34288142

### DIFF
--- a/lib/gui/models/store.js
+++ b/lib/gui/models/store.js
@@ -22,6 +22,7 @@ const redux = require('redux');
 const persistState = require('redux-localstorage');
 const uuidV4 = require('uuid/v4');
 const constraints = require('../../shared/drive-constraints');
+const supportedFormats = require('../../shared/supported-formats');
 const errors = require('../../shared/errors');
 const fileExtensions = require('../../shared/file-extensions');
 
@@ -338,22 +339,36 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         });
       }
 
-      if (!_.isString(action.data.extension)) {
+      if (_.some([
+        !_.isString(action.data.extension),
+        !_.includes(supportedFormats.getAllExtensions(), action.data.extension)
+      ])) {
         throw errors.createError({
           title: `Invalid image extension: ${action.data.extension}`
         });
       }
 
-      if (fileExtensions.getLastFileExtension(action.data.path) !== action.data.extension) {
+      const lastImageExtension = fileExtensions.getLastFileExtension(action.data.path);
+
+      if (lastImageExtension !== action.data.extension) {
         if (!action.data.archiveExtension) {
           throw errors.createError({
             title: 'Missing image archive extension'
           });
         }
 
-        if (!_.isString(action.data.archiveExtension)) {
+        if (_.some([
+          !_.isString(action.data.archiveExtension),
+          !_.includes(supportedFormats.getAllExtensions(), action.data.archiveExtension)
+        ])) {
           throw errors.createError({
             title: `Invalid image archive extension: ${action.data.archiveExtension}`
+          });
+        }
+
+        if (lastImageExtension !== action.data.archiveExtension) {
+          throw errors.createError({
+            title: `Image archive extension mismatch: ${action.data.archiveExtension} and ${lastImageExtension}`
           });
         }
       }
--- a/tests/gui/models/selection-state.spec.js
+++ b/tests/gui/models/selection-state.spec.js
@@ -512,6 +512,73 @@ describe('Browser: SelectionState', function() {
           }).to.throw('Invalid image archive extension: 1');
         });
 
+        it('should throw if the archive extension doesn\'t match the last path extension in a compressed image', function() {
+          m.chai.expect(function() {
+            SelectionStateModel.setImage({
+              path: 'foo.img.xz',
+              extension: 'img',
+              archiveExtension: 'gz',
+              size: {
+                original: 999999999,
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
+            });
+          }).to.throw('Image archive extension mismatch: gz and xz');
+        });
+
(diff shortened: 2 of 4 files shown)


lib/gui/models/store.js
```diff
@@ -344,7 +344,9 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         });
       }
 
-      if (fileExtensions.getLastFileExtension(action.data.path) !== action.data.extension) {
+      const imageLastExtension = fileExtensions.getLastFileExtension(action.data.path);
```
lurch: Only a minor nit, but I think `lastImageExtension` sounds more natural than `imageLastExtension`.

tests/shared/file-extensions.spec.js
```diff
@@ -29,62 +29,66 @@ describe('Shared: fileExtensions', function() {
...
-        extension: [ 'img', 'bz2' ]
+        extensions: [ 'img', 'bz2' ]
       },
       {
         file: 'path/to/filename.img.xz',
-        extension: [ 'img', 'xz' ]
+        extensions: [ 'img', 'xz' ]
+      },
+      {
+        file: 'path/to/filename.img.gz',
```
lurch: You've now got `'path/to/filename.img.gz'` on line 47 and line 59 ??

tests/gui/models/selection-state.spec.js
```diff
@@ -512,6 +512,40 @@ describe('Browser: SelectionState', function() {
...
+              archiveExtension: 'xz',
+              size: {
+                original: 999999999,
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
+            });
+          }).to.throw('Invalid image archive extension: foo');
```
lurch: Huh?

lurch: Although from the look of the CI tests, it seems this isn't failing as expected?!?

lib/gui/models/store.js

tests/gui/models/selection-state.spec.js
jviotti: Uh, I was testing this out, then got dragged to a meeting, and then though it as complete :P I'll fix it in a second.

tests/shared/file-extensions.spec.js
jviotti: Fixed!

jviotti: I should seriously stop multi tasking, otherwise I'll keep committing non sense stuff :P

lurch: Well as long as you're happy for me to keep catching the little slip-ups you make ;-)

tests/gui/models/selection-state.spec.js
```diff
@@ -512,6 +512,40 @@ describe('Browser: SelectionState', function() {
...
+                original: 999999999,
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
+            });
+          }).to.throw('Invalid image extension: ifg');
+        });
+
```
lurch: From what I can tell of https://github.com/resin-io/etcher/pull/1347/files#diff-74bcbd72e100aecd2f68e833e7a169f4R342 this is a newly-caught error? Bonus!
Could we also add a test for `{ path: 'foo.ifg', extension: 'ifg' }` ?  (and perhaps `{ path: 'foo.ifg.gz', extension: 'ifg', archiveExtension: 'gz' }` too?)

tests/shared/file-extensions.spec.js
jviotti: Haha, sounds good to me! :P

tests/gui/models/selection-state.spec.js
jviotti: Done!

jviotti:
@lurch I noticed an issue with the wording on this PR. We use `archiveExtension` to also set compressed extensions that are not archives, like `gz`. We could rename `archiveExtension` to `compressedExtension` but again not every "wrapper" format is compressed (think `tar`).

Looks like we need a generic word that means that this is the extension of the format that wraps the image. What do you think?

lurch:
> @lurch I noticed an issue with the wording on this PR. We use archiveExtension to also set compressed extensions that are not archives, like gz. We could rename archiveExtension to compressedExtension but again not every "wrapper" format is compressed (think tar).

Yeah, I noticed that too, but given that this is only for the analytics, I think `archiveExtension` is suitably vague to cover both 'archive' and 'compressed' formats.
Ping @stefan-mihaila and @alexandrosm in case they have any comments.

tests/gui/models/selection-state.spec.js
```diff
@@ -512,6 +512,73 @@ describe('Browser: SelectionState', function() {
...
+              size: {
+                original: 999999999,
+                final: {
+                  estimation: false,
+                  value: 999999999
+                }
+              }
+            });
+          }).to.throw('Invalid image extension: ifg');
+        });
```
lurch: Are there any unit-tests (of physical files, and not just filenames) that check that a `foo.ifg` wrapped inside a `foo.zip` "fails as expected" ?

jviotti: The redux store is only concerned with the information an image object provides. The image format archive handlers are in charge of setting both `extension` and `archiveExtension`, and that is covered by the `image-stream` tests.

jviotti:
> but given that this is only for the analytics, I think archiveExtension is suitably vague to cover both 'archive' and 'compressed' formats.

I still feel it sounds awkward. What do you think about `wrapperExtension`?

lurch:
I'd be happy with `wrapperExtension`, but maybe it feels a bit clunky? Lets see what the other guys say.

jviotti:
I'll merge this, but happy to send another PR if needed after we agree on the name.

PR #1351, opened 2017-04-24
refactor: extract elevation routines to lib/shared/permissions.js

jviotti:
The elevation mechanism currently embedded in
`lib/child-writer/writer-proxy.js` is extracted as a separate re-usable
function in `lib/shared/permissions.js`.

This change hugely simplifies the writer proxy, while allowing us to
iterate faster on the elevation core functionality.

### DIFF
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 resin.io
+ * Copyright 2017 resin.io
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,15 +18,12 @@
 
 const Bluebird = require('bluebird');
 const childProcess = require('child_process');
-const commandJoin = require('command-join');
 const ipc = require('node-ipc');
 const _ = require('lodash');
 const os = require('os');
 const path = require('path');
-const sudoPrompt = Bluebird.promisifyAll(require('sudo-prompt'));
 const utils = require('./utils');
 const EXIT_CODES = require('../shared/exit-codes');
-const errors = require('../shared/errors');
 const robot = require('../shared/robot');
 const permissions = require('../shared/permissions');
 const packageJSON = require('../../package.json');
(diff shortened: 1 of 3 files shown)


jviotti:
OS testing:

- [x] macOS
- [ ] Windows
- [ ] GNU/Linux
- [ ] GNU/Linux on AppImage

lib/child-writer/writer-proxy.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 resin.io
+ * Copyright 2017 resin.io
```
lurch: By contrast, the files in this PR were created in 2016, so I wonder if their copyright-year should stay as 2016? I'm afraid I'm not a copyright lawyer, so I dunno what the "correct" solution is ;-)

EDIT: Did a quick search, and found http://stackoverflow.com/a/2391555 and https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html which at least seem to be consistent with each other.

tests/shared/permissions.spec.js
```diff
@@ -0,0 +1,246 @@
...
+
+      afterEach(function() {
+        this.osPlatformStub.restore();
+      });
+
+      it('should return an empty array if no environment', function() {
+        m.chai.expect(permissions.getEnvironmentCommandPrefix()).to.deep.equal([]);
+      });
+
+      it('should return an empty array environment is an empty object', function() {
```
lurch: Did your keyboard-gremlin eat an `if` from this description? ;)

lib/child-writer/writer-proxy.js
```diff
@@ -135,44 +79,32 @@ return permissions.isElevated().then((elevated) => {
...
-
-    return sudoPrompt.execAsync(command, {
-      name: packageJSON.displayName
-    }).then((stdout, stderr) => {
-      if (!_.isEmpty(stderr)) {
-        throw errors.createError({
-          title: stderr
-        });
-      }
+    console.log(`Running: ${commandArguments.join(' ')}`);
```
lurch: I wonder if this should be moved to `permissions.js` ?

tests/shared/permissions.spec.js
```diff
@@ -0,0 +1,246 @@
...
```
lurch: if

tests/shared/permissions.spec.js
```diff
@@ -0,0 +1,246 @@
...
+          '&&',
+          'call'
+        ]);
+      });
+
+      it('should ignore undefined and null variable values', function() {
+        m.chai.expect(permissions.getEnvironmentCommandPrefix({
+          FOO: null,
+          BAR: undefined,
+          BAZ: 'qux'
```
lurch: Only a minor nit, but for unit-testing purposes perhaps you should rearrange the values as `null, 'qux', undefined` which checks that valid-values work after null, and also checks that undefined values are ignored _after_ a valid value?

tests/shared/permissions.spec.js
```diff
@@ -0,0 +1,246 @@
...
+
+      afterEach(function() {
+        this.osPlatformStub.restore();
+      });
+
+      it('should return an empty array if no environment', function() {
+        m.chai.expect(permissions.getEnvironmentCommandPrefix()).to.deep.equal([]);
+      });
+
+      it('should return an empty array environment is an empty object', function() {
```
lurch: Pesky gremlin!

lurch:
On Windows, did you test both portable and non-portable versions?
For all platforms, did you check the behaviour when Etcher is already launched with sudo/admin rights? (which it seems that some users still do, even though it's not necessary)

jviotti:
Yep, it all seems to work fine :)

jviotti:
Oh damn, I merged, but didn't notice all the comments that where there! I'll send another PR with the fixes!

lib/child-writer/writer-proxy.js
jviotti: Oh, great find :+1:

lib/child-writer/writer-proxy.js
jviotti: Can we do it once we get analytics support outside the main Electron process, given that will take care of not printing the lines during tests, etc?

tests/shared/permissions.spec.js
jviotti: Haha, fixed!

tests/shared/permissions.spec.js
jviotti: Fixed!

tests/shared/permissions.spec.js
jviotti: Fixed!

tests/shared/permissions.spec.js
jviotti: Haha, fixed!

jviotti:
See https://github.com/resin-io/etcher/pull/1355. Sorry about that; the approval confused me!

lib/child-writer/writer-proxy.js
lurch: Perhaps something to add to the Contributing guide? And maybe the contributing guide should also say that by submitting a PR to the Etcher repo, any 3rd-party authors are explicitly assigning their copyright over to resin.io?

lib/child-writer/writer-proxy.js
lurch: Entirely up to you, it was only a suggestion ;-)

lurch:
Okay, do I need to be less liberal with my approvals in future? ;)

jviotti:
Haha, don't worry, it was totally my fault.

PR #1352, opened 2017-04-24
refactor: move OSNotificationService out of the angular world

jviotti:


lurch:
LGTM, but CI tests are all failing

lib/gui/os/notification.js
```diff
@@ -0,0 +1,47 @@
...
+ * notification.send('Hello', 'Foo Bar Bar');
+ */
+exports.send = (title, body) => {
+
+  // `app.dock` is only defined in OS X
+  if (electron.remote.app.dock) {
+    electron.remote.app.dock.bounce();
+  }
+
+  new Notification(title, {
```
Shou: Looks like it's complaining about the lack of a `return` before `new` here.

jviotti: > Do not use 'new' for side effects

Thanks a lot [HTML5](https://www.html5rocks.com/en/tutorials/notifications/quick/) :P I'll make it return so the linter doesn't complain

PR #1353, opened 2017-04-25
refactor(utils): make PERCENTAGE_ constants public, and re-use them in flash.js

lurch:


lib/shared/utils.js
```diff
@@ -22,18 +22,20 @@ const errors = require('./errors');
 /**
  * @summary Minimum percentage value
  * @constant
- * @private
+ * @public
  * @type {Number}
  */
 const PERCENTAGE_MINIMUM = 0;
+exports.PERCENTAGE_MINIMUM = PERCENTAGE_MINIMUM;
```
jviotti: Can we do `exports.PERCENTAGE_MINIMUM = 0` directly?

jviotti: Also, can we add some unit tests to check that this constants are positive integers?

lurch: > Can we do exports.PERCENTAGE_MINIMUM = 0 directly?

Dunno, I just went for the "simplest" option from http://stackoverflow.com/questions/8595509/how-do-you-share-constants-in-nodejs-modules
Me is still JS newbie ;-)  (and so didn't fully understand all the suggestions that SO post was making)
My initial attempt was `const exports.PERCENTAGE_MINIMUM = 0` but that didn't work.

> Also, can we add some unit tests to check that this constants are positive integers?

Hmmm, I can't imagine there'd ever be any scenario (at least in _this_ universe) where `PERCENTAGE_MINIMUM` is anything other than `0` and `PERCENTAGE_MAXIMUM` is anything other than `100`? :wink:  So I _could_ add some unit tests that `utils.PERCENTAGE_MINIMUM === 0` and `utils.PERCENTAGE_MAXIMUM === 100` but I dunno if that's a bit superfluous?

jviotti: Haha, fair enough. Regarding the declaration, `exports.PERCENTAGE_MINIMUM = 0` should work (without the `const`).

lurch: Done! Took a bit of fiddling, but I eventually managed to get `npm test` passing again :)

PR #1354, opened 2017-04-25
chore: make use of browserify as a development dependency

jviotti:
This is causing all the snapshot builds to fail when a PR is merged to
master.

### DIFF
--- a/scripts/build/concatenate-javascript.sh
+++ b/scripts/build/concatenate-javascript.sh
@@ -19,7 +19,8 @@
 set -u
 set -e
 
-./scripts/build/check-dependency.sh browserify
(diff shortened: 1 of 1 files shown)


lurch:
Erm, do we want to version-lock this?

jviotti:
Turns out we already had browserify locked down as a development dependency, but we were still asking the environment to have it. It should be fine now.

lurch:
https://github.com/resin-io/etcher/commits/master is still full of red crosses :-(  Looks like other fixes are needed too...

jviotti:
Hm, I'll make the CI servers build the whole thing on pull requests but not deploy anything, so we catch these things before merging breaking changes.

PR #1359, opened 2017-04-25
refactor(GUI): turn the update notifier modal into a native dialog

jviotti:
Electron v1.6.1 introduced checkbox support to the native message
dialog, giving us everything that was needed to implement the update
notifier modal using a native dialog.

This change allows us to get rid of a lot code.

screenshot 2017-04-25 14 09 02

See: https://github.com/electron/electron/pull/8590
Change-Type: patch
Changelog-Entry: Turn the update notifier modal into a native dialog.


lib/gui/components/update-notifier.js
```diff
@@ -138,17 +166,48 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_DAYS, M
...
+      settings.set('sleepUpdateCheck', false);
+
+      return new Bluebird((resolve) => {
+        const BUTTONS = [
+          'Download',
+          'Skip'
+        ];
+
+        const BUTTON_CONFIRMATION_INDEX = _.indexOf(BUTTONS, _.first(BUTTONS));
+        const BUTTON_REJECTION_INDEX = _.indexOf(BUTTONS, _.last(BUTTONS));
```
lurch: lolz

tests/gui/components/update-notifier.spec.js
```diff
@@ -72,7 +70,8 @@ describe('Browser: UpdateNotifier', function() {
         describe('given the `lastUpdateNotify` was updated long ago', function() {
 
           beforeEach(function() {
-            const SLEEP_MS = units.daysToMilliseconds(UPDATE_NOTIFIER_SLEEP_DAYS);
+            console.log(UpdateNotifierService);
```
lurch: Debug code that should have been removed?

lurch:
Lol, I left my comment before noticing Juan had already requested a view from Benedict ;-)

tests/gui/components/update-notifier.spec.js
jviotti: Oops, fixed!

lib/gui/components/update-notifier.js
```diff
@@ -138,17 +166,48 @@ module.exports = function($http, $q, ModalService, UPDATE_NOTIFIER_SLEEP_DAYS, M
...
-      return ModalService.open({
-        template: './components/update-notifier/templates/update-notifier-modal.tpl.html',
-        controller: 'UpdateNotifierController as modal',
-        size: 'update-notifier',
-        resolve: {
-          options: _.constant({
-            version
-          })
+      settings.set('lastUpdateNotify', Date.now());
+      settings.set('sleepUpdateCheck', false);
```
lurch: I keep struggling to work out what the `sleepUpdateCheck` setting is supposed to do. Is it just me that's confused? :-/

Shou: Looking at it `lastUpdateNotify` is set before anytime the dialog appears so it can't be relied upon alone as-is, because anytime the dialog appears it'd pause it for seven days without the help of `sleepUpdateCheck`. However it doesn't need to be set at the start of the dialog code, we can instead set it if `results.sleepUpdateCheck` is `true`. We don't actually use `lastUpdateNotify` for anything else so this should work fine, unless there are plans to do something like log the space between each update check. I believe this should work:

```diff
diff --git a/lib/gui/components/update-notifier.js b/lib/gui/components/update-notifier.js
index 9fad229..92dfbf0 100644
--- a/lib/gui/components/update-notifier.js
+++ b/lib/gui/components/update-notifier.js
@@ -142,12 +142,11 @@ UpdateNotifier.service('UpdateNotifierService', function(
   this.shouldCheckForUpdates = () => {
     const lastUpdateNotify = settings.get('lastUpdateNotify');
 
-    if (!settings.get('sleepUpdateCheck') || !lastUpdateNotify) {
+    if (!lastUpdateNotify) {
       return true;
     }
 
     if (lastUpdateNotify - Date.now() > units.daysToMilliseconds(this.UPDATE_NOTIFIER_SLEEP_DAYS)) {
-      settings.set('sleepUpdateCheck', false);
       return true;
     }
 
@@ -166,8 +165,6 @@ UpdateNotifier.service('UpdateNotifierService', function(
    */
   this.notify = () => {
     return this.getLatestVersion().then((version) => {
-      settings.set('lastUpdateNotify', Date.now());
-      settings.set('sleepUpdateCheck', false);
 
       return new Bluebird((resolve) => {
         const BUTTONS = [
@@ -194,7 +191,9 @@ UpdateNotifier.service('UpdateNotifierService', function(
           });
         });
       }).then((results) => {
-        settings.set('sleepUpdateCheck', results.sleepUpdateCheck);
+        if (results.sleepUpdateCheck) {
+          settings.set('lastUpdateNotify', Date.now());
+        }
 
         AnalyticsService.logEvent('Close update modal', {
           sleepUpdateCheck: results.sleepUpdateCheck,
```

jviotti: I don't think I get it. The `lastUpdateNotify` option represents the date where the update notification last appeared, so it makes sense that it has to be set to the current time right before showing the dialog, and it will only get set again once we determine that the dialog should be shown again (e.g: after 7 days if the sleep update check was set).

> anytime the dialog appears it'd pause it for seven days without the help of sleepUpdateCheck

Can you capture this as a unit test so I can better understand the issue? I might be missing something, but we only call `.notify()` if `.shouldCheckForUpdates()` returns `true`.

If we show the dialog once, then `lastUpdateNotify` has the current date. Lets say the user didn't set `sleepUpdateCheck`. The user will then open Etcher again the day afterwards. At this point, `lastUpdateNotify` contains the day before time, and `sleepUpdateCheck` is `false`. The `(!settings.get('sleepUpdateCheck') || !lastUpdateNotify)` is `true`, which makes `.shouldCheckForUpdates()` return `true`, and then the dialog is shown (note that the `lastUpdateNotify` is not even considered, given that `!settings.get('sleepUpdateCheck')` is already `true`).

Now lets consider the same scenario if the user did set `sleepUpdateCheck`. The first conditonal, `(!settings.get('sleepUpdateCheck') || !lastUpdateNotify)` is false, because `sleepUpdateCheck` is set and `lastUpdateNotify` has a value, which makes `!lastUpdateNotify` evaluate to `false`.

So the next condition applies: 

```js
if (lastUpdateNotify - Date.now() > units.daysToMilliseconds(this.UPDATE_NOTIFIER_SLEEP_DAYS)) {
  settings.set('sleepUpdateCheck', false);
  return true;
}
```

If the date set on `lastUpdateNotify` is greater than 7 days, then the dialog is shown, and only then the `lastUpdateNotify` variable gets set again. Otherwise, the function returns `false`, and the update notifier is not shown.

Am I missing something?

lurch: I think I can say what @Shou is saying, although without thinking about it in more depth (which I don't have time for right now) I can't 100% follow the code-arguments from either @Shou or @jviotti 

I _think_ @Shou is saying that it only makes sense to update `lastUpdateNotify` **if** the user ticked "remind me again in 7 days". If the user doesn't tick the box, then although we _could_ update `lastUpdateNotify` there's not strictly any need to. Even @jviotti mentioned above "(note that the lastUpdateNotify is not even considered, given that !settings.get('sleepUpdateCheck') is already true)" - I guess @Shou was just taking that one step further.

Shou: Oh sorry, the only thing my diff does is reduce the usage of two settings down to one. There's no bug in the current code, my diff and the current code should behave identically. I was just trying to explain to @lurch why `sleepUpdateCheck` was used, and then during this I realized it isn't necessarily needed hence the diff and all.

jviotti: Oh, I see, thanks a lot for the clarification. I really like the idea of reducing the variables. The only thing that makes me a bit uncomfortable is that `lastUpdateNotify` will no longer represent the time where the update notification was last shown, but only the last time is was shown and the user put it to sleep, but we can fix that by renaming such setting.

jviotti:
@Shou Let me know how it looks now!

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,145 @@
...
+
+    if (_.get(options, [ 'allowSleepUpdateCheck' ], true)) {
+      _.merge(dialogOptions, {
+        checkboxLabel: `Remind me again in ${this.UPDATE_NOTIFIER_SLEEP_DAYS} days`,
+        checkboxChecked: false
+      });
+    }
+
+    return new Bluebird((resolve) => {
+      const currentWindow = electron.remote.getCurrentWindow();
```
Shou: This could be made a constant at the top of the service, I believe.

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,145 @@
...
+   */
+  this.shouldCheckForUpdates = (options = {}) => {
+    if (_.some([
+      !options.lastSleptUpdateNotify,
+      _.get(options, [ 'ignoreSleepUpdateCheck' ], false)
+    ])) {
+      return true;
+    }
+
+    if (options.lastSleptUpdateNotify - Date.now() > units.daysToMilliseconds(this.UPDATE_NOTIFIER_SLEEP_DAYS)) {
```
Shou: So `options.lastSleptUpdateNotify - Date.now()` will always be negative because `options.lastSleptUpdateNotify` will be an older `Date.now()`, hence `options.lastSleptUpdateNotify <= Date.now()` is always true and this results in a negative number that decreases over time, instead of eventually growing past the seven days worth of seconds. So we should swap this to `Date.now() - options.lastSleptUpdateNotify > ...`. I think this was my bad for suggesting this is how it should be in the diff I made, sorry about that.

lib/gui/components/update-notifier.js
jviotti: Sounds like a good idea!

lib/gui/components/update-notifier.js
jviotti: You're totally right. This function really confuses me for some reason, so I'm glad you're on top of it! :D I updated the test cases accordingly. Do you mind checking that those make sense as well?

tests/gui/components/update-notifier.spec.js
```diff
@@ -3,146 +3,112 @@
...
+          const result = UpdateNotifierService.shouldCheckForUpdates({
+            ignoreSleepUpdateCheck: true,
+            lastSleptUpdateNotify: Date.now() - 1000
           });
 
-          describe('given the `lastUpdateNotify` was updated long ago', function() {
-
-            beforeEach(function() {
-              const SLEEP_MS = units.daysToMilliseconds(UPDATE_NOTIFIER_SLEEP_DAYS);
-              settings.set('lastUpdateNotify', Date.now() + SLEEP_MS + 1000);
```
lurch: Hmmm, looks like the unit-tests were only ever passing because `lastUpdateNotify` was being set to a timestamp _in the future_?? (which of course would never actually happen in reality)
So I guess the description of `describe('given the lastUpdateNotify was updated long ago',` was wrong too? Whoops!

In practical terms, does this mean that if somebody ever ticked the "remind me again in 7 days" tickbox, then they'll _never_ get notified about any future Etcher versions? (or am I just getting muddled up myself?) :-S

tests/gui/components/update-notifier.spec.js
```diff
@@ -3,146 +3,112 @@
...
+          m.chai.expect(result).to.be.true;
+        });
 
-              m.chai.expect(result).to.be.true;
-            });
+      });
 
-            it('should unset the `sleepUpdateCheck` setting', function() {
-              m.chai.expect(settings.get('sleepUpdateCheck')).to.be.true;
+      describe('given the `lastSleptUpdateNotify` was updated in the future', function() {
```
lurch: Would there be any harm in adding a `describe('given the lastSleptUpdateNotify was updated far in the future',` test, with `lastSleptUpdateNotify: Date.now() + SLEEP_MS + 1000` ?

lurch:
Hmmm, what happens if the user actually ticks "Remind me again in 7 days" but then clicks "Download" anyway? Perhaps we should only update `lastSleptUpdateNotify` if `(results.sleepUpdateCheck && !results.agreed)` ?

And I wonder if installing a new Etcher version should automatically 'reset' the `lastSleptUpdateNotify` too? (which I guess we could do by saving the version of Etcher used to do the last update-check)

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,145 @@
...
+      _.get(options, [ 'ignoreSleepUpdateCheck' ], false)
+    ])) {
+      return true;
+    }
+
+    if (Date.now() - options.lastSleptUpdateNotify > units.daysToMilliseconds(this.UPDATE_NOTIFIER_SLEEP_DAYS)) {
+      return true;
+    }
+
+    return false;
```
jviotti: Right, I'll fix it!

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,145 @@
...
+        });
+      });
+    }).then((results) => {
+      if (results.sleepUpdateCheck) {
+        settings.set('lastSleptUpdateNotify', Date.now());
+      }
+
+      AnalyticsService.logEvent('Close update modal', {
+        sleepUpdateCheck: results.sleepUpdateCheck,
+        notifyVersion: version
```
jviotti: Done!

lib/gui/components/update-notifier/services/update-notifier.js
```diff
@@ -1,89 +0,0 @@
...
-
-    if (_.some([
-      !settings.get('sleepUpdateCheck'),
-      !lastUpdateNotify,
-      _.get(options, [ 'ignoreSleepUpdateCheck' ], false)
-    ])) {
-      return true;
-    }
-
-    if (lastUpdateNotify - Date.now() > units.daysToMilliseconds(UPDATE_NOTIFIER_SLEEP_DAYS)) {
```
jviotti: Yeah, looks like this whole thing was broken :(

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,145 @@
...
+   *   ignoreSleepUpdateCheck: false,
+   *   lastSleptUpdateNotify: Date.now()
+   * })) {
+   *   console.log('We should check for updates!');
+   * }
+   */
+  this.shouldCheckForUpdates = (options = {}) => {
+    if (_.some([
+      !options.lastSleptUpdateNotify,
+      _.get(options, [ 'ignoreSleepUpdateCheck' ], false)
```
jviotti: Done!

jviotti:
> Hmmm, what happens if the user actually ticks "Remind me again in 7 days" but then clicks "Download" anyway? Perhaps we should only update lastSleptUpdateNotify if (results.sleepUpdateCheck && !results.agreed) ?

Sounds reasonable!

> And I wonder if installing a new Etcher version should automatically 'reset' the lastSleptUpdateNotify too? (which I guess we could do by saving the version of Etcher used to do the last update-check)

Interesting. Let's store the version in the settings, and use it in `lib/gui/app.js` to set `ignoreSleepUpdateCheck` if so.

tests/gui/components/update-notifier.spec.js
jviotti: Sounds like a good idea. I'll add it!

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,145 @@
...
+
+UpdateNotifier.service('UpdateNotifierService', function(AnalyticsService, OSOpenExternalService) {
+
+  /**
+   * @summary The number of days the update notifier can be put to sleep
+   * @constant
+   * @private
+   * @type {Number}
+   */
+  this.UPDATE_NOTIFIER_SLEEP_DAYS = 7;
```
jviotti: Sounds like a good idea!

tests/gui/components/update-notifier.spec.js
jviotti: Indeed, which is terrible :/ . I clearly have major issues reasoning about this component, so lets join forces to make sure I get it right this time (and that the tests really test that).

lib/gui/app.js
```diff
@@ -94,7 +94,8 @@ app.run((AnalyticsService, ErrorService, UpdateNotifierService, SelectionStateMo
   const currentVersion = packageJSON.version;
   const currentReleaseType = release.getReleaseType(currentVersion);
   const shouldCheckForUpdates = UpdateNotifierService.shouldCheckForUpdates({
-    ignoreSleepUpdateCheck: currentReleaseType !== release.RELEASE_TYPE.PRODUCTION
+    ignoreSleepUpdateCheck: currentReleaseType !== release.RELEASE_TYPE.PRODUCTION,
+    lastSleptUpdateNotify: settings.get('lastSleptUpdateNotify')
```
jviotti: Sounds good. Regarding the second suggestion, I propose going the opposite direction: move the actual logic inside the `shouldCheckForUpdates` function, so we can unit test it there (and remove the meaningless `ignoreSleepUpdateChec` option)

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,148 @@
...
+   */
+  this.UPDATE_NOTIFIER_SLEEP_DAYS = packageJSON.updates.sleepDays;
+
+  /**
+   * @summary Determine if it's time to check for updates
+   * @function
+   * @public
+   *
+   * @param {Object} [options] - options
+   * @param {Number} [options.lastSleptUpdateNotifier] - last slept update notify time
```
lurch: Minor nit, but could we change 'notify' -> 'notifier' in the comment?

lib/gui/components/update-notifier.js
lurch: Did you forget to make this change?

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,148 @@
...
+      electron.remote.dialog.showMessageBox(currentWindow, dialogOptions, (response, checkboxChecked) => {
+        return resolve({
+          agreed: response === BUTTON_CONFIRMATION_INDEX,
+          sleepUpdateCheck: checkboxChecked || false
+        });
+      });
+    }).then((results) => {
+
+      // Only update the last update timestamp if the
+      // user ticked the "Reming me again in ..." checkbox,
```
lurch: Reming -> Remind
(perhaps you could add that to our codespell dictionary too)

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,148 @@
...
+      const currentWindow = electron.remote.getCurrentWindow();
+      electron.remote.dialog.showMessageBox(currentWindow, dialogOptions, (response, checkboxChecked) => {
+        return resolve({
+          agreed: response === BUTTON_CONFIRMATION_INDEX,
+          sleepUpdateCheck: checkboxChecked || false
+        });
+      });
+    }).then((results) => {
+
+      // Only update the last update timestamp if the
```
lurch: Could you change "last update timestamp" to "last slept update timestamp" ?

tests/gui/components/update-notifier.spec.js
```diff
@@ -1,148 +1,203 @@
...
 
-            it('should return true', function() {
-              const result = UpdateNotifierService.shouldCheckForUpdates({
-                ignoreSleepUpdateCheck: this.ignoreSleepUpdateCheck
-              });
+      describe('given the `lastSleptUpdateNotifier` was updated in the future', function() {
 
-              m.chai.expect(result).to.be.true;
-            });
+        it('should return false if releaseType is production', function() {
```
lurch: Hmmm, thinking about it, does it make sense for `shouldCheckForUpdates` to return `false` if `lastSleptUpdateNotifier` was updated in the future? Perhaps we should treat a future `lastSleptUpdateNotifier` as an 'error condition' and always return `true`?

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,148 @@
...
+   * @type {Number}
+   */
+  this.UPDATE_NOTIFIER_SLEEP_DAYS = packageJSON.updates.sleepDays;
+
+  /**
+   * @summary Determine if it's time to check for updates
+   * @function
+   * @public
+   *
+   * @param {Object} [options] - options
```
lurch: Given that `options.releaseType` isn't in square brackets (indicating that it's non-optional?), should `options` here also not be in square brackets?

lib/gui/app.js
```diff
@@ -94,13 +94,13 @@ app.run((AnalyticsService, ErrorService, UpdateNotifierService, SelectionStateMo
   const currentVersion = packageJSON.version;
   const currentReleaseType = release.getReleaseType(currentVersion);
   const shouldCheckForUpdates = UpdateNotifierService.shouldCheckForUpdates({
-    ignoreSleepUpdateCheck: currentReleaseType !== release.RELEASE_TYPE.PRODUCTION
```
lurch: Yay, nice to see the removal of the confusing `ignoreSleepUpdateCheck` parameter :+1:

lurch:
> Interesting. Let's store the version in the settings, and use it in lib/gui/app.js to set ignoreSleepUpdateCheck if so.

Are you planning to make that change in this PR, or in a separate PR?

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,148 @@
...
+   * if (UpdateNotifierService.shouldCheckForUpdates({
+   *   lastSleptUpdateNotifier: Date.now(),
+   *   releaseType: release.RELEASE_TYPE.PRODUCTION
+   * })) {
+   *   console.log('We should check for updates!');
+   * }
+   */
+  this.shouldCheckForUpdates = (options = {}) => {
+    if (_.some([
+      !_.get(options, [ 'lastSleptUpdateNotifier' ], false),
```
lurch: Would it be more consistent with the unit-tests if this defaulted to `undefined` rather than `false`?

lurch:
I just installed rc5 from the bintray apt-get, and noticed that the settings screen still has the "Include unstable update channel", which looks a bit odd. Perhaps if `ETCHER_DISABLE_UPDATES` is set (as it is for apt-get installed versions of Etcher), then that setting should be automatically hidden?

lib/gui/components/update-notifier.js
jviotti: Done!

lib/gui/components/update-notifier.js
jviotti: Done!

lib/gui/components/update-notifier.js
jviotti: Done!

lib/gui/components/update-notifier.js
jviotti: Done!

lib/gui/components/update-notifier.js
jviotti: Done!

jviotti:
> Are you planning to make that change in this PR, or in a separate PR?

I just pushed this change :)

jviotti:
> I just installed rc5 from the bintray apt-get, and noticed that the settings screen still has the "Include unstable update channel", which looks a bit odd. Perhaps if ETCHER_DISABLE_UPDATES is set (as it is for apt-get installed versions of Etcher), then that setting should be automatically hidden?

Done!

lib/gui/app.js
```diff
@@ -94,13 +94,14 @@ app.run((AnalyticsService, ErrorService, UpdateNotifierService, SelectionStateMo
...
+    releaseType: currentReleaseType,
+    lastSleptUpdateNotifier: settings.get('lastSleptUpdateNotifier'),
+    lastSleptUpdateNotifierVersion: settings.get('lastSleptUpdateNotifierVersion') || packageJSON.version
```
lurch: I wonder if it might make sense to invert the logic, and always pass in the `packageJSON.version` here, and then have `.shouldCheckForUpdates` check that against `settings.get('lastSleptUpdateNotifierVersion')` ?

And then potentially `.shouldCheckForUpdates` could itself also get the `releaseType` from the version string?

lib/gui/app.js
```diff
@@ -94,13 +94,14 @@ app.run((AnalyticsService, ErrorService, UpdateNotifierService, SelectionStateMo
...
   });
 
   if (_.some([
     !shouldCheckForUpdates,
-    process.env.ETCHER_DISABLE_UPDATES,
-    currentReleaseType === release.RELEASE_TYPE.UNKNOWN
+    process.env.ETCHER_DISABLE_UPDATES
```
lurch: Is this change "correct"? Previously if we had a `release.RELEASE_TYPE.UNKNOWN` then it would always skip the update check, but now it doesn't do that? If we have a `release.RELEASE_TYPE.UNKNOWN` then we can't know what to update to, can we?

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,159 @@
...
+   *   lastSleptUpdateNotifier: Date.now(),
+   *   lastSleptUpdateNotifierVersion: '1.0.0',
+   *   releaseType: release.RELEASE_TYPE.PRODUCTION
+   * })) {
+   *   console.log('We should check for updates!');
+   * }
+   */
+  this.shouldCheckForUpdates = (options) => {
+    if (_.some([
+      !options.lastSleptUpdateNotifier,
```
lurch: As the docstring indicates these as optional, should they be using the `_.get(...` syntax?
(I'm just trying to understand why sometimes you use `_.get`, and other times you don't)

lib/gui/pages/settings/controllers/settings.js
```diff
@@ -55,6 +55,13 @@ module.exports = function(WarningModalService, ErrorService, AnalyticsService) {
   this.model = settings;
 
   /**
+   * @summary Environment
+   * @type {Object}
+   * @public
+   */
+  this.environment = process.env;
```
lurch: I wonder if it would be better to _just_ expose `ETCHER_DISABLE_UPDATES`, rather than exposing the whole `process.env` ?

lib/gui/components/update-notifier.js
```diff
@@ -0,0 +1,159 @@
...
+      if (results.sleepUpdateCheck && !results.agreed) {
+        settings.set('lastSleptUpdateNotifier', Date.now());
+        settings.set('lastSleptUpdateNotifierVersion', packageJSON.version);
+      }
+
+      AnalyticsService.logEvent('Close update modal', {
+        sleepUpdateCheck: results.sleepUpdateCheck,
+        notifyVersion: version,
+        agreed: results.agreed
+      });
```
lurch: Hmmm, I wonder if it'd be worth this logging both the `notifyVersion` _and_ the `currentVersion` ? ;-)

lib/gui/app.js
lurch: (or perhaps we should move this logic directly into `.shouldCheckForUpdates`?)

jviotti: Note that whether we should update or not based on the release type is now being handled by `UpdateNotifierService.shouldCheckForUpdates()`, so the logic should remain the same.

lib/gui/components/update-notifier.js
jviotti: The `options` object is not optional anymore, so its safe to use the dot notation. I usually only use `_.get` if we don't know whether the base object exists or not (`_.get` takes care of that automatically).

lib/gui/components/update-notifier.js
jviotti: Sounds like a great suggestion, will do!

lib/gui/pages/settings/controllers/settings.js
jviotti: Makes sense, done!

lib/gui/app.js
jviotti: Yeah, sounds like a good idea. I'll refactor it!

jviotti:
@Shou @lurch How do you feel about this?

lurch:
I still want to add some comments to this, but just busy looking at the filename-escaping issues ATM :-/

jviotti:
Re-ping @lurch @Shou

PR #1363, opened 2017-04-26
fix(image-stream): Remove inability to fallback to octet-stream

jhermsmeier:
Using `mime-types` in that place made it impossible to use other
other file extensions like `.sdcard` and have them treated as `application/octet-stream`
when drag-and-dropping images into Etcher.

This moves the fallback logic in `lib/image-stream` to `.getFromFilePath()`,
to still facilitate proper MIME type detection, while allowing it to fall back
to the octet-stream handler, if there's no available handler for a specific type.

Change-Type: patch
Changelog-Entry: Fix not treating unknown images as octet-stream


tests/image-stream/utils.spec.js
```diff
@@ -61,13 +61,6 @@ describe('ImageStream: Utils', function() {
       });
     });
 
-    it('should resolve application/octet-stream for an uncompressed iso', function() {
```
jviotti: Can we keep this test case with the right mime type now?

lib/image-stream/utils.js
```diff
@@ -41,10 +41,6 @@ exports.getArchiveMimeType = (filename) => {
   const mimeType = mime.lookup(filename);
 
   if (mimeType) {
-    if (mimeType === 'application/x-iso9660-image') {
-      return Bluebird.resolve(MIME_TYPE_RAW_IMAGE);
```
jviotti: Can `MIME_TYPE_RAW_IMAGE` be removed from this function now?

lib/image-stream/index.js
```diff
@@ -76,14 +76,13 @@ exports.getFromFilePath = (file) => {
     }
 
     return utils.getArchiveMimeType(file).then((type) => {
+      const MIME_TYPE_RAW_IMAGE = 'application/octet-stream';
+      let mimeType = type;
```
jviotti: This can be simplified to:

```js
const MIME_TYPE_RAW_IMAGE = 'application/octet-stream';
const mimeType = _.get(handlers, type, MIME_TYPE_RAW_IMAGE);
return _.invoke(handlers, mimeType, file, {
    size: fileStats.size
});
```

jviotti: Ah, wait, I got the logic wrong... Let me fix it...

jviotti: This should do it, so we avoid the `let`:

```js
const MIME_TYPE_RAW_IMAGE = 'application/octet-stream';
const mimeType = _.has(handlers, type) ? type : MIME_TYPE_RAW_IMAGE;
return _.invoke(handlers, mimeType, file, {
  size: fileStats.size
});
```

jviotti:
Can we add new tests that cover this new behavior?

lib/image-stream/utils.js
jhermsmeier: I't still used in the fallback below – should we just pass the literal string there, or keep the var?

tests/image-stream/utils.spec.js
jhermsmeier: Yup, makes more sense, done

lib/image-stream/index.js
jhermsmeier: 👍  Updated

jhermsmeier:
> Can we add new tests that cover this new behavior?

Yes, added a new test file and test for an unrecognized extension :)

lib/image-stream/utils.js
jviotti: Ah, the GitHub diff didn't reveal that. Ignore me then :)

lurch:
Is the `random.rpi-sdcard` file that this PR adds actually used by any of the unit-tests? (perhaps it is, but I can't tell from the Travis logs)

And it looks like we need to add `*.rpi-sdcard` to the list of extensions that codespell should ignore?

jviotti:
I sent a PR to ignore the extension. The image file is used by this test: https://github.com/resin-io/etcher/blob/master/tests/image-stream/utils.spec.js#L86

lurch:
Doesn't that test just operate on strings though, and so not actually need a physical file?

jhermsmeier:
@lurch `utils.getArchiveMimeType()` will do I/O to figure out the type if it doesn't match known stuff via `mime.lookup()` – which is the case with `.rpi-sdcard` (because it's not a registered extension) – so needs a corresponding file.

lurch:
Ah, didn't realise that. Thanks for the clarification :+1:

PR #1364, opened 2017-04-26
chore: build on Ubuntu 12.04 Docker base images

jviotti:
We recently discovered that users not running latest GNU/Linux
distributions were having problems when Etcher tried to load native
modules. We used to build packages on Ubuntu 16.04, which ships with
GLIBC 2.23.

In order to solve that issue, we moved our build Docker containers to
Debian Jessie, which ships with GLIBC 2.19, however we noticed that
users running even older Linux distributions were still having the same
issue, so this commit switched the Docker containers to Ubuntu 12.04,
which ships with GLIBC 2.15.

See: https://github.com/resin-io/etcher/pull/1319

### DIFF
--- a/scripts/build/docker/Dockerfile-i686
+++ b/scripts/build/docker/Dockerfile-i686
@@ -1,25 +1,34 @@
-FROM 32bit/debian:jessie
+FROM erwinchang/ubuntu-12.04-32bit-build
 
 # Install dependencies
-RUN apt-get update && apt-get install -y \
-  build-essential \
-  curl \
-  fuse \
-  git \
-  jq \
-  libasound2 \
-  libgconf-2-4 \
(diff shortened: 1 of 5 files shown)


lurch:
Rather than committing near-identical changes to 3 separate files, each time `Dockerfile.template` gets modified, I guess we could remove the arch-specific Dockerfiles from github, and have `run-command.sh` create them on demand?

jviotti:
Yeah, could be. We're going to introduce ARM Dockerfiles soon, which look like will need some different steps, so I wonder if we should maintain them separate, rather than introducing a complex template mechanism.

lurch:
Fair enough, I'll wait to see what happens with the ARM versions then before making any drastic changes ;)

PR #1365, opened 2017-04-27
chore: create installers (but don't publish) on every pull request

jviotti:
This allows us to catch changes that break our installer builds before
merging the problematic changes.

As a way to simplify the CI configuration files, this commit introduces
an `installers-all` Makefile target that builds all installers.

This commit also replaces all the `cp -rf` calls with `cp -RLf` in
Makefile to avoid some weird hard link Appveyor issues.

See: https://github.com/resin-io/etcher/pull/1078
See: https://github.com/resin-io/etcher/pull/1354

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,8 @@ TARGET_ARCH_DEBIAN = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET
 PRODUCT_NAME = etcher
 APPLICATION_NAME_LOWERCASE = $(shell echo $(APPLICATION_NAME) | tr A-Z a-z)
 APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
+
+# Fix hard link Appveyor issues
 CPRF = cp -RLf
 
 # ---------------------------------------------------------------------
@@ -430,6 +432,7 @@ TARGETS = \
 	package-electron \
 	package-cli \
 	cli-develop \
+	installers-all \
 	electron-develop
 
 package-electron: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)
@@ -477,6 +480,8 @@ PUBLISH_AWS_S3 += \
 	$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe \
 	$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
 endif
+
+installers-all: $(PUBLISH_AWS_S3) $(PUBLISH_BINTRAY_DEBIAN)
 
 ifdef PUBLISH_AWS_S3
 publish-aws-s3: $(PUBLISH_AWS_S3)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,10 +35,11 @@ build: off
 test_script:
   - node --version
   - npm --version
-  - cmd: .\scripts\ci\appveyor-test.bat
+  - .\scripts\ci\appveyor-test.bat
+  - .\scripts\ci\appveyor-build-installers.bat
 
 deploy_script:
-  - cmd: .\scripts\ci\appveyor-deploy.bat
+  - .\scripts\ci\appveyor-deploy.bat
 
 notifications:
 
--- a/scripts/ci/travis-build-installers.sh
+++ b/scripts/ci/travis-build-installers.sh
@@ -24,15 +24,13 @@ if [ -z "$TRAVIS_OS_NAME" ]; then
   exit 1
 fi
 
-./scripts/build/check-dependency.sh npm
 ./scripts/build/check-dependency.sh make
 
 if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
   ./scripts/build/docker/run-command.sh \
     -r "$TARGET_ARCH" \
     -s "$(pwd)" \
-    -c 'make sanity-checks && xvfb-run --server-args=$XVFB_ARGS npm test'
+    -c 'make installers-all'
 else
-  make sanity-checks
-  npm test
+  make installers-all
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ install:
 
 script:
   - ./scripts/ci/travis-test.sh
+  - ./scripts/ci/travis-build-installers.sh
 
 deploy:
   provider: script
--- a/scripts/ci/appveyor-build-installers.bat
+++ b/scripts/ci/appveyor-build-installers.bat
@@ -1,0 +1,24 @@
+@echo off
+
+:: Copyright 2017 resin.io
+::
+:: Licensed under the Apache License, Version 2.0 (the "License");
+:: you may not use this file except in compliance with the License.
+:: You may obtain a copy of the License at
+::
+::   http://www.apache.org/licenses/LICENSE-2.0
+::
+:: Unless required by applicable law or agreed to in writing, software
+:: distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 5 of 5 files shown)


jviotti:
And the culprit was a `wget` call in one of the CLI build scripts :)

lurch:
Still failing on Appveyor, and apparently there's a conflict in `Makefile` :-(

Also, a minor nit, but could you change the commit / PR title to "chore: create installers (but don't publish) on every pull request" ? The current title makes it sound like you're not creating installers when publishing ;-)

Makefile
```diff
@@ -476,6 +482,7 @@ PUBLISH_AWS_S3 += \
 	$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip \
 	$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe \
 	$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
+installers-all: $(PUBLISH_AWS_S3)
```
lurch: IIRC if the platform isn't Linux, then `$(PUBLISH_BINTRAY_DEBIAN)` will just be an empty variable, so you might be able to do `installers-all: $(PUBLISH_AWS_S3) $(PUBLISH_BINTRAY_DEBIAN)` just once (at the 'global' level), instead of having it inside multiple `ifeq` blocks?

appveyor.yml
```diff
@@ -43,6 +43,7 @@ test_script:
   - npm --version
   - make sanity-checks
   - cmd: npm test
+  - cmd: make installers-all
```
lurch: Just out of curiosity, why is there a `cmd: ` in front of `make installers-all` but not in front of `make sanity-checks` ?

Makefile
jviotti: Indeed, fixed!

appveyor.yml
jviotti: No reason, really. Appveyor defaults to `cmd` if that's not set. I'll remove them in this PR.

Makefile
```diff
@@ -245,7 +248,7 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERS
 	mkdir $@
 	cp $(word 1,$^) $@
 	cp $(word 2,$^) $@
```
lurch: Based on the CI failures, I wonder if this might need to be using `$(CPRF)` too?

lurch:
Probably worth merging #1378 first, and then rebasing this one afterwards?

lurch:
Still failing on Appveyor, but at least it's a slightly different error this time ;-)

jviotti:
What do you think about this @lurch ?

lurch:
CI tests *seem* to be passing, but are printing:
```
> lzma-native@1.5.2 install /Users/travis/build/resin-io/etcher/release/node-darwin-x64-dependencies/node_modules/lzma-native
> node-pre-gyp install --fallback-to-build
node-pre-gyp ERR! UNCAUGHT EXCEPTION 
node-pre-gyp ERR! stack Error: Unsupported target version: 6.10.3
```
?

jviotti:
The issue is gone! Looks like passing `-f` to `npm install` (which we did because of optional dependencies), causes npm to continue even if errors are encountered along the way. I got rid of this option and everything is back to normal.

lurch:
"back to normal" with all the CI builds failing again? :rofl:

scripts/ci/travis-test.sh
```diff
@@ -28,8 +28,9 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
   ./scripts/build/docker/run-command.sh \
     -r "$TARGET_ARCH" \
     -s "$(pwd)" \
-    -c 'make sanity-checks && xvfb-run --server-args=$XVFB_ARGS npm test'
+    -c 'make sanity-checks && xvfb-run --server-args=$XVFB_ARGS npm test && make installers-all'
 else
   make sanity-checks
   npm test
+  make installers-all
```
lurch: Not sure if it makes sense to have `make installers-all` inside the "test" scripts?

jhermsmeier:
Why is gyp trying to fetch `iojs-1.6.6` (https://travis-ci.org/resin-io/etcher/jobs/233069196#L6131)?

Makefile
```diff
@@ -19,7 +19,7 @@ BUILD_OUTPUT_DIRECTORY = $(BUILD_DIRECTORY)/out
 # ---------------------------------------------------------------------
 
 ELECTRON_VERSION = $(shell jq -r '.devDependencies["electron"]' package.json)
-NODE_VERSION = 6.10.3
+NODE_VERSION = 6.1.0
```
lurch: What impact does switching to a lower nodeJS version have?
ISTR you saying in gitter that the version of `node-pre-gyp` we're using doesn't work with newer node versions - would upgrading to a newer version of `node-pre-gyp` fix that (and allow us to use Node 6.10.3), or does upgrading `node-pre-gyp` come with other problems?

jviotti: I tried upgrading, but not even the latest `node-pre-gyp` version supports 6.10.3. 6.1 has the same ABI as 6.10.3, so everything works.

scripts/ci/travis-test.sh
jviotti: Its the CI test scripts. I agree it'd look weird to have it in our main development test commands, but looks fine here. What do you think? Do you have any other alternatives in mind?

jviotti:
> "back to normal" with all the CI builds failing again? 

Precisely :P

>Why is gyp trying to fetch iojs-1.6.6 (https://travis-ci.org/resin-io/etcher/jobs/233069196#L6131)?

Hm, it should be downloading Electron 1.6.6 headers instead. I'll take a look.

jviotti:
Hm, the target is correctly set to `electron` (https://travis-ci.org/resin-io/etcher/jobs/233069196#L6123), and our npm dependencies script is configured exactly as https://github.com/electron/electron/blob/master/docs/tutorial/using-native-node-modules.md :/

scripts/ci/travis-test.sh
lurch: I was thinking it probably ought to be in a separate `travis-build-installers.sh` script?

I just found it very odd that the TravisCI log said something like `scripts/ci/travis-test.sh exited with code 2`, and then scrolling up a bit further the error was actually coming from NPM.

\*shrug\* But if you think it's not worth it I guess the current approach works fine.

Makefile
lurch: Fair enough. So it doesn't matter that `appveyor.yml` still says `nodejs_version: "6.10.3"` ?

jviotti: It shouldn't matter, since the ABI is the same. The only advantage of running 6.10.3 is that it comes with the exact npm version that we're interested in, so we don't have to manually install it afterwards.

scripts/ci/travis-test.sh
jviotti: Sounds good. I don't have any strong opinions, so let's go for it!

scripts/ci/travis-build-installers.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit 1
+fi
+
+if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+  ./scripts/build/docker/run-command.sh \
+    -r "$TARGET_ARCH" \
+    -s "$(pwd)" \
+    -c 'make installers-all'
+else
+  make installers-all
```
lurch: You could add a dependency-check for the `make` command? ;)

jviotti: Good catch, done!

lurch:
:tada:

jhermsmeier:
> Hm, the target is correctly set to electron (https://travis-ci.org/resin-io/etcher/jobs/233069196#L6123), and our npm dependencies script is configured exactly as https://github.com/electron/electron/blob/master/docs/tutorial/using-native-node-modules.md :/

@jviotti did you find out what caused this?

PR #1366, opened 2017-04-27
feat(GUI): implement Windows elevation using a native module

jviotti:
Sentry error reports showcase that elevation errors on Windows are one
of the most frequent Windows errors.

In order to perform Windows elevation, we ship compiled EXEs of a third
party CLI elevation application (http://code.kliu.org/misc/elevate/)
that has several limitations:

- We have the scan the output of the script to determine if a user
  cancelled the elevation request, which causes all sorts of issues on
  computers where English is not the main language

- The application displays a `cmd.exe` window for some milliseconds,
  which is bad UX, that we have to workaround by distributing a patched
  version of the tool

- The CLI application has to be spawned, which seems to be problematic
  if users have anti-virus software, leading to hard to debug issues

- We don't have any control if something goes wrong

For these reasons, we decided to implement our own elevation mechanism
in C++ as a Node.js add-on, based on the `elevate.exe` code we where
previously using.

Misc changes:

- Introduce a `lib/shared/bindings.js` module to easily require local native
  add-ons

- Install `cpplint` and configure it to lint C++ files

Note that for practical reasons, the C++ code lives in this repository
rather than in a separate module. We will release this functionality in
a more accessible way in the future as part of the Etcher SDK project.

Change-Type: patch
Changelog-Entry: Fix uncaught errors when cancelling elevation requests on Windows when the system's language is not English.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,40 +174,40 @@ $(BUILD_DIRECTORY):
 $(BUILD_TEMPORARY_DIRECTORY): | $(BUILD_DIRECTORY)
 	mkdir $@
 
-$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: | $(BUILD_DIRECTORY)
-	mkdir $@
-
-$(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: | $(BUILD_DIRECTORY)
-	mkdir $@
-
 $(BUILD_OUTPUT_DIRECTORY): | $(BUILD_DIRECTORY)
 	mkdir $@
 
-$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json \
-	| $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: package.json npm-shrinkwrap.json \
+	| $(BUILD_DIRECTORY)
+	mkdir $@
+	cp -rf src $@
 	./scripts/build/dependencies-npm.sh -p \
 		-r "$(TARGET_ARCH)" \
 		-v "$(ELECTRON_VERSION)" \
-		-x $| \
+		-x $@ \
 		-t electron \
 		-s "$(TARGET_PLATFORM)"
+	rm -rf $@/src
 
-$(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json \
-	| $(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
+$(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: package.json npm-shrinkwrap.json \
+	| $(BUILD_DIRECTORY)
+	mkdir $@
+	cp -rf src $@
 	./scripts/build/dependencies-npm.sh -p -f \
 		-r "$(TARGET_ARCH)" \
 		-v "$(NODE_VERSION)" \
-		-x $| \
+		-x $@ \
 		-t node \
 		-s "$(TARGET_PLATFORM)"
+	rm -rf $@/src
 
 $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH)-app: \
-	$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules \
+	$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies \
 	| $(BUILD_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
 	./scripts/build/electron-create-resources-app.sh -s . -o $@ \
 		-v $(APPLICATION_VERSION) \
 		-f "$(APPLICATION_FILES)"
-	cp -RLf $< $@
+	cp -RLf $</* $@
 
 ifdef ANALYTICS_SENTRY_TOKEN
 	./scripts/build/jq-insert.sh \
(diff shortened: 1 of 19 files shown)


jviotti:
I'm sending the PR so we can start the review process in the mid-time, but there are still some rough edges I have the address (like C++ linting warnings).

jviotti:
Testing plan:

- Windows 10 from `npm start`
- Windows 10 packaged
- Windows 8 from `npm start`
- Windows 8 packaged
- Windows 7 from `npm start`
- Windows 7 packaged
- macOS from `npm start`
- macOS packaged
- GNU/Linux from `npm start`
- GNU/Linux packaged (AppImage)
- GNU/Linux from `npm start`
- GNU/Linux packaged (debian)

jviotti:
I tried to add many comments to demystify the Windows C APIs that we're using here. Please let me know if we can make things clearer somehow!

src/elevator_init.cpp
```diff
@@ -0,0 +1,88 @@
...
+    break;
+  case etcher::ELEVATE_RESULT::ELEVATE_NOT_ENOUGH_MEMORY:
+    YIELD_ERROR(callback,
+                "There is not enough memory to perform the specified action");
+    break;
+  case etcher::ELEVATE_RESULT::ELEVATE_SHARING_VIOLATION:
+    YIELD_ERROR(callback, "A sharing violation occurred");
+    break;
+  default:
+    YIELD_ERROR(callback, "Unknown error");
```
jhermsmeier: Can we create a method like `etcher::get_error_message( DWORD error )` that directly translates an error code? That way we could squash this to only one `YIELD_ERROR(callback, get_error_message(result))`, and potentially also get rid of the YIELD_* macros.

jviotti: Sounds good, done!

src/os/elevate.cpp
```diff
@@ -0,0 +1,155 @@
...
+  delete[] file;
+  delete[] argv;
+
+  // Finally, let's try to elevate the command
+  if (!executeResult) {
+    DWORD executeError = GetLastError();
+
+    switch (executeError) {
+    case ERROR_FILE_NOT_FOUND:
+      return etcher::ELEVATE_RESULT::ELEVATE_FILE_NOT_FOUND;
```
jhermsmeier: Why not directly translate the `DWORD executeError`, and void remapping these?

jviotti: I don't want to return strings from this function, but constants that we can easily check for later on. We could make this function return a `DWORD`, however not all errors coming from the Windows API are represented using `DWORD`, so its nice to have an extra layer (the `enum class`) for normalization purposes (both to encapsulate Windows specific details and present a uniform error interface).

jviotti:
I'm going to merge this so we can release today, but as always, I'm happy to address any comments later on.

src/os/elevate.cpp
jhermsmeier: I see, makes sense – would it be worth to add a comment about that, so we don't forget?

binding.gyp
```diff
@@ -0,0 +1,25 @@
...
+      "include_dirs" : [
+        "src",
+        "<!(node -e \"require('nan')\")"
+      ],
+      'conditions': [
+
+        [ 'OS=="win"', {
+          "sources": [
+            "src/utils/v8utils.cpp",
+            "src/os/elevate.cpp",
```
lurch: From looking through the code further down, it looks like this is all Windows-specific? Should "src/os/elevate.cpp" be at "src/win32/elevate.cpp" instead?

lib/shared/permissions.js
```diff
@@ -145,22 +146,13 @@ exports.elevateCommand = (command, options) => {
...
-
-    return elevator.executeAsync(prefixedCommand, {
-      hidden: true,
-      terminating: true,
-      doNotPushdCurrentDirectory: true,
-      waitForTermination: true
-    }).then(() => {
+    const elevator = Bluebird.promisifyAll(bindings.load('elevator'));
+    return elevator.elevateAsync(_.concat([
+      'cmd.exe',
```
lurch: Why do we still need to spawn a `cmd.exe`?

lib/shared/permissions.js
```diff
@@ -145,22 +146,13 @@ exports.elevateCommand = (command, options) => {
...
   if (os.platform() === 'win32') {
-    const elevator = Bluebird.promisifyAll(require('elevator'));
-
-    return elevator.executeAsync(prefixedCommand, {
-      hidden: true,
-      terminating: true,
-      doNotPushdCurrentDirectory: true,
-      waitForTermination: true
-    }).then(() => {
+    const elevator = Bluebird.promisifyAll(bindings.load('elevator'));
```
lurch: Would it be worth me adapting my CI bash script which checks all the `require` calls, to also check all the `bindings.load` calls?

lurch:
Have you also tested with Etcher being run in a directory-name containing spaces, and trying to flash an image from a directory-name containing spaces and ampersands? (and you could test directory names containing single-quotes too, like `C:\foo\Juanchi's files & folders\another dir\some image.img` ?)

jviotti:
I checked the ampersand/spaces/quotes test idea and it works fine on Windows 10. Regarding unit tests, I guess that will be covered by https://github.com/resin-io/etcher/pull/1376

lib/shared/permissions.js
jviotti: The thing is that the location of the final add-on depends on the platform, and on the gyp version, so you would have to implement some non-trivial logic to validate this.

lib/shared/permissions.js
jviotti: We need to pass environment variables to the child process, and the only way I found to do that on Windows is to spawn a `cmd.exe`, set the environment variables in that context, and run the application afterwards.

binding.gyp
jviotti: Makes sense, I'll fix it!

lib/shared/permissions.js
lurch: Right, that makes sense. The only downside is that the UAC prompt asks to elevate "Command Prompt" which I've always thought looks a bit weird :-/
How hard would it be to pass the "necessary info" as command-line arguments instead of environment variables? (which would mean we wouldn't then need to rely on `cmd.exe`?)

lib/shared/permissions.js
lurch: Fair enough.

src/os/elevate.cpp
jviotti: Yep, will do!

lib/shared/permissions.js
```diff
@@ -17,6 +17,7 @@
 'use strict';
 
 const os = require('os');
+const bindings = require('./bindings');
```
jviotti: Makes sense. I renamed this to `native-module.js`

lib/shared/permissions.js
jviotti: We can refactor the writer proxy to accept most of them, however `ELECTRON_RUN_AS_NODE` will need to be passed as an env var :(

Makefile
```diff
@@ -174,40 +174,40 @@ $(BUILD_DIRECTORY):
...
-
 $(BUILD_OUTPUT_DIRECTORY): | $(BUILD_DIRECTORY)
 	mkdir $@
 
-$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies/node_modules: package.json npm-shrinkwrap.json \
-	| $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies
+$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: package.json npm-shrinkwrap.json \
+	| $(BUILD_DIRECTORY)
+	mkdir $@
+	cp -rf src $@
```
jviotti: Fixed!

lib/shared/permissions.js
lurch: Hmmm, what a pain :(

Makefile
```diff
@@ -174,40 +174,40 @@ $(BUILD_DIRECTORY):
...
 	./scripts/build/dependencies-npm.sh -p \
 		-r "$(TARGET_ARCH)" \
 		-v "$(ELECTRON_VERSION)" \
-		-x $| \
+		-x $@ \
 		-t electron \
 		-s "$(TARGET_PLATFORM)"
+	rm -rf $@/src
```
jviotti: The issue is that installing the dependencies require the source code to be present in the same directory, however we can't make any assumption in the script that `src` is the correct directory, or that its even the only one.

Regarding naming, `src` and `lib` are pretty standard directory names in the node community. `src` contains native code, and `lib` contains JavaScript code.

Hm, actually, given that `src` is pretty standard I think we *can* make that assumption.

lurch: I'd be perfectly happy with `if [ -d src ]; then` checks inside `dependencies-npm.sh` ;-)

PR #1367, opened 2017-04-28
chore(gitattributes): Add "hex" diff tag to binary files

jhermsmeier:
This adds a `diff` tag to binary files in `.gitattributes`,
enabling git to use an external tool to generate diffable output for binary files
by adding a handler to the local or global git config, i.e:

```
[diff "hex"]
  textconv = hexdump
  binary = true
```

screen shot 2017-04-13 at 03 04 23

Change-Type: patch

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -22,14 +22,14 @@ Makefile text
 *.yml text
 
 # Binary files (no line-ending conversions)
-*.bz2 binary
-*.gz binary
-*.icns binary
-*.ico binary
-*.img binary
-*.iso binary
-*.png binary
-*.xz binary
-*.zip binary
-*.dmg binary
-*.rpi-sdcard binary
+*.bz2 binary diff=hex
+*.gz binary diff=hex
+*.icns binary diff=hex
+*.ico binary diff=hex
+*.img binary diff=hex
+*.iso binary diff=hex
+*.png binary diff=hex
+*.xz binary diff=hex
+*.zip binary diff=hex
+*.dmg binary diff=hex
+*.rpi-sdcard binary diff=hex


jviotti:
Amazing! I must confess I'm really curious about what you're using this for :)

jhermsmeier:
> Amazing! I must confess I'm really curious about what you're using this for :)

Hehe 😄 

The story behind it: I got a bit frustrated with the test .dmg images in #1248 differing from their source image (the raspberry test image file), and no nice way to diff them. So I started looking for ways to get git to do binary diffs, and found the gitattribute tags and textconv settings deep in the docs.
Then I got a bit frustrated with how slow `hexdump` was, and wrote [hxd](https://github.com/jhermsmeier/hxd), which streams, is pageable and 2x as fast as hexdump – and that's what I'm using now to do bindiffs on test images in Etcher :)

jviotti:
Cool stuff. I'm thinking if we could somehow make use of these diffs for something else. Do you it would make sense to display something like that (or allow it to save it somehow) to the user when there is a validation error? Alternatively, if most users won't be able to make sense out of it, would it make sense to send it to us so we can inspect it?

lurch:
@jhermsmeier Would it be worth adding hxd as a devDependency, and then setting up the unit-tests so that if one of the image-writing-tests fails, it automatically outputs the hex diff of the failed file? (or is that best still left as a manual step, due to the large amount of output it'd be likely to generate?)

jhermsmeier:
I don't think so – we'd have to write a bunch more code in the tests to even display the diffs – it'll be a lot easier & faster to just run the failing ones locally and inspect the output.

lurch:
Perhaps you could add some notes to one of the docs, explaining how to do this hex-diff for any image that fails the unit-tests? ;)

jhermsmeier:
@lurch yup, good idea! Added in https://github.com/resin-io/etcher/pull/1380

Issue #993, opened 2016-12-30
Add an information tooltip for the CRC32 checksum


jviotti:
From http://www.linuxlinks.com/article/20161223100458799/Etcher.html

> The user is presented with a CRC32 checksum for the image, but there is no explanation what this means or the purpose of the checksum. A little bit of research reveals the CRC is an acronym for Cyclic Redundancy Code, and the number 32 specifies the size of the hash value - 32 bits. The CRC is an error-detecting code.

Maybe we can add an "i" tooltip explaining how to manually calculate the CRC32 checksum of the image and validate it yourself.

jviotti:
Pinging @alexandrosm , @lurch , @Shou , and @konmouz for input on what/how to explain.

konmouz:
Or you could click the 'CRC32 CHECKSUM' itself to get more info (the way you click 'others' under 'select image'). @jviotti can you estimate the size of this explanation? Can we fit it in a tooltip or we need more space? Is it possible to have a tooltip for a basic description and a link to redirect to an online resource?

lurch:
This is obviously also related to #643 
Given that the usage of MD5 for checksums is _much_ more common than CRC32, maybe if we switched from CRC32 to MD5 (TBH I've got no idea why CRC32 was chosen originally?), then maybe it'd be more obvious and not actually need any explanation?

jviotti:
> @jviotti can you estimate the size of this explanation? Can we fit it in a tooltip or we need more space? Is it possible to have a tooltip for a basic description and a link to redirect to an online resource?

It depends on how we want to do it. We can either:

- Show a very small tooltip that describes what this is for in a few words (a short sentence)
- Show a modal with an extended explanation (I estimate about 2 or 3 sentences at most)
- Redirect to an article online (could be part of the user documentation)

I wonder if we should redirect somewhere else though... I have the feeling that any information should be shown in place unless there is a convincing reason to point the user somewhere else.

> Given that the usage of MD5 for checksums is much more common than CRC32, maybe if we switched from CRC32 to MD5 (TBH I've got no idea why CRC32 was chosen originally?), then maybe it'd be more obvious and not actually need any explanation?

I think we should still provide some explanation. Using MD5 might mean that we need to provide *less* explanation than with CRC32, but we should still provide some IMHO.

alexandrosm:
Or maybe this is another case where we need the pattern of "short
explanation + learn more link"

konmouz:
[...]

PR #1368, opened 2017-04-28
[WIP] feat(GUI): dynamic finish page

Shou:
We implement an externally loaded dynamic finish page. If the Internet
connection is unreliable or unavailable, or a non-200 HTTP response is returned
we display a fallback default finish banner.

Changelog-Entry: Add a dynamic finish page.


lib/gui/app.js
```diff
@@ -262,7 +262,7 @@ app.run(($rootScope, AnalyticsService) => {
 });
 
 app.config(($urlRouterProvider) => {
-  $urlRouterProvider.otherwise('/main');
+  $urlRouterProvider.otherwise('success');
```
Shou: This should only be here while I work on it.

lib/gui/pages/finish/directives/banner-view.js
```diff
@@ -0,0 +1,93 @@
...
+
+const showWebview = (elem) => {
+  elem.style.flex = null;
+  elem.style.width = null;
+  elem.style.height = null;
+};
+
+module.exports = ($timeout) => {
+  return {
+    template: '<div><webview></webview><div ng-show="showDefaultBanner" ng-transclude></div></div>',
```
jviotti: I think that the concept that something else gets shown if the webview didn't load shouldn't belong here. Can we extract it out by making this directive report its own state and using `ngIf`s on the template?

lib/gui/pages/finish/directives/banner-view.js
```diff
@@ -0,0 +1,93 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+const hideWebview = (elem) => {
+
+  // We hide with width and height instead of display as that may be buggy
+  // with <webview>s.
```
jviotti: > as that may be buggywith <webview>s.

Maybe worth elaborating on the buggy behavior you see?

lib/gui/pages/finish/directives/banner-view.js
```diff
@@ -0,0 +1,93 @@
...
+
+      let loadingTimeout;
+
+      webviewElem.addEventListener('did-start-loading', (event) => {
+        loadingTimeout = $timeout(() => {
+          webviewElem.src = null;
+          scope.showDefaultBanner = true;
+
+        // Seconds to wait before cancelling the page-load.
+        }, 2000);
```
jviotti: Can we move this into a configuration property that can be passed to the directive?

lib/gui/pages/finish/directives/banner-view.js
```diff
@@ -0,0 +1,93 @@
...
+          });
+
+          hideWebview(webviewElem);
+
+        } else {
+          console.log(httpResponseCode);
+        }
+      });
+
+      webviewElem.src = 'https://etcher.io/banner/';
```
jviotti: Can we also make this a property of the directive?

lib/gui/pages/finish/directives/banner-view.js
```diff
@@ -0,0 +1,93 @@
...
+    },
+    link: (scope, $element) => {
+      const webviewElem = $element[0].querySelector('webview');
+
+      let loadingTimeout;
+
+      webviewElem.addEventListener('did-start-loading', (event) => {
+        loadingTimeout = $timeout(() => {
+          webviewElem.src = null;
+          scope.showDefaultBanner = true;
```
jviotti: We shouldn't tie this to the concept of a "banner". Can we make it a generic AngularJS Electron "safe web view" directive?

jviotti: Actually, how do you feel about making this in React, and wrapping it out with the ngReact module? The template is so simple that we can probably get away without using JSX.

lib/gui/pages/finish/directives/banner-view.js
Shou: Oh yeah I should've linked this: https://electron.atom.io/docs/api/webview-tag/#css-styling-notes

lib/gui/pages/finish/directives/banner-view.js
Shou: That sounds like a good idea, though ngReact seems to recommend react2angular over itself—should we go with that instead?

jviotti: Can you ping the UI team, so we make sure all use the same React-Angular integration module during the transitions?

lib/gui/pages/finish/directives/banner-view.js
lurch: 2000 seconds?!? Or did you mean milliseconds? ;-)

Does this mean that if the computer isn't internet-connected, the user will see a blank screen for 2 seconds at the end of validation? If so, could we hide that by downloading the banner in the background at the same time as flashing?

Shou: Oops, yeah milliseconds!

Indeed, which isn't ideal. I'll have to look into an idiomatic way to get around this, because with pages I don't think we can keep the `<webview>` element around—we'd have to insert it outside the page and later move it inside the Finish page, but maybe there's caching or something better suited for this purpose.

Shou:
This is the React rewrite of this PR using `react2angular`. Let me know what you think!

@lurch The design is WIP but should be done soon.

lib/gui/components/safe-webview/components/safe-webview.js
```diff
@@ -0,0 +1,176 @@
...
+          // https://electron.atom.io/docs/api/webview-tag/#css-styling-notes
+          flex: this.state.shouldLoad ? null : '0 1',
+          width: this.state.shouldLoad ? null : '0',
+          height: this.state.shouldLoad ? null : '0'
+        }
+      }, []);
+
+    }
+
+    return null;
```
Shou: Just FYI we have to explicitly return `null` here because `undefined` is an error in React.

package.json
```diff
@@ -95,6 +95,10 @@
     "node-ipc": "^8.9.2",
     "node-stream-zip": "^1.3.4",
     "path-is-inside": "^1.0.2",
+    "prop-types": "^15.5.4",
```
jviotti: Why is the shrinkwrap file not getting updated for these dependencies?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,30 @@
...
+'use strict';
+
+const angular = require('angular');
+const {
+  react2angular
+} = require('react2angular');
+const MODULE_NAME = 'Etcher.Components.SafeWebview';
+const SafeWebview = angular.module(MODULE_NAME, []);
+
+SafeWebview.component('safeWebview',
```
jviotti: Given that the whole component is just one file, can we join everything as `lib/gui/components/safe-webview.js`?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const angular = require('angular');
+const {
+  react2angular
```
jviotti: This looks very weird. Can we do something like `const react2angular = require('react2angular').react2angular` instead?

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,45 @@
...
+        </button>
+      </div>
+    </div>
 
-        <div class="col-xs separator-xs"></div>
+    <div class="box center">
+      <safe-webview src="'https://etcher.io/'"
+        ng-show="finish.shouldShowWebview"
+        timeout="2000"
+        callback="finish.setShowWebview"></safe-webview>
```
jviotti: Can we move the logic that makes the `safe-webview` only appear if it has fully loaded to the component itself? The directive can then simply take an attribute that points to a boolean variable that will be set once its finally visible. For example:

```html
<safe-webview src="'https://etcher.io/'" timeout="2000" visible="myVisiblePropertyWhichIsInitiallySetToFalse"></safe-webview>

div class="fallback-banner" ng-hide="myVisiblePropertyWhichIsInitiallySetToFalse">
```

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,45 @@
...
+          Show Details
+        </button>
+      </div>
+    </div>
 
-        <div class="col-xs separator-xs"></div>
+    <div class="box center">
+      <safe-webview src="'https://etcher.io/'"
+        ng-show="finish.shouldShowWebview"
+        timeout="2000"
```
lurch: Perhaps the `src` and `timeout` here should both be configured in `package.json` ?

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,45 @@
...
+        callback="finish.setShowWebview"></safe-webview>
+      <div class="fallback-banner" ng-hide="finish.shouldShowWebview">
```
lurch: Would it be more flexible if this fallback-banner was actually specified in a separate template file (which would still be included within the Etcher bundle), rather than being declared 'inline'?

Shou:
The changes in this commit include:

- styles done
- ipc proxying for corvus
- expose version through GET parameter
- open urls marked with external hash in browser

Feedback so far?

This only leaves the seamless loading, which I'm not sure how to do in a nice fashion. I don't believe we can cache/preload it without touching the DOM, i.e. we have to add the `<webview>` somewhere else and move it to the finish page once that's available.

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+     * @type Number
+     * @public
+     */
+    this.loadingTimeoutId = null;
+    this.state = {
+      shouldLoad: true
+    };
+
+    // Events steal 'this'
+    this.didStartLoading = this.didStartLoading.bind(this);
```
jviotti: This is purely optional, but can we use `_.bind` for consistency purposes?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+   * @param {Event} event - Event object
+   */
+  didGetResponseDetails(event) {
+
+    const {
+      httpResponseCode
+    } = event;
+    const HTTP_OK = 200;
+
+    if (httpResponseCode !== HTTP_OK) {
```
jviotti: Can we do `event.httpResponseCode`?

```js
const {
   httpResponseCode
} = event;
```

Looks very weird and this is the only place where this variable is used.

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+
+  ipcMessage(event) {
+    const maybeArg = _.get(event.args, 0);
+
+    if (_.get(maybeArg, 'type') === 'error') {
+      analytics.logException(_.get(maybeArg, 'message'));
+
+    } else if (_.get(maybeArg, 'type') === 'event') {
+      analytics.logEvent(_.get(maybeArg, 'message'));
+    }
```
jviotti: Can we log a debugging message if the type doesn't match, just to make sure we don't silently swallow anything.

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+      url.hash = '';
+
+      if (url.protocol === 'http:' || url.protocol === 'https:') {
+        electron.shell.openExternal(url.href);
+      }
+    }
+  }
+
+  ipcMessage(event) {
+    const maybeArg = _.get(event.args, 0);
```
jviotti: I think `_.nth` is what you want here:

```js
const maybeArg = _.nth(event.args, 0);
```

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+
+    if (_.get(maybeArg, 'type') === 'error') {
```
jviotti: Can we extract the type and the message as variables given that we're using them twice here?

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,45 @@
...
+    </div>
 
-            <button class="button button-primary button-brick" ng-click="finish.restart({ preserveImage: true })">
-              Use <b>same</b> image
-            </button>
-          </div>
+    <div class="box center">
+      <safe-webview src="'https://etcher.io/etcher_banner/'"
+        timeout="2000"
+        callback="finish.setShowWebview"></safe-webview>
```
jviotti: I think I mentioned at some point, but can we change this to a `visible` boolean property that gets set or unset accordingly?

lib/gui/components/safe-webview/preload/safe-webview.js
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+const electron = require('electron');
+
+/**
+ * @summary The SafeWebview functionality to be exposed to the guest
+ */
+window.safeWebview = {
+  sendToHost: (object) => {
```
jviotti: What is this needed for?

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,45 @@
...
+        </button>
+      </div>
+    </div>
 
-            <button class="button button-primary button-brick" ng-click="finish.restart({ preserveImage: true })">
-              Use <b>same</b> image
-            </button>
-          </div>
+    <div class="box center">
+      <safe-webview src="'https://etcher.io/etcher_banner/'"
```
jviotti: Are we passing the current version as query parameters?

lib/gui/components/safe-webview/preload/safe-webview.js
Shou: Basically, the Electron objects are only temporary to be safe, so we define the `sendToHost` IPC function wrapper in a custom object on `window` so we can access it in the guest frame and send things over IPC.

lib/gui/pages/finish/templates/success.tpl.html
Shou: Yeah, but it's implicit—we append a GET parameter `etcher-version` to the URL in the component code.

lib/gui/components/safe-webview/safe-webview.js
lurch: The naked zero is making eslint cry ;-)  (amongst other eslint errors)

lib/gui/components/safe-webview/safe-webview.js
Shou: So we can just remove the `if` part of the else? That means the `type` and `message` strings are only used once, too.

lib/gui/components/safe-webview/safe-webview.js
Shou: Yup!

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+   * @returns {react.Element}
+   */
+  render() {
+    if (this.state.shouldLoad) {
+      const url = new URL(this.props.src);
+
+      // We set the 'etcher-version' GET parameter here.
+      url.searchParams.set('etcher-version', packageJSON.version);
```
lurch: Does this mean that the `etcher-version` gets appended to _every_ URL that the webview displays, or only the first one?

Shou: Only the first one. The page changing doesn't cause a React re-render.

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+    if (url.hash === '#external') {
+      this.refs.webview.stop();
+
+      url.hash = '';
+
+      if (url.protocol === 'http:' || url.protocol === 'https:') {
+        electron.shell.openExternal(url.href);
+      }
+    }
+  }
```
lurch: Does there also need to be something to stop the webview trying to also navigate to the 'external' URL?

lib/gui/pages/finish/templates/success.tpl.html
Shou: Is the `attribute="variableToSet"` thing an Angular thing I've missed or am I misunderstanding?

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,45 @@
...
+        timeout="2000"
+        callback="finish.setShowWebview"></safe-webview>
+      <div class="fallback-banner" ng-hide="finish.shouldShowWebview">
```
lurch: Given that `setShowWebview` updates `shouldShowWebview`, should there be a `getShowWebview` instead of accessing `shouldShowWebview` directly?

lib/gui/components/safe-webview/safe-webview.js
Shou: Luckily there's `_.head` which requires no zeroes

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,45 @@
...
+        </div>
+        <div class="section-footer">
+          <span class="caption footer-right"
+            manifest-bind="version"
+            os-open-external="https://github.com/resin-io/etcher/blob/master/CHANGELOG.md"></span>
         </div>
       </div>
-
-      <span class="label label-big label-inset"
-        ng-if="finish.checksum">CRC32 CHECKSUM : <b>{{ ::finish.checksum }}</b></span>
```
lurch: I guess this change means this PR also impacts on #643 and #993 ? ;-)

lib/gui/components/safe-webview/preload/safe-webview.js
lurch: And what's this IPC used for? I can't see any other references to `sendToHost`, or is it some kind of built-in mechanism?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * @summary Webviews that time out
+ * @type {Object}
+ * @public
+ */
+class SafeWebview extends react.PureComponent {
+
+  /**
+   * @param {Object} props - React element properties
```
lurch: I'm afraid I know nothing about React, but should this document the properties available here? Or at least the properties that are of interest to us?

Shou: The `propTypes` section specifies the types so maybe it's better to do there? But yes we should add descriptive comments for the properties.

lib/gui/components/safe-webview/preload/safe-webview.js
Shou: Oh right I forgot to mention that, it's so we can pass Corvus errors and events through to Etcher. The guest window would use `sendToHost`. If we ever decide to expand the `safeWebview` object I've decided to add a version field so we can check what's supported.

lurch: From a security point of view, I guess we need to be very careful what this IPC channel is used for?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -167,13 +167,9 @@ class SafeWebview extends react.PureComponent {
...
-    const {
-      httpResponseCode
-    } = event;
     const HTTP_OK = 200;
 
-    if (httpResponseCode !== HTTP_OK) {
+    if (event.httpResponseCode !== HTTP_OK) {
       this.setState({
         shouldLoad: false
       });
```
lurch: Does this happen for just the first load, or for all page-loads? i.e. if a user clicks on a link which for some reason is invalid or non-responsive, will the "default banner" get shown again after two seconds?

Shou: Oh yeah, good catch. This does happen for all page loads.

Shou:
I saw the `console-message` event and thought it'd be better to just forward them directly instead of using the ad-hoc forwarding with IPC, so that's done now.

lib/gui/pages/finish/templates/success.tpl.html
lurch: I still think it'd be nice if this URL was stored in `package.json` and retrieved from there somehow ;-)

Shou:
Let me know what you think of how the preloading works in this commit.

lib/gui/index.html
```diff
@@ -30,7 +30,10 @@
 
     <main class="wrapper" ui-view></main>
 
-    <footer class="section-footer">
+    <footer class="section-footer" ng-controller="StateController as state"
+      ng-style="{
```
jviotti: Can we use `ngIf`, or `ngHide` here?

lib/gui/app.js
```diff
@@ -296,3 +297,12 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
   };
 
 });
+
+app.controller('StateController', function($state) {
+
+  /**
+   * @param {string} state - state page
+   */
+  this.is = $state.is;
```
jviotti: Can we make use of the `angular-if-state` directives (https://www.npmjs.com/package/angular-if-state) that we already have to replace this controller?

Shou: Unfortunately they don't work well here because the footer's `display: flex` property is overridden with `display: initial` and changes the footer structure.

Shou:
Here we go

jviotti:
@Shou Looks like the whole page content should be a bit below (e.g. its all too close to the top edge)

screenshot 2017-05-11 19 26 34

Compare it with one of the mockups:

screenshot 2017-05-11 19 26 34

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -1,35 +1,42 @@
...
-              Use <b>same</b> image
-            </button>
-          </div>
+    <div class="box center">
+      <div class="fallback-banner">
+        <div class="caption caption-big">Thanks for using
+          <svg-icon path="../../../assets/etcher.svg"
+            width="150px"
+            height="auto"
+            os-open-external="https://etcher.io?ref=etcher_footer"></svg-icon>
```
jviotti: We should update the `ref` param here to something like `etcher_offline_banner`

jviotti:
I don't think this is working. I edited the URL to something like `https://etcher.io` and the default offline banner is still being shown

jviotti:
Same with `https://google.com`

jviotti:
Increasing the timeout doesn't help

jviotti:
Links inside the banner are not opening. If I click them, the banner hangs for like a second, then I see some white flashes, and finally the offline banner is shown :(

jviotti:
Now nothing opens anymore for me, and this is what I see on DevTools at startup:

screenshot 2017-05-11 20 01 44

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * @summary Set the element state depending on the HTTP response code
+   * @param {Event} event - Event object
+   */
+  didGetResponseDetails(event) {
+    const HTTP_OK = 200;
+    const HTTP_ERR = 400;
+
+    if (event.httpResponseCode < HTTP_OK || event.httpResponseCode >= HTTP_ERR) {
```
lurch: Does this imply that the webview _does_ support 3xx redirects? Has that been tested? ;)

lib/gui/index.html
```diff
@@ -49,5 +50,15 @@
...
         os-open-external="https://github.com/resin-io/etcher/blob/master/CHANGELOG.md"></span>
     </footer>
+
+    <div class="section-loader"
+      ng-controller="StateController as state"
+      ng-class="{
+        isFinish: state.is('success')
+      }">
+      <safe-webview
+        src="'https://etcher.io/success-banner/'"
```
lurch: Does this definitely need to be a single-quoted string inside a double-quoted string? What happens if somebody edits it to only be quoted by single-quotes or double-quotes?

Shou: It's treated as code so it's probably going to become a syntax error.

jviotti: It looks like the `src` attribute of the `safe-webview` AngularJS directive is configured to expect an expression rather than a value. This allows the user to pass a variable that is defined in the controller, for example.

I'm happy with both approaches. Taking an interpolation rather than an expression would remove the quotes, but it wouldn't allow a variable, although we're not using that at the moment.

jviotti: Actually, this value is not supposed to change on the fly, so lets make it an interpolation @Shou

Shou: AFAIK this is basically just how `react2angular` works and we can't change it.

lib/gui/pages/finish/controllers/finish.js
```diff
@@ -53,4 +53,20 @@ module.exports = function($state, SelectionStateModel) {
...
+   * @summary Set the `shouldShowWebview` property
+   * @function
+   * @public
+   *
+   * @param {Boolean} bool - preserve image
+   *
+   * @example
+   * FinishController.setShowWebview(true);
+   */
+  this.setShowWebview = (bool) => {
```
jviotti: Can this be removed?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -0,0 +1,215 @@
...
+   */
+  didStartLoading() {
+    if (this.props.timeout && this.loadingTimeoutId === null) {
+      this.loadingTimeoutId = setTimeout(() => {
+        this.setState({
+          shouldLoad: false
+        });
+
+      // Milliseconds to wait before cancelling the page-load.
+      }, this.props.timeout);
```
jviotti: Should this timeout go away now?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -173,7 +173,10 @@ class SafeWebview extends react.PureComponent {
   static newWindow(event) {
     const url = new URL(event.url);
 
-    if (url.protocol === 'http:' || url.protocol === 'https:') {
+    if (_.every([
+      url.protocol === 'http:' || url.protocol === 'https:',
```
lurch: Perhaps it's just me, but using a logical-or inside an `_.every` looks a bit convoluted? Why not just do:
```js
if ((url.protocol === 'http:' || url.protocol === 'https:') && (event.disposition === 'foreground-tab')) {
```
?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -173,7 +173,10 @@ class SafeWebview extends react.PureComponent {
...
+      event.disposition === 'foreground-tab'
```
lurch: Is there somewhere that lists these different dispositions, and documents what actions trigger each disposition?

lib/gui/components/safe-webview/safe-webview.js
lurch: Is that already handled by `if (this.props.timeout`? i.e. the code isn't triggered currently, but is still there if we want to re-enable it in future?

lib/gui/index.html
lurch: So, if somebody _did_ accidentally change it to be only single-quoted or double-quoted, would that be picked up by the CI builds, or would it only be triggered as an error at runtime?

lurch:
In flowdock I asked what would happen if a user navigated to a different 'internal' page in the webview, and then flashed another image. What page would be displayed after the second image finished flashing?

lib/gui/components/safe-webview/safe-webview.js
lurch: If the `newWindow` callback _doesn't_ trigger an `openExternal` action, does the webview remain unchanged, i.e. it ignores that particular action?

Shou: Yeah, the page remains.

jviotti:
@Shou Are you still around to test that?

Shou:
Yeah, it's something that needs to be handled as well, which is another feature that I guess I need to do. We want it to reset because the user may have browsed away from the success screen, and also to obtain any potential new pages.

I think I can latch onto the $state changes and assume if the `fromPage` is `'success'` then we reset the webview in whatever clean way does it with `react2angular`, which I believe has to update itself from the inside with `this.setState` because changing the props seemingly has no effect?

jviotti:
Good catch @lurch. Yeah, let's get that fixed.

jviotti:
Let's also fix the content position, as mentioned in https://github.com/resin-io/etcher/pull/1368#issuecomment-300943601

lib/gui/components/safe-webview/safe-webview.js
jviotti: Yep, works fine!

jviotti:
I addressed some of the final comments myself. I left out resetting the webview url given that I have no clue about React, and I don't really want to dive into it right before the release. Our banners will most probably not include links that don't open in the user's browser, so this can wait until the next release.

Issue #1342, opened 2017-04-21
Stall while flashing Mac OS X Snow Leopard


jhermsmeier:
- **Etcher version:** 1.0.0-rc.3
- **Operating system and architecture:** macOS Sierra

The flashing process halts at a certain percentage while flashing a **Mac OS X Snow Leopard** disk image (`.dmg`) and does not resume, while a dialog from Mac OS pops up, saying the disk can't be recognised mid-flash.

![screen shot 2017-04-21 at 19 54 16](https://cloud.githubusercontent.com/assets/244907/25290163/face353e-26cc-11e7-9c67-3c57b38eb986.png)

jhermsmeier:
This appears to be a bug in https://github.com/jhermsmeier/node-udif – it parses just fine, but comes up short with `Bytes written: 0.96 GB (0.19 GB/s)` (uncompressed size is 7.8 GB I think, so ~7 GB missing).

jhermsmeier:
Fixed in #1346

PR #1369, opened 2017-04-28
feat(image-stream): Make invalid images user errors

jhermsmeier:
This displays a user error if the reading the image causes an error,
instead of letting it fall through and get reported.
This is to avoid reporting errors that are not due to malfunction of the software,
but due to malformatted images.

|   |   |
|---|---|
| screen shot 2017-04-28 at 18 19 08 | screen shot 2017-04-28 at 18 19 20 |

Change-Type: minor
Changelog-Entry: Display nicer error dialog when reading an invalid image


jhermsmeier:
As this will catch any error coming from the file type handler, this has the potential to mask real bugs as user errors, and thus cause them to not be reported.
A mechanism to tell bug errors from malformed image errors would be needed, but might be non-trivial.

jviotti:
> As this will catch any error coming from the file type handler, this has the potential to mask real bugs as user errors, and thus cause them to not be reported.

Yeah, that's my worry here. What about only reacting to individual errors as they come rather than adding a catch-all block? 

As far as I can see, this PR was mainly conceived to deal with the "invalid footer" DMG error. Can we make the module that causes that error on the first place (udif, right?) include a code that can catch here?

jhermsmeier:
> Can we make the module that causes that error on the first place (udif, right?) include a code that can catch here?

I've been fiddeling with errors a bit now, and think that if we check `error.name` and only create user errors for when `error.name == 'Error'`, we should be able to still track errors resulting from bugs, as those would be either a `ReferenceError`, `TypeError`, `RangeError`, or `SyntaxError` etc, since they're thrown by the VM.

jviotti:
Normal thrown errors have an `Error` `name` property:

```js
> const foo = new Error('My error');
undefined
> foo.name
'Error'
>
```

Adding such check means that we automatically consider any error thrown using `throw new Error()` as a user error. Note that we do use these types of errors in various places to signify real application errors.

lurch:
I agree with @jviotti that we should be **very** careful when filtering out any errors that _might_ be actual errors. Because even with the invalid-zip / invalid-dmg errors shown in your screenshots, I guess we can't be 100% certain that it's due to invalid files, and not due to some subtle bug in our handling of zip or dmg files?

@jviotti or @stefan-mihaila : how many "invalid image" errors do we _actually_ get logged?

jviotti:
> @jviotti or @stefan-mihaila : how many "invalid image" errors do we actually get logged?

The invalid image error is a user error, so we don't have information about it on Sentry. We should definitely add a Mixpanel event for it once we port corvus to node.

jhermsmeier:
> Normal thrown errors have an Error name property:

Yeah, that's what I was suggesting to detect whether it's an exception coming from previously known error conditions in the written library code (i.e. `if(signature != magic) throw new Error('message')`), which should indicate with fairly high proficiency that it's an error due to malformed input.

Bugs on the other hand, would usually cause everything but a plain Error, i.e. a ReferenceError as they're thrown by the VM, not by lib code – if that makes sense?

Another failure mode is of course that the lib doesn't error but still behaves buggy (see #1342 for an example), but we're not going to catch that through these means either way.

> I agree with @jviotti that we should be very careful when filtering out any errors that might be actual errors.

I also agree that we should be very careful with this, but I also don't think that adding stuff to third party libraries' errors will save us here. Relying on that seems more fragile to me.

> We should definitely add a Mixpanel event for it once we port corvus to node.

👍  That'd be pretty neat in general, to know how many images cause issues one way or another.

jviotti:
> Bugs on the other hand, would usually cause everything but a plain Error, i.e. a ReferenceError as they're thrown by the VM, not by lib code – if that makes sense?

Sure, JavaScript syntax errors, etc will not cause plain errors, so your logic handles those fine, however a library might use `new Error()` to signify things that shouldn't happen. For example `if (!options) { throw new Error('Missing options'); }`. I don't think we should rely on third party libraries to only throw plain errors on user errors.

> I also agree that we should be very careful with this, but I also don't think that adding stuff to third party libraries' errors will save us here. Relying on that seems more fragile to me.

Yeah, makes sense. The only thing I can think of is capturing errors we know about using regular expression over the message string. That way don't make assumptions over error objects nor require external projects to add error codes. Of course the dependency can change its error messages at any point, but we're shrinkwrapping things, so we'll get a chance of resolving those every time we perform an upgrade.

jhermsmeier:
> I don't think we should rely on third party libraries to only throw plain errors on user errors.
> The only thing I can think of is capturing errors we know about using regular expression over the message string. 

Yeah, true! I'll change this to known error detection in the particular handlers then

lurch:
...and hopefully we won't have any dependencies which "helpfully" localise their error messages to the user's preferred language ;-)

lib/image-stream/handlers.js
```diff
@@ -161,6 +163,17 @@ module.exports = {
         },
         transform: new PassThroughStream()
       };
+    }).catch((error) => {
+      if (/invalid footer/i.test(error.message)) {
+        throw errors.createUserError({
+          title: 'Invalid image',
+          description: `There was an error reading "${path.basename(file)}". `
+            + 'The image does not appear to be a valid Apple Disk Image, or may have the wrong filename extension.\n\n'
```
lurch: Are we sure that any error messages containing `/invalid footer/i` will always be related to Apple Disk Images? Is this the same thing as `.dmg`? Would it be worth tweaking the message slightly to `... a valid Apple Disk Image (DMG), ...`? I'm not a Mac user, so I've got no idea how familiar people are likely to be with the terminology.

jhermsmeier: > Are we sure that any error messages containing /invalid footer/i will always be related to Apple Disk Images?

In this case, yes, because this catch is in the dmg handler ;)

> Would it be worth tweaking the message slightly to ... a valid Apple Disk Image (DMG), ...?

Yeah, that's probably better

lurch: > this catch is in the dmg handler ;)

Oh yeah, I hadn't expanded enough of the Github diff to spot that ;-)

lib/image-stream/handlers.js
```diff
@@ -161,6 +163,17 @@ module.exports = {
...
       };
+    }).catch((error) => {
+      if (/invalid footer/i.test(error.message)) {
+        throw errors.createUserError({
+          title: 'Invalid image',
+          description: `There was an error reading "${path.basename(file)}". `
+            + 'The image does not appear to be a valid Apple Disk Image (dmg), or may have the wrong filename extension.\n\n'
+            + `Error: ${error.description || error.message}`
+        });
+      } else {
```
jviotti: Just a minor nit, but the `else` is not necessary here. We can do:

```js
if (...) {

}

throw error;
```

tests/image-stream/tester.js
```diff
@@ -43,11 +43,14 @@ const deleteIfExists = (file) => {
   });
 };
 
-exports.expectError = function(file, errorMessage) {
+exports.expectError = function(file, errorMessage, errorDetail) {
```
jviotti: Is this function being used anywhere?

lib/image-stream/handlers.js
```diff
@@ -161,6 +163,17 @@ module.exports = {
         },
         transform: new PassThroughStream()
       };
+    }).catch((error) => {
+      if (/invalid footer/i.test(error.message)) {
```
jviotti: Can we add a test case to cover this?

tests/image-stream/dmg.spec.js
```diff
@@ -105,4 +105,14 @@ describe('ImageStream: DMG', function() {
 
   });
 
+  context('invalid', function() {
+
+    describe('given an invalid dmg file', function() {
+      tester.expectError(
+        path.join(DATA_PATH, 'unrecognized', 'invalid.dmg'),
```
lurch: Does / should this file physically exist?

jhermsmeier: Oops. Yes, it should now :)

tests/image-stream/tester.js
jhermsmeier: Yup, by the added test

lib/image-stream/handlers.js
jhermsmeier: Yeah, just haven't gotten around to it yet – added now.

lib/image-stream/handlers.js
jhermsmeier: Done!

tests/image-stream/tester.js
lurch: Should it have a default value, for tests that don't use it? Or do unused parameters in JS automatically default to `null` or something?

jhermsmeier: Unused arguments default to `undefined`, hence it won't use this if not given

PR #1372, opened 2017-05-01
chore: remove `etcher-latest-version` from package.json

jviotti:
This version is no longer used, and was removed from the shrinkwrap
file, but for some reason, we forgot to remove it from package.json.

See: https://github.com/resin-io/etcher/pull/1183


lurch:
Ah, whoops! Perhaps we should do a `npm shrinkwrap` check on the CI servers to check for uncommited inconsistencies, just like we do with sass?

jviotti:
Yeah, I'll do it!

PR #1373, opened 2017-05-01
chore: make `isstream` a top level production development

jviotti:
Our `npm-shrinkwrap.json` file contains two entries for the `isstream`
dependency. One of them is in the top level of the file, and marks
`isstream` as a development dependency.

The `request` module makes use of `isstream`, so there is another
instance of `isstream` nested inside `request` which is marked as a
production dependency.

Most NPM versions will realise that there are two instances of
`isstream`, and since one of them is a production development, it will
include it in the final package.

However, there are some NPM version, like 4.6.1, which will only
consider the top level entry, and completely omit `isstream` on the
final package, causing a "Cannot find module 'isstream'" error.

The solution was to manually update the shrinkwrap file to remove the
nested instance, and make the top level instance a production
development.

I don't think there is a way we can protect ourselves from these edge
cases. NPM should have recognised that the dependencies were compatible,
and not create a nested instance on the first place.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3514,8 +3514,7 @@
     "isstream": {
       "version": "0.1.2",
       "from": "isstream@>=0.1.1 <0.2.0",
-      "resolved": "https://registry.npmjs.org/isstream/-/isstream-0.1.2.tgz",
-      "dev": true
+      "resolved": "https://registry.npmjs.org/isstream/-/isstream-0.1.2.tgz"
     },
     "jju": {
       "version": "1.3.0",
@@ -4246,11 +4245,6 @@
               "version": "1.0.0",
               "from": "isarray@>=1.0.0 <1.1.0",
               "resolved": "https://registry.npmjs.org/isarray/-/isarray-1.0.0.tgz"
-            },
-            "isstream": {
-              "version": "0.1.2",
-              "from": "isstream@>=0.1.2 <0.2.0",
-              "resolved": "https://registry.npmjs.org/isstream/-/isstream-0.1.2.tgz"
             },
             "jodid25519": {
               "version": "1.0.2",


lurch:
> However, there are some NPM version, like 4.6.1, which will only
consider the top level entry, and completely omit isstream on the
final package, causing a "Cannot find module 'isstream'" error.

Is that a bug in NPM? Should we try locking down the version of NPM that Etcher can be built with?

> make the top level instance a production development.

Huh? Was that supposed to be "production dependency" ?

lurch:
" jviotti added a commit that referenced this pull request 2 days ago
 @jviotti	Revert "chore: make `isstream` a top level production development (#1373  …"

Hmmm, whassat all about?

lurch:
Should this PR have included a modified `package.json` too??

jviotti:
> Is that a bug in NPM? Should we try locking down the version of NPM that Etcher can be built with?

We discussed that recently, but we still need to find a good way to do so. Maybe we can add npm as a git submodule?

> Huh? Was that supposed to be "production dependency" ?

See what happens when you're not around? :P

> " jviotti added a commit that referenced this pull request 2 days ago
> @jviotti	Revert "chore: make isstream a top level production development (#1373 …"

> Hmmm, whassat all about?

I accidentally clicked the "revert" button on the GitHub UI. I didn't get into master, since I didn't create a PR for it.

> Should this PR have included a modified package.json too??

This is a dependency very down the dependency tree, so no `package.json` change required.

lurch:
> See what happens when you're not around? :P

:rofl: 

> > Should this PR have included a modified `package.json` too??

> This is a dependency very down the dependency tree, so no `package.json` change required.

Oh? From what little I understand, isn't it supposed to be theoretically possible to regenerate the `npm-shrinkwrap.json` from the `package.json`? So shouldn't `isstream` be listed in the `dependencies` in `package.json`, to make sure it stays a production dependency ? Or have I grabbed the wrong end of the stick again?

> I didn't get into master

Sorry to hear that... ;)

I was going to ask if a relevant bug exists for this issue against NPM, but then I saw that https://github.com/npm/npm/issues lists nearly 3000 open issues... eek!

jviotti:
> Oh? From what little I understand, isn't it supposed to be theoretically possible to regenerate the npm-shrinkwrap.json from the package.json? So shouldn't isstream be listed in the dependencies in package.json, to make sure it stays a production dependency ? Or have I grabbed the wrong end of the stick again?

By production dependency, I mean a dependency that is not a development dependency. We don't depend on `isstream` directly, which is the reason its not in out package.json (its a dependency of a dependency).

> I was going to ask if a relevant bug exists for this issue against NPM, but then I saw that https://github.com/npm/npm/issues lists nearly 3000 open issues... eek!

lol

lurch:
Right. So basically the whole `npm-shrinkwrap.json` is quite fragile then?

jviotti:
Yeah, looks like we can't even trust npm to keep it on a sane state.

PR #1376, opened 2017-05-03
WIP fix(GUI): correctly handle quoting and escaping on elevation 

jviotti:

### DIFF
--- a/lib/shared/permissions.js
+++ b/lib/shared/permissions.js
@@ -17,13 +17,12 @@
 'use strict';
 
 const os = require('os');
-const bindings = require('./bindings');
 const Bluebird = require('bluebird');
 const childProcess = Bluebird.promisifyAll(require('child_process'));
 const sudoPrompt = Bluebird.promisifyAll(require('sudo-prompt'));
-const commandJoin = require('command-join');
 const _ = require('lodash');
 const errors = require('./errors');
+const bindings = require('./bindings');
 
 /**
  * @summary The user id of the UNIX "superuser"
@@ -73,6 +72,32 @@ exports.isElevated = () => {
 };
 
 /**
+ * @summary Quote and escape a string
+ * @function
+ * @private
+ *
+ * @param {String} string - input string
+ * @returns {String} quoted and escaped string
+ *
+ * @example
+ * const result = quoteAndEscapeString('foo');
+ * console.log(result);
+ *
+ * // Windows
+ * "foo"
+ *
+ * // Unix
+ * 'foo'
+ */
+const quoteAndEscapeString = (string) => {
+  if (os.platform() === 'win32') {
+    return `"${string}"`;
+  }
+
+  return `'${_.replace(_.toString(string), /'/g, '\\\'')}'`;
+};
+
+/**
  * @summary Get environment command prefix
  * @function
  * @private
@@ -101,10 +126,10 @@ exports.getEnvironmentCommandPrefix = (environment) => {
     }
 
     if (isWindows) {
-      return [ 'set', `${key}=${value}`, '&&' ];
+      return [ 'set', quoteAndEscapeString(`${key}=${value}`), '&&' ];
     }
 
-    return [ `${key}=${value}` ];
+    return [ `${key}=${quoteAndEscapeString(value)}` ];
   });
 
   if (isWindows) {
@@ -116,6 +141,45 @@ exports.getEnvironmentCommandPrefix = (environment) => {
   }
 
   return _.concat([ 'env' ], argv);
+};
+
+/**
+ * @summary Build the elevate command
+ * @function
+ * @private
+ *
+ * @param {String[]} command - command and arguments
+ * @param {Object} [environment] - environment variables
+ * @returns {String[]} elevate command
+ *
+ * @example
+ * const command = permissions.buildElevateCommand([ 'foo', 'bar' ], {
+ *   ELECTRON_RUN_AS_NODE: 1
+ * });
+ *
+ * console.log(command);
+ *
+ * // Windows
+ * [ 'cmd.exe', '/c', '"set "ELECTRON_RUN_AS_NODE=1" && call foo bar"' ]
+ *
+ * // Unix
+ * [ 'env', 'ELECTRON_RUN_AS_NODE="1"', 'foo', 'bar' ]
+ */
+exports.buildElevateCommand = (command, environment = {}) => {
+  const prefixedCommand = _.concat(
+    exports.getEnvironmentCommandPrefix(environment),
+    _.map(command, quoteAndEscapeString)
+  );
+
+  if (os.platform() === 'win32') {
+    return [
+      'cmd.exe',
+      '/c',
+      quoteAndEscapeString(_.join(prefixedCommand, ' '))
+    ];
+  }
+
+  return prefixedCommand;
 };
 
 /**
@@ -143,21 +207,18 @@ exports.getEnvironmentCommandPrefix = (environment) => {
  * });
  */
 exports.elevateCommand = (command, options) => {
-  const prefixedCommand = _.concat(exports.getEnvironmentCommandPrefix(options.environment), command);
+  const elevateCommand = exports.buildElevateCommand(command, options.environment);
 
   if (os.platform() === 'win32') {
     const elevator = Bluebird.promisifyAll(bindings.load('elevator'));
-    return elevator.elevateAsync(_.concat([
-      'cmd.exe',
-      '/c'
-    ], prefixedCommand)).then((results) => {
+    return elevator.elevateAsync(elevateCommand).then((results) => {
       return {
         cancelled: results.cancelled
       };
     });
   }
 
-  return sudoPrompt.execAsync(commandJoin(prefixedCommand), {
+  return sudoPrompt.execAsync(_.join(elevateCommand, ''), {
     name: options.applicationName
   }).then((stdout, stderr) => {
     if (!_.isEmpty(stderr)) {
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -918,11 +918,6 @@
       "resolved": "http://registry.npmjs.org/combined-stream/-/combined-stream-0.0.7.tgz",
       "dev": true
     },
-    "command-join": {
-      "version": "2.0.0",
-      "from": "command-join@latest",
-      "resolved": "https://registry.npmjs.org/command-join/-/command-join-2.0.0.tgz"
-    },
     "command-line-args": {
       "version": "2.1.6",
       "from": "command-line-args@>=2.1.6 <3.0.0",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "bluebird": "^3.0.5",
     "bootstrap-sass": "^3.3.5",
     "chalk": "^1.1.3",
-    "command-join": "^2.0.0",
     "drivelist": "^5.0.19",
     "electron-is-running-in-asar": "^1.0.0",
     "etcher-image-write": "^9.1.2",
--- a/tests/shared/permissions.spec.js
+++ b/tests/shared/permissions.spec.js
@@ -17,6 +17,7 @@
 'use strict';
 
 const m = require('mochainon');
+const _ = require('lodash');
 const os = require('os');
 const permissions = require('../../lib/shared/permissions');
 
@@ -43,12 +44,23 @@ describe('Shared: permissions', function() {
         m.chai.expect(permissions.getEnvironmentCommandPrefix({})).to.deep.equal([]);
       });
 
+      it('should quote environment variables with spaces', function() {
+        m.chai.expect(permissions.getEnvironmentCommandPrefix({
+          FOO: 'bar baz'
+        })).to.deep.equal([
+          'set',
+          '"FOO=bar baz"',
+          '&&',
+          'call'
+        ]);
+      });
+
       it('should create an environment command prefix out of one variable', function() {
         m.chai.expect(permissions.getEnvironmentCommandPrefix({
           FOO: 'bar'
         })).to.deep.equal([
           'set',
-          'FOO=bar',
+          '"FOO=bar"',
           '&&',
           'call'
         ]);
@@ -61,13 +73,13 @@ describe('Shared: permissions', function() {
           BAZ: 'qux'
         })).to.deep.equal([
           'set',
-          'FOO=bar',
+          '"FOO=bar"',
           '&&',
           'set',
-          'BAR=baz',
+          '"BAR=baz"',
           '&&',
           'set',
-          'BAZ=qux',
+          '"BAZ=qux"',
           '&&',
           'call'
         ]);
@@ -80,7 +92,7 @@ describe('Shared: permissions', function() {
           BAZ: undefined
         })).to.deep.equal([
           'set',
-          'BAR=qux',
+          '"BAR=qux"',
           '&&',
           'call'
         ]);
@@ -93,13 +105,13 @@ describe('Shared: permissions', function() {
           BAZ: -1
         })).to.deep.equal([
           'set',
-          'FOO=1',
+          '"FOO=1"',
           '&&',
           'set',
-          'BAR=0',
+          '"BAR=0"',
           '&&',
           'set',
-          'BAZ=-1',
+          '"BAZ=-1"',
           '&&',
           'call'
         ]);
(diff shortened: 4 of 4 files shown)


lib/shared/permissions.js
```diff
@@ -143,14 +158,21 @@ exports.getEnvironmentCommandPrefix = (environment) => {
...
-  if (os.platform() === 'win32') {
+  if (isWindows) {
     const elevator = Bluebird.promisifyAll(bindings.load('elevator'));
-    return elevator.elevateAsync(_.concat([
+    return elevator.elevateAsync([
       'cmd.exe',
-      '/c'
-    ], prefixedCommand)).then((results) => {
+      '/c',
+      `"${_.join(prefixedCommand, ' ')}"`
```
lurch: Should this be using `quoteString` too?

lib/shared/permissions.js
```diff
@@ -143,14 +158,21 @@ exports.getEnvironmentCommandPrefix = (environment) => {
  * });
  */
 exports.elevateCommand = (command, options) => {
-  const prefixedCommand = _.concat(exports.getEnvironmentCommandPrefix(options.environment), command);
+  const isWindows = os.platform() === 'win32';
+  const prefixedCommand = _.concat(
+    exports.getEnvironmentCommandPrefix(options.environment),
+    _.map(command, (argument) => {
+      return isWindows ? quoteString(argument) : argument;
```
lurch: Out of curiosity, why is the quoting here only applicable to Windows, and not all OSes?

lurch:
Perhaps the logic to create an actual platform-specific command-line should be extracted out into a separate method, which we could then add appropriate unit-tests for?

lib/shared/permissions.js
jviotti: Indeed, done!

jviotti:
> Perhaps the logic to create an actual platform-specific command-line should be extracted out into a separate method, which we could then add appropriate unit-tests for?

Sounds like a great idea!

lib/shared/permissions.js
jviotti: The quoting was done later on for non-windows, on the `commandJoin` call. I refactored the whole thing so that quoting happens in a single place!

jviotti:
@lurch I heavily reworked this PR, so I'll need to carefully test everywhere again. Can I make use of your bash/linux ninja skills to double check the UNIX side of things? :)

jviotti:
Testing plan:

- [ ] Windows
- [ ] macOS
- [ ] GNU/Linux

lurch:
> Can I make use of your bash/linux ninja skills to double check the UNIX side of things? :)

Yup, I'll take a look tomorrow :)

lurch:
As I mentioned on #1366 I think we should also add tests for all kinds of "strange" characters - as well as spaces we should check single quotes, double-quotes, ampersands, brackets, slashes, etc.  And where possible we should check these in both the environment variable values, and also in the command-path, command-arguments, etc. etc. (Linux also allows double-quotes in filenames, which IIRC Windows doesn't).
It means adding a lot of tedious tests, but a comprehensive test suite is the only way to be robust, as we have no way of knowing what kind of weird and wonderful filenames the users might give their images ;-)

Perhaps we should also check that the env-var keys match an expected regex http://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names

lurch:
To help with testing, I whipped up this little script, which I named `path_test.js`:
```js
#!/usr/bin/env node

const process = require('process');

console.log("Envvars:");
for (key in process.env) {
    if (key.startsWith('ENVVAR_TEST_')) {
        console.log(`${key}={${process.env[key]}}`);
    }
}

console.log("Args:");
for (arg in process.argv) {
    console.log(`${arg}={${process.argv[arg]}}`);
}
```

And on Linux I'm successfully able to run it like:
```bash
$ ENVVAR_TEST_2="two & three" ENVVAR_TEST_1= ENVVAR_TEST_3="" ENVVAR_TEST_="=" ENVVAR_TEST_4="foo" ENVVAR_TEST_3="\"&\"\$bar'cat(){}[]/\\" ./path_test.js "" "two & \\three" "foo" "\"'&(1)\${foo}"
Envvars:
ENVVAR_TEST_2={two & three}
ENVVAR_TEST_3={"&"$bar'cat(){}[]/\}
ENVVAR_TEST_4={foo}
ENVVAR_TEST_={=}
ENVVAR_TEST_1={}
Args:
0={/usr/bin/nodejs}
1={/home/andrew/github/resin-io/path_test.js}
2={}
3={two & \three}
4={foo}
5={"'&(1)${foo}}
```

So from these _initial_ tests it looks like as long as you double-quote everything, on Bash on Linux you only need to backslash-escape double-quotes, backslashes and dollar signs (to stop them being interpreted as variables). But you'd obviously need to check how much of that is already done by `commandJoin`. Feel free to include that test-script in the unit-tests, in case it's useful ;-)
And I've not checked to see if things work differently on non-Bash shells.

lib/shared/permissions.js
```diff
@@ -101,7 +116,22 @@ exports.getEnvironmentCommandPrefix = (environment) => {
...
+            return quoteString(definition);
+          }
+
+          return definition;
+        }),
+        '&&'
+      ];
+    }
+
+    if (/\s/.test(value)) {
```
jhermsmeier: @jviotti what do you think about reusing the `command-join` module to escape these, instead of custom code?

lib/shared/permissions.js
```diff
@@ -73,6 +73,21 @@ exports.isElevated = () => {
...
+ * @function
+ * @private
+ *
+ * @param {String} string - input string
+ * @returns {String} quoted string
+ *
+ * @example
+ * const result = quote('foo');
+ */
+const quoteString = (string) => {
```

lib/shared/permissions.js
jviotti: Sounds good, I'll make sure they are always quoted. @jhermsmeier sadly `command-join` doesn't play nice with environment variable declarations. In the case of Windows, for example, `set` shouldn't be quoted, otherwise `cmd.exe` complains.

lib/shared/permissions.js
```diff
@@ -119,6 +149,37 @@ exports.getEnvironmentCommandPrefix = (environment) => {
 };
 
 /**
+ * @summary Build the elevate command
+ * @function
+ * @private
+ *
+ * @param {String[]} command - command arguments
```

lib/shared/permissions.js
```diff
@@ -119,6 +149,37 @@ exports.getEnvironmentCommandPrefix = (environment) => {
...
+ * @param {Object} [environment] - environment variables
+ * @returns {String[]} elevate command
+ *
+ * @example
+ * const command = permissions.buildElevateCommand([ 'foo', 'bar' ], {
+ *   ELECTRON_RUN_AS_NODE: 1
+ * });
```
jviotti: Yeah, I'll update the example. The elevation programs require an array of strings. What this function basically does is transform the command so that it embeds the environment variable definitions.

lib/shared/permissions.js
jhermsmeier: I think that might be because it'll treat that everything after the command as an argument – should work if only called on the value though (instead of `quoteString()`)?
Also, why aren't the env vars passed through the `env` option to `sudoPrompt.execAsync()`? Does that not work properly on Windows?

jviotti: Ah, I see what you mean now. Makes a lot of sense! What `env` option?

jhermsmeier: I was under the impression that `sudo-prompt` would pass options to the [ChildProcess](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback), but a quick look at the source told me otherwise.
Forking now, and will see if I can throw a PR together that will allow for it.

lib/shared/permissions.js
```diff
@@ -73,6 +73,21 @@ exports.isElevated = () => {
 };
 
 /**
+ * @summary Quote a string
+ * @function
+ * @private
+ *
+ * @param {String} string - input string
+ * @returns {String} quoted string
```
lurch: Minor nit, but could you update the `@summary` and `@returns` to clarify that it's double-quoting?
I guess for the sake of completion, (if you really wanted to) you could move this to `utils.js` and add `singleQuoteString` too ;-)

lib/shared/permissions.js
```diff
@@ -73,6 +73,21 @@ exports.isElevated = () => {
...
+ *
+ * @example
+ * const result = doubleQuoteString('foo');
```
lurch: Again, showing the contents of `result` might be useful in this example ;-)

lib/shared/permissions.js
```diff
@@ -73,6 +73,21 @@ exports.isElevated = () => {
...
+ */
+const doubleQuoteString = (string) => {
+  return `"${string}"`;
```
lurch: Hmmmm, I wonder if this should additionally be escaping `"` to `\"` inside the string, in case the string already contains a double-quote?  Quoting and string-escaping always causes headaches ;)

lib/shared/permissions.js
```diff
@@ -119,6 +134,45 @@ exports.getEnvironmentCommandPrefix = (environment) => {
...
+ *
+ * @example
+ * const command = permissions.buildElevateCommand([ 'foo', 'bar' ], {
+ *   ELECTRON_RUN_AS_NODE: 1
+ * });
+ *
+ * console.log(command);
+ *
+ * // Windows
+ * [ 'cmd.exe', '/c', '"set "ELECTRON_RUN_AS_NODE=1" && call foo baz"' ]
```
lurch: Haha, I think you meant `... foo bar...` ;-)

lib/shared/permissions.js
lurch: Forgot to say earlier, but when I had a look at `command-join` it was doing all the escaping we needed inside a for loop, and not inside a separate function that we could re-use here :(

lurch:
It would be great to see lots more unit-tests, for all the other "weird and wonderful" characters that we might get inside env-vars or command-line args - see my earlier comment for examples ;)

jviotti:
I will add more tests once I finish doing some manual checks (so we don't test for the wrong thing). I discovered that `command-join` attempts to escape various characters on Windows by prepending them with a caret (e.g. `^&`), however that type of escaping doesn't work at all when executing commands (!!). Getting rid of that makes everything work.

It looks like we'll have to manually check for those characters, and if so, roll our own thing.

lib/shared/permissions.js
jviotti: I'll test this changes on Linux and macOS first. At least in Windows, un-escaped nested quotes are what makes everything work :P

lib/shared/permissions.js
jviotti: Good catch!

jviotti:
I ended up ditching `command-join`, because it throws JavaScript errors on GNU/Linux :S

lib/shared/permissions.js
```diff
@@ -73,6 +72,32 @@ exports.isElevated = () => {
...
+ *
+ * // Unix
+ * 'foo'
+ */
+const quoteAndEscapeString = (string) => {
+  if (os.platform() === 'win32') {
+    return `"${string}"`;
+  }
+
+  return `'${_.replace(_.toString(string), /'/g, '\\\'')}'`;
```
lurch: Hmmm, perhaps it'd be better to have `string = _.toString(string);` as the first line of this function?

lib/shared/permissions.js
```diff
@@ -119,6 +144,45 @@ exports.getEnvironmentCommandPrefix = (environment) => {
...
+ * // Unix
+ * [ 'env', 'ELECTRON_RUN_AS_NODE="1"', 'foo', 'bar' ]
+ */
+exports.buildElevateCommand = (command, environment = {}) => {
+  const prefixedCommand = _.concat(
+    exports.getEnvironmentCommandPrefix(environment),
+    _.map(command, quoteAndEscapeString)
+  );
+
+  if (os.platform() === 'win32') {
```
lurch: Given how many times we do this check, would it be worth doing a single `const isWindows = os.platform() === 'win32';` at the module-level scope? I think it's unlikely that the OS platform will change during the execution-lifetime of the script ;-)

lurch:
Given how comprehensive the `command-join` tests are, if it's not working properly then I suspect it might be because you're using it wrong? :-S

jviotti:
I don't think I'm doing anything weird. See https://travis-ci.org/resin-io/etcher/jobs/230489399 for an example. I get:

```js
TypeError: command[Symbol.iterator] is not a function
    at joinNix (node_modules/command-join/command-join.js:24:22)
    at commandJoin (node_modules/command-join/command-join.js:102:12)
```

I can reproduce locally, even though the same code works fine on macOS and Windows.

jviotti:
I also feel tempted to not use the module, given that we're basically adding custom code to revert what the module is doing on Windows :/

lurch:
> I don't think I'm doing anything weird. See https://travis-ci.org/resin-io/etcher/jobs/230489399 for an example

Looks like I can't clone the git-hash used by that job.

> I can reproduce locally, even though the same code works fine on macOS and Windows.

If you can give me some examples, I'd love to use my ninja-skills to dig into that ;-) :desktop_computer:

jviotti:
Oh, damn. I have the older commit that exhibits the problem in one of my NUCs. I'll create a branch for it tomorrow, and you should probably be able to reproduce from your laptop.

jviotti:
@lurch Check the `windows-elevate-quoting-linux-js-error` branch. That fails on my NUC with Ubuntu 17.04.

lurch:
Ignoring the unit-tests (which would just need their "expected" quote-characters updated to suit) I was able to get command-join working properly on that `windows-elevate-quoting-linux-js-error` branch with:
```diff
--- a/lib/shared/permissions.js
+++ b/lib/shared/permissions.js
@@ -119,7 +119,7 @@ exports.getEnvironmentCommandPrefix = (environment) => {
       return [ 'set', commandJoin([ `${key}=${value}` ]), '&&' ];
     }
 
-    return [ `${key}=${commandJoin([ value ])}` ];
+    return [ `${key}=${commandJoin([ _.toString(value) ])}` ];
   });
 
   if (isWindows) {
```

jviotti:
Oh, interesting. Looks like you're becoming a JavaScript ninja :D

lurch:
I'm getting there... still haven't quite got to grips with the syntax of Promises, but I definitely understand the idea, and I'm even getting to grips with Lodash now :grin:

lurch:
Finally got around to playing with my `path_test.js` script from above on Windows, and I'm seeing some really weird behaviour :-/
```
# empty argument
C:\Users\IEUser>node path_test.js "" # 2 quotes
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={}

# argument containing a single double-quote
C:\Users\IEUser>node path_test.js """ # 3 quotes
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={"}

C:\Users\IEUser>node path_test.js "\"" # also 3 quotes
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={"}

# argument containing two double-quotes
C:\Users\IEUser>node path_test.js """"" # actually 5 quotes, wtf
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={""}

C:\Users\IEUser>node path_test.js "\"\"" # 4 quotes
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={"}

# argument containing two double-quotes with a space between them
C:\Users\IEUser>node path_test.js """ """ # actually 6 quotes
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={" "}

C:\Users\IEUser>node path_test.js "\" \""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={" "}

# argument containing an ampersand
C:\Users\IEUser>node path_test.js ^&
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={&}

C:\Users\IEUser>node path_test.js "&"
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={&}

# argument containing ampersand and double-quote
C:\Users\IEUser>node path_test.js "&""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={&"}

C:\Users\IEUser>node path_test.js "&\""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={&"}

# argument containing a double-quote and ampersand
C:\Users\IEUser>node path_test.js "\"^&" # AFAICT this is the only way to get what I want, bizarrely
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={"&}
```

Everything all seems logical (escape `"` to `\"` and enclose the whole thing in double-quotes), but then the last example seems to blow that out of the water :-S
And similarly, if I want ` ^&"` (i.e. space, caret, ampersand, double-quote) I can get that with a (logical) parameter of `" ^&\""`, but if I want `" ^&` then I actually have to use `"\" ^^^&"` as my parameter!?!?!? :frowning_face: 
Yes, these are obviously _very_ contrived examples, but IME it's in being able to handle the corner-cases that you get proper robustness for the general case ;)

And when it comes to escaping in env-vars, it looks like you can only have an _even_ number of double-quotes in the value?
```
C:\Users\IEUser>set "ENVVAR_TEST_1=a" && call node path_test.js
Envvars:
ENVVAR_TEST_1={a}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1="a"" && call node path_test.js
Envvars:
ENVVAR_TEST_1={"a"}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1=""a" && call node path_test.js
Envvars:
ENVVAR_TEST_1={""a}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1=""" && call node path_test.js
Envvars:
ENVVAR_TEST_1={""}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1=a\""""" && call node path_test.js
Envvars:
ENVVAR_TEST_1={a\""""}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1=\"\"" && call node path_test.js
Envvars:
ENVVAR_TEST_1={\"\"}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1=a^\""" && call node path_test.js
Envvars:
ENVVAR_TEST_1={a^\""}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1=\"a^^"" && call node path_test.js # looks like the weird needing to double ^ after " rule applies to setting env-vars too
Envvars:
ENVVAR_TEST_1={\"a^"}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
```
Seems like no matter what escaping mechanism I try, I can't get ENVVAR_TEST_1 to contain an odd number of double-quote characters! (But obviously it's very unlikely that we'd ever have an env-var value with an odd number of double-quote characters)

But I guess the examples with multiple double-quotes also shows that the escaping rules for env-vars are actually different to the escaping rules for command-line parameters, and so we **can't** re-use commandJoin to properly escape env-vars (at least on Windows). Which I guess counts as a step in the right direction..... will come back to this tomorrow.

LOL! I was just about to click on the `Comment` button to submit this massive diatribe, when I accidentally discovered how to have an odd number of double-quotes in an env-var - you need to proceed the _last_ double-quote by a caret! (wtf?)
```
C:\Users\IEUser>set "ENVVAR_TEST_1="^" && call node path_test.js
Envvars:
ENVVAR_TEST_1={"}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1=a"^" && call node path_test.js
Envvars:
ENVVAR_TEST_1={a"}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1="a^" && call node path_test.js
Envvars:
ENVVAR_TEST_1={"a}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}

C:\Users\IEUser>set "ENVVAR_TEST_1="a" "^" && call node path_test.js
Envvars:
ENVVAR_TEST_1={"a" "}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
```
But if you try any of that with a string containing an even number of double-quotes, then the trailing caret appears in the variable:
```
C:\Users\IEUser>set "ENVVAR_TEST_1="a"^" && call node path_test.js
Envvars:
ENVVAR_TEST_1={"a"^}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
```
Obviously.

lurch:
Oh great, and it seems like adding "call" on the front, like we need to do in order to set env-vars in the same one-liner, messes things up even further too :(
This works as expected from before:
```
C:\Users\IEUser>call node path_test.js "&\""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={&"}
```
but I've not _yet_ found which magic incantation is needed if I want a parameter to contain `"&` when using `call node ...` instead of just `node ...` :-( :-(

jviotti:
I was hoping to get this merged for today's release, but it looks like there is a huge amount of work left for this. I'll extract the absolutely necessary Windows fixes from this as another PR, and we can continue to work on this after the v1.0.0 release.

lurch:
Lots more \*\*\*\*-ing around, and I've found that I can get `"&` when `call` is also involved with 
```
C:\Users\IEUser>call node path_test \"&
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test}
2={"&}
```

and to get ` ^&"` I need to do:
```
C:\Users\IEUser>call node path_test " "^^"&\""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test}
2={ ^&"}
```
I _think_ if I can figure out how to get `" ^&` then I _might_ be all set... (to turn even more insane :wink: )

lurch:
Ooooh, I think I _might_ have finally discovered a reliable and consistent parameter-escaping syntax, that works the same both with and without `call`... ?
```
C:\Users\IEUser>node path_test.js "&"""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={&"}

C:\Users\IEUser>call node path_test.js "&"""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={&"}

C:\Users\IEUser>node path_test.js """&"
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={"&}

C:\Users\IEUser>call node path_test.js """&"
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={"&}

C:\Users\IEUser>node path_test.js "\ "" & "^^" &"
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={\ " & ^ &}

C:\Users\IEUser>call node path_test.js "\ "" & "^^" &"
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={\ " & ^ &}
```

Yippee, so far so good! The trick appears to be replacing double-quotes with double-double-quotes (`"` -> `""`), then doing some weird escaping for carets (`^` -> `"^^"`) and then finally enclosing the whole result in double quotes. Wow!

```
C:\Users\IEUser>node path_test.js " `!""£$%"^^"&*()-_=+[]{};:'@#~\|,<.>/? ` ! "" £ $ % "^^" & * ( ) - _ = + [ ] { } ; : ' @ # ~ \ | , < . > / ? " ""^^"param2 ""2"" & "^^""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={ `!"£$%^&*()-_=+[]{};:'@#~\|,<.>/? ` ! " £ $ % ^ & * ( ) - _ = + [ ] { } ; : ' @ # ~ \ | , < . > / ? }
3={^param2 "2" & ^}

C:\Users\IEUser>call node path_test.js " `!""£$%"^^"&*()-_=+[]{};:'@#~\|,<.>/? ` ! "" £ $ % "^^" & * ( ) - _ = + [ ] { } ; : ' @ # ~ \ | , < . > / ? " ""^^"param2 ""2"" & "^^""
Envvars:
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={ `!"£$%^&*()-_=+[]{};:'@#~\|,<.>/? ` ! " £ $ % ^ & * ( ) - _ = + [ ] { } ; : ' @ # ~ \ | , < . > / ? }
3={^param2 "2" & ^}
```
:grinning: :tada:

lurch:
WIP on trying to figure out a consistent escaping scheme for env-vars (i.e. so that we don't have to count the number of double-quotes)...
```
C:\Users\IEUser>set "ENVVAR_TEST_1a="^" && set "ENVVAR_TEST_1b= "^" && set "ENVVAR_TEST_1c=" ^" && set "ENVVAR_TEST_1d= " ^" && set "ENVVAR_TEST_2a="^"^" && set  "ENVVAR_TEST_2b=""" && set "ENVVAR_TEST_2c= """  && set "ENVVAR_TEST_2d= "^"^" && set "ENVVAR_TEST_2e= " " " && set "ENVVAR_TEST_2f= "^ " " && set "ENVVAR_TEST_X=X" && call node path_test.js " `!""£$%"^^"&*()-_=+[]{};:'@#~\|,<.>/? ` ! "" £ $ % "^^" & * ( ) - _ = + [ ] { } ; : ' @ # ~ \ | , < . > / ? " ""^^"param2 ""2"" & "^^""
Envvars:
ENVVAR_TEST_1a={"}
ENVVAR_TEST_1b={ "}
ENVVAR_TEST_1c={" }
ENVVAR_TEST_1d={ " }
ENVVAR_TEST_2a={""}
ENVVAR_TEST_2b={""}
ENVVAR_TEST_2c={ ""}
ENVVAR_TEST_2d={ ""}
ENVVAR_TEST_2e={ " " }
ENVVAR_TEST_2f={ " " }
ENVVAR_TEST_X={X}
Args:
0={C:\Program Files\nodejs\node.exe}
1={C:\Users\IEUser\path_test.js}
2={ `!"£$%^&*()-_=+[]{};:'@#~\|,<.>/? ` ! " £ $ % ^ & * ( ) - _ = + [ ] { } ; : ' @ # ~ \ | , < . > / ? }
3={^param2 "2" & ^}```

jviotti:
This PR is stale. Closing...

PR #1371, opened 2017-05-01
chore: ensure there are no extraneous or invalid npm dependencies

jviotti:
The `npm ls` command will exit with an error code if there are invalid
or extraneous dependencies.

This commits adds that command to `make sanity-checks`, so we can catch
those cases early one.

Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

PR #1378, opened 2017-05-03
chore: add ensure-npm-valid-dependencies.sh to sanity-checks make target

jviotti:
We forgot to do so on the PR that introduces this script.

This commit also moves all the CI code to separate scripts.

See: https://github.com/resin-io/etcher/pull/1371

### DIFF
--- a/scripts/ci/appveyor-test.bat
+++ b/scripts/ci/appveyor-test.bat
@@ -19,8 +19,7 @@ IF "%APPVEYOR_REPO_BRANCH%"=="" (
   EXIT /B 1
 )
 
-IF %APPVEYOR_REPO_BRANCH%==master (
-  make publish-aws-s3
-)
+call make sanity-checks
+call npm test
 
 EXIT /B %ERRORLEVEL%
--- a/scripts/ci/travis-install.sh
+++ b/scripts/ci/travis-install.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+###
+# Copyright 2017 resin.io
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###
+
+set -e
+set -u
+
+if [ -z "$TRAVIS_OS_NAME" ]; then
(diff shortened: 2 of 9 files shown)


lurch:
Works on Travis but fails on Appveyor - weird...

jviotti:
Hm, must be the npm version running on Appveyor. I'll make sure they are synced.

scripts/ci/travis-install.sh
```diff
@@ -0,0 +1,44 @@
...
+
+if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+  ./scripts/build/docker/run-command.sh \
+    -r "$TARGET_ARCH" \
+    -s "$(pwd)" \
+    -c "make info && make electron-develop"
+else
+  ./scripts/build/check-dependency.sh pip
+  ./scripts/build/check-dependency.sh brew
+  ./scripts/build/check-dependency.sh make
```
lurch: Could check for `npm` too?

scripts/ci/appveyor-test.bat
```diff
@@ -0,0 +1,25 @@
...
+:: WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+:: See the License for the specific language governing permissions and
+:: limitations under the License.
+
+IF "%APPVEYOR_REPO_BRANCH%"=="" (
+  ECHO This script is only meant to run in Appveyor CI 1>&2
+  EXIT /B 1
+)
+
+make sanity-checks
```
lurch: I guess you'll need to add MinGW to the %PATH% before `make` will work?

scripts/ci/travis-install.sh
jviotti: Done!

scripts/ci/appveyor-test.bat
jviotti: I moved the PATH declarations to `appveyor.yml` so they get re-used everywhere.

jviotti:
> A shame there's so much redundancy (repeated info) between all the scripts, but given the platform-specific and CI-service-specific nature to all this, I guess there's not much that can be done about that.

Yeah, I want to at least try moving everything to bash scripts we can call with MinGW. What do you think? I might do it in another PR soon :)

jviotti:
Looks like we have to prefix any external command we call in a `.bat` file with `call`, otherwise the script exits successfully, but the commands don't actually run :/

jviotti:
All tests finally green :D

scripts/ci/appveyor-test.bat
lurch: Haha, _back_ to `appveyor.yml` ;-)

lurch:
> Yeah, I want to at least try moving everything to bash scripts we can call with MinGW. What do you think?

You could, but there'll still be differences between the setup needed on Appveyor and Travis (and/or differences between OSes), and I was thinking also about the duplication of the "make" commands in the Travis scripts for the Docker and non-Docker versions.

PR #1379, opened 2017-05-03
chore: check that there are no unstaged shrinkwrap changes

jviotti:
See: https://github.com/resin-io/etcher/pull/1372

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,7 +14,7 @@ cache:
 environment:
   global:
     ELECTRON_NO_ATTACH_CONSOLE: true
-    nodejs_version: 6.1.0
+    nodejs_version: "6.10.3"
   matrix:
     - TARGET_ARCH: x64
     - TARGET_ARCH: x86
@@ -23,7 +23,7 @@ matrix:
   fast_finish: true
 
 install:
-  - ps: Install-Product node $env:nodejs_version x64
+  - ps: Update-NodeJsInstallation $env:nodejs_version $env:TARGET_ARCH
   - set PATH=C:\Program Files (x86)\Windows Kits\8.1\bin\x86;%PATH%
   - set PATH=C:\Program Files (x86)\NSIS;%PATH%
   - set PATH=C:\MinGW\bin;%PATH%
--- a/package.json
+++ b/package.json
@@ -64,9 +64,6 @@
       "verbosity": 1
     }
   },
-  "optionalDependencies": {
-    "electron-installer-debian": "0.5.1"
-  },
   "dependencies": {
     "angular": "1.6.3",
     "angular-if-state": "^1.0.0",
--- a/scripts/build/electron-installer-debian-linux.sh
+++ b/scripts/build/electron-installer-debian-linux.sh
@@ -25,9 +25,7 @@ if [[ "$OS" != "Linux" ]]; then
   exit 1
 fi
 
-ELECTRON_INSTALLER_DEBIAN="./node_modules/.bin/electron-installer-debian"
-
-./scripts/build/check-dependency.sh "$ELECTRON_INSTALLER_DEBIAN"
+./scripts/build/check-dependency.sh electron-installer-debian
 
 function usage() {
   echo "Usage: $0"
@@ -67,7 +65,7 @@ fi
 DEBIAN_ARCHITECTURE=$(./scripts/build/architecture-convert.sh -r "$ARGV_ARCHITECTURE" -t debian)
 
 cp scripts/build/debian/etcher-electron.sh "$ARGV_DIRECTORY"
-"$ELECTRON_INSTALLER_DEBIAN" \
+electron-installer-debian \
   --src "$ARGV_DIRECTORY" \
   --dest "$ARGV_OUTPUT" \
   --config "$ARGV_DEBIAN_CONFIGURATION" \
--- a/scripts/ci/ensure-all-node-requirements-available.sh
+++ b/scripts/ci/ensure-all-node-requirements-available.sh
@@ -31,7 +31,7 @@ PACKAGE_JSON=package.json
 # workaround for path-length bug in jq that only affects Windows https://github.com/stedolan/jq/issues/1155
 NODE_MODULES=($(cat "$HERE/builtin-modules.json" | jq -r '.[]'))
 NPM_MODULES=($(jq -r '.dependencies | keys | .[]' "$PACKAGE_JSON"))
-NPM_OPTIONAL_MODULES=($(jq -r '.optionalDependencies | keys | .[]' "$PACKAGE_JSON"))
+NPM_OPTIONAL_MODULES=($(jq -r '.optionalDependencies // {} | keys | .[]' "$PACKAGE_JSON"))
 NPM_DEV_MODULES=($(jq -r '.devDependencies | keys | .[]' "$PACKAGE_JSON"))
 
 DEV_FILES_REGEX=^\(tests\|scripts\)/
@@ -91,7 +91,7 @@ git ls-tree -r HEAD | while IFS='' read line; do
             done
           fi
           # Check optionalDependencies from package.json
-          if [[ $requirement_found -eq 0 ]]; then
+          if [[ $requirement_found -eq 0 ]] && [[ -n "${NPM_OPTIONAL_MODULES:-}" ]]; then
             for module in "${NPM_OPTIONAL_MODULES[@]}"; do
               if [[ "$required" == "$module" ]]; then
                 requirement_found=1
--- a/scripts/ci/ensure-staged-shrinkwrap.sh
+++ b/scripts/ci/ensure-staged-shrinkwrap.sh
@@ -19,13 +19,18 @@
 set -e
 set -u
 
-npm run sass
+echo "node version: $(node --version)"
+echo "npm version: $(npm --version)"
 
-# From http://stackoverflow.com/a/9393642/1641422
-if [[ -n $(git status -s | grep "\\.css$" || true) ]]; then
-  echo "There are unstaged sass changes. Please commit the result of:" 1>&2
+SHRINKWRAP_FILE=npm-shrinkwrap.json
+
+npm shrinkwrap --dev
+
+if [[ -n $(git status -s "$SHRINKWRAP_FILE") ]]; then
+  echo "There are unstaged $SHRINKWRAP_FILE changes. Please commit the result of:" 1>&2
   echo ""
-  echo "    npm run sass" 1>&2
+  echo "    npm shrinkwrap --dev" 1>&2
   echo ""
+  git --no-pager diff "$SHRINKWRAP_FILE"
   exit 1
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 sudo: false
 node_js:
-  - "6.1.0"
+  - "6.10.3"
 
 services:
   - docker
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ BUILD_OUTPUT_DIRECTORY = $(BUILD_DIRECTORY)/out
 # ---------------------------------------------------------------------
 
 ELECTRON_VERSION = $(shell jq -r '.devDependencies["electron"]' package.json)
-NODE_VERSION = 6.1.0
+NODE_VERSION = 6.10.3
 COMPANY_NAME = $(shell jq -r '.companyName' package.json)
 APPLICATION_NAME = $(shell jq -r '.displayName' package.json)
 APPLICATION_DESCRIPTION = $(shell jq -r '.description' package.json)
@@ -540,6 +540,7 @@ info:
 sanity-checks:
 	./scripts/ci/ensure-all-node-requirements-available.sh
 	./scripts/ci/ensure-staged-sass.sh
+	./scripts/ci/ensure-staged-shrinkwrap.sh
 	./scripts/ci/ensure-npm-dependencies-compatibility.sh
 	./scripts/ci/ensure-npm-valid-dependencies.sh
 	./scripts/ci/ensure-npm-shrinkwrap-versions.sh
--- a/scripts/build/docker/Dockerfile-i686
+++ b/scripts/build/docker/Dockerfile-i686
@@ -34,7 +34,7 @@ RUN add-apt-repository ppa:ubuntu-toolchain-r/test \
 RUN curl -sL https://deb.nodesource.com/setup_6.x | bash - \
   && apt-get install -y nodejs
 RUN npm config set spin=false
-RUN npm install -g uglify-es@3.0.3
+RUN npm install -g uglify-es@3.0.3 electron-installer-debian@0.5.1
 
 # Python
 RUN pip install codespell==1.9.2 awscli cpplint
--- a/scripts/build/docker/Dockerfile-x86_64
+++ b/scripts/build/docker/Dockerfile-x86_64
@@ -34,7 +34,7 @@ RUN add-apt-repository ppa:ubuntu-toolchain-r/test \
 RUN curl -sL https://deb.nodesource.com/setup_6.x | bash - \
   && apt-get install -y nodejs
 RUN npm config set spin=false
-RUN npm install -g uglify-es@3.0.3
+RUN npm install -g uglify-es@3.0.3 electron-installer-debian@0.5.1
 
 # Python
 RUN pip install codespell==1.9.2 awscli cpplint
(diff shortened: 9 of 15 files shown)


scripts/ci/ensure-staged-shrinkwrap.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
lurch: No license header block?

lurch:
Should this link up to `clean-shrinkwrap.js` in any way, or are they entirely different stages?

lurch:
Failing on both Appveyor and Travis :-/

scripts/ci/ensure-staged-shrinkwrap.sh
jviotti: I copied from the sass staged files script, which doesn't have a license neither. I added it to both in this PR.

jviotti:
> Should this link up to clean-shrinkwrap.js in any way, or are they entirely different stages?

That script is defined in `package.json` as `"preshrinkwrap": "node ./scripts/clean-shrinkwrap.js"`, which means that npm will automatically call it before running shrinkwrap.

scripts/ci/ensure-staged-sass.sh
```diff
@@ -1,5 +1,21 @@
 #!/bin/bash
 
+###
+# Copyright 2017 resin.io
```
lurch: Haha, I had to double-check and indeed this file was only created in 2017! Feels like much longer ago ;-)

lurch:
In a change of circumstances, it's now working on Appveyor but failing on Travis!
Out of curiosity, I cloned this branch on my Ubuntu laptop, and running `make sanity-checks` also tells me that there's uncommited changes. Doing `git diff` reveals loads of entries like:
```diff
diff --git a/npm-shrinkwrap.json b/npm-shrinkwrap.json
index f78f4fd..c4b9c56 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -100,17 +100,17 @@
     },
     "angular-moment": {
       "version": "1.0.1",
-      "from": "angular-moment@1.0.1",
+      "from": "angular-moment@>=1.0.1 <2.0.0",
       "resolved": "https://registry.npmjs.org/angular-moment/-/angular-moment-1
     },
     "angular-seconds-to-date": {
       "version": "1.0.0",
-      "from": "angular-seconds-to-date@latest",
+      "from": "angular-seconds-to-date@>=1.0.0 <2.0.0",
       "resolved": "https://registry.npmjs.org/angular-seconds-to-date/-/angular
     },
     "angular-ui-bootstrap": {
       "version": "2.5.0",
-      "from": "angular-ui-bootstrap@>=2.5.0 <2.6.0",
+      "from": "angular-ui-bootstrap@>=2.5.0 <3.0.0",
       "resolved": "https://registry.npmjs.org/angular-ui-bootstrap/-/angular-ui
     },
     "angular-ui-router": {
@@ -892,7 +892,7 @@
     },
     "colors": {
       "version": "1.1.2",
-      "from": "colors@>=1.1.2 <2.0.0",
+      "from": "colors@*",
       "resolved": "https://registry.npmjs.org/colors/-/colors-1.1.2.tgz"
     },
     "column-layout": {
@@ -920,7 +920,7 @@
     },
     "command-join": {
       "version": "2.0.0",
-      "from": "command-join@latest",
+      "from": "command-join@>=2.0.0 <3.0.0",
       "resolved": "https://registry.npmjs.org/command-join/-/command-join-2.0.0
     },
     "command-line-args": {
@@ -948,7 +948,7 @@
       "dependencies": {
         "commander": {
           "version": "2.9.0",
-          "from": "commander@^2.5.0",
+          "from": "commander@>=2.5.0 <3.0.0",
           "resolved": "https://registry.npmjs.org/commander/-/commander-2.9.0.t
           "dev": true
         },
@@ -998,7 +998,7 @@
       "dependencies": {
         "isarray": {
           "version": "1.0.0",
-          "from": "isarray@~1.0.0",
+          "from": "isarray@>=1.0.0 <1.1.0",
           "resolved": "https://registry.npmjs.org/isarray/-/isarray-1.0.0.tgz",
           "dev": true
         },
@@ -1188,7 +1188,7 @@
       "dependencies": {
         "debug": {
           "version": "0.7.4",
-          "from": "debug@~0.7.2",
+          "from": "debug@>=0.7.2 <0.8.0",
           "resolved": "https://registry.npmjs.org/debug/-/debug-0.7.4.tgz",
           "dev": true
         }
```
etc. etc.

My NPM version is 3.10.0 and my nodejs version is v6.10.3
Given how fragile `npm-shrinkwrap.json` appears to be, do we need to ensure that all the Etcher developers are running the identical NPM version? :-/

jviotti:
Hm, that is insane. We'll unify the npm version on the CI servers, and that can be the "recommended" npm version.

lurch:
> Hm, that is insane.

Well, you know NPM much better than I do ;-)

Appveyor seems to be all working, but Travis on Linux x64 is failing with `Error: Cannot find module 'npmlog'` and Travis on Linux-x86 and OSX still thinks npm-shrinkwrap has unstaged changes?
Downloading and comparing the Travis logfiles for the (failed) Linux-x86 and Linux-x64 Travis builds, I can't see why they're giving different errors :-/

> that can be the "recommended" npm version.

Maybe this PR should also update part of the documentation?

lurch:
Hmmm, and I've just deleted and then re-cloned my local copy of this branch (on Ubuntu 14.04), and after doing `make electron-develop`, `make sanity-checks` tells me npm-shrinkwrap has changed, and `git diff` again shows me that `npm-shrinkwrap.js` has the identical diff to what I reported in my previous message! And yet Appveyor is apparently using the same npm 3.10.10 that I am, and yet that passes `make sanity-checks` okay??? :cry:  (I have no idea what's going on)

Perhaps the 'fix' is to filter out all the 'from:' lines in npm-shrinkwrap (e.g. by creating some temporary files using `grep -v`) before comparing them in `ensure-staged-shrinkwrap.sh` ? :confused:

lurch:
Pinging @jhermsmeier in case he has any good ideas or can help diagnose what's going on ;-)

jhermsmeier:
The `Error: Cannot find module 'npmlog'` is npm exploding because it's missing a dependency (meaning the npm CLI is actually missing a dep, not Etcher) – not sure how exactly that's happening, but the Travis builds are using npm@3.8.6. Maybe we could resort to doing an `npm install --global npm@latest` (or a version we know works) beforehand...

lurch:
The Linux builds on Travis are actually running inside Docker containers, so the version of npm on Travis is (I believe) only relevant on OSX.

jviotti:
This whole thing is crazy, but we should definitely find a single version that works everywhere. @lurch Can you help test npm versions to find one that makes the test pass on Ubuntu 14.04?

jviotti:
I made the script print the actual diff, so at least we get a better chance of figuring out what's going on.

lurch:
Wow, we're now getting different errors / output on all 3 Travis platforms :sob:

jviotti:
Damn, Windows shows errors but its finishing successfully: https://ci.appveyor.com/project/resin-io/etcher/build/1.0.4011/job/4214mu6tio1ttwxx

jviotti:
At least OS X is finally green :P

lurch:
Finally green, or temporarily green? :stuck_out_tongue_winking_eye:

lurch:
Just noticed that if my `npm-shrinkwrap.js` becomes modified, then if I run `make sanity-checks` a second time, it tells me `There are unstaged sass changes.`.
Perhaps these ensure-staged-_foo_ scripts should only check the status of the files they actually care about?

jviotti:
Makes sense. I'll tweak the scripts.

Looks like the Linux ones are failing because of the `electron-installer-debian` optional dependency.

jviotti:
I'll lock it in the Dockerfiles. Optional development dependencies are a risky combination for these kinds of checks.

npm-shrinkwrap.json
```diff
@@ -4233,6 +4238,11 @@
               "from": "isarray@>=1.0.0 <1.1.0",
               "resolved": "https://registry.npmjs.org/isarray/-/isarray-1.0.0.tgz"
             },
+            "isstream": {
```
lurch: I'm not doubting that this change was needed, but just out of curiosity how was this not here before? Or is it just another side-effect of all the madness of the different NPM versions?

scripts/build/electron-installer-debian-linux.sh
```diff
@@ -25,9 +25,7 @@ if [[ "$OS" != "Linux" ]]; then
   exit 1
 fi
 
-ELECTRON_INSTALLER_DEBIAN="./node_modules/.bin/electron-installer-debian"
-
-./scripts/build/check-dependency.sh "$ELECTRON_INSTALLER_DEBIAN"
+./scripts/build/check-dependency.sh electron-installer-debian
```
lurch: Haha, funny how things go around in circles ;-) ( https://github.com/resin-io/etcher/commit/880ca257a673da94d6c58c67c43d124a0be74a35#diff-d3e903f23548d60d01b1c17f94913c6e )

scripts/ci/ensure-all-node-requirements-available.sh
```diff
@@ -31,7 +31,7 @@ PACKAGE_JSON=package.json
 # workaround for path-length bug in jq that only affects Windows https://github.com/stedolan/jq/issues/1155
 NODE_MODULES=($(cat "$HERE/builtin-modules.json" | jq -r '.[]'))
 NPM_MODULES=($(jq -r '.dependencies | keys | .[]' "$PACKAGE_JSON"))
-NPM_OPTIONAL_MODULES=($(jq -r '.optionalDependencies | keys | .[]' "$PACKAGE_JSON"))
+NPM_OPTIONAL_MODULES=($(jq -r '.optionalDependencies // {} | keys | .[]' "$PACKAGE_JSON"))
```
lurch: Probably worth adding this additional syntax to the lines above and below too? (for extra robustness)

npm-shrinkwrap.json
jviotti: This dependency is declared on the root of the shrinkwrap file, however it looks like npm 3.10.10 insists that it should be duplicated inside the nested dependency that also requires it.

scripts/ci/ensure-staged-shrinkwrap.sh
```diff
@@ -0,0 +1,34 @@
...
+
+set -e
+set -u
+
+echo "node version: $(node --version)"
+echo "npm version: $(npm --version)"
+
+npm shrinkwrap --dev
+
+if [[ -n $(git status -s npm-shrinkwrap.json) ]]; then
```
lurch: Could we make `npm-shrinkwrap.json` a variable please? (given that it's used in 3 separate locations)
It'll make other `ensure-staged-blah.sh` scripts easier to write in future :)

scripts/ci/ensure-all-node-requirements-available.sh
```diff
@@ -91,7 +91,7 @@ git ls-tree -r HEAD | while IFS='' read line; do
             done
           fi
           # Check optionalDependencies from package.json
-          if [[ $requirement_found -eq 0 ]]; then
+          if [[ $requirement_found -eq 0 ]] && [[ -n "${NPM_OPTIONAL_MODULES:-}" ]]; then
```
lurch: Small nit, but I think I'd rather revert this line to how it was, and then use http://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u to change the line below to:
```
for module in "${NPM_OPTIONAL_MODULES[@]+${NPM_OPTIONAL_MODULES[@]}}"; do
```
(and as with my earlier comment, for extra robustness similar changes could be applied to lines 86 and 104)

scripts/ci/ensure-staged-sass.sh
```diff
@@ -1,12 +1,28 @@
...
+###
+
 set -e
 set -u
 
 npm run sass
 
 # From http://stackoverflow.com/a/9393642/1641422
-if [[ -n $(git status -s) ]]; then
+if [[ -n $(git status -s | grep "\\.css$" || true) ]]; then
```
lurch: Could this be simplified to `if [[ -n $(git status -s lib/gui/*.css) ]]; then` ?
Although I see in `package.json` we have:
```
"sass": "node-sass ./lib/gui/scss/main.scss > ./lib/gui/css/main.css",
```
so potentially this line could even be `if [[ -n $(git status -s lib/gui/main.css) ]]; then` ?

appveyor.yml
```diff
@@ -14,7 +14,7 @@ cache:
 environment:
   global:
     ELECTRON_NO_ATTACH_CONSOLE: true
-    nodejs_version: 6.1.0
+    nodejs_version: "6.10.3"
```
lurch: I still think that we ought to document this 'required' nodejs version somewhere, even if it's only the Etcher developers who actually need to lock themselves to this version. (because we want the updated `npm-shrinkwrap.json` files to stay compatible with all our lovely CI checks)

appveyor.yml
```diff
@@ -23,7 +23,7 @@ matrix:
   fast_finish: true
 
 install:
-  - ps: Install-Product node $env:nodejs_version x64
+  - ps: Update-NodeJsInstallation $env:nodejs_version $env:TARGET_ARCH
```
lurch: Hmmm? How come trying to install the x64 version of node on x86 Windows wasn't previously causing problems?

npm-shrinkwrap.json
lurch: ...Obviously! Thanks NPM :-S

lurch:
BTW, any idea if:
```
warning: LF will be replaced by CRLF in npm-shrinkwrap.json.
The file will have its original line endings in your working directory.
```
that Appveyor is reporting is anything to worry about?
https://github.com/resin-io/etcher/blob/master/.gitattributes reports that `*.json` files should be treated as `text`, so I've no idea 'who' or 'where' the above warning is coming from.

jhermsmeier:
@lurch Looks like that's just git's auto-CRLF setting being set to change to Windows line endings on Appveyor, see https://www.git-scm.com/docs/gitattributes#_checking_out_and_checking_in – maybe we should turn autocrlf off, to avoid any hard to track down problems.

lurch:
I think we should leave autocrlf alone for now (see #1192 ), but bear it in mind in case it causes any problems in the future.

appveyor.yml
jviotti: We were setting x86 on the node-gyp settings, so it all just worked fine, however I decided to make this change as a nicety (I didn't know Appveyor supported x86 Node builds)

scripts/ci/ensure-staged-sass.sh
jviotti: I'm going to be off for a couple of hours, and given you're the bash expert, do you mind sending a new PR with all these changes, and then I'll rebase this one?

scripts/ci/ensure-all-node-requirements-available.sh
jviotti: Yeah, I think you should send a different PR with these changes (I have no idea what I'm doing), and I'll rebase afterwards.

scripts/ci/ensure-staged-shrinkwrap.sh
jviotti: Done!

scripts/ci/ensure-staged-sass.sh
lurch: Sure :-)

package.json
```diff
@@ -64,9 +64,6 @@
       "verbosity": 1
     }
   },
-  "optionalDependencies": {
```
lurch: So, if the entirety of `optionalDependencies` has been removed, does that mean the entirety of `shrinkwrapIgnore` should be removed too?

scripts/ci/ensure-staged-sass.sh
lurch: I started modifying the bash scripts yesterday, but got a bit carried away with making _all_ our bash scripts more uniform (and then got side-tracked with the separate CLI-path-escaping stuff), so feel free to merge this if the CI tests are passing, and I can PR my bash-changes later.

package.json
jviotti: This is sadly still needed. There are optional dependencies of some development dependencies that when added to the shrinkwrap file, will cause the other platforms to complain about extraneous dependencies.

lurch: Ahhhhh, right.

lurch:
As a postscript to all this: It seems that I was still getting different `from: ` lines in my `npm-shrinkwrap.json` (see above) because of "something" in my `node_modules` folder. When I deleted my `node_modules` folder, running `make electron-develop` and then `make sanity-checks` showed my `npm-shrinkwrap.json` as remaining unchanged :-)

Issue #1356, opened 2017-04-25
Store settings in an `.etcherrc` file instead of in local storage


jviotti:
We could also allow overriding these options with environment variables, but that's a bonus item.

lurch:
So we're actually "downgrading" from a database to a flat text file, instead of the other way around? ;-D

jviotti:
Haha, exactly. This is particular useful for the hard etcher project, since automatically tweaking certain settings is very difficult/impossible, while a plain text configuration file makes it simpler.

PR #1382, opened 2017-05-03
WIP feat(GUI): move settings to an `.etcherrc` JSON file

jviotti:
This commit makes Etcher stop storing settings in localStorage, and
instead moves them to a settings file (`.etcherrc`) in the home
directory. This change is motivated by the following reasons:

- The Hard Etcher project needs an easy way to pre-configure Etcher, and
  localStorage is an inaccessible place

- The Redux store is not supposed to contain impure code (e.g: updating
  localStorage)

The below changes can be summarised as:

- Introduce `lib/shared/settings.js`, which handles the gory details of
  reading and writing to/from the setting file in the home directory

- Remove `redux-localstorage`

- Delete the `SET_SETTING` Redux action and add a new `SET_SETTINGS`
  action that is able to set the whole object

- Make the settings model merge the current changes with the new ones
  when the `.set()` function is called

- Make the settings model write to the configuration file before
  updating the Redux store. If there is any error during the write, the
  Redux store is not updated

- Since now we're doing I/O, the `.set()` setting function is now async.
  Update its usage everywhere

See: https://github.com/resin-io/etcher/issues/1356

### DIFF
--- a/lib/gui/models/settings.js
+++ b/lib/gui/models/settings.js
@@ -20,28 +20,11 @@
  * @module Etcher.Models.Settings
  */
 
+const _ = require('lodash');
+const Bluebird = require('bluebird');
 const Store = require('./store');
-
-/**
- * @summary Set a setting value
- * @function
- * @public
- *
- * @param {String} key - setting key
- * @param {*} value - setting value
- *
- * @example
- * settings.set('unmountOnSuccess', true);
- */
-exports.set = (key, value) => {
-  Store.dispatch({
-    type: Store.Actions.SET_SETTING,
-    data: {
-      key,
-      value
-    }
-  });
-};
+const errors = require('../../shared/errors');
+const settings = require('../../shared/settings');
 
 /**
  * @summary Get a setting value
@@ -71,4 +54,47 @@ exports.get = (key) => {
  */
 exports.getAll = () => {
   return Store.getState().get('settings').toJS();
+};
+
+exports.setAll = (object) => {
+  return settings.writeAll(settings.getConfigurationFilePath(), object).then(() => {
+    Store.dispatch({
+      type: Store.Actions.SET_SETTINGS,
+      data: object
+    });
+  });
+};
+
+/**
+ * @summary Set a setting value
+ * @function
+ * @public
+ *
+ * @param {String} key - setting key
+ * @param {*} value - setting value
+ * @returns {Promise}
+ *
+ * @example
+ * settings.set('unmountOnSuccess', true).then(() => {
+ *   console.log('Done!');
+ * });
+ */
+exports.set = (key, value) => {
+  if (_.isNil(key)) {
+    return Bluebird.reject(errors.createError({
+      title: 'Missing setting key'
+    }));
+  }
+
+  if (!_.isString(key)) {
+    return Bluebird.reject(errors.createError({
+      title: `Invalid setting key: ${key}`
+    }));
+  }
+
+  const newSettings = _.merge(exports.getAll(), _.fromPairs([
+    [ key, value ]
+  ]));
+
+  return exports.setAll(newSettings);
 };
--- a/lib/shared/settings.js
+++ b/lib/shared/settings.js
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2017 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const path = require('path');
+const _ = require('lodash');
+const Bluebird = require('bluebird');
+const fs = Bluebird.promisifyAll(require('fs'));
+const os = require('os');
+const errors = require('./errors');
+const packageJSON = require('../../package.json');
+
+/**
+ * @summary Get the file path of the application configuration file
+ * @function
+ * @private
+ *
+ * @returns {String} configuration file path
+ *
+ * @example
+ * const configurationFilePath = settings.getConfigurationFilePath();
+ */
+exports.getConfigurationFilePath = () => {
+  const hiddenFilePrefix = os.platform() === 'win32' ? '_' : '.';
+  return path.join(os.homedir(), `${hiddenFilePrefix}${packageJSON.name}rc`);
+};
+
+/**
+ * @summary Check if a setting value is invalid
+ * @function
+ * @private
+ *
+ * @description
+ * Our settings implementation only supports primitive values.
+ *
+ * @param {*} value - settings value
+ * @returns {Boolean} whether the setting value is invalid
+ *
+ * @example
+ * if (isInvalidSettingValue([ 1, 2, 3 ])) {
+ *   console.log('This setting value is invalid');
+ * }
+ */
+const isInvalidSettingValue = _.overSome([ _.isArray, _.isObject ]);
+
+/**
+ * @summary Read all settings from a configuration file
+ * @function
+ * @public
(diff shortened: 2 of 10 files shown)


jviotti:
The tests are failing because the update notifier modal relies heavily on the synchronicity of the settings module. This problem is resolved by https://github.com/resin-io/etcher/pull/1359, so we'll rebase this PR once that one is merged.

jviotti:
TODO:

- [ ] Read configuration values from `.etcherrc` during startup

lurch:
So, does this also mean a whole bunch of these files can be moved from `lib/gui/...` to `lib/shared/...`?
For diff-reviewing purposes, perhaps this PR should be done as two separate commits (with the second commit being the one that moves files around).

tests/shared/settings.spec.js
```diff
@@ -0,0 +1,419 @@
...
+
+describe('Shared: Settings', function() {
+
+  describe('.getConfigurationFilePath()', function() {
+
+    it('should return an absolute path', function() {
+      m.chai.expect(path.isAbsolute(settings.getConfigurationFilePath())).to.be.true;
+    });
+
+    it('should be a hidden file', function() {
```
lurch: I don't believe filenames starting with an underscore are actually hidden on Windows?

tests/shared/settings.spec.js
```diff
@@ -0,0 +1,419 @@
...
+      it('should resolve an empty object', function(done) {
+        settings.readAll(this.configurationFilePath).then((data) => {
+          m.chai.expect(data).to.deep.equal({});
+          done();
+        }).catch(done);
+      });
+
+    });
+
+    describe('given the file contains invalid data', function() {
```
lurch: Could we change this description to 'invalid JSON' ?

tests/shared/settings.spec.js
```diff
@@ -0,0 +1,419 @@
...
+        fs.writeFileSync(this.configurationFilePath, JSON.stringify({
+          foo: 'bar',
+          bar: {
+            baz: 'hey'
+          },
+          baz: [ 1, 2, 3 ]
+        }));
+      });
+
+      it('should ignore the scalar types', function(done) {
```
lurch: ignore the non-scalar types?

tests/shared/settings.spec.js
```diff
@@ -0,0 +1,419 @@
...
+            bar: 'baz',
+            baz: 'qux'
+          });
+          done();
+        }).catch(done);
+      });
+
+    });
+
+    describe('given a file with nested settings', function() {
```
lurch: with nested settings and an array?

tests/shared/settings.spec.js
```diff
@@ -0,0 +1,419 @@
...
+        settings.writeAll(this.configurationFilePath, _.noop).catch((error) => {
+          m.chai.expect(errors.isUserError(error)).to.be.false;
+          m.chai.expect(error.message).to.equal('Invalid settings');
+          done();
+        });
+      });
+
+    });
+
+    describe('given the settings are a string', function() {
```
lurch: Might also be worth adding a test for 'given the settings are a JSON string' ?

lib/gui/models/settings.js
```diff
@@ -72,3 +55,46 @@ exports.get = (key) => {
...
+  });
+};
+
+/**
+ * @summary Set a setting value
+ * @function
+ * @public
+ *
+ * @param {String} key - setting key
+ * @param {*} value - setting value
```
lurch: Should this mention that only primitive values are supported?

lib/gui/models/settings.js
```diff
@@ -72,3 +55,46 @@ exports.get = (key) => {
 exports.getAll = () => {
   return Store.getState().get('settings').toJS();
 };
+
+exports.setAll = (object) => {
```
lurch: Missing docstring?

npm-shrinkwrap.json
```diff
@@ -5705,11 +5705,6 @@
       "from": "redux@>=3.5.2 <4.0.0",
       "resolved": "https://registry.npmjs.org/redux/-/redux-3.5.2.tgz"
     },
-    "redux-localstorage": {
```
lurch: Blimey, surprised that this only results in the removal of one package ;)

lurch:
Spotted an issue today in rc5, which is mainly related to #1183 but possibly partially also related to this PR.

I upgraded from rc4 to rc5, and was somewhat surprised to find the "Include unstable update channel" setting disabled by default! I did some investigation, and here's what I found:
* If I manually set "Include unstable update channel" and then quit and re-launch Etcher rc5, then it stays enabled
* If I delete the localStorage (`rm ~/.config/etcher/Local\ Storage/file__0.localstorage` on Ubuntu) and then launch Etcher rc5, then "Include unstable update channel" _does_ default to on, as I'd expect it to
* If I delete the localStorage and then launch Etcher rc4, quit and then launch Etcher rc5, then "Include unstable update channel" defaults to off :-(

So it seems like the default-values we set in `lib/gui/models/store.js` are only used when there are _no_ previous settings at all. If _some_ previous settings exist (e.g. from running an rc4 or earlier version of Etcher) then the default values aren't used for any "missing values" (e.g. `includeUnstableUpdateChannel`) when upgrading from rc4 to rc5.
Could we fix that?

I _think_ the current behaviour means that if somebody had never used Etcher before, and installed rc5 as their first version, then they would get notified (as expected) when rc6 is released. Whereas if somebody had Etcher rc4 installed and then upgraded to rc5, then they _won't_ get notified when rc6 is released? :-(

lib/gui/models/settings.js
```diff
@@ -72,3 +55,46 @@ exports.get = (key) => {
...
+      title: `Invalid setting key: ${key}`
+    }));
+  }
+
+  const newSettings = _.merge(exports.getAll(), _.fromPairs([
+    [ key, value ]
+  ]));
+
+  return exports.setAll(newSettings);
+};
```
lurch: In the spirit of CRUD, does there need to be / should there be a way of _deleting_ settings too?

tests/gui/pages/main.spec.js
```diff
@@ -245,11 +245,13 @@ describe('Browser: MainPage', function() {
...
-          settings.set('unmountOnSuccess', true);
-          m.chai.expect(controller.getProgressButtonLabel()).to.equal('0%');
+          settings.set('unmountOnSuccess', true).then(() => {
+            m.chai.expect(controller.getProgressButtonLabel()).to.equal('0%');
+            done();
+          }).catch(done);
         });
 
-        it('should handle percentage == 0, type = write, unmountOnSuccess', function() {
+        it('should handle percentage == 0, type = write, unmountOnSuccess', function(done) {
```
jhermsmeier: This is correct, as `.catch(done)` handles errors, but could be shortened to
```js
it('should report 0% if percentage == 0 but speed != 0', function() {
  return settings.set('unmountOnSuccess', true).then(() => {
    m.chai.expect(controller.getProgressButtonLabel()).to.equal('0%');
  })
})
```

lib/shared/settings.js
```diff
@@ -0,0 +1,150 @@
...
+ * }).then(() => {
+ *   console.log('Done!');
+ * });
+ */
+exports.writeAll = (configurationFilePath, settings) => {
+  const invalidSettings = _.pickBy(settings, isInvalidSettingValue);
+  const JSON_INDENTATION_SPACES = 2;
+  const serialisedData = JSON.stringify(settings, null, JSON_INDENTATION_SPACES);
+
+  if (!_.isString(serialisedData) || !_.isPlainObject(settings) || !_.isEmpty(invalidSettings)) {
```
lurch: @jviotti Do we have unit-tests for the cases @Shou just mentioned? ;)

Shou: Actually I'm wrong! I tried `_.isPlainObject(function(){})` again and it returns `false`—I must've prefixed it with `!` by mistake, copying the code. This means we *can* remove the first predicate in the or-chain because `JSON.stringify(() => {})` → `undefined`, `!_.isString(undefined)` → `false`, and `!_.isPlainObject(() => {})` → `false`, hence `false || false` → `false`.

jviotti:
I'll close this since its mostly addressed by:

- https://github.com/resin-io/etcher/pull/1515
- https://github.com/resin-io/etcher/pull/1569

And other small PRs I'm working on.

PR #1258, opened 2017-04-05
fix(GUI): limit progress button label to 100

Shou:
We limit the progress button's label to 100.

Changelog-Type: Limit the progress button's label to 100.

PR #1390, opened 2017-05-09
refactor: address code review comments from #1366

jviotti:
See: https://github.com/resin-io/etcher/pull/1366


lurch:
`elevate.h` also includes Windows-specific code, so should probably be moved to the same location as `elevate.cpp`? (either that, or the Windows-specific code should be removed from `elevate.h`)

jviotti:
I surrounded the Windows specific includes with a preprocessor conditional.

lurch:
Hmmm, another (single) failed Travis build because of a timeout error in the html validator... (see also #1258 )

Issue #1348, opened 2017-04-24
Add support for .sdcard extension


sebastient:
The .sdcard extension is a common raw image format and supporting it would be handy.  Currently one can rename a .sdcard to .img and it works fine with Etcher but direct support would be even better.

lurch:
Interesting, I'd not heard of that before. Can you link us to some examples, that we can use for testing?

sebastient:
For example Yocto builds will generate .sdcard and .sdcard.gz images.  They're just normal disk images that would get dd'ed to an /dev/DRIVE on Linux/Mac.

alexandrosm:
It's funny you mention that, because resinOS is of course a yocto distro..
so we ourselves make .sdcard files :)

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lurch:
Ping @agherzan - how come you never told us about this? ;-)

agherzan:
These files are not coming from the oe-core but BSPs provide sd card generation mechanisms and most of them use different names for them as there is no standardisation around it. For example TI uses sdimg, RPi uses rpi-sdimg, FSL uses sdcard etc. Even resin uses another extension: resinos-img. I'm not sure about the internal rationale about this filename restriction but if we can't avoid it we could add support for all these variations for the major BSP layers.

adcurtin:
we use .usb, so this is still an issue. The issue should be renamed to reflect what was fixed, or the actual problem should be fixed. IMO the best solution is a warning about unsupported file format, but let me do it anyway. Preventing this because of a file extension just seems wrong.

lurch:
@adcurtin see #2106 and #1360

jhermsmeier:
@adcurtin yeah, that sounds like a good compromise, we could warn & treat it as a flat octet-stream if continued.

Issue #1360, opened 2017-04-26
Allow selection of any file type by using a custom filter


jviotti:
We should have two extensions filters in the open file dialog:

- Supported images: containing all the images formats we know about
- All Files: which allows the user to select any file

lurch:
We believe this should fix #1348

lurch:
Haha, I've just realised that this is a duplicate pf #833  ;-)

gromain:
It is, but it is definitely needed.

My build system generates `.wic` and `.sdcard`, it would be nice to be able to add these "custom" extensions somehow.

jviotti:
@gromain @lurch What do you think about adding these here: https://github.com/resin-io/etcher/blob/master/lib/image-stream/supported.js? We've encountered `.sdcard` multiple times already, so I think that at least that one deserves an entry on that file.

lurch:
@jviotti  Yeah, could do, as long as we still plan to also allow a "*.*" filter.
Should `.etch` be removed from that list too, given that we've kind of abandoned that idea in favour of the online catalog?

jviotti:
Yeah, I agree with both points!

gromain:
Also, wic is the new tool provided by yocto to generate images (akin to .sdcard).
Its default output is a `.wic` file. It would be good to add it too I think.
@lurch, do you take care of the changes, or should I put a pull request?

lurch:
We discussed this previously in #1348 - I'd be a bit hesitant to add support for every weird and wonderful file extension that various image-backup/image-creation tools happen to create. What do you think @jviotti ?

> or should I put a pull request?

Sure, that would be very helpful thank you :+1:

gromain:
No, I completely understand! I was just stating that wic is becoming more and more used in the yocto community and is slowly replacing `.sdcard` .

lurch:
Just out of curiosity, what does `.wic` stand for?

gromain:
Wic is a contraction of oeic which stands for Open-Embedded Image Creator.

See chapter 5.9.2.1 [here](https://www.yoctoproject.org/docs/2.3/dev-manual/dev-manual.html#create-wic-images-wic) in the Yocto Manual.

lurch:
O, I C
(oh, I see) 
:laughing:

jviotti:
> I'd be a bit hesitant to add support for every weird and wonderful file extension that various image-backup/image-creation tools happen to create. What do you think @jviotti ?

[...]

PR #1391, opened 2017-05-09
feat(gui): Revert locking down image file extensions

jhermsmeier:
This reverts minor changes made in https://github.com/resin-io/etcher/pull/1347 and https://github.com/resin-io/etcher/pull/1343 which resulted in user being unable to select images with unrecognised extensions by drag 'n' dropping them into Etcher – which initially was intended.

As it turned out though, there are plenty of other non-standard extensions being used in the wild, like `.rpi-sdimg`, `.sdcard`, etc.

Connects To: #1348
Changelog-Entry: Allow selecting an image of any file extension via drag 'n' drop

### DIFF
--- a/lib/gui/models/store.js
+++ b/lib/gui/models/store.js
@@ -337,21 +337,6 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         });
       }
 
-      if (!action.data.extension) {
-        throw errors.createError({
-          title: 'Missing image extension'
-        });
-      }
-
-      if (_.some([
-        !_.isString(action.data.extension),
-        !_.includes(supportedFormats.getAllExtensions(), action.data.extension)
-      ])) {
-        throw errors.createError({
-          title: `Invalid image extension: ${action.data.extension}`
-        });
-      }
-
       const lastImageExtension = fileExtensions.getLastFileExtension(action.data.path);
 
       if (lastImageExtension !== action.data.extension) {
--- a/lib/gui/pages/main/controllers/image-selection.js
+++ b/lib/gui/pages/main/controllers/image-selection.js
@@ -20,7 +20,6 @@ const _ = require('lodash');
 const Bluebird = require('bluebird');
 const path = require('path');
 const messages = require('../../../../shared/messages');
-const errors = require('../../../../shared/errors');
 const imageStream = require('../../../../image-stream');
 const supportedFormats = require('../../../../shared/supported-formats');
 const analytics = require('../../../modules/analytics');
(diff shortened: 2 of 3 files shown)


lib/gui/pages/main/controllers/image-selection.js
```diff
@@ -67,18 +66,6 @@ module.exports = function(
...
-    if (!supportedFormats.isSupportedImage(image.path)) {
-      const invalidImageError = errors.createUserError({
-        title: 'Invalid image',
-        description: messages.error.invalidImage({
-          image
-        })
-      });
-
-      OSDialogService.showError(invalidImageError);
-      AnalyticsService.logEvent('Invalid image', image);
```
lurch: Even if we don't show an Error dialog, maybe it'd still be worth logging an "Unrecognised image format" event?

And instead of getting rid of the Error dialog completely, maybe we could/should change it to a "Warning" dialog, with a "Proceed anyway" button, like we do for Windows images?

lib/gui/models/store.js
```diff
@@ -336,21 +336,6 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         });
       }
 
-      if (!action.data.extension) {
-        throw errors.createError({
-          title: 'Missing image extension'
```
lurch: If this is no longer an error, I guess we should add some extension-less test files to the unit-tests?

lurch:
So I'm assuming this PR is also a pre-requisite for fixing #1360 ?

lib/gui/pages/main/controllers/image-selection.js
jhermsmeier: Yeah, that's a good idea! I'll see to it.

lib/gui/models/store.js
jhermsmeier: There are a few in the unit tests (which is why CI is failing) – just wanted to get this out quick, because I felt I was missing another, nicer way to this, as you pointed out above.

jhermsmeier:
> So I'm assuming this PR is also a pre-requisite for fixing #1360 ?

Yup, I think so – unless that takes a different code path – I think I've come across two or three places that each check filename extensions so far.

lurch:
> I think I've come across two or three places that each check filename extensions so far.

Another place for refactoring?

jhermsmeier:
Closing for now, as this is too stale at this point.

lurch:
It would obviously be nice to see something similar to this come back in future though, so that we don't have to keep adding all kinds of weird & wonderful extensions to Etcher's list of "recognised extensions" ;-)

Issue #1383, opened 2017-05-04
Slightly ugly settings screen in 1.0.0-rc.5


lurch:
![image](https://cloud.githubusercontent.com/assets/476186/25703881/4b246966-30cf-11e7-9ab9-ce7891cbc8ae.png)

Just IMHO, but it looks a little "bottom heavy", and I don't like the way the red "Dangerous" box touches the line below it.
Pinging @konmouz and @Shou to do their prettification magic :-)

konmouz:
@lurch, settings will be transferred to a modal window and unsafe mode will be moved to the drives' modal. I'll try to find the relevant issues.

lurch:
Are those changes still planned to be squeezed in before the release of 1.0.0 ? @jviotti ?

Shou:
Maybe we can push it into the empty space on the right? Or remove the 'advanced' header because that's sort of communicated with the 'dangerous' label already?

lurch:
Oh, and I keep forgetting to suggest it, but perhaps "Report errors" should be renamed to "Report errors back to resin.io" ? Because otherwise users might not realise what it actually means (I confess when I first saw it, I thought that unticking it meant that any write-errors would be silently ignored).
And this button actually controls both the error-reporting _and_ the statistics-collection, right @stefan-mihaila ?

jviotti:
@Shou Let's make a quick fix to solve this, and then we can focus on the new settings screen.

lurch:
I've just jun the Fedora 25 LiveCD in a VM (VirtualBox), and after installing `libXScrnSaver` ( #835 ) the Etcher v1.0.0 AppImage runs fine, but the (bottom of the) settings-screen looks even worse than it does on my Ubuntu screenshot above :-(  Perhaps this difference is caused by different Linux distros including slightly different fonts or something?

![image](https://cloud.githubusercontent.com/assets/476186/26132417/15997250-3a98-11e7-95b3-5da26b0c94ec.png)

PR #1392, opened 2017-05-09
style(settings): Clarify meaning of the "Report Errors" button

lurch:
To make it clear that this actually controls the reporting of errors back to resin.io, and not the reporting of errors to the user.

See #1383 

If we wanted to be even more explicit, we could even change this to "Report errors and statistics back to resin.io" or "Report errors and anonymous statistics back to resin.io" ?  This would better protect the users' privacy, but also _might_ result in more people unticking the box, leading to less reliable statistics?

### DIFF
--- a/lib/gui/pages/settings/templates/settings.tpl.html
+++ b/lib/gui/pages/settings/templates/settings.tpl.html
@@ -7,7 +7,7 @@
         ng-model="settings.currentData.errorReporting"
         ng-change="settings.toggle('errorReporting')">
 
-      <span>Report errors</span>
+      <span>Anonymously report errors and usage statistics to resin.io</span>
     </label>
   </div>
 


lib/gui/pages/settings/templates/settings.tpl.html
```diff
@@ -7,7 +7,7 @@ <h1 class="title space-bottom-large">Settings</h1>
         ng-model="settings.currentData.errorReporting"
         ng-change="settings.toggle('errorReporting')">
 
-      <span>Report errors</span>
+      <span>Report errors back to resin.io</span>
```
jhermsmeier: How about "Enable sending error reports [to resin.io]"? That'd be a similar wording to what browsers and other programs use, and the "sending" should make it clear that there's stuff going over the wire.

Maybe we could also split it up into error reports & telemetry (just a thought, but that's out of the scope of this PR)?

lurch: > Maybe we could also split it up into error reports & telemetry (just a thought, but that's out of the scope of this PR)?

We _could_, but I find it hard to imagine situations in which a user would enable one but disable the other?

lurch:
@jviotti @konmouz @alexandrosm do you want to give your thoughts on the wording?

konmouz:
Sounds clearer to me as well.

lurch:
I just had a look in the Advanced Settings, and Google Chrome labels this option "Automatically send usage statistics and crash reports to Google".
Should we borrow that language too? (except obviously we'd replace 'crash' with 'error')

lurch:
IMHO this PR should make it into Etcher v1.0.0, but I'd still like to hear from @alexandrosm (or @alisondavis17 ?) before merging this.

alexandrosm:
how about "Anonymously report errors and usage statistics to resin.io"

alisondavis17:
Would we want to add any language around "help us improve Etcher..." i.e. "Help us improve Etcher by anonymously reporting errors and usage statistics to resin.io"

I'm fine with just "Anonymously report errors and usage statistics to resin.io" as well though

lurch:
I think if we make the text too long it looks odd, especially next to all the other options (which have short labels).

PR #1393, opened 2017-05-09
Suggest macOS alternative to Rufus

da2x:
* Specify that Rufus is only an option for Windows.
* Suggest Apple Boot Camp Assistant (built-in) as an alternative for macOS users.

Not sure what to suggest for Linux other than possibly the following online search:
[`Linux "ms-sys" bootable Windows ISO`](https://duckduckgo.com/?q=Linux%20%22ms-sys%22%20bootable%20Windows%20ISO)

### DIFF
--- a/lib/shared/messages.js
+++ b/lib/shared/messages.js
@@ -57,7 +57,7 @@ module.exports = {
       'It looks like you are trying to burn a Windows image.\n\n',
       'Unlike other images, Windows images require special processing to be made bootable.',
       'We suggest you use a tool specially designed for this purpose, such as',
-      '<a href="https://rufus.akeo.ie">Rufus</a>.'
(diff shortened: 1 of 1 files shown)


jviotti:
As far as I've seen, BootCamp is only a solution in a minority of cases (it refuses to create an external USB drive on later versions, handling everything external), but @jhermsmeier can probably tell.

jhermsmeier:
Yeah, Boot Camp Assistant only flashes to USB thumb drives (no USB HDDs, no SD cards, etc afaik), so that makes it a bit of a limited alternative.

da2x:
Still better than suggestion a Windows only alternative.

jviotti:
Cool, yeah, up to @jhermsmeier really.

jhermsmeier:
> Still better than suggestion a Windows only alternative.

Yeah definitely, I think so to – was just confirming the limitations.

jviotti:
Awesome, thanks a lot @da2x !

PR #1400, opened 2017-05-10
upgrade: `etcher-image-write` to v9.1.3

jviotti:
This version gets rid of the `diskpart` optional dependency:

- https://github.com/resin-io-modules/etcher-image-write/pull/101

which will simplify the way we manage our `npm-shrinkwrap.json` file.

See: https://github.com/resin-io/etcher/pull/1379


jviotti:
Amazing catch @lurch ! I updated this PR with the change.

PR #1401, opened 2017-05-10
chore(CLI): minify Etcher CLI code

jviotti:
UglifyJS v3 was recently released, which includes support for
compressing ES6 code.

This commit introduces the following changes:

- Add a boolean `-m` option to `concatenate-javascript.sh`, which when
  set, will minify the final concatenated JavaScript file

- Install `uglify-es` in the CI servers

See: https://news.ycombinator.com/item?id=14285179
See: https://github.com/mishoo/UglifyJS2/tree/harmony

### DIFF
--- a/scripts/build/concatenate-javascript.sh
+++ b/scripts/build/concatenate-javascript.sh
@@ -21,6 +21,7 @@ set -e
 
 BROWSERIFY="./node_modules/.bin/browserify"
 ./scripts/build/check-dependency.sh "$BROWSERIFY"
+./scripts/build/check-dependency.sh uglifyjs
 
 function usage() {
   echo "Usage: $0"
@@ -29,16 +30,19 @@ function usage() {
   echo ""
   echo "    -e <entry point (.js)>"
   echo "    -o <output>"
+  echo "    -m minify"
   exit 1
 }
 
 ARGV_ENTRY_POINT=""
 ARGV_OUTPUT=""
+ARGV_MINIFY=false
 
-while getopts ":e:o:" option; do
+while getopts ":e:o:m" option; do
   case $option in
     e) ARGV_ENTRY_POINT=$OPTARG ;;
     o) ARGV_OUTPUT=$OPTARG ;;
+    m) ARGV_MINIFY=true ;;
     *) usage ;;
   esac
 done
@@ -48,3 +52,8 @@ if [ -z "$ARGV_ENTRY_POINT" ] || [ -z "$ARGV_OUTPUT" ]; then
 fi
 
 "$BROWSERIFY" "$ARGV_ENTRY_POINT" --node --outfile "$ARGV_OUTPUT"
+
+if [ "$ARGV_MINIFY" == "true" ]; then
+  uglifyjs --compress --output "$ARGV_OUTPUT.MIN" -- "$ARGV_OUTPUT"
+  mv "$ARGV_OUTPUT.MIN" "$ARGV_OUTPUT"
+fi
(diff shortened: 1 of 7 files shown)


scripts/build/concatenate-javascript.sh
```diff
@@ -21,6 +21,7 @@ set -e
 
 BROWSERIFY="./node_modules/.bin/browserify"
 ./scripts/build/check-dependency.sh "$BROWSERIFY"
+./scripts/build/check-dependency.sh uglifyjs
```
lurch: Strictly speaking, uglifyjs is only a dependency if `-m` was passed ;)

scripts/build/concatenate-javascript.sh
```diff
@@ -48,3 +52,8 @@ if [ -z "$ARGV_ENTRY_POINT" ] || [ -z "$ARGV_OUTPUT" ]; then
 fi
 
 "$BROWSERIFY" "$ARGV_ENTRY_POINT" --node --outfile "$ARGV_OUTPUT"
+
+if [ "$ARGV_MINIFY" == "true" ]; then
+  uglifyjs --compress --output "$ARGV_OUTPUT.MIN" -- "$ARGV_OUTPUT"
+  mv "$ARGV_OUTPUT.MIN" "$ARGV_OUTPUT"
```
lurch: Are we sure that `$ARGV_OUTPUT.MIN` _definitely_ contains all we need it to, before we clobber the previous `$ARGV_OUTPUT` directory?

P.S. I just did a test using some temp directrories in my /tmp folder on Linux, and `mv dir1 dir2` actually moves dir1 inside dir2? Just checking... ;-)

scripts/build/concatenate-javascript.sh
jviotti: I'll address this in a new PR.

scripts/build/concatenate-javascript.sh
jviotti: Not sure I follow. Do you mean that we should check if `$ARGV_OUTPUT.MIN` before moving it?

lurch: Sorry, I guess that was a bit of a confusing comment.
I've just checked, and I now see that `uglifyjs` actually outputs a file - given that it takes a directory as its input-argument, I'd mistakenly assumed that it also outputted a directory, and that the `$ARGV_OUTPUT.MIN` output-directory was then being renamed over the `$ARGV_OUTPUT` input directory ;-)
Does `uglifyjs --compress --output "$ARGV_OUTPUT/$ARGV_OUTPUT.MIN" -- "$ARGV_OUTPUT"` not work correctly then?

Also, what is it that actually makes use of the `"$ARGV_OUTPUT/$ARGV_OUTPUT.MIN"` file?

scripts/ci/travis-install.sh
```diff
@@ -35,7 +35,7 @@ else
   ./scripts/build/check-dependency.sh make
   ./scripts/build/check-dependency.sh npm
 
-  npm install -g npm@3.10.10
+  npm install -g npm@3.10.10 uglify-es@3.0.3
```
lurch: So, there seems to be a number of packages which we're now installing into each of Docker, Travis and Appveyor. To reduce duplication (and to ensure that all the version-numbers stay in sync), perhaps we should add some kind of meta-file(s) that all three processes read? (And perhaps that/those same meta-file(s) would also make the Etcher development-environment easier to set up?)

lurch: E.g. for `pip` python packages we can use https://pip.readthedocs.io/en/1.1/requirements.html

jviotti: This is great idea. I'll send a PR for it :)

PR #1409, opened 2017-05-11
fix(CLI): workaround absolute paths in Browserify bundles

jviotti:
For some strange reason, Browserify will hardcode absolute paths from
the machine that generated the bundle to be able to resolve `__dirname`
and `__filename` calls. This makes no sense, given that it means that
the Browserify bundle will not work when we move it to another machine,
which went undetected probably for months.

The Browserify community apparently makes modules to fix this particular
issue (like `bundle-collapser`, and `intreq`), however none of this seem
to solve the problem for the Etcher CLI bundle.

I also gave https://github.com/zeit/pkg a go, however I gave up after
not being able to make use of native modules (nothing seems to work; the
packager result will simply not find the addons).

Finally, I ended up making the following workarounds:

- Edit the Browserify bundle file to use its own `__dirname` to
  dynamically resolve the values of `__dirname` and `__filename` of the
  files it contains

- Patch `lzma-native` to statically require its add-on rather than
  relying on dynamic requires from `node-pre-gyp`, which makes it
  impossible to resolve on the final bundle

See: https://github.com/resin-io/etcher/issues/355
See: http://stackoverflow.com/questions/21993073/browserify-with-paths-to-folders-in-my-system

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,7 @@ $(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: package.
 		-x $@ \
 		-t node \
 		-s "$(TARGET_PLATFORM)"
+	git apply --directory $@/node_modules/lzma-native patches/cli/lzma-native-index-static-addon-require.patch
 
 $(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH)-app: \
 	$(BUILD_DIRECTORY)/electron-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies \
@@ -247,7 +248,7 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERS
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH).js: \
 	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH)-app \
 	| $(BUILD_DIRECTORY)
-	./scripts/build/concatenate-javascript.sh -e $</lib/cli/etcher.js -o $@ -m
+	./scripts/build/concatenate-javascript.sh -e lib/cli/etcher.js -b $< -o $@ -m
 
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH): \
 	$(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies \
--- a/patches/cli/lzma-native-index-static-addon-require.patch
+++ b/patches/cli/lzma-native-index-static-addon-require.patch
@@ -1,0 +1,17 @@
+diff --git a/index.js b/index.js
+index 1e7c766..e42c513 100644
+--- a/index.js
++++ b/index.js
+@@ -7,11 +7,8 @@ var extend = require('util-extend');
+ var assert = require('assert');
+ var fs = require('fs');
+
+-// node-pre-gyp magic
+-var nodePreGyp = require('node-pre-gyp');
+ var path = require('path');
+-var binding_path = nodePreGyp.find(path.resolve(path.join(__dirname,'./package.json')));
+-var native = require(binding_path);
++var native = require(path.join(__dirname, 'binding', 'lzma_native.node'));
+
+ extend(exports, native);
+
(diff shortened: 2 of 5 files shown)


lurch:
hardcore -> hardcode ;)

lurch:
Two problems:
 - Travis is failing because you need to add `*.patch` to `.gitattributes`
 - Appveyor spots this issue too, and prints 
```
No wildcards match lzma-native-index-static-addon-require.patch
```
in the log, but it doesn't fail the build! :-(

patches/cli/lzma-native-index-static-addon-require.patch
```diff
@@ -0,0 +1,17 @@
...
+@@ -7,11 +7,8 @@ var extend = require('util-extend');
+ var assert = require('assert');
+ var fs = require('fs');
+ 
+-// node-pre-gyp magic
+-var nodePreGyp = require('node-pre-gyp');
+ var path = require('path');
+-var binding_path = nodePreGyp.find(path.resolve(path.join(__dirname,'./package.json')));
+-var native = require(binding_path);
++var native = require(path.join(__dirname, 'build', 'Release', 'lzma_native.node'));
```
lurch: Does this _ever_ get built in non-Release mode?

scripts/build/concatenate-javascript.sh
```diff
@@ -29,29 +29,56 @@ function usage() {
...
 
-"$BROWSERIFY" "$ARGV_ENTRY_POINT" --node --outfile "$ARGV_OUTPUT"
+"$BROWSERIFY" "$ARGV_BASE_DIRECTORY/$ARGV_ENTRY_POINT" --node --outfile "$ARGV_OUTPUT"
+
+# This hack workarounds the fact the Browserify stores absolute paths
+# of the machine that was used to produce the bundle, giving "not found"
+# module errors when executing it in another computer.
+# The fix is to replace absolute paths with `__dirname`
+node <<EOF > "$ARGV_OUTPUT.TMP"
+const separator = process.platform === 'win32' ? '\\\\\\\\\\\\\\\\' : '\\/';
```
lurch: Whhheeeeee..... ;-)

scripts/build/concatenate-javascript.sh
```diff
@@ -29,29 +29,56 @@ function usage() {
...
 fi
 
-"$BROWSERIFY" "$ARGV_ENTRY_POINT" --node --outfile "$ARGV_OUTPUT"
+"$BROWSERIFY" "$ARGV_BASE_DIRECTORY/$ARGV_ENTRY_POINT" --node --outfile "$ARGV_OUTPUT"
+
+# This hack workarounds the fact the Browserify stores absolute paths
+# of the machine that was used to produce the bundle, giving "not found"
+# module errors when executing it in another computer.
+# The fix is to replace absolute paths with `__dirname`
+node <<EOF > "$ARGV_OUTPUT.TMP"
```
lurch: I think I'd rather have this as a separate script, rather than inlining it as a heredoc

Makefile
```diff
@@ -247,7 +248,7 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERS
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH).js: \
 	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH)-app \
 	| $(BUILD_DIRECTORY)
-	./scripts/build/concatenate-javascript.sh -e $</lib/cli/etcher.js -o $@ -m
+	./scripts/build/concatenate-javascript.sh -e lib/cli/etcher.js -b $< -o $@
```
lurch: Was the `-m` supposed to have been removed here? Or was that just a temporary debugging step?

Makefile
```diff
@@ -197,6 +197,7 @@ $(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies: package.
 		-x $@ \
 		-t node \
 		-s "$(TARGET_PLATFORM)"
+	git apply --directory $@/node_modules/lzma-native patches/cli/lzma-native-index-static-addon-require.patch
```
lurch: Is this effectively the same as `cd`ing to `$@/node_modules/lzma-native` and then applying the patch directly in there?

jviotti:
Fixed!

Makefile
jviotti: Yeah, it does the same thing.

scripts/build/concatenate-javascript.sh
jviotti: Can we address that after the release? I really don't want to go through another round of testing to ensure I got the backslashes right :P

lurch: Lol, fair enough ;-)

jviotti:
Check https://github.com/resin-io/etcher/pull/1379/files. I solved that problem over there already (we need to manually exit on errors on every command). Hopefully it all gets better if we manage to port those Appveyor scripts to bash.

PR #1411, opened 2017-05-11
fix(GUI): handle spaces in installation path when elevating on Windows

jviotti:
Etcher will get stuck at "Starting..." when executing the application on
a directory that contains spaces, like "C:\Program Files (x86)".

The problem is that the command is not quoted correctly when passed to
`cmd.exe /c`. This commit addresses the following specific problems:

- Quote the whole argument to `cmd.exe /c`
- Quote each individual argument after `call`

Change-Type: patch
Changelog-Entry: Fix application stuck at "Starting..." on Windows.
See: https://github.com/resin-io/etcher/pull/1376

### DIFF
--- a/lib/shared/permissions.js
+++ b/lib/shared/permissions.js
@@ -119,6 +119,21 @@ exports.getEnvironmentCommandPrefix = (environment) => {
 };
 
 /**
+ * @summary Quote a string
+ * @function
+ * @private
+ *
+ * @param {String} string - input string
+ * @returns {String} quoted string
+ *
+ * @example
+ * const result = quote('foo');
+ */
+const quoteString = (string) => {
+  return `"${string}"`;
+};
+
+/**
  * @summary Elevate a command
  * @function
  * @public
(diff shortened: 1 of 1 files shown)


lib/shared/permissions.js
```diff
@@ -143,14 +158,18 @@ exports.getEnvironmentCommandPrefix = (environment) => {
...
 
   if (os.platform() === 'win32') {
     const elevator = Bluebird.promisifyAll(nativeModule.load('elevator'));
-    return elevator.elevateAsync(_.concat([
+    return elevator.elevateAsync([
       'cmd.exe',
-      '/c'
-    ], prefixedCommand)).then((results) => {
+      '/c',
+      `"${_.join(prefixedCommand, ' ')}"`
```
lurch: In an earlier version of this PR, I suggested this could be using `quoteString` too? I can't see why not to do that here?

lib/shared/permissions.js
```diff
@@ -143,14 +158,18 @@ exports.getEnvironmentCommandPrefix = (environment) => {
  * });
  */
 exports.elevateCommand = (command, options) => {
-  const prefixedCommand = _.concat(exports.getEnvironmentCommandPrefix(options.environment), command);
+  const prefixedCommand = _.concat(
+    exports.getEnvironmentCommandPrefix(options.environment),
+    _.map(command, quoteString)
```
lurch: I'll have to test, but I think this might cause problems on Linux in the (unlikely) scenario that a path contains a `"` ?

lib/shared/permissions.js

lib/shared/permissions.js
jviotti: Let me check that here.

jviotti: You were right. After some experimentation, the quoting here is only needed for Windows, given that `command-join` takes care of it for OS X later on.

Issue #1385, opened 2017-05-05
SD card not ejected after verification step.


ilovetogetspamed:
_- **Etcher version:** 1.0.0-rc.4
- **Operating system and architecture:** iMac17,1running OS X 10.11.6 (15G1421)  
- **Do you see any meaningful error information on DevTools?** No

Nice, simple to use product.

It does not actually eject the SD card that it just created, as it states at the end of the verification process.

It's not a big deal, just wanted to bring it to your attention.  In my case it actually helps my workflow since I need to add/edit files on the SD card image.  Maybe that should be a checkbox option?

Thanks for your work on Etcher.

lurch:
There _is_ a checkbox option for it, in the settings screen ;-)  (cog icon in the upper-right)
Please try updating to 1.0.0-rc.5 and let us know if it's still not working as expected.

ilovetogetspamed:
Still behaves the same after updating to 1.0.0-rc.5.  Note the 'boot' volume still visible on the desktop above the Etcher window.

I also made sure that 'Auto-unmount on success' was selected.

<img width="808" alt="etcher1" src="https://cloud.githubusercontent.com/assets/3266972/25746671/8d01fc7e-3172-11e7-9fb3-97e6cbc74b88.png">
<img width="264" alt="etcher2" src="https://cloud.githubusercontent.com/assets/3266972/25746672/8d179bd8-3172-11e7-836c-2b677b59847a.png">

jviotti:
That's weird. Maybe there is an edge case on Mountutils that we're not catching? 

@ilovetogetspamed What kind of drive do you have? Can you send us the output of `diskutil info /dev/diskN`?

jviotti:
Looks like I'm not able to reproduce this on a MacBook Pro. This is definitely not an issue on mountutils. We need to unmount the drive before being able to flash to it, so we know that its getting unmounted correctly. 

@ilovetogetspamed Can you link us to the exact image you're trying to flash? Could it be that the operating system is re-mounting the drive after we unmount it for some reason (e.g. the image contains certain partition types)?

ilovetogetspamed:
diskutil info /dev/disk2
   Device Identifier:        disk2
   Device Node:              /dev/disk2
   Whole:                    Yes
   Part of Whole:            disk2
   Device / Media Name:      Built In SDXC Reader

   Volume Name:              Not applicable (no file system)

   Mounted:                  Not applicable (no file system)

   File System:              None

[...]

PR #1414, opened 2017-05-11
fix(writer): Fully eject device after flashing

jhermsmeier:
This updates `mountutils` from 1.0.6 to 1.1.0 (adding ejection functionality),
and ejects the device after flashing & verifying instead of only unmounting it.

**TODO:**
- [ ] Create a user-error if the ejection is unsuccessful

Connects To: #1385

### DIFF
--- a/lib/cli/writer.js
+++ b/lib/cli/writer.js
@@ -26,6 +26,13 @@ const errors = require('../shared/errors');
 const constraints = require('../shared/drive-constraints');
 
 /**
+ * @summary Time to wait, in milliseconds, before ejecting the device
+ * @constant
+ * @type {Number}
+ */
+const EJECT_TIMEOUT_MS = 2000;
+
+/**
  * @summary Write an image to a disk drive
  * @function
  * @public
@@ -37,7 +44,7 @@ const constraints = require('../shared/drive-constraints');
  * @param {String} imagePath - path to image
  * @param {Object} drive - drive
  * @param {Object} options - options
- * @param {Boolean} [options.unmountOnSuccess=false] - unmount on success
+ * @param {Boolean} [options.ejectOnSuccess=false] - eject on success
  * @param {Boolean} [options.validateWriteOnSuccess=false] - validate write on success
  * @param {Function} onProgress - on progress callback (state)
  *
@@ -48,7 +55,7 @@ const constraints = require('../shared/drive-constraints');
  * writer.writeImage('path/to/image.img', {
  *   device: '/dev/disk2'
  * }, {
- *   unmountOnSuccess: true,
+ *   ejectOnSuccess: true,
  *   validateWriteOnSuccess: true
  * }, (state) => {
  *   console.log(state.percentage);
@@ -104,11 +111,16 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
       // right afterwards in some Windows 7 systems.
       return fs.closeAsync(driveFileDescriptor).then(() => {
 
-        if (!options.unmountOnSuccess) {
+        if (!options.ejectOnSuccess) {
           return Bluebird.resolve();
         }
 
-        return mountutils.unmountDiskAsync(drive.device);
+        // Because Mac OS doesn't like immediate unmounting / ejection
+        // after closing the file handle, as it's busy remounting,
+        // we have to wait a moment for it to be able to succeed
+        return Bluebird.delay(EJECT_TIMEOUT_MS)
+          .return(drive.device)
+          .then(mountutils.ejectAsync);
       });
     });
   });
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4877,14 +4877,14 @@
       "resolved": "https://registry.npmjs.org/moment-duration-format/-/moment-duration-format-1.3.0.tgz"
     },
     "mountutils": {
-      "version": "1.0.6",
-      "from": "mountutils@1.0.6",
-      "resolved": "https://registry.npmjs.org/mountutils/-/mountutils-1.0.6.tgz",
+      "version": "1.1.0",
+      "from": "mountutils@latest",
+      "resolved": "https://registry.npmjs.org/mountutils/-/mountutils-1.1.0.tgz",
       "dependencies": {
         "nan": {
-          "version": "2.6.1",
+          "version": "2.6.2",
           "from": "nan@>=2.5.1 <3.0.0",
-          "resolved": "https://registry.npmjs.org/nan/-/nan-2.6.1.tgz"
+          "resolved": "https://registry.npmjs.org/nan/-/nan-2.6.2.tgz"
         }
       }
     },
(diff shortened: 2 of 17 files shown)


npm-shrinkwrap.json
```diff
@@ -4877,14 +4877,14 @@
       "resolved": "https://registry.npmjs.org/moment-duration-format/-/moment-duration-format-1.3.0.tgz"
     },
     "mountutils": {
-      "version": "1.0.6",
-      "from": "mountutils@1.0.6",
-      "resolved": "https://registry.npmjs.org/mountutils/-/mountutils-1.0.6.tgz",
+      "version": "1.1.0",
+      "from": "mountutils@latest",
```
lurch: ?

npm-shrinkwrap.json
```diff
@@ -4877,14 +4877,14 @@
...
+      "resolved": "https://registry.npmjs.org/mountutils/-/mountutils-1.1.0.tgz",
       "dependencies": {
         "nan": {
-          "version": "2.6.1",
+          "version": "2.6.2",
```
lurch: Was this supposed to be upgraded too?

lurch:
Hmmm, if this is now actually ejecting rather than unmounting, should the corresponding label in the settings page be updated too?

npm-shrinkwrap.json
jhermsmeier: That's the install's input - it'll contain whatever you put after `npm install`

npm-shrinkwrap.json
jhermsmeier: Since that version conforms to it's dependency version range (`^`), yes

jhermsmeier:
Yeah, was thinking about that as well, especially since there's also an issue about that floating around...

jhermsmeier:
Had to add a timeout (again), because Mac OS really doesn't want to release devices after just closing the file handle :/ At least it works now.

lib/cli/writer.js
```diff
@@ -107,8 +110,20 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
         if (!options.unmountOnSuccess) {
           return Bluebird.resolve();
         }
-
-        return mountutils.unmountDiskAsync(drive.device);
```
jviotti: This can be simplified to:

```js
return Bluebird.delay(EJECT_TIMEOUT)
    .return(drive.device)
    .then(mountutils.ejectAsync);
```

lib/cli/writer.js
```diff
@@ -107,8 +110,20 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
...
+        return new Bluebird((resolve, reject) => {
+          // Because Mac OS doesn't like immediate unmounting / ejection
```
jviotti: Very minor nit, but can you add a space above the comment for consistency purposes?

lib/cli/writer.js
```diff
@@ -25,6 +25,9 @@ const imageStream = require('../image-stream');
 const errors = require('../shared/errors');
 const constraints = require('../shared/drive-constraints');
 
+/** @const {Number} Time to wait before ejecting the device */
```
jviotti: Can we use the following JSDoc style, just for consistency purposes?

```js
/**
 * @summary Time to wait before ejecting the device
 * @constant
 * @type {Number}
 */
```

lib/cli/writer.js
jhermsmeier: Mended!

lib/cli/writer.js
jhermsmeier: Oh okay, compacted it

lib/cli/writer.js
jhermsmeier: Aye :)

lib/cli/writer.js
```diff
@@ -26,6 +26,13 @@ const errors = require('../shared/errors');
 const constraints = require('../shared/drive-constraints');
 
 /**
+ * @summary Time to wait before ejecting the device
```
lurch: Maybe specify what units this is in? ;)

lib/cli/writer.js
```diff
@@ -108,7 +115,12 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
           return Bluebird.resolve();
         }
 
-        return mountutils.unmountDiskAsync(drive.device);
+        // Because Mac OS doesn't like immediate unmounting / ejection
+        // after closing the file handle, as it's busy remounting,
```
lurch: Did we ever track down _why_ Mac OS is doing this remounting?  (and is it for all images, or only particular images?)
I think I asked somewhere else if you were able to verify that writing an image with `dd` also results in the same auto-remounting behaviour?

npm-shrinkwrap.json
lurch: Ahhh, so does that somehow come from the `node_modules` directory, rather than `package.json`?
If so, that might explain some of the weirdness I was seeing with #1379

lib/cli/writer.js
jviotti: It seems to do it with every image that contains mountable partitions. I can confirm I can reproduce with `dd`

jviotti:
I'll take over this PR tonight.

lib/cli/writer.js
jviotti: Done!

npm-shrinkwrap.json
jviotti: @lurch Looks like if you do `npm install --save dependency@x.y.z`, then the shrinkwrap file includes that `x.y.z` version in the `from` property. If you simply do `npm install --save dependency`, the last version is implicit, and the shrinkwrap file will contains "latest"

lib/cli/writer.js
lurch: Thanks :)

lurch:
A thought that's just occurred to me, along related lines...
If we're changing this from unmounting to ejecting, and we're going to be doing the same on Linux and Windows, I guess it means that if a user flashes Raspbian to their SD card but then wants to edit e.g. `/boot/config.txt` before booting it for the first time, then they'd need to physically remove and then re-insert the SD card from their PC's reader, before being able to do so?

jviotti:
> I guess it means that if a user flashes Raspbian to their SD card but then wants to edit e.g. /boot/config.txt before booting it for the first time, then they'd need to physically remove and then re-insert the SD card from their PC's reader, before being able to do so?

Yeah, that's right. If you want to edit the drive, disabling the eject after success setting is your best bet, given that the drive will even get mounted after the flash finishes.

lurch:
> given that the drive will even get mounted

But only on OSX, right?

jviotti:
Yeah, both `dd` and Etcher behave fine on GNU/Linux. It seems that macOS has some logic to attempt to re-mount drives after a file descriptor on them has been closed.

jhermsmeier:
@jviotti what's the status of this now?

jviotti:
@jhermsmeier I'll continue investigating this during this week. I have the impression that we're covering up a subtle issue on the macOS mountutils unmount implement. I'll see what I find, and we can take it from there (this task is not urgent anymore though)

jhermsmeier:
@jviotti I think this one can be closed, as mountutils has been updated to the version that fixes these issues. Guess we can add the actual ejection later, once it's implemented for all operating systems...

Issue #1388, opened 2017-05-07
ECONNRESET from Corvus on unstable connections


Geabulek:
- **Etcher version:**
1.0.0.rc-5
- **Operating system and architecture:**
Windows 7 64 bits
- **Do you see any meaningful error information on DevTools?**

```
C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 Sun May 07 2017 12:01:39 GMT+0200 (Paris, Madrid (heure d’été)) Application start
C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 Sun May 07 2017 12:01:39 GMT+0200 (Paris, Madrid (heure d’été)) Checking for updates ({"currentVersion":"1.0.0-rc.5","releaseType":"PRODUCTION","updateSemverRange":"<2.0.0"})
C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 {"cause":{"code":"ECONNRESET"},"isOperational":true,"code":"ECONNRESET"}
console.(anonymous function) @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51
logException @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\resin-corvus\src\resin-corvus.js:226
reportException @ C:\Program Files (x86)\Etcher\resources\app.asar\lib\gui\modules\error.js:50
tryCatcher @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\bluebird\js\release\util.js:16
Promise._settlePromiseFromHandler @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\bluebird\js\release\promise.js:504
Promise._settlePromise @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\bluebird\js\release\promise.js:561
Promise._settlePromise0 @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\bluebird\js\release\promise.js:606
Promise._settlePromises @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\bluebird\js\release\promise.js:681
Async._drainQueue @ C:\Program Files (x86)\Etcher\resources\app.asar\node_modules\bluebird\js\release\async.js:138
[...]

PR #1415, opened 2017-05-12
fix(GUI): ignore ECONNRESET and ECONNREFUSED when querying S3

jviotti:
Querying S3 to determine the latest available versions might throw
`ECONNRESET` and `ECONNREFUSED`. This commit extends the
`s3Packages.getRemoteVersions()` function to handle these errors and
return no available version if so, like we already do with other similar
HTTP errors.

Change-Type: patch
Changelog-Entry: Fix `ECONNRESET` and `ECONNREFUSED` errors when checking for updates on unstable connections.
Fixes: https://github.com/resin-io/etcher/issues/1396
Fixes: https://github.com/resin-io/etcher/issues/1388

### DIFF
--- a/lib/shared/s3-packages.js
+++ b/lib/shared/s3-packages.js
@@ -149,6 +149,10 @@ exports.getRemoteVersions = _.memoize((bucketUrl) => {
     .catch({
       code: 'ENOTFOUND'
     }, {
+      code: 'ECONNRESET'
+    }, {
+      code: 'ECONNREFUSED'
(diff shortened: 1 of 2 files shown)


lurch:
So it looks like this took quite a while to track down...? Anything we could change to make similar problems easier to track down in future?

jviotti:
Better stack traces from corvus. That module needs to be vastly improved. We'll hopefully have time to invest into it after this release.

PR #1419, opened 2017-05-12
doc(CLI-INSTALLATION): Make headings h3

jhermsmeier:
Due to display on the website, the headings need to be level 3, rather than level 2.

### DIFF
--- a/docs/CLI-INSTALLATION.md
+++ b/docs/CLI-INSTALLATION.md
@@ -1,5 +1,4 @@
-macOS and GNU/Linux
--------------------
+### macOS and GNU/Linux
 
 - Extract the `.tar.gz` package by running:
 
@@ -16,8 +15,7 @@ tar fvx path/to/cli.tar.gz
 export PATH="$PATH:/opt/etcher-cli"
 ```
 
-Windows
--------
+### Windows
 
 - Unzip the `.zip` package by right-clicking on it and selecting "Extract All"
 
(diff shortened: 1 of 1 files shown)


jviotti:
This is an independent file, so it should have top level titles. If the website needs a different thing, then the website itself should do the transformation. @craig-mulligan ?

craig-mulligan:
@jviotti yea we can of course do that, but I didn't have time because I was given the doc on short notice. I've created an issue - https://github.com/resin-io/etcher-homepage/issues/47. 

I'll ping you when it's implemented so you can update here.

lurch:
I guess this also means that `CLI-INSTALLATION.md` should have its own "proper" H1 title too, which the website should then strip?

Hmmm, or perhaps `CLI.md` could be reverted back to how it was, and then the website-building-code could pull out _only_ the "Installation" and "Running" sections from the markdown?

craig-mulligan:
@lurch, We can do that too, but it does make things slightly less robust. We'd have to ensure that the document isn't edited without taking into account it's affect on the website. For instance if we split the page by regex matching a heading then we couldn't change that heading, without checking that it still matches. 

What about laying it like this? 
```
docs/
  cli/
     intro.md
     installation.md
  anotherDocsPage.md
  ...
```

I suppose this is another case to include the website in the main repo.

lurch:
I'd be :-1: on getting rid of the existing `docs/CLI.md` in the Etcher repo, for fear of breaking any existing links to it.

I'll create a PR this afternoon to make `CLI-INSTALLATION.md` "fit better" with the existing Etcher docs, and ping you on it @craig-mulligan before merging so that you can tweak your website-build-script to suit :-)

jviotti:
Remember that there are various modules out there to parse Markdown, so parsing the document, changing the title levels, and regenerate it should be an easy task.

PR #1422, opened 2017-05-12
feat(GUI): reset webview after navigating away

Shou:
We reload and reset the webview to its original URL when the user
navigates away from the success screen.

Changelog-Entry: Reset webview after navigating away from success
screen.

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -59,7 +59,7 @@ const app = angular.module('Etcher', [
   // Components
   require('./components/svg-icon/svg-icon'),
   require('./components/warning-modal/warning-modal'),
-  require('./components/safe-webview/safe-webview'),
+  require('./components/safe-webview'),
 
   // Pages
   require('./pages/main/main'),
@@ -300,12 +300,13 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
 
 });
 
-app.controller('StateController', function($rootScope) {
-
-  $rootScope.$on('$stateChangeSuccess', (event, toState, toParams, fromState) => {
+app.controller('StateController', function($rootScope, $scope) {
+  const unregisterStateChange = $rootScope.$on('$stateChangeSuccess', (event, toState, toParams, fromState) => {
     this.previousName = fromState.name;
     this.currentName = toState.name;
   });
+
+  $scope.$on('$destroy', unregisterStateChange);
 
   /**
    * @summary Get the previous state name
@@ -321,6 +322,18 @@ app.controller('StateController', function($rootScope) {
    */
   this.previousName = null;
 
+  /**
+   * @summary Get the current state name
+   * @function
+   * @public
+   *
+   * @returns {String} current state name
+   *
+   * @example
+   * if (StateController.currentName === 'main') {
+   *   console.log('We are on the main screen!');
+   * }
+   */
   this.currentName = null;
 
 });
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -300,11 +300,18 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
 
 });
 
-app.controller('StateController', function($state) {
+app.controller('StateController', function($rootScope, $state) {
 
   /**
    * @param {string} state - state page
    */
   this.is = $state.is;
+
+  /**
+   * @param {function} func - function to pass $rootScope to
+   */
+  this.onStateChange = (func) => {
+    $rootScope.$on('$stateChangeSuccess', func);
+  };
 
 });
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -303,15 +303,19 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
 app.controller('StateController', function($rootScope, $state) {
 
   /**
-   * @param {string} state - state page
+   * @summary Get the current state name
+   * @function
+   * @public
+   *
+   * @returns {String} current state name
+   *
+   * @example
+   * if (StateController.currentName() === 'main') {
+   *   console.log('We are on the main screen!');
+   * }
    */
-  this.is = $state.is;
-
-  /**
-   * @param {function} func - function to pass $rootScope to
-   */
-  this.onStateChange = (func) => {
-    $rootScope.$on('$stateChangeSuccess', func);
+  this.currentName = () => {
+    return $state.current.name;
   };
 
 });
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -300,22 +300,27 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
 
 });
 
-app.controller('StateController', function($rootScope, $state) {
+app.controller('StateController', function($rootScope) {
+
+  $rootScope.$on('$stateChangeSuccess', (event, toState, toParams, fromState) => {
+    this.previousName = fromState.name;
+    this.currentName = toState.name;
+  });
 
   /**
-   * @summary Get the current state name
+   * @summary Get the previous state name
    * @function
    * @public
    *
-   * @returns {String} current state name
+   * @returns {String} previous state name
    *
    * @example
-   * if (StateController.currentName() === 'main') {
-   *   console.log('We are on the main screen!');
+   * if (StateController.previousName === 'main') {
+   *   console.log('We left the main screen!');
    * }
    */
-  this.currentName = () => {
-    return $state.current.name;
-  };
+  this.previousName = null;
+
+  this.currentName = null;
 
 });
--- a/lib/gui/components/safe-webview.js
+++ b/lib/gui/components/safe-webview.js
@@ -22,8 +22,8 @@ const angular = require('angular');
 const react = require('react');
 const propTypes = require('prop-types');
 const react2angular = require('react2angular').react2angular;
-const analytics = require('../../modules/analytics');
-const packageJSON = require('../../../../package.json');
+const analytics = require('../modules/analytics');
+const packageJSON = require('../../../package.json');
 
 const MODULE_NAME = 'Etcher.Components.SafeWebview';
 const angularSafeWebview = angular.module(MODULE_NAME, []);
@@ -35,6 +35,14 @@ const angularSafeWebview = angular.module(MODULE_NAME, []);
  * @type {String}
  */
 const VERSION_PARAM = 'etcher-version';
+
+/**
+ * @summary Electron session identifier
+ * @constant
+ * @private
+ * @type {String}
+ */
+const ELECTRON_SESSION = 'persist:success-banner';
 
 /**
  * @summary Webviews that hide/show depending on the HTTP status returned
@@ -56,14 +64,28 @@ class SafeWebview extends react.PureComponent {
       shouldShow: true
     };
 
+    const url = new URL(props.src);
+
+    // We set the version GET parameter here.
+    url.searchParams.set(VERSION_PARAM, packageJSON.version);
+
+    this.entryHref = url.href;
+
     // Events steal 'this'
     this.didFailLoad = _.bind(this.didFailLoad, this);
     this.didGetResponseDetails = _.bind(this.didGetResponseDetails, this);
 
-    // Make a persistent electron session called 'success-banner'
-    electron.remote.session.fromPartition('persist:success-banner', {
+    this.eventTuples = [
+      [ 'did-fail-load', this.didFailLoad ],
+      [ 'did-get-response-details', this.didGetResponseDetails ],
+      [ 'new-window', this.constructor.newWindow ],
+      [ 'console-message', this.constructor.consoleMessage ]
+    ];
 
-      // Disable the cache for the 'success-banner' session used by the webview
+    // Make a persistent electron session for the webview
+    electron.remote.session.fromPartition(ELECTRON_SESSION, {
+
+      // Disable the cache for the session
       cache: false
     });
   }
@@ -88,17 +110,27 @@ class SafeWebview extends react.PureComponent {
   componentDidMount() {
 
     // Events React is unaware of have to be handled manually
-    this.refs.webview.addEventListener('did-fail-load', this.didFailLoad);
-    this.refs.webview.addEventListener('did-get-response-details', this.didGetResponseDetails);
-    this.refs.webview.addEventListener('new-window', this.constructor.newWindow);
-    this.refs.webview.addEventListener('console-message', this.constructor.consoleMessage);
+    _.map(this.eventTuples, (tuple) => {
+      this.refs.webview.addEventListener(...tuple);
+    });
 
     // Use the 'success-banner' session
-    this.refs.webview.partition = 'persist:success-banner';
+    this.refs.webview.partition = ELECTRON_SESSION;
 
     // It's important that this comes after the partition setting, otherwise it will
     // use another session and we can't change it without destroying the element again
-    this.refs.webview.src = this.makeURL();
+    this.refs.webview.src = this.entryHref;
+  }
+
+  /**
+   * @summary Remove the Webview events
+   */
+  componentWillUnmount() {
+
+    // Events that React is unaware of have to be handled manually
+    _.map(this.eventTuples, (tuple) => {
+      this.refs.webview.removeEventListener(...tuple);
+    });
   }
 
   /**
@@ -111,29 +143,17 @@ class SafeWebview extends react.PureComponent {
       // Reload the page if it hasn't changed, otherwise reset the source URL,
       // because reload interferes with 'src' setting, resetting the 'src' attribute
       // to what it was was just prior.
-      if (this.refs.webview.src === this.makeURL()) {
+      if (this.refs.webview.src === this.entryHref) {
         this.refs.webview.reload();
 
       } else {
-        this.refs.webview.src = this.makeURL();
+        this.refs.webview.src = this.entryHref;
       }
 
       this.setState({
         shouldShow: true
       });
     }
-  }
-
-  /**
-   * @summary Remove the Webview events
-   */
-  componentWillUnmount() {
-
-    // Events that React is unaware of have to be handled manually
-    this.refs.webview.removeEventListener('did-fail-load', this.didFailLoad);
-    this.refs.webview.removeEventListener('did-get-response-details', this.didGetResponseDetails);
-    this.refs.webview.removeEventListener('new-window', this.constructor.newWindow);
-    this.refs.webview.removeEventListener('console-message', this.constructor.consoleMessage);
   }
 
   /**
@@ -186,22 +206,6 @@ class SafeWebview extends react.PureComponent {
     } else if (event.level === ERROR_LEVEL) {
       analytics.logException(event.message);
     }
-  }
-
-  /**
-   * @summary Parse the 'src' prop into an URL and append metadata as GET parameters
-   * @function
-   * @public
-   *
-   * @returns {URL} url object
-   */
-  makeURL() {
-    const url = new URL(this.props.src);
-
-    // We set the version GET parameter here.
-    url.searchParams.set(VERSION_PARAM, packageJSON.version);
-
-    return url.href;
   }
 
 }
--- a/lib/gui/components/safe-webview.js
+++ b/lib/gui/components/safe-webview.js
@@ -24,6 +24,7 @@ const propTypes = require('prop-types');
 const react2angular = require('react2angular').react2angular;
 const analytics = require('../modules/analytics');
 const packageJSON = require('../../../package.json');
+const robot = require('../../shared/robot');
 
 const MODULE_NAME = 'Etcher.Components.SafeWebview';
 const angularSafeWebview = angular.module(MODULE_NAME, []);
@@ -85,7 +86,7 @@ class SafeWebview extends react.PureComponent {
     // Make a persistent electron session for the webview
     electron.remote.session.fromPartition(ELECTRON_SESSION, {
 
-      // Disable the cache for the session
+      // Disable the cache for the session such that new content shows up when refreshing
       cache: false
     });
   }
@@ -198,26 +199,29 @@ class SafeWebview extends react.PureComponent {
    * @param {Event} event - event object
    *
    * @example
+   *
+   * // In the webview
    * console.log({
-   *   type: 0,
-   *   message: 'Hello, World!'
+   *   command: 'error',
+   *   data: 'Good night!'
    * });
+   *
+   * // or with robot
+   * robot.printError('Good night!');
    */
   static consoleMessage(event) {
-    const LOG_LEVEL = 0;
-    const WARN_LEVEL = 1;
-    const ERROR_LEVEL = 2;
+    const {
+      command,
+      data
+    } = _.attempt(robot.parseMessage, event.message);
 
-    _.attempt(() => {
-      const data = JSON.parse(event.message);
+    if (command === robot.COMMAND.LOG) {
+      analytics.logEvent(data);
+    }
 
-      if (data.type === LOG_LEVEL || data.type === WARN_LEVEL) {
-        analytics.logEvent(data.message);
-
-      } else if (data.type === ERROR_LEVEL) {
-        analytics.logException(data.message);
-      }
-    });
+    if (command === robot.COMMAND.ERROR) {
+      analytics.logException(data);
+    }
   }
 
 }
--- a/lib/gui/components/safe-webview.js
+++ b/lib/gui/components/safe-webview.js
@@ -195,32 +195,34 @@ class SafeWebview extends react.PureComponent {
   }
 
   /**
-   * @summary Forward console messages from the webview to analytics module
+   * @summary Forward specially-formatted console messages from the webview
    * @param {Event} event - event object
    *
    * @example
    *
    * // In the webview
-   * console.log({
+   * console.log(JSON.stringify({
    *   command: 'error',
    *   data: 'Good night!'
-   * });
+   * }));
    *
-   * // or with robot
-   * robot.printError('Good night!');
+   * console.log(JSON.stringify({
+   *   command: 'log',
+   *   data: 'Hello, Mars!'
+   * }));
    */
   static consoleMessage(event) {
-    const {
-      command,
-      data
-    } = _.attempt(robot.parseMessage, event.message);
-
-    if (command === robot.COMMAND.LOG) {
-      analytics.logEvent(data);
+    if (!robot.isMessage(event.message)) {
+      return;
     }
 
-    if (command === robot.COMMAND.ERROR) {
-      analytics.logException(data);
+    const message = robot.parseMessage(event.message);
+
+    if (robot.getCommand(message) === robot.COMMAND.LOG) {
+      analytics.logEvent(robot.getData(message));
+
+    } else if (robot.getCommand(message) === robot.COMMAND.ERROR) {
+      analytics.logException(robot.getData(message));
     }
   }
 
--- a/lib/gui/components/safe-webview.js
+++ b/lib/gui/components/safe-webview.js
@@ -196,16 +196,28 @@ class SafeWebview extends react.PureComponent {
   /**
    * @summary Forward console messages from the webview to analytics module
    * @param {Event} event - event object
+   *
+   * @example
+   * console.log({
+   *   type: 0,
+   *   message: 'Hello, World!'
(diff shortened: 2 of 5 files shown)


lib/gui/index.html
```diff
@@ -57,7 +57,8 @@
         isFinish: state.is('success')
       }">
       <safe-webview
-        src="'https://etcher.io/success-banner/'"></safe-webview>
+        src="'https://etcher.io/'"
```
lurch: Huh?

lib/gui/index.html
```diff
@@ -57,7 +57,8 @@
...
+        on-state-change="state.onStateChange"></safe-webview>
```
lurch: Out of curiosity, what triggers this `on-state-change` to happen?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -84,6 +84,14 @@ class SafeWebview extends react.PureComponent {
       this.refs.webview.addEventListener('new-window', this.constructor.newWindow);
       this.refs.webview.addEventListener('console-message', this.constructor.consoleMessage);
 
+      if (this.props.onStateChange) {
+        this.props.onStateChange((event, toState, toParams, fromState) => {
+          if (fromState.name === 'success' && this.state.shouldLoad) {
+            this.refs.webview.src = this.props.src;
```
lurch: I guess it could also check that `this.refs.webview.src != this.props.src`, to avoid an unnecessary `reload()` ?

Shou: My thought here was in case the page has been updated, I assume some people leave Etcher open constantly to burn many cards, so this would make them see updated pages.

lib/gui/index.html
Shou: It happens whenever the `$state` changes, e.g. user navigating to the `'settings'`, `'success'`, or back to the `'main'` page.

lib/gui/index.html
Shou: My mistake, that's for testing.

lib/gui/index.html
lurch: Does that mean that changing from the 'flashing' page to the 'finished' page will also trigger a state-change, and thus a user-visible delay while the webview reloads?
Perhaps it only makes sense to reload the webview when _leaving_ the 'finished' page? (because IMHO merely going back and forth to the settings page shouldn't cause a reload)

lib/gui/components/safe-webview/safe-webview.js
lurch: @alexandrosm Can we tell from the analytics how many people leave Etcher open for an extended period of time? (my gut instinct would be not many people at all)  How important is it that users always see the 'latest' banner, rather than the banner that they downloaded when they started Etcher?

lib/gui/index.html
Shou: Oh sorry, that's what it does. So `onStateChange` is a higher-order function that passes the state changes to its argument function, and in there we check if the state change is from `'success'` before reloading, visible in the comment code block below.

lib/gui/app.js
```diff
@@ -298,11 +298,18 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
...
-app.controller('StateController', function($state) {
+app.controller('StateController', function($rootScope, $state) {
 
   /**
    * @param {string} state - state page
    */
   this.is = $state.is;
 
+  /**
+   * @param {function} func - function to pass $rootScope to
```
jviotti: This is missing the `@summary`, `@public`, `@example`, etc. Also `function` should be `Function`

lib/gui/app.js
```diff
@@ -298,11 +298,18 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
 
 });
 
-app.controller('StateController', function($state) {
+app.controller('StateController', function($rootScope, $state) {
 
   /**
    * @param {string} state - state page
```
jviotti: Looks like I failed to catch it last time, but this should also include all the JSDoc tags mentioned below.

lib/gui/index.html
jviotti: It looks weird that the `safe-webview` is exposing this attribute. Can this all be handled internally?

lib/gui/app.js
```diff
@@ -298,11 +298,18 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
...
    */
   this.is = $state.is;
 
+  /**
+   * @param {function} func - function to pass $rootScope to
+   */
+  this.onStateChange = (func) => {
```
jviotti: If this function is called multiple times, then multiple event listeners will be registered for this event.

lib/gui/app.js
lurch: > Also `function` should be `Function`

Does / should eslint pick that up?

lib/gui/index.html
lurch: Ahhhh, thanks for the explanation. So I guess it's more like `onLeaveSuccessPage` than `onStateChange`? ;-)

Shou: @jviotti It's the 'cleanest' way I've found so far, given the React/Angular mixing situation we're in. Basically, the React element isn't re-rendered when the attributes change, so we can't just pass an Angular variable and cause a re-render whenever that changes, and I suspect we won't want to do that anyway because if attribute changes caused a re-render it would refresh the webview. I'll look into if there's a better way though!

@lurch `onStateChange` is what it says on the tin, but way it's used is more like `onLeaveSuccessPage` because the first-class function we pass to `onStateChange` checks if the state change is from `'success`', so we could call the function `onStateChange` takes `onLeaveSuccessPage`.

Shou:
What do you think of the way this is handled now?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -54,63 +57,64 @@ class SafeWebview extends react.PureComponent {
...
 
   /**
-   * @summary Add the Webview events if there is an element
+   * @summary Add the Webview events
    */
   componentDidMount() {
 
-    // There is no element to add events to if 'shouldLoad' is false.
-    if (this.state.shouldLoad) {
+    // Events React is unaware of have to be handled manually
```
lurch: minor nit, but could you change this to "Events that React is ...", otherwise "Events React" might sound like another framework ;)

jviotti:
Sorry for the huge delay on this. I'll do a review tomorrow :)

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -157,14 +161,35 @@ class SafeWebview extends react.PureComponent {
...
+   *
+   * @returns {URL} url object
+   */
+  makeURL() {
+    const url = new URL(this.props.src);
+
+    // We set the 'etcher-version' GET parameter here.
+    url.searchParams.set('etcher-version', packageJSON.version);
+
+    return url;
```
jviotti: Can we do `url.href` here, given that its the only property we use?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -157,14 +161,35 @@ class SafeWebview extends react.PureComponent {
...
+   * @function
+   * @public
+   *
+   * @returns {URL} url object
+   */
+  makeURL() {
+    const url = new URL(this.props.src);
+
+    // We set the 'etcher-version' GET parameter here.
+    url.searchParams.set('etcher-version', packageJSON.version);
```
jviotti: Can we extract `etcher-version` to a constant at the top of this component?

jviotti:
Minor comment (and doesn't need to be addressed here), but since the safe webview element is just a single file, can we move it to `lib/gui/components/safe-webview.js`?

jviotti:
I don't think this is working. I'm doing the following:

- Check the URL of the `webview` to https://etcher.io
- Flash an image
- In the webview, click a link on https://etcher.io (e.g. the changelog)
- Wait for it to load
- Click "Flash Another"
- Flash the image again

I now see the webview on the changelog page, so it hasn't been reset.

Shou:
@jviotti I looked into it and it has to do with the reloading being too fast and rolling back the URL change. However, I went further, and it seems like the caching is a bit aggressive here and the reloading has no effect. I believe we can disable the caching as you can see with the closing PR [here](https://github.com/electron/electron/issues/1720).

jviotti:
I don't think I get it. Where do we pass `--disable-http-cache switch`? Also, can we disable caching just for the webview, but keep it enabled for the actual main window?

Shou:
@jviotty Here's a working implementation. The webview now uses a specific session where the cache is disabled, so yes that's possible and done. Unfortunately we can't set the `'partition'` property directly in the React attributes because it's an unknown attribute to React, and it strips unknown attributes. [There's a big issue about it.](https://github.com/facebook/react/issues/140) This is why the `src` setting was moved to `componentDidMount`.

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -28,10 +28,15 @@ const packageJSON = require('../../../../package.json');
 const MODULE_NAME = 'Etcher.Components.SafeWebview';
 const angularSafeWebview = angular.module(MODULE_NAME, []);
 
+const VERSION_PARAM = 'etcher-version';
```
jviotti: Can we document this constant?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -54,63 +59,76 @@ class SafeWebview extends react.PureComponent {
...
-    if (this.state.shouldLoad) {
+      // Reset URL to initial
+      this.refs.webview.src = this.makeURL();
 
-      // Events React is unaware of have to be handled manually
-      this.refs.webview.removeEventListener('did-fail-load', this.didFailLoad);
-      this.refs.webview.removeEventListener('did-get-response-details', this.didGetResponseDetails);
-      this.refs.webview.removeEventListener('new-window', this.constructor.newWindow);
-      this.refs.webview.removeEventListener('console-message', this.constructor.consoleMessage);
+      const DECISECOND = 100;
```
jviotti: Can we describe what this variable is for instead? e.g. `WEBVIEW_RELOAD_TIMEOUT_MS` or something like that?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -157,14 +175,35 @@ class SafeWebview extends react.PureComponent {
...
+   * @summary Parse the 'src' prop into an URL and append metadata as GET parameters
+   * @function
+   * @public
+   *
+   * @returns {URL} url object
+   */
+  makeURL() {
+    const url = new URL(this.props.src);
+
+    // We set the 'etcher-version' GET parameter here.
```
jviotti: Can we update this comment to not hard-code the string `etcher-version`?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -54,63 +59,76 @@ class SafeWebview extends react.PureComponent {
...
+    this.refs.webview.addEventListener('new-window', this.constructor.newWindow);
+    this.refs.webview.addEventListener('console-message', this.constructor.consoleMessage);
 
-      // Events React is unaware of have to be handled manually
-      this.refs.webview.addEventListener('did-fail-load', this.didFailLoad);
-      this.refs.webview.addEventListener('did-get-response-details', this.didGetResponseDetails);
-      this.refs.webview.addEventListener('new-window', this.constructor.newWindow);
-      this.refs.webview.addEventListener('console-message', this.constructor.consoleMessage);
+    // Use the 'success-banner' session
+    this.refs.webview.partition = 'persist:success-banner';
```
jviotti: Looks like we're duplicating this same string here and in the session definition, which means it might get out of sync. Can we declare it only in one place (e.g. with a constant)?

jviotti: Actually, should we define the electron session in this file instead?

jviotti:
I'm getting this when clicking "Flash Another", after navigating to another page in the webview:

screenshot 2017-05-24 09 29 30

I still keep seeing the new page after a retry though. The webview is not getting reset.

lib/gui/components/safe-webview/safe-webview.js
Shou: `electron.session` is only available in the main process so I don't think so unfortunately.

Shou: Also what's a good location to share this constant? `lib/shared/constants.js` maybe?

Shou:
Strange, I'm not getting any errors, but I can confirm that the resetting bug was back though. This commit should fix it.

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -95,15 +101,14 @@ class SafeWebview extends react.PureComponent {
   componentWillReceiveProps(nextProps) {
     if (this.props.currentStateName === 'success' && nextProps.currentStateName !== 'success') {
 
-      // Reset URL to initial
-      this.refs.webview.src = this.makeURL();
+      // Only reset the URL to initial if it has changed, otherwise reload the page,
```
lurch: Hahaha, isn't that what I originally suggested? ;)

Shou: Yeah, if only I had listened lol

lib/gui/components/safe-webview/safe-webview.js
jviotti: You can do `electron.remote.session` from the renderer process.

lurch:
Unit-tests are failing with:
```
> etcher@1.0.0 jslint /etcher
> eslint lib tests scripts bin versionist.conf.js
/etcher/lib/gui/components/safe-webview/safe-webview.js
  106:7  error  Unexpected negated condition  no-negated-condition
```

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -54,63 +65,75 @@ class SafeWebview extends react.PureComponent {
...
   }
 
   /**
-   * @summary Remove the Webview events if there is an element
+   * @summary Refresh the webview if we are navigating away from the success page
+   * @param {Object} nextProps - upcoming properties
    */
-  componentWillUnmount() {
+  componentWillReceiveProps(nextProps) {
+    if (this.props.currentStateName === 'success' && nextProps.currentStateName !== 'success') {
```
jviotti: This component is hard-coding a state name that is set outside of the component (in the Angular.js app), which doesn't look right. It looks like we should make the component take a boolean attribute to determine when it should be refreshed, and move the fact that in our use case that depends on the `success` state outside of this component.

lib/gui/etcher.js
```diff
@@ -39,6 +39,13 @@ electron.app.on('ready', () => {
   // No menu bar
   electron.Menu.setApplicationMenu(null);
 
+  // Make a persistent electron session called 'success-banner'
+  electron.session.fromPartition('persist:success-banner', {
```
jviotti: I thing I said this before, but can we move this to the safe webview component by using `electron.remote.session`?

Shou: Yeah I forgot about that, we should be able to use that.

lib/gui/components/safe-webview/safe-webview.js
Shou: Yeah true, it should be as agnostic as possible to the rest of the program.

Shou:
Updated

lib/gui/components/safe-webview/safe-webview.js
jviotti: Let's move `'persist:success-banner'` to a constant, so we don't type it twice here, and when declaring the session.

lib/gui/index.html
```diff
@@ -54,10 +54,11 @@
     <div class="section-loader"
       ng-controller="StateController as state"
       ng-class="{
-        isFinish: state.is('success')
+        isFinish: state.currentName === 'success'
       }">
       <safe-webview
-        src="'https://etcher.io/success-banner/'"></safe-webview>
+        src="'https://etcher.io/success-banner/'"
+        refresh-now="state.previousName === 'success'"></safe-webview>
```
jviotti: I'm going to test this later, but what happens if this always evaluates to `true`? Will the webview keep refreshing?

lib/gui/app.js
```diff
@@ -298,11 +298,27 @@ app.controller('HeaderController', function(SelectionStateModel, OSOpenExternalS
 
 });
 
-app.controller('StateController', function($state) {
+app.controller('StateController', function($rootScope) {
+
+  $rootScope.$on('$stateChangeSuccess', (event, toState, toParams, fromState) => {
```
jviotti: This is probably pedantic, but a standard practice is to register for the `$destroy` event of the controller and un-register any registered event.

jviotti:
The webview now gets successfully reset, and the uncaught error I was getting is also gone. Great work @Shou ! The only thing is that timeout errors are still printed inline in a scary way:

screenshot 2017-05-30 11 03 50

Can we catch these error and send a nice "Success banner timeout error" Mixpanel event instead?

lib/gui/index.html
Shou: I made sure to check if the previous prop was `false` and the current one `true`, so that a refresh is only triggered when the change from `false` → `true` happens and not `true` → `true`. React will sometimes trigger `componentWillReceiveProps()` when it's not just a prop update, so that's also why I took this precaution.

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -42,75 +53,95 @@ class SafeWebview extends react.PureComponent {
...
+      if (this.refs.webview.src === this.makeURL()) {
+        this.refs.webview.reload();
 
-      // Events React is unaware of have to be handled manually
-      this.refs.webview.removeEventListener('did-fail-load', this.didFailLoad);
-      this.refs.webview.removeEventListener('did-get-response-details', this.didGetResponseDetails);
-      this.refs.webview.removeEventListener('new-window', this.constructor.newWindow);
-      this.refs.webview.removeEventListener('console-message', this.constructor.consoleMessage);
+      } else {
+        this.refs.webview.src = this.makeURL();
```
lurch: Should we assign the result of `this.makeURL()` to a const variable, to avoid calling the function multiple times?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -42,75 +53,95 @@ class SafeWebview extends react.PureComponent {
...
+  /**
+   * @summary Remove the Webview events
+   */
+  componentWillUnmount() {
 
+    // Events that React is unaware of have to be handled manually
+    this.refs.webview.removeEventListener('did-fail-load', this.didFailLoad);
+    this.refs.webview.removeEventListener('did-get-response-details', this.didGetResponseDetails);
+    this.refs.webview.removeEventListener('new-window', this.constructor.newWindow);
+    this.refs.webview.removeEventListener('console-message', this.constructor.consoleMessage);
```
lurch: Any (easy) way to avoid the "repetition" between the addEventListener and removeEventListener calls?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -157,14 +188,35 @@ class SafeWebview extends react.PureComponent {
     }
   }
 
+  /**
+   * @summary Parse the 'src' prop into an URL and append metadata as GET parameters
+   * @function
+   * @public
+   *
+   * @returns {URL} url object
```
lurch: Is this `@returns` docstring now incorrect?

lib/gui/components/safe-webview/safe-webview.js
```diff
@@ -42,75 +53,95 @@ class SafeWebview extends react.PureComponent {
...
 
   /**
-   * @summary Add the Webview events if there is an element
+   * @summary Add the Webview events
    */
   componentDidMount() {
 
-    // There is no element to add events to if 'shouldLoad' is false.
-    if (this.state.shouldLoad) {
+    // Events React is unaware of have to be handled manually
```
lurch: Could you add the `that` in here too? ;-)

lurch:
I think this PR also needs to be rebased on master to prevent the Ubuntu 12.04 Linux x86 builds failing issue...

Shou:
I've followed the advice above now. @jviotti I'm still a bit bewildered by that error since I've never gotten it myself, but I suspect it's an error that can happen on the Etcher homepage, which is forwarded over `console.error` and caught by Corvus inside Etcher, which means everything's working as intended, as banner pages are only supposed to use the console for Etcher-related debugging and errors intended to reach Corvus.

lurch:
Yeah, I also wonder if the error might be coming from the Etcher homepage itself, as it's very javascript-heavy. @Shou is it possible to redirect any JS errors created by the pages displayed inside the webview into a blackhole, so that they don't end up spamming the DevTools console?

jviotti:
I agree with @lurch. Let's prevent any error from the banner from spamming DevTools, however please catch this particular timeout error from the Etcher side (from the `webview`) and transform it into a "Success banner timeout" Mixpanel event, since I'd like to know how often that happens.

Shou:
So I've changed it from accepting all console messages to only specific JSON objects containing `type` and `message` now, what do you think?

lib/gui/components/safe-webview.js
```diff
@@ -196,16 +196,28 @@ class SafeWebview extends react.PureComponent {
   /**
    * @summary Forward console messages from the webview to analytics module
    * @param {Event} event - event object
+   *
+   * @example
+   * console.log({
```
lurch: Should this example be using `consoleMessage` rather than `console.log` ?

lib/gui/components/safe-webview.js
```diff
@@ -196,16 +196,28 @@ class SafeWebview extends react.PureComponent {
...
-    } else if (event.level === ERROR_LEVEL) {
-      analytics.logException(event.message);
-    }
+      if (data.type === LOG_LEVEL || data.type === WARN_LEVEL) {
+        analytics.logEvent(data.message);
+
+      } else if (data.type === ERROR_LEVEL) {
+        analytics.logException(data.message);
+      }
+    });
```
lurch: So, just to check I'm understanding this diff correctly:
`consoleMessage` takes an `event` object, which is expected to have a (possibly JSON-encoded) `.message` attribute? (and also previously a `.level` attribute, which is now ignored)
And then if the JSON message is decoded correctly (i.e. into `data`), it's then expected to have both `.type` and `.message` attributes?

Which I guess means the expected usage is something like:
```
consoleMessage({level:0, message: "{type: 0, message: 'Hello, World!'}"});
```

lib/gui/components/safe-webview.js
lurch: Ahhh, so it'd actually be in the banner page that we do the `console.log` shown here, and `consoleMessage` then forwards that to our analytics services? Now I understand :-)
(so perhaps the example could clarify that?)

Also, would it be worth the analytics-logging clarifying (perhaps by adding an extra property?) that these are messages being generated by whatever JS is running in the page being displayed in the webview, and not messages being generated by the core Etcher application itself?

lib/gui/components/safe-webview.js
```diff
@@ -0,0 +1,241 @@
...
+   * @param {Event} event - event object
+   *
+   * @example
+   * console.log({
+   *   type: 0,
+   *   message: 'Hello, World!'
+   * });
+   */
+  static consoleMessage(event) {
+    const LOG_LEVEL = 0;
```
jviotti: Can we re-use our robot existing commands? See https://github.com/resin-io/etcher/blob/master/lib/shared/robot/index.js#L27. We only have "error" and "log" at the moment, but we could also add a "warning" one.

lib/gui/components/safe-webview.js
```diff
@@ -0,0 +1,241 @@
...
+    const WARN_LEVEL = 1;
+    const ERROR_LEVEL = 2;
+
+    _.attempt(() => {
+      const data = JSON.parse(event.message);
```
jviotti: Let's also re-use our robot module to parse messages.

lib/gui/components/safe-webview.js
```diff
@@ -0,0 +1,241 @@
...
+      [ 'did-fail-load', this.didFailLoad ],
+      [ 'did-get-response-details', this.didGetResponseDetails ],
+      [ 'new-window', this.constructor.newWindow ],
+      [ 'console-message', this.constructor.consoleMessage ]
+    ];
+
+    // Make a persistent electron session for the webview
+    electron.remote.session.fromPartition(ELECTRON_SESSION, {
+
+      // Disable the cache for the session
```
jviotti: Can we extend this message to describe why we want to disable the cache?

lib/gui/components/safe-webview.js
lurch: Nice thinking @jviotti :+1:

lib/gui/components/safe-webview.js
lurch: Unless there's an explicit need for 'warnings', it might suffice to just stick with "error" and "log"?

jviotti: True, that should be more than enough.

Shou:
Here we go. Let me know what you think.

lib/gui/components/safe-webview.js
```diff
@@ -198,26 +199,29 @@ class SafeWebview extends react.PureComponent {
...
+   * // In the webview
    * console.log({
-   *   type: 0,
-   *   message: 'Hello, World!'
+   *   command: 'error',
+   *   data: 'Good night!'
    * });
+   *
+   * // or with robot
+   * robot.printError('Good night!');
```
lurch: But the `robot` isn't exposed to the webview, is it?

lib/gui/components/safe-webview.js
```diff
@@ -198,26 +199,29 @@ class SafeWebview extends react.PureComponent {
...
-        analytics.logEvent(data.message);
+    if (command === robot.COMMAND.LOG) {
+      analytics.logEvent(data);
+    }
 
-      } else if (data.type === ERROR_LEVEL) {
-        analytics.logException(data.message);
-      }
-    });
+    if (command === robot.COMMAND.ERROR) {
```
lurch: This should be using `else if` ?

lib/gui/components/safe-webview.js
```diff
@@ -198,26 +199,29 @@ class SafeWebview extends react.PureComponent {
...
+   * robot.printError('Good night!');
    */
   static consoleMessage(event) {
-    const LOG_LEVEL = 0;
-    const WARN_LEVEL = 1;
-    const ERROR_LEVEL = 2;
+    const {
+      command,
+      data
+    } = _.attempt(robot.parseMessage, event.message);
```
lurch: It's up to @jviotti which style he prefers, but why 'unpack' the object here, instead of just checking e.g. `parsedMessage.command` instead of `command` later on?

lib/gui/components/safe-webview.js
```diff
@@ -198,26 +199,29 @@ class SafeWebview extends react.PureComponent {
    * @param {Event} event - event object
    *
    * @example
+   *
+   * // In the webview
    * console.log({
-   *   type: 0,
-   *   message: 'Hello, World!'
+   *   command: 'error',
```
lurch: Given that the Robot constants aren't available to the webview, perhaps the examples could show both 'error' and 'log' being used?
And maybe the `@summary` should be "Forward specially-formatted console messages..."?

lib/gui/components/safe-webview.js
lurch: Although according to https://github.com/resin-io/etcher/blob/master/lib/shared/robot/README.md you're 'supposed' to use `robot.getCommand(message)` and `robot.getData(message)` ;-)
And it looks like you can use `robot.isMessage` to avoid using `_.attempt`.

lib/gui/components/safe-webview.js
Shou: Nope, but maybe we could/should?

lurch: I don't think we should (for security reasons, it makes sense to keep the webview as isolated as possible). And therefore this `robot.printError` example probably isn't useful?

Shou: Oh, I don't mean through injection, more like decoupling `robot` from Etcher into its own module, but I guess if/until then I'll just remove this example.

Shou:
How about now?

lurch:
Looks great :-)

jviotti:
Works amazingly well!

PR #1430, opened 2017-05-15
feat(GUI): add success page resin referrer

Shou:


lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -29,7 +29,7 @@ <h3 class="title"><span class="tick tick--success space-right-tiny"></span> Flas
           <svg-icon path="../../../assets/resin.svg"
             width="100px"
             height="auto"
-            os-open-external="https://resin.io?ref=etcher"></svg-icon>
+            os-open-external="https://resin.io?ref=etcher_success"></svg-icon>
```
lurch: Would it make sense for this to use the same referrer-string as the other URL in this page? i.e. `https://resin.io?ref=etcher_offline_banner`

PR #1434, opened 2017-05-15
chore: unify CI scripts

jviotti:
Currently we have almost equal CI scripts implemented in bash (for
Travis), and batch (for Appveyor). This commit takes advantage of MinGW
bash to unify all the scripts as bash scripts.


scripts/ci/install.sh
```diff
@@ -19,26 +19,34 @@
...
     -r "$TARGET_ARCH" \
     -s "$(pwd)" \
     -c "make info && make electron-develop"
 else
   ./scripts/build/check-dependency.sh pip
-  ./scripts/build/check-dependency.sh brew
   ./scripts/build/check-dependency.sh make
   ./scripts/build/check-dependency.sh npm
 
+  if [ -n "$APPVEYOR_REPO_BRANCH" ]; then
```
lurch: Hmmm, rather than checking for Travis-specific or Appveyor-specific variables throughout the script, maybe it'd be better to set e.g. a `$CI_SERVICE` variable at the top of the script, and then we can add checks for `if [ "$CI_SERVICE"  == "APPVEYOR" ]; then` etc.?

Although it seems the test you're doing here (to determine whether to use `choco` or `brew`) should perhaps also/instead be testing the host OS? Perhaps it'd make sense to pass that in as a cmdline variable?
Or again, perhaps we can automatically determine that from the different env-vars set by the different CI services, so we could have e.g. a `$CI_OS_NAME` variable.

lurch:
Doh! CI tests failing with e.g.
`./scripts/ci/install.sh: line 37: APPVEYOR_REPO_BRANCH: unbound variable`

See http://stackoverflow.com/questions/11362250 for some solutions.

jviotti:
I made the scripts accept the OS and architecture as parameters. Let's see how it goes :)

scripts/ci/deploy.sh
```diff
@@ -16,19 +16,36 @@
 # limitations under the License.
 ###
 
-set -e
-set -u
```
lurch: There's no reason to get rid of these `set` lines, is there?

scripts/ci/deploy.sh
```diff
@@ -16,19 +16,36 @@
...
+    *) usage ;;
+  esac
+done
+
+if [ -z "$ARGV_OPERATING_SYSTEM" ] || [ -z "$ARGV_ARCHITECTURE" ]; then
+  usage
 fi
 
-if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+./scripts/build/check-dependency.sh make
```
lurch: Just to be pedantic, this could be moved to the `else` block ;)

.travis.yml
```diff
@@ -19,6 +19,7 @@ env:
     - CCACHE_COMPRESS=1
     - CC="clang"
     - CXX="clang++"
+    - HOST_OS="$(if [ "$TRAVIS_OS_NAME" == "osx" ]; then echo "darwin"; else echo "$TRAVIS_OS_NAME")"
```
lurch: You've forgotten the `; fi` on the end of this if-block (assuming TravisCI even supports bash-style expressions in the yml file?)

scripts/ci/deploy.sh
jviotti: Oops, that was accidental. Fixed!

scripts/ci/deploy.sh
jviotti: Indeed, fixed!

.travis.yml
jviotti: Do you know of an alternative? To set `HOST_OS` to `darwin` of `TRAVIS_OS_NAME` is `osx`?

scripts/ci/deploy.sh
```diff
@@ -19,18 +19,37 @@
 set -e
```
lurch: Perhaps this script should either be renamed to `deploy-aws-s3.sh`, or it should also make the `publish-bintray-debian` target?

appveyor.yml
```diff
@@ -28,18 +28,18 @@ install:
...
 
 build: off
 
 test_script:
   - node --version
   - npm --version
-  - .\scripts\ci\appveyor-test.bat
-  - .\scripts\ci\appveyor-build-installers.bat
+  - bash .\scripts\ci\test.sh -o win32 -r %TARGET_ARCH%
+  - bash .\scripts\ci\appveyor-build-installers.sh -o win32 -r %TARGET_ARCH%
```
lurch: The CI build has just failed - looks like somebody forgot to remove the word `appveyor-` from this line ;-)

.travis.yml
lurch: Travis is green so I guess you got this working :+1:

appveyor.yml
jviotti: lol, who is that guy? :P

scripts/ci/deploy.sh
jviotti: I renamed it to `deploy-aws-s3`. I'll rename it back to `deploy` once I manage to configure bintray snapshot builds.

appveyor.yml
lurch: Dunno, but I reckon we ought to forgive him for such a simple mistake :)

scripts/ci/deploy.sh
lurch: Ah yes, I forgot that this deploys both production and snapshot builds. In that case perhaps `deploy-snapshot.sh` would be a better name than `deploy-aws-s3.sh` ? \*shrug\*

jviotti: The same script is able to deploy to both snapshot or production, depending on the environment.

lurch: Right. Except that our production builds are code-signed, which the CI-servers don't currently do? Or am I misremembering?

jviotti: Exactly, that's the only difference.

lurch:
CI tests are now all green (after I restarted a failed Appveyor build) but there's a couple of conflicts.

jviotti:
Rebased!

lurch:
The appveyor branch-build passed, but the pr-build failed with
```
mkdir release/.tmp
./scripts/build/electron-create-resources-app.sh -s . -o release/electron-win32-1.0.0+9895cea-x86-app \
		-v 1.0.0+9895cea \
		-f "lib,assets"
cp -RLf release/electron-win32-x86-dependencies/* release/electron-win32-1.0.0+9895cea-x86-app
cp: cannot copy cyclic symbolic link `release/electron-win32-x86-dependencies/node_modules/@types/lodash/toPath'
make: *** [release/electron-win32-1.0.0+9895cea-x86-app] Error 1
Command exited with code 2
```
again :-/
Let's see if a rebuild fixes it...

scripts/ci/install.sh
```diff
@@ -19,26 +19,58 @@
...
   ./scripts/build/docker/run-command.sh \
-    -r "$TARGET_ARCH" \
+    -r "$ARGV_ARCHITECTURE" \
     -s "$(pwd)" \
     -c "make info && make electron-develop"
 else
   ./scripts/build/check-dependency.sh pip
-  ./scripts/build/check-dependency.sh brew
   ./scripts/build/check-dependency.sh make
   ./scripts/build/check-dependency.sh npm
```
lurch: Personally, I'd re-order these lines to match the order the commands are used in, i.e. `npm, pip, make`. But that's obviously a very minor thing and doesn't really matter.

jviotti: Done!

jviotti:
> `cp: cannot copy cyclic symbolic link`

Hm, that's weird. I'll investigate this and see if I can send a fix in a future PR.

lurch:
Seems to be an intermittent error, so it's probably going to be very hard to reproduce? :-/
(especially if it's an error which only ever pops up on Appveyor)

Issue #1427, opened 2017-05-14
Feature request: show flashing percents in the window headbar


TotalCaesar659:
Hello, could you add flashing percents to the window headbar? It's not comfortable to check progress manually.
![1](https://cloud.githubusercontent.com/assets/14265316/26037646/18981d60-3900-11e7-8d08-0496d78c6766.png)

lurch:
I guess you mean dynamically changing the title bar to e.g. `Etcher - Flashing 32%` or `Etcher - Validating 64%` ?
What do you think from a UX perspective @konmouz ?

jviotti:
Sounds good to me, but I wonder if this is a common practice for desktop apps. Let's get feedback from the design team.

konmouz:
@lurch & @jviotti This is a common practice in many applications. It is pretty much what we already do with the app icon and the progress bar (at least on mac). Sounds a nice addition to me.

jviotti:
OK, cool, let's go for it!

PR #1438, opened 2017-05-16
feat(GUI): add progress and status to window title

Shou:
We add the progress percentage to the window alongside the status
(validating, flashing).

Closes: #1427
Fixes: #1439
Changelog-Entry: Add the progress and status to the window title.

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -171,11 +171,11 @@ app.run(() => {
 
 app.run(() => {
   store.subscribe(() => {
-    const currentFlashState = flashState.getFlashState()
-
     if (!flashState.isFlashing()) {
       return
     }
+
+    const currentFlashState = flashState.getFlashState()
 
     // There is usually a short time period between the `isFlashing()`
     // property being set, and the flashing actually starting, which
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -173,6 +173,10 @@ app.run(() => {
   store.subscribe(() => {
     const currentFlashState = flashState.getFlashState()
 
+    if (!flashState.isFlashing()) {
+      return
+    }
+
     // There is usually a short time period between the `isFlashing()`
     // property being set, and the flashing actually starting, which
     // might cause some non-sense flashing state logs including
@@ -180,17 +184,15 @@ app.run(() => {
     //
     // We use the presence of `.eta` to determine that the actual
     // writing started.
-    if (!flashState.isFlashing() || !currentFlashState.eta) {
-      return
+    if (!currentFlashState.eta) {
+      analytics.logDebug([
+        `Progress (${currentFlashState.type}):`,
+        `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s`,
+        `(eta ${currentFlashState.eta}s)`
+      ].join(' '))
     }
 
-    analytics.logDebug([
-      `Progress (${currentFlashState.type}):`,
-      `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s`,
-      `(eta ${currentFlashState.eta}s)`
-    ].join(' '))
-
-    windowProgress.set(currentFlashState.percentage)
+    windowProgress.set(currentFlashState)
   })
 })
 
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -185,11 +185,11 @@ app.run(() => {
     // We use the presence of `.eta` to determine that the actual
     // writing started.
     if (!currentFlashState.eta) {
-      analytics.logDebug([
-        `Progress (${currentFlashState.type}):`,
-        `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s`,
+      analytics.logDebug(
+        `Progress (${currentFlashState.type}): ` +
+        `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s ` +
         `(eta ${currentFlashState.eta}s)`
-      ].join(' '))
+      )
     }
 
     windowProgress.set(currentFlashState)
--- a/lib/gui/modules/progress-status.js
+++ b/lib/gui/modules/progress-status.js
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2017 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const settings = require('../models/settings')
+const utils = require('../../shared/utils')
+
+/**
+ * @summary Make the progress status subtitle string
+ *
+ * @param {Object} state - flashing metadata
+ *
+ * @returns {String}
+ *
+ * @example
+ * const status = progressStatus.fromFlashState({
+ *   type: 'write',
+ *   percentage: 55,
+ *   speed: 2049
+ * })
+ *
+ * console.log(status)
+ * // '55% Flashing'
+ */
+exports.fromFlashState = (state) => {
+  const isChecking = state.type === 'check'
+  const shouldValidate = settings.get('validateWriteOnSuccess')
+  const shouldUnmount = settings.get('unmountOnSuccess')
+
+  if (state.percentage === utils.PERCENTAGE_MINIMUM && !state.speed) {
+    if (isChecking) {
+      return 'Validating...'
+    }
+
+    return 'Starting...'
+  } else if (state.percentage === utils.PERCENTAGE_MAXIMUM) {
+    if ((isChecking || !shouldValidate) && shouldUnmount) {
+      return 'Unmounting...'
+    }
+
+    return 'Finishing...'
+  } else if (state.type === 'write') {
+    return `${state.percentage}% Flashing`
+  } else if (state.type === 'check') {
+    return `${state.percentage}% Validating`
+  }
+
+  throw new Error(`Invalid state: ${JSON.stringify(state)}`)
+}
--- a/lib/gui/os/window-progress.js
+++ b/lib/gui/os/window-progress.js
@@ -18,6 +18,39 @@
 
 const electron = require('electron')
 const utils = require('../../shared/utils')
+const progressStatus = require('../modules/progress-status')
+
+/**
+ * @summary The title of the main window upon program launch
+ * @type {String}
+ * @private
+ * @constant
+ */
+const initialTitle = document.title
+
+/**
+ * @summary Make the full window status title
+ * @private
+ *
+ * @param {Object} state - flash state object
+ *
+ * @returns {String}
+ *
+ * @example
+ * const title = makeTitle({
+ *   type: 'write',
+ *   percentage: 55,
+ *   speed: 2049
+ * });
+ *
+ * console.log(title);
+ * // 'Etcher \u2013 55% Flashing'
+ */
+const makeTitle = (state) => {
+  const subtitle = progressStatus.fromFlashState(state)
+  const DASH_UNICODE_CHAR = '\u2013'
+  return `${initialTitle} ${DASH_UNICODE_CHAR} ${subtitle}`
+}
 
 /**
  * @summary A reference to the current renderer Electron window
@@ -37,13 +70,18 @@ exports.currentWindow = electron.remote.getCurrentWindow()
  * @description
  * Show progress inline in operating system task bar
  *
- * @param {Number} percentage - percentage
+ * @param {Number} state - flash state object
  *
  * @example
- * windowProgress.set(85);
+ * windowProgress.set({
+ *   type: 'write',
+ *   percentage: 55,
+ *   speed: 2049
+ * })
  */
-exports.set = (percentage) => {
-  exports.currentWindow.setProgressBar(utils.percentageToFloat(percentage))
+exports.set = (state) => {
+  exports.currentWindow.setProgressBar(utils.percentageToFloat(state.percentage))
+  exports.currentWindow.setTitle(makeTitle(state))
 }
 
 /**
@@ -59,4 +97,5 @@ exports.clear = () => {
   const ELECTRON_PROGRESS_BAR_RESET_VALUE = -1
 
   exports.currentWindow.setProgressBar(ELECTRON_PROGRESS_BAR_RESET_VALUE)
+  exports.currentWindow.setTitle(initialTitle)
 }
--- a/lib/gui/os/window-progress.js
+++ b/lib/gui/os/window-progress.js
@@ -26,7 +26,7 @@ const progressStatus = require('../modules/progress-status')
  * @private
  * @constant
  */
-const initialTitle = document.title
+const INITIAL_TITLE = document.title
 
 /**
  * @summary Make the full window status title
@@ -37,7 +37,7 @@ const initialTitle = document.title
  * @returns {String}
  *
  * @example
- * const title = makeTitle({
+ * const title = getWindowTitle({
  *   type: 'write',
  *   percentage: 55,
  *   speed: 2049
@@ -46,10 +46,14 @@ const initialTitle = document.title
  * console.log(title);
  * // 'Etcher \u2013 55% Flashing'
  */
-const makeTitle = (state) => {
-  const subtitle = progressStatus.fromFlashState(state)
-  const DASH_UNICODE_CHAR = '\u2013'
-  return `${initialTitle} ${DASH_UNICODE_CHAR} ${subtitle}`
+const getWindowTitle = (state) => {
+  if (state) {
+    const subtitle = progressStatus.fromFlashState(state)
+    const DASH_UNICODE_CHAR = '\u2013'
+    return `${INITIAL_TITLE} ${DASH_UNICODE_CHAR} ${subtitle}`
+  }
+
+  return INITIAL_TITLE
 }
 
 /**
@@ -81,7 +85,7 @@ exports.currentWindow = electron.remote.getCurrentWindow()
  */
 exports.set = (state) => {
   exports.currentWindow.setProgressBar(utils.percentageToFloat(state.percentage))
-  exports.currentWindow.setTitle(makeTitle(state))
+  exports.currentWindow.setTitle(getWindowTitle(state))
 }
 
 /**
@@ -97,5 +101,5 @@ exports.clear = () => {
   const ELECTRON_PROGRESS_BAR_RESET_VALUE = -1
 
   exports.currentWindow.setProgressBar(ELECTRON_PROGRESS_BAR_RESET_VALUE)
-  exports.currentWindow.setTitle(initialTitle)
+  exports.currentWindow.setTitle(getWindowTitle(null))
 }
--- a/lib/gui/pages/main/controllers/flash.js
+++ b/lib/gui/pages/main/controllers/flash.js
@@ -17,10 +17,9 @@
 'use strict'
 
 const messages = require('../../../../shared/messages')
-const settings = require('../../../models/settings')
 const flashState = require('../../../../shared/models/flash-state')
 const driveScanner = require('../../../modules/drive-scanner')
-const utils = require('../../../../shared/utils')
+const progressStatus = require('../../../modules/progress-status')
 const notification = require('../../../os/notification')
 const exceptionReporter = require('../../../modules/exception-reporter')
 const path = require('path')
@@ -113,23 +112,10 @@ module.exports = function (
    * const label = FlashController.getProgressButtonLabel();
    */
   this.getProgressButtonLabel = () => {
-    const currentFlashState = flashState.getFlashState()
-    const isChecking = currentFlashState.type === 'check'
-
     if (!flashState.isFlashing()) {
       return 'Flash!'
-    } else if (currentFlashState.percentage === utils.PERCENTAGE_MINIMUM && !currentFlashState.speed) {
-      return 'Starting...'
-    } else if (currentFlashState.percentage === utils.PERCENTAGE_MAXIMUM) {
-      if (isChecking && settings.get('unmountOnSuccess')) {
-        return 'Unmounting...'
-      }
-
-      return 'Finishing...'
-    } else if (isChecking) {
-      return `${currentFlashState.percentage}% Validating...`
     }
 
-    return `${currentFlashState.percentage}%`
+    return progressStatus.fromFlashState(flashState.getFlashState())
   }
 }
--- a/tests/gui/modules/progress-status.spec.js
+++ b/tests/gui/modules/progress-status.spec.js
@@ -1,0 +1,104 @@
+'use strict'
+
+const m = require('mochainon')
+const settings = require('../../../lib/gui/models/settings')
+const progressStatus = require('../../../lib/gui/modules/progress-status')
+
+describe('Browser: progressStatus', function () {
+  describe('.fromFlashState()', function () {
+    beforeEach(function () {
+      this.state = {
+        type: 'write',
+        percentage: 0,
+        eta: 15,
+        speed: 100000000000000
+      }
+
+      settings.set('unmountOnSuccess', true)
+      settings.set('validateWriteOnSuccess', true)
+    })
+
+    it('should report 0% if percentage == 0 but speed != 0', function () {
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('0% Flashing')
+    })
+
+    it('should handle percentage == 0, type == write, unmountOnSuccess', function () {
+      this.state.speed = 0
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Starting...')
+    })
+
+    it('should handle percentage == 0, type == write, !unmountOnSuccess', function () {
+      this.state.speed = 0
+      settings.set('unmountOnSuccess', false)
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Starting...')
+    })
+
+    it('should handle percentage == 0, type == check, unmountOnSuccess', function () {
+      this.state.speed = 0
+      this.state.type = 'check'
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Validating...')
+    })
+
+    it('should handle percentage == 0, type == check, !unmountOnSuccess', function () {
+      this.state.speed = 0
+      this.state.type = 'check'
+      settings.set('unmountOnSuccess', false)
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Validating...')
+    })
+
+    it('should handle percentage == 50, type == write, unmountOnSuccess', function () {
+      this.state.percentage = 50
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('50% Flashing')
+    })
+
+    it('should handle percentage == 50, type == write, !unmountOnSuccess', function () {
+      this.state.percentage = 50
+      settings.set('unmountOnSuccess', false)
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('50% Flashing')
+    })
+
+    it('should handle percentage == 50, type == check, unmountOnSuccess', function () {
+      this.state.type = 'check'
+      this.state.percentage = 50
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('50% Validating')
+    })
+
+    it('should handle percentage == 50, type == check, !unmountOnSuccess', function () {
+      this.state.type = 'check'
+      this.state.percentage = 50
+      settings.set('unmountOnSuccess', false)
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('50% Validating')
+    })
+
+    it('should handle percentage == 100, type == write, unmountOnSuccess, validateWriteOnSuccess', function () {
+      this.state.percentage = 100
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Finishing...')
+    })
+
+    it('should handle percentage == 100, type == write, unmountOnSuccess, !validateWriteOnSuccess', function () {
+      this.state.percentage = 100
+      settings.set('validateWriteOnSuccess', false)
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Unmounting...')
+    })
+
+    it('should handle percentage == 100, type == write, !unmountOnSuccess, !validateWriteOnSuccess', function () {
+      this.state.percentage = 100
+      settings.set('unmountOnSuccess', false)
+      settings.set('validateWriteOnSuccess', false)
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Finishing...')
+    })
+
+    it('should handle percentage == 100, type == check, unmountOnSuccess', function () {
+      this.state.type = 'check'
+      this.state.percentage = 100
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Unmounting...')
+    })
+
+    it('should handle percentage == 100, type == check, !unmountOnSuccess', function () {
+      this.state.type = 'check'
+      this.state.percentage = 100
+      settings.set('unmountOnSuccess', false)
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Finishing...')
+    })
+  })
+})
--- a/tests/gui/os/window-progress.spec.js
+++ b/tests/gui/os/window-progress.spec.js
@@ -24,38 +24,79 @@ describe('Browser: WindowProgress', function () {
     describe('given a stubbed current window', function () {
       beforeEach(function () {
         this.setProgressBarSpy = m.sinon.spy()
+        this.setTitleSpy = m.sinon.spy()
 
         windowProgress.currentWindow = {
-          setProgressBar: this.setProgressBarSpy
+          setProgressBar: this.setProgressBarSpy,
+          setTitle: this.setTitleSpy
+        }
+
+        this.state = {
+          percentage: 85,
+          speed: 100,
+          type: 'write'
         }
       })
 
       describe('.set()', function () {
         it('should translate 0-100 percentages to 0-1 ranges', function () {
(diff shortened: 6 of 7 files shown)


Shou:
image

tests/gui/os/window-progress.spec.js
```diff
@@ -62,6 +64,16 @@ describe('Browser: WindowProgress', function() {
...
+        it('should set the flashing title', function() {
+          windowProgress.set(100, 'write');
+          m.chai.expect(this.setTitleSpy).to.have.been.calledWith('Etcher – Flashing 100%');
+        });
+
+        it('should set the validating title', function() {
+          windowProgress.set(100, 'check');
+          m.chai.expect(this.setTitleSpy).to.have.been.calledWith('Etcher – Validating 100%');
+        });
+
```
lurch: Can we also have a test that after Validation, the title gets set back to just "Etcher" ?

lib/gui/os/window-progress.js
```diff
@@ -38,12 +38,15 @@ exports.currentWindow = electron.remote.getCurrentWindow();
...
+ * @param {String} status - flashing/validating status
  *
  * @example
  * windowProgress.set(85);
  */
-exports.set = (percentage) => {
+exports.set = (percentage, status) => {
   exports.currentWindow.setProgressBar(utils.percentageToFloat(percentage));
+  const statusString = status === 'check' ? 'Validating' : 'Flashing';
+  exports.currentWindow.setTitle(`Etcher – ${statusString} ${percentage}%`);
```
lurch: Rather than hardcoding `Etcher` here (and below), we should probably get this from `displayName` in `package.json` ?

lib/gui/os/window-progress.js
```diff
@@ -38,12 +38,15 @@ exports.currentWindow = electron.remote.getCurrentWindow();
...
  * @param {Number} percentage - percentage
+ * @param {String} status - flashing/validating status
  *
  * @example
  * windowProgress.set(85);
  */
-exports.set = (percentage) => {
+exports.set = (percentage, status) => {
   exports.currentWindow.setProgressBar(utils.percentageToFloat(percentage));
+  const statusString = status === 'check' ? 'Validating' : 'Flashing';
```
lurch: Are we _sure_ that this function only ever gets called during Flashing or Validating?

lurch:
CI tests are failing with
```
./scripts/ci/ensure-all-text-files-only-ascii.sh
lib/gui/os/window-progress.js contains non-ASCII characters
```
:-/

lurch:
Hmmm, Appveyor is (correctly) failing on pr-build but passing on branch-build?? @jviotti

lib/gui/os/window-progress.js
Shou: For now yes. If it ever changes I left the argument `status` as a string instead of using a boolean so we can just update this function to check for other things.

lib/gui/os/window-progress.js
```diff
@@ -46,7 +47,8 @@ exports.currentWindow = electron.remote.getCurrentWindow();
 exports.set = (percentage, status) => {
   exports.currentWindow.setProgressBar(utils.percentageToFloat(percentage));
   const statusString = status === 'check' ? 'Validating' : 'Flashing';
-  exports.currentWindow.setTitle(`Etcher – ${statusString} ${percentage}%`);
+  const dash = '\u2013';
```
lurch: What's wrong with a regular hyphen? Does using a unicode-dash have some special significance?

lib/gui/os/window-progress.js
jviotti: We have some very similar logic to show the content of the main progress bar. Can we extract that somewhere and reuse?

lib/gui/os/window-progress.js
```diff
@@ -38,12 +39,16 @@ exports.currentWindow = electron.remote.getCurrentWindow();
  * Show progress inline in operating system task bar
  *
  * @param {Number} percentage - percentage
+ * @param {String} status - flashing/validating status
  *
  * @example
  * windowProgress.set(85);
  */
-exports.set = (percentage) => {
+exports.set = (percentage, status) => {
```
jviotti: Can we decouple this module from Etcher's use case by naming it accept a generic title, and calculate what that title should actually be from outside?

jviotti: Hm, although that now raises the question of whether this functionality should live in this module at all. Maybe worth creating a new function that simply sets the window title, and nothing else?

lib/gui/os/window-progress.js
lurch: I wondered that too, which is what lead to #1439 ;-)

jviotti: Haha, good catch!

lib/gui/os/window-progress.js
Shou: Sure! I like this idea.

lib/gui/os/window-progress.js
Shou: It's more aligned with typography, where hyphens are used to link words together, and dashes used to separate words alike commas or parentheses.

lib/gui/os/window-progress.js
Shou: Maybe we can make a separate function that's also used within the `set` function? That way, in the future, should we want to set the title to something arbitrary somewhere, we can do that.

lib/gui/os/window-progress.js
lurch: Fair enough. Have you been able to check that all the platforms that Etcher runs on support unicode characters in their window-titles?

Shou: It definitely works on Linux with any semi-modern DE/panel, and Windows. I think it's safe to assume it works on OS X too but I can't test this myself, @jviotti? As long as it supports unicode we're pretty much good because the dash used is in a lower and not very obscure character block, which most modern fonts support.

lib/gui/os/window-progress.js
jviotti: Not sure I follow the `set` function idea. Can you give me an example?

lib/gui/os/window-progress.js
jviotti: I'll give it a go in a second.

jviotti: BTW, should this variable be in uppercase (e.g. `DASH`), given that its statically determined? Also, can we call it something like `DASH_UNICODE_CHARACTER`, `SEPARATOR_UNICODE_CHARACTER`, or something like that?

jviotti: So this is not shown at all in macOS, given we don't even have a title bar on that platform, but I guess that's "expected".

jhermsmeier: It should show up on macOS when <kbd>Cmd</kbd>+<kbd>Tab</kbd>-ing through open apps though, I think

Shou:
How about now?

jviotti:
I'm getting a weird timeout error sometimes:

screenshot 2017-05-25 20 54 21

Let's capture the error, avoid a huge scary stacktrace, and send an event that says "Success banner timeout", or something like that.

jviotti:
BTW, I'm getting that timeout about 5 out of 6 times. I suspect that a lot of people are on not very good internet connections, so we must ensure it works fine on those cases.

jviotti:
I still get the same error right after clicking "Flash again":

screenshot 2017-05-25 20 57 46

Can you try navigating to another link on the webview before clicking the "Flash Again" button?

jviotti:
I confirm that despite the errors, the webview is correctly reset.

Shou:
I think this may be the wrong PR but I'll take it to mind hah

jviotti:
Yeah, you can do it in a separate PR

Shou:
Oh sorry, I mean this is the PR for the window progress title, not the webview :)

Does it look good?

lurch:
Yeah, looks like poor @jviotti is getting his PRs confused ;-)

@Shou this PR currently has conflicts?

lib/gui/app.js
```diff
@@ -176,6 +176,11 @@ app.run(() => {
     ].join(' '));
 
     windowProgress.set(currentFlashState.percentage);
+
+    const DASH_UNICODE_CHAR = '\u2013';
+    const flashProgress = messages.info.flashProgress(currentFlashState);
+    const title = `${packageJSON.displayName} ${DASH_UNICODE_CHAR} ${flashProgress}`;
+    windowProgress.setTitle(title);
```
lurch: IMHO it looks very odd to do `windowProgress.set()` and then `windowProgress.setTitle()`. IMHO either the `.set` should be renamed to something like `.setPercentage` or perhaps a better option is for the `.set` method itself to directly also update the title-bar?

EDIT: Which I guess would also have a nice symmetry with the `.clear` method directly updating the title-bar.

lib/gui/app.js
```diff
@@ -176,6 +176,11 @@ app.run(() => {
     ].join(' '));
 
     windowProgress.set(currentFlashState.percentage);
+
+    const DASH_UNICODE_CHAR = '\u2013';
```
lurch: Should this constant be moved to a 'higher level', or does it make sense to keep it so localised?

jviotti:
lol, sorry about that :P

lib/gui/app.js
jviotti: Good point. Here's my proposal:

- Let's ditch the `windowProgress.setTitle()` function
- Store the initial title in this module at the top of the file (that is, read it from the actual window)
- Make `windowProgress.set()` also set the window title, by making use of the initial title we found out before, and appending the dash plus the percentage
- Make `windowProgress.clear()` set the title back to the initial title we calculated

The reason why I don't want `windowProgress.set()` to calculate the title internally is that the way the title is laid out (and the fact that we keep the application name on the left), is not its responsibility. 

I also want to fix `windowProgress.clear()` to not set the title back to `packageJSON.name`, since we're currently assuming the initial value of the window title, which is actually set from the `index.html` file.

lurch: But then how would we know whether to display "validating" or "flashing" in the title bar? Or perhaps `windowProgress.set()` should also have an extra parameter, specifying what state we're currently in? And thinking about it a bit more, maybe it makes sense for the title to also display "Finishing..." too, so that the user doesn't think that Etcher has actually finished before it already has, if they're _only_ watching the title-bar?

jviotti: > EDIT: i.e. maybe FlashController.getProgressButtonLabel() and windowProgress.set() should both now call a new shared method.

Yeah, that sounds like the way to go. Also, we should make this window progress function accept a "progress state" object (the one coming from `etcher-image-write`, which has the percentage, type, speed, etc) rather than just a percentage, so the function has enough context to know if the percentage is avoid writing or validation.

lurch: > Also, we should make this window progress function accept a "progress state" object (the one coming from etcher-image-write, which has the percentage, type, speed, etc) rather than just a percentage

Yeah, that's exactly what I was thinking with my `EDIT:` comment, but forgot to explicitly mention it :+1:

lurch:
This'll need to be refactored after #1504

Shou:
What do you think of this?

Also, it seems the `Store` doesn't dispatch enough events for `Starting...`, `Finishing...`, and `Umounting...` to appear in the window title. I was thinking of using the Angular cycles instead, but is there maybe a better way, or should I just go with that?

lib/gui/os/window-progress.js
```diff
@@ -18,6 +18,14 @@
 
 const electron = require('electron');
 const utils = require('../../shared/utils');
+const progressStatus = require('../models/progress-status');
+
+/**
+ * @summary The title of the main window upon program launch
+ * @type {String}
+ * @protected
```
jviotti: I guess this should be `@private`

jviotti: Can also mention that this is a `@constant`?

lib/gui/models/progress-status.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @summary Make the progress status subtitle string
+ *
+ * @param {Object} currentFlashState - flashing metadata
+ *
+ * @returns {String}
+ *
+ * @example
+ * const fullStatus = progressStatus.makeSubtitle({
```
jviotti: Can we move this whole file to `lib/gui/modules`, given that its not strictly a model?

jviotti:
> Also, it seems the Store doesn't dispatch enough events for Starting..., Finishing..., and Umounting... to appear in the window title. I was thinking of using the Angular cycles instead, but is there maybe a better way, or should I just go with that?

The store dispatches events every-time its own data is changed. While flashing, the store is changed every time the progress information changes (like each 500ms) or so. 

The "Starting" and "Finishing" events only appear for a brief amount of time in the progress button. Is that not the case in the window title?

lib/gui/pages/main/controllers/flash.js
```diff
@@ -117,23 +117,12 @@ module.exports = function(
    */
   this.getProgressButtonLabel = () => {
     const currentFlashState = flashState.getFlashState();
```
jviotti: This is very minor, but can we move this line above `return progressStatus.makeSubtitle(currentFlashState);` so that we only get the current flash state if we need to (e.g. if we're really flashing)

tests/gui/pages/main.spec.js
```diff
@@ -338,7 +338,7 @@ describe('Browser: MainPage', function() {
           });
 
           settings.set('unmountOnSuccess', false);
-          m.chai.expect(controller.getProgressButtonLabel()).to.equal('50%');
+          m.chai.expect(controller.getProgressButtonLabel()).to.equal('50% Flashing');
```
jviotti: Can we make all the tests that test `getProgressButtonLabel` test `makeSubtitle` instead? Once that's done, the ``getProgressButtonLabel` can be greatly reduced to only test the "Flash" edge case.

lib/gui/models/progress-status.js
```diff
@@ -0,0 +1,70 @@
...
+ *   type: 'write',
+ *   percentage: 55,
+ *   speed: 2049
```
lurch: Perhaps this example should also show what String would be returned?

lib/gui/models/progress-status.js
```diff
@@ -0,0 +1,70 @@
...
+ * });
+ */
+exports.makeSubtitle = (currentFlashState) => {
```
lurch: I wonder if something like `convertFlashStatusToString` would be a better name than `makeSubtitle`?

lib/gui/models/progress-status.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  if (subtitle === '') {
+    throw new Error(`Invalid state: ${JSON.stringify(currentFlashState)}`);
+  }
+
+  return subtitle;
+};
+
+exports.makeTitle = (baseTitle, currentFlashState) => {
```
lurch: Should this also be documented with an example?
And AFAICT this function is quite specific, so perhaps it 'belongs' in `window-progress.js` instead? \*shrug\*

lib/gui/os/window-progress.js
```diff
@@ -18,7 +18,14 @@
...
 const utils = require('../../shared/utils');
-const packageJSON = require('../../../package.json');
+const progressStatus = require('../models/progress-status');
+
+/**
+ * @summary The title of the main window upon program launch
+ * @type {String}
+ * @protected
+ */
+const entryTitle = document.title;
```
lurch: Minor nit, but `entryTitle` doesn't sound right to me. Perhaps `initialTitle` or `defaultTitle` instead?

lib/gui/os/window-progress.js
```diff
@@ -38,29 +45,15 @@ exports.currentWindow = electron.remote.getCurrentWindow();
  * @description
  * Show progress inline in operating system task bar
  *
- * @param {Number} percentage - percentage
+ * @param {Number} state - flashing metadata
  *
  * @example
  * windowProgress.set(85);
```
lurch: I guess the example needs updating ;-)

tests/gui/os/window-progress.spec.js
```diff
@@ -27,41 +27,67 @@ describe('Browser: WindowProgress', function() {
...
           m.chai.expect(function() {
-            windowProgress.set(-1);
+            windowProgress.set(state);
           }).to.throw('Invalid percentage: -1');
         });
 
+        it('should set the flashing title', function() {
+          this.state.type = 'write';
+          windowProgress.set(this.state);
+          m.chai.expect(this.setTitleSpy).to.have.been.calledWith(' \u2013 85% Flashing');
```
lurch: Hmmm, so the `entryTitle` is a blank string when the unit tests are being run? Is that expected / deliberate?
I wonder if 'something' could be done to make the unit-tests behave more similarly to the application itself?
(since at the moment, I think changing
```
return `${baseTitle} ${DASH_UNICODE_CHAR} ${subtitle}`;
```
to
```
return ` ${DASH_UNICODE_CHAR} ${subtitle}`;
```
or changing `const entryTitle = document.title;` to `const entryTitle = '';` would still allow all these unit-tests pass?)

tests/gui/pages/main.spec.js
lurch: ...and there should probably be a test for `makeTitle` too (see my previous comment).

lurch:
There's also a failing CI check:
```
/etcher/lib/gui/pages/main/controllers/flash.js
  22:7  error  'utils' is assigned a value but never used  no-unused-vars
```

tests/gui/modules/progress-status.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      };
+
+      settings.set('unmountOnSuccess', true);
+    });
+
+    it('should report 0% if percentage == 0 but speed != 0', function() {
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('0% Flashing');
+    });
+
+    it('should handle percentage == 0, type = write, unmountOnSuccess', function() {
```
lurch: For consistency, should these descriptions use `type ==  ` instead of `type = ` ?

tests/gui/modules/progress-status.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      this.state.speed = 0;
+      settings.set('unmountOnSuccess', false);
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Starting...');
+    });
+
+    it('should handle percentage == 0, type = check, unmountOnSuccess', function() {
+      this.state.speed = 0;
+      this.state.type = 'check';
+      settings.set('unmountOnSuccess', false);
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Starting...');
```
lurch: Hmmm, looks weird to have the text reset back to `Starting...` when `type == check` ? :-/
Perhaps in this case it should just say `Validating...`? What do you think @jviotti ?

tests/gui/modules/progress-status.spec.js
```diff
@@ -0,0 +1,86 @@
...
+
+    it('should handle percentage == 100, type = write', function() {
+      this.state.percentage = 100;
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Finishing...');
+    });
+
+    it('should handle percentage == 100, type = check, unmountOnSuccess', function() {
+      this.state.type = 'check';
+      this.state.percentage = 100;
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('Unmounting...');
```
lurch: If `percentage == 100, type == write, unmountOnSuccess` and the user has turned off validation (i.e. we'll never get to `type == check`), should it also display `Unmounting...` ?

And just for the sake of completion, should there also be a test for `should handle percentage == 100, type = check, !unmountOnSuccess`

tests/gui/modules/progress-status.spec.js
```diff
@@ -0,0 +1,86 @@
...
+        eta: 15,
+        speed: 100000000000000
+      };
+
+      settings.set('unmountOnSuccess', true);
+    });
+
+    it('should report 0% if percentage == 0 but speed != 0', function() {
+      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('0% Flashing');
+    });
```
lurch: The tests in here look much neater than those you removed from `main.spec.js` - good job :+1:

tests/gui/modules/progress-status.spec.js
Shou: Yeah I had to double-check I was doing things right when I saw that, but I didn't think to change it. So I agree!

tests/gui/modules/progress-status.spec.js
Shou: Those are some good suggestions, let me add them.

Shou:
Here we go.

@jviotti 
> The store dispatches events every-time its own data is changed. While flashing, the store is changed every time the progress information changes (like each 500ms) or so.
> The "Starting" and "Finishing" events only appear for a brief amount of time in the progress button. Is that not the case in the window title?

Yeah it goes straight to the percentage labels instead.

lib/gui/modules/progress-status.js
```diff
@@ -38,14 +38,19 @@ const utils = require('../../shared/utils');
...
 
   let subtitle = '';
 
-  if (currentFlashState.percentage === utils.PERCENTAGE_MINIMUM && !currentFlashState.speed) {
+  if (!isChecking && currentFlashState.percentage === utils.PERCENTAGE_MINIMUM && !currentFlashState.speed) {
     subtitle = 'Starting...';
 
+  } else if (isChecking && currentFlashState.percentage === utils.PERCENTAGE_MINIMUM && !currentFlashState.speed) {
+    subtitle = 'Validating...';
+
```
Shou: That's better, thanks!

Shou:
Done that now.

tests/gui/os/window-progress.spec.js
jviotti: I think that a good default that doesn't require any extra knowledge is to just display `85% Flashing`, without any dash.

tests/gui/pages/main.spec.js
```diff
@@ -223,222 +222,6 @@ describe('Browser: MainPage', function() {
         m.chai.expect(controller.getProgressButtonLabel()).to.equal('Flash!');
       });
 
-      describe('given there is a flash in progress', function() {
```
jviotti: Can we keep just one test case that tests something other than "Flash!" to cover the other side of the conditional?

lib/gui/modules/progress-status.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * console.log(fullStatus);
+ * // '55% Flashing'
+ */
+exports.fromFlashState = (currentFlashState) => {
+  const isChecking = currentFlashState.type === 'check';
+  const shouldValidate = settings.get('validateWriteOnSuccess');
+  const shouldUnmount = settings.get('unmountOnSuccess');
+
+  let subtitle = '';
```
jviotti: Can we try to refactor this function to avoid mutability? I haven't test this, but I think you can simply replace every `subtitle = <value>` assignment with `return <value>`.

lib/gui/os/window-progress.js
```diff
@@ -37,13 +71,20 @@ exports.currentWindow = electron.remote.getCurrentWindow();
...
+ * windowProgress.set({
+ *   type: 'write',
+ *   percentage: 55,
+ *   speed: 2049
+ * });
  */
-exports.set = (percentage) => {
-  exports.currentWindow.setProgressBar(utils.percentageToFloat(percentage));
+exports.set = (state) => {
+  const title = makeTitle(initialTitle, state);
```
lurch: But I guess `makeTitle` will become (a tiny bit) more complex if it incorporates your suggestion to "I think that a good default that doesn't require any extra knowledge is to just display `85% Flashing`, without any dash." ?

tests/gui/os/window-progress.spec.js
lurch: Just to clarify - I wasn't 'complaining' about ` ${DASH_UNICODE_CHAR} ${subtitle}` being displayed when `initialTitle` is an empty string; I was 'complaining' about `initialTitle` being an empty string when the unit-tests are being run ;-)
(I hope that makes sense)

lib/gui/modules/progress-status.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+exports.fromFlashState = (currentFlashState) => {
+  const isChecking = currentFlashState.type === 'check';
+  const shouldValidate = settings.get('validateWriteOnSuccess');
+  const shouldUnmount = settings.get('unmountOnSuccess');
+
+  if (currentFlashState.percentage === utils.PERCENTAGE_MINIMUM && !currentFlashState.speed) {
+    if (isChecking) {
+      return 'Validating...';
+
```
lurch: minor nit: Do we need so much whitespace here?

Shou: Thanks, that was unintended.

tests/gui/os/window-progress.spec.js
```diff
@@ -27,41 +27,82 @@ describe('Browser: WindowProgress', function() {
...
+          const state = this.state;
           m.chai.expect(function() {
-            windowProgress.set(101);
+            windowProgress.set(state);
           }).to.throw('Invalid percentage: 101');
         });
 
         it('should throw if given a percentage less than 0', function() {
+          this.state.percentage = -1;
+          const state = this.state;
```
lurch: I've just noticed that some unit-tests do:
```
this.state.foo = blah;
windowProgress.set(this.state);
```
but others do:
```
this.state.foo = blah;
const state = this.state;
windowProgress.set(state);
```

Any reason for that?

lurch:
Things are looking great :+1: 
Could you update the screenshot at the top of this PR? :)

And I believe this PR should also explicitly state that it fixes #1439 too?

tests/gui/os/window-progress.spec.js
Shou: Apparently `this` is `undefined` inside the exception tests, so I had to work around it.

Shou:
Done!

lib/gui/modules/progress-status.js
```diff
@@ -0,0 +1,66 @@
...
+ * @returns {String}
+ *
+ * @example
+ * const status = progressStatus.fromFlashState({
+ *   type: 'write',
+ *   percentage: 55,
+ *   speed: 2049
+ * });
+ *
+ * console.log(fullStatus);
```
jviotti: Should this be `status`?

lib/gui/modules/progress-status.js
```diff
@@ -0,0 +1,66 @@
...
+ * // '55% Flashing'
+ */
+exports.fromFlashState = (currentFlashState) => {
```
jviotti: Can we name the argument `flashState` or `state`? I guess this function shouldn't care if the flash state is the current one or not.

lib/gui/os/window-progress.js
jviotti: Fair enough :)

tests/gui/os/window-progress.spec.js
jviotti: Oh, I see. I guess the title is never set because `electron-mocha` never loads any HTML. Let's set one manually here then, just for testing purposes.

lib/gui/modules/progress-status.js
Shou: Sure!

tests/gui/os/window-progress.spec.js
lurch: Then again, `initialTitle` is a private constant, so I dunno how you'd set it manually just for the unit-tests?

Shou: We'd have to set the `document.title` before `initialTitle` is set. Is that practically possible?

jviotti: @Shou How do you feel about sending a small PR to `electron-mocha` to set a default title in the HTML file they load (https://github.com/jprichardson/electron-mocha/blob/master/renderer/index.html)? We could make it something like "Electron Mocha", and that should be more than enough to test our changes here.

lurch:
Is this the PR where Juanchi was suggesting adding an additional "Elevating..." status?

Shou:
I made a separate local branch for that, but I could include it here because they're so closely related.

Shou:
Actually on second thought I think it's better if we just get this merged, because there's still some work to be done for the "Elevation..." changes – they're not so straightforward. Unless there are any changes needed here, this should be done.

lib/gui/app.js
```diff
@@ -160,23 +160,25 @@ app.run(() => {
...
+      return;
+    }
+
     // There is usually a short time period between the `isFlashing()`
     // property being set, and the flashing actually starting, which
     // might cause some non-sense flashing state logs including
     // `undefined` values.
     //
     // We use the presence of `.eta` to determine that the actual
     // writing started.
```
lurch: @Shou Is `(flashState.isFlashing() && !currentFlashState.eta)` a reliable indicator of being in the 'elevating' stage? (or does `flashState.isFlashing()` only get set post-elevation?)

tests/gui/os/window-progress.spec.js
Shou: Somehow I missed this reply, and this is a good idea so I'll do that. I've submitted an issue on their repo to serve as a reminder for myself.

lib/gui/app.js
Shou: We need to delve into `child-writer` to get accurate results for the "Elevating..." label, which unfortunately lives in another process so I've been setting up `electron-mocha` hoping that will help.

Shou:
I've rebased this to master now. This should be good to merge.

lib/gui/os/window-progress.js
```diff
@@ -18,6 +18,40 @@
...
+ * @summary Make the full window status title
+ * @private
+ *
+ * @param {String} baseTitle - base window title
+ * @param {Object} currentFlashState - flash state object
+ *
+ * @returns {String}
+ *
+ * @example
+ * const title = makeTitle({
```
lurch: This example is missing the first `baseTitle` parameter

lib/gui/os/window-progress.js
```diff
@@ -18,6 +18,40 @@
...
+ *   type: 'write',
+ *   percentage: 55,
+ *   speed: 2049
+ * });
+ *
+ * console.log(title);
+ * // 'Etcher \u2013 55% Flashing'
+ */
+const makeTitle = (baseTitle, currentFlashState) => {
```
lurch: Like with `progressStatus.fromFlashState`, I wonder if it makes sense to rename this `currentFlashState` parameter to just `state`?

And a minor nitpick, but the jsdoc for `progressStatus.fromFlashState` refers to `flashing metadata` but the jsdoc here refers to `flash state object`. AFAICT these are actually the same thing, so a bit of consistency might be nice ;)

lib/shared/messages.js
```diff
@@ -35,7 +35,9 @@ module.exports = {
     flashComplete: _.template([
       '<%= imageBasename %> was successfully written to',
       '<%= drive.description %> (<%= drive.displayName %>)'
-    ].join(' '))
+    ].join(' ')),
+
+    flashProgress: _.template('<%= type === "check" ? "Validating" : "Flashing" %> <%= percentage %>%')
```
lurch: Hmmm? This isn't actually used anywhere??

lurch: Also, this has the %age after the word, whereas `progress-status.js` currently has the %age before the word?

lib/gui/os/window-progress.js
Shou: I now noticed we can simplify the function to not need that argument, so let me make that change.

lib/gui/os/window-progress.js
```diff
@@ -59,4 +97,5 @@ exports.clear = () => {
   const ELECTRON_PROGRESS_BAR_RESET_VALUE = -1
 
   exports.currentWindow.setProgressBar(ELECTRON_PROGRESS_BAR_RESET_VALUE)
+  exports.currentWindow.setTitle(initialTitle)
```
lurch: TBH, I'm not sure if I like the "asymmetry" of `.setTitle` here taking `initialTitle` as a parameter, but then the other `.setTitle` having `makeTitle(state)` as a parameter, with `makeTitle` itself retrieving `initialTitle`.
Instead of adding `initialTitle` back as a parameter `makeTitle`, I guess another approach would be to change _this_ line to `.setTitle(makeTitle(null))` (or `.setTitle(makeTitle({}))`), with `makeTitle` also modified as necessary, and that way the usage of `initialTitle` is entirely self-contained inside `makeTitle`? IYSWIM ;-)

lib/gui/os/window-progress.js
lurch: See my other comment - and feel free to ignore me if I'm being too picky.

lib/gui/os/window-progress.js
Shou: I feel like the cons outweigh the pros here unless I'm missing something, we're basically just moving the logic that's already present in the separation of `set` and `clear` into `makeTitle` by adding an if-statement.

lib/gui/app.js
```diff
@@ -173,24 +173,26 @@ app.run(() => {
   store.subscribe(() => {
     const currentFlashState = flashState.getFlashState()
 
+    if (!flashState.isFlashing()) {
```
jviotti: Should we get the `currentFlashState` *after* this check, just to make it more natural?

Shou: Yes that's a great call!

Shou:
Is there anything I need to change here or is this good to go now?

lib/gui/os/window-progress.js
jhermsmeier: I agree with @lurch here, it'd be a little less confusing to just have `makeTitle(null)` returning the `initialTitle`, and would get rid of duplication of similar logic.
Also, what do you think of changing the name of `makeTitle()` to `getTitle()` or `getWindowTitle()` to clarify a bit what it's doing?

lib/gui/app.js
```diff
@@ -180,17 +184,15 @@ app.run(() => {
...
     // We use the presence of `.eta` to determine that the actual
     // writing started.
-    if (!flashState.isFlashing() || !currentFlashState.eta) {
-      return
+    if (!currentFlashState.eta) {
+      analytics.logDebug([
+        `Progress (${currentFlashState.type}):`,
+        `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s`,
+        `(eta ${currentFlashState.eta}s)`
+      ].join(' '))
```
jhermsmeier: Can we change this to use string concatenation, so we don't allocate a new array on every call?

Shou: Done this now!

lib/gui/os/window-progress.js
jhermsmeier: @Shou do you want to change this ^ still?

PR #1441, opened 2017-05-16
WIP: docs(CLI-INSTALLATION): Move headings back to h2

lurch:
Also add some h1 headers

Ping @craig-mulligan @jviotti 

This is WIP because it depends on https://github.com/resin-io/etcher-homepage/issues/47 (and shouldn't be merged until after that's been fixed).

BTW Juanchi - why is the CLI executable named just `etcher` and not `etcher-cli` or `etcher-node` as a counterpart to `etcher-electron` ?

### DIFF
--- a/docs/CLI-INSTALLATION.md
+++ b/docs/CLI-INSTALLATION.md
@@ -1,4 +1,8 @@
-### macOS and GNU/Linux
+CLI Installation
+================
+
+macOS and GNU/Linux
+--------------------
 
 - Extract the `.tar.gz` package by running:
 
@@ -15,7 +19,8 @@ tar fvx path/to/cli.tar.gz
 export PATH="$PATH:/opt/etcher-cli"
 ```
 
-### Windows
+Windows
+-------
 
 - Unzip the `.zip` package by right-clicking on it and selecting "Extract All"
 
@@ -43,13 +48,15 @@ export PATH="$PATH:/opt/etcher-cli"
 
   - Re-open `cmd.exe`, or PowerShell
 
-### Running
+Running
+=======
 
 ```sh
 etcher -v
 ```
 
-### Options
+Options
+-------
 
 ```
   --help, -h     show help


jviotti:
> BTW Juanchi - why is the CLI executable named just etcher and not etcher-cli or etcher-node as a counterpart to etcher-electron ?

No particular reason. I never seen a CLI include the `-cli` prefix, but maybe its just me. I'm happy to reconsider the binary name.

lurch:
Perhaps `etcher-cli` would feel less weird if `etcher-electron` was renamed to `etcher-gui`? ;-)
What does the rest of the team think? @alexandrosm @konmouz @Shou @jhermsmeier 

Just to clarify, I have no real objection to the Etcher CLI being named just `etcher`, I'm just playing devil's advocate :)  (to ensure that all options get explored)

craig-mulligan:
LGTM, @lurch I'll ping you when I have done [resin-io/etcher-homepage#47](https://github.com/resin-io/etcher-homepage/issues/47).

lurch:
@craig-mulligan Looking at this again, I've noticed that there's still duplicated content between `docs/CLI-INSTALLATION.md` and `docs/CLI.md` (the Running and Options sections), which obviously isn't ideal.

What would everyone think about moving the Running and Options sections into a separate `docs/CLI-USAGE.md` page?

craig-mulligan:
I'm reverting my initial opinion, if we are parsing the doc to change the headings then we may as well do all the adaptations there rather than splitting it unnaturally.

This would impose some restrictions on updating the doc. For instance we could say we will remove the first paragraph after the `installing` heading, this covers the `Head over to etcher.io/cli, download the package ...` bit. And replace that paragraph with the download table.

We should comment in the markdown warning about that restriction, but seems like the cleanest approach? @lurch

jviotti:
@craig-mulligan Hard-coding something like "ignore the first paragraph" sounds like an assumption that will likely break someday. I wonder if we can add certain comments on the Markdown file that get parsed by the current website generator (and landr in the future), to identify which parts should be excluded. 

For example:

```
<!-- landr:exclude -->
My Document
==========

Foo Bar.
<!-- landr:include -->
My Section
-----------

Foo Bar Baz.
```

jviotti:
Oh, this raises a very interesting idea. Could landr exploit the fact that can embed a small templating engine inside markdown documents as comments?

lurch:
Yeah, or rather than having `<!-- landr:exclude -->` and `<!-- landr:include -->` tags, maybe @craig-mulligan 's job would be easier if we had stuff like:
```
lorem ipsum
<!-- landr:section:installing:start -->
blah blah blah
<!-- landr:section:installing:end -->
foo foo bar
```
?

jhermsmeier:
What's the state of this? Can it be closed, or does it need reviving?

lurch:
I suspect it probably needs reviving, if anyone ever gets any time to give https://etcher.io/cli/ some love? See also #1419  - but I think most of the discussion around this happened in Flowdock.

lurch:
@zvin I guess this can be closed too, given that Etcher CLI no longer exists?

PR #1447, opened 2017-05-17
chore: track python dependencies with pip

jviotti:
We're currently hardcoding various pip dependencies in
`appveyor-install.bat` and `travis-install.sh`.

This commit moves all the dependencies to a `requirements.txt` file in
the root of the project, and makes every install script run `pip install
-r requirements.txt`.

See: https://github.com/resin-io/etcher/pull/1401#discussion_r116547053

### DIFF
--- a/docs/RUNNING-LOCALLY.md
+++ b/docs/RUNNING-LOCALLY.md
@@ -12,9 +12,13 @@ Prerequisites
 - [NodeJS](https://nodejs.org) (at least v6)
 - [Python](https://www.python.org)
 - [jq](https://stedolan.github.io/jq/)
-- [Codespell](https://github.com/lucasdemarchi/codespell)
 - [curl](https://curl.haxx.se/)
-- [cpplint](https://github.com/cpplint/cpplint)
+
+```sh
+pip install -r requirements.txt
+```
+
+You might need to run this with `sudo` or administrator permissions.
 
 ### Windows
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 *
+!requirements.txt
--- a/.gitattributes
+++ b/.gitattributes
@@ -24,6 +24,7 @@ Makefile text
 *.svg text
 *.yml text
 *.patch text
+*.txt text
 
 # Binary files (no line-ending conversions)
 *.bz2 binary diff=hex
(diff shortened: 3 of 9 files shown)


docs/RUNNING-LOCALLY.md
```diff
@@ -12,9 +12,11 @@ Prerequisites
 - [NodeJS](https://nodejs.org) (at least v6)
 - [Python](https://www.python.org)
 - [jq](https://stedolan.github.io/jq/)
-- [Codespell](https://github.com/lucasdemarchi/codespell)
 - [curl](https://curl.haxx.se/)
-- [cpplint](https://github.com/cpplint/cpplint)
+
+```sh
+pip install -r requirements.txt
```
lurch: Should this be `sudo pip ...`?

lurch:
Looks like you need to add `*.txt` to `.gitattributes`. Haven't managed to work out yet why the Docker builds are failing though - something to do with `requirements.txt` not being picked up from the right directory? Perhaps it'd work if you changed the Dockerfiles to use `RUN pip install -r /etcher/requirements.txt` ?

docs/RUNNING-LOCALLY.md
jviotti: I added a clarification at the bottom, since `sudo` is very unix-y

jviotti:
Fixed!

lurch:
Could we do something similar for the NPM modules e.g. https://github.com/resin-io/etcher/search?q=uglify-es&type=Code&utf8=%E2%9C%93

With an eye towards the upcoming RPM support, maybe there could be one 'npm-requirements.txt' that installs `uglify-es` into both Docker and the CI services, and then a second 'npm-requirements.txt' that installs `etcher-installer-debian` and `etcher-installer-redhat` just into Docker?

Or am I just trying to overcomplicate things? :-S

jviotti:
The requirements.txt node equivalent is the package.json. I suggest to find a way to get them listed there without confusing the shrinkwrap file. Maybe we should try NPM v5?

lurch:
I dunno - you're obviously much more familiar with NodeJS packaging than I am. It was only a suggestion, so if this isn't possible without making the `package.json` / `npm-shrinkwrap.json` more "bodged" than they already are, feel free to ignore this suggestion.

PR #1455, opened 2017-05-18
feat(GUI): add metadata and icon to notifications

Shou:
We add the image filename, its destination drive, and application icon
to the notifications.

See: #1443
Changelog-Entry: Add image name, drive name, and icon to notifications.

### DIFF
--- a/lib/gui/os/notification.js
+++ b/lib/gui/os/notification.js
@@ -29,20 +29,23 @@ const electron = require('electron');
  * http://electron.atom.io/docs/v0.37.5/tutorial/desktop-environment-integration/
  *
  * @param {String} title - notification title
- * @param {String} body - notification body
+ * @param {Object} options - options object
+ * @param {String} options.body - notification body
+ * @param {String} options.icon - supported icon path
  * @returns {Object} HTML5 notification instance
  *
  * @example
- * notification.send('Hello', 'Foo Bar Bar');
+ * notification.send('Hello', {
+ *   body: 'Foo Bar Bar',
+ *   icon: 'icon.png'
+ * });
  */
-exports.send = (title, body) => {
+exports.send = (title, options) => {
 
   // `app.dock` is only defined in OS X
   if (electron.remote.app.dock) {
     electron.remote.app.dock.bounce();
   }
 
-  return new Notification(title, {
-    body
-  });
+  return new Notification(title, options);
 };
--- a/lib/gui/pages/main/controllers/flash.js
+++ b/lib/gui/pages/main/controllers/flash.js
@@ -21,6 +21,7 @@ const settings = require('../../../models/settings');
 const flashState = require('../../../models/flash-state');
 const utils = require('../../../../shared/utils');
 const notification = require('../../../os/notification');
+const path = require('path');
 
 module.exports = function(
   $state,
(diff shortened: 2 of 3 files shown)


lib/gui/os/notification.js
```diff
@@ -30,19 +30,21 @@ const electron = require('electron');
...
  * @param {String} body - notification body
+ * @param {String} icon - supported icon path
  * @returns {Object} HTML5 notification instance
  *
  * @example
- * notification.send('Hello', 'Foo Bar Bar');
+ * notification.send('Hello', 'Foo Bar Bar', 'icon.png');
  */
-exports.send = (title, body) => {
+exports.send = (title, body, icon) => {
```
jviotti: Can we take an options object, which optionally may contain an `icon` property instead?

lib/shared/messages.js
```diff
@@ -32,7 +32,10 @@ module.exports = {
    */
   info: {
 
-    flashComplete: _.template('Your flash is complete!')
+    flashComplete: _.template([
+      '<%= imageBasename %> was written to',
```
jviotti: Maybe add the work "successfully"? What do you think @lurch ?

lib/gui/pages/main/controllers/flash.js
```diff
@@ -61,7 +62,13 @@ module.exports = function(
 
     ImageWriterService.flash(image.path, drive).then(() => {
       if (!flashState.wasLastFlashCancelled()) {
-        notification.send('Success!', messages.info.flashComplete());
+        notification.send('Success!',
+          messages.info.flashComplete({
+            imageBasename: path.basename(image.path),
+            drive
+          }),
+          '../../../../../assets/icon.png'
```
jviotti: Oh my god :P

lib/shared/messages.js
lurch: Could do, but would it look weird if the success window gets too big? Given that it contains the filename of the image and the description of the drive, it might already be quite long?
Mind sharing a screenshot @Shou ? ;)

lib/gui/pages/main/controllers/flash.js
lurch: Still not as bad as your megabackslashosaurus ;-P

lib/gui/os/notification.js
lurch: Or would it be simpler if the `icon` parameter just defaulted to `null`?

jviotti: Three positional arguments seem to much. Two are good, but if we want to add another one, lets make it a property of an object.

lurch: \*shrug\* You know JS better than I do :)

lib/shared/messages.js
Shou: It's very context-sensitive, but here's GNOME 3. As you can see it gets truncated, but you can expand it by hovering over it.
image

Maybe we can utilize the title space?

jviotti: Fair enough, lets leave `successfully` out for now.

jviotti:
I'm approving, but the linter is complaining.

Shou:
Added that missing options parameter JSDoc and squashed the commits, should pass tests now.

PR #1464, opened 2017-05-22
feat(GUI): rewrite svg-icon directive in react

Shou:
We change from using an Angular directive to using React through
react2angular in the SVGIcon module.

Changelog-Entry: Use React instead of Angular for the SVGIcon directive.


Shou:
Most noteworthy is probably that we can no longer set arbitrary attributes directly on the directive element, and instead have to use the surrounding familial elements.

lib/gui/pages/main/styles/_main.scss
```diff
@@ -18,7 +18,16 @@
...
+svg {
+  width: 100%;
+  height: 100%;
+}
+
+div.center-block > svg-icon {
+  display: inline-block;
+}
+
+div[disabled] > svg-icon path {
```
jviotti: A disabled `div` looks weird. Can we somehow move it to `svg-icon`?

lib/gui/pages/main/styles/_main.scss
```diff
@@ -18,7 +18,16 @@
...
 }
 
-.svg-icon[disabled] path {
+svg {
+  width: 100%;
+  height: 100%;
+}
+
+div.center-block > svg-icon {
+  display: inline-block;
```
jviotti: I wonder if `svg-icon` instances should always be `inline-block`

lib/gui/pages/main/styles/_main.scss
Shou: We could implement the disabled property on the React component, but we'd have to do that for every component. To get around it we could make a 'core' component which implements all the 'convenience' functionality we'd expect, and extend any other new classes with that.

lib/gui/pages/main/styles/_main.scss
Shou: Yeah that sounds like a good idea. None of the SVGs are affected by the change in master, so I'd say it's safe.

lib/gui/pages/main/styles/_main.scss
Shou: Alternatively, maybe a `disable` component?

```html
<disable when={ true }>
  <svg-icon ...></svg-icon>
</disable>
```
```css
disable[when] > svg { color: gray; }
```

lib/gui/components/svg-icon/svg-icon.js
```diff
@@ -21,8 +21,94 @@
...
+    if (_.first(this.props.path) === '<') {
+      contents = this.props.path;
+
+    } else {
+
+      // This means the path to the icon should be
+      // relative to *this directory*.
+      // TODO: There might be a way to compute the path
+      // relatively to the `index.html`.
+      const imagePath = path.join(__dirname, this.props.path);
```
jhermsmeier: If you know where `index.html` is – which would be `path.join(__dirname, '../../../')` in this case – you can use [`path.relative(from, to)`](https://nodejs.org/api/path.html#path_path_relative_from_to) here:

```js
const indexPath = path.join(__dirname, '../../../');
const imagePath = path.relative(indexPath, this.props.path);
```

lib/gui/index.html
```diff
@@ -32,18 +32,18 @@
...
       <span class="caption">
         is <span class="caption" os-open-external="https://github.com/resin-io/etcher">an open source project</span> by
       </span>
 
-      <svg-icon path="../../../assets/resin.svg"
-        width="79px"
-        height="23px"
+      <svg-icon path="'../../assets/resin.svg'"
+        width="'79px'"
+        height="'23px'"
```
jhermsmeier: Possibly uninformed question: Why are these single-quoted additionally? Is that a react thing?

Shou: Yeah, React basically treats data in attributes like regular Javascript, so to get strings you have to encapsulate them in quotes, and everything else you'd normally do. Looks odd because it's in quotes already, but this is temporary while we still have Angular in our codebase—once we're pure React we can use JSX and the syntax is a bit cleaner:

```jsx
<Svg-Icon path={ '../../assets/resin.svg' } ...></Svg-Icon>
```

lurch:
```
> etcher@1.0.0 jslint C:\projects\etcher
> eslint lib tests scripts bin versionist.conf.js
C:\projects\etcher\tests\gui\components\svg-icon.spec.js
  21:71  error  '$templateCache' is defined but never used  no-unused-vars
```

lib/gui/index.html
jhermsmeier: Ahh, thanks for the explanation!

jviotti:
Can we change the commit type to "refactor"?

lib/gui/components/svg-icon/svg-icon.js
```diff
@@ -21,8 +21,94 @@
...
+
+  /**
+   * @summary Render the SVG
+   * @returns {react.Element}
+   */
+  render() {
+    let contents = null;
+
+    // The path contains SVG contents
+    if (_.first(this.props.path) === '<') {
```
jviotti: Can we use `_.startsWith`, which better expresses the intention?

jviotti: Also, should we trim the string, and add a test case that passes an SVG string with leading whitespace?

lib/gui/components/svg-icon/svg-icon.js
jviotti: I don't think we should hardcore the path to the main HTML file from this directive, since this component shouldn't care about that. I agree that we can improve the situation, but looks like it needs to be done when passing the path to this component, rather than on this component itself.

lib/gui/components/svg-icon/svg-icon.js
```diff
@@ -21,8 +21,94 @@
...
+      // TODO: There might be a way to compute the path
+      // relatively to the `index.html`.
+      const imagePath = path.join(__dirname, this.props.path);
+
+      contents = fs.readFileSync(imagePath, {
+        encoding: 'utf8'
+      });
+    }
+
+    const width = this.props.width || '40px';
```
jviotti: Can we move 40px to a constant at the top of this file?

lib/gui/pages/main/styles/_main.scss
jviotti: Hm, what is the common way to solve this problem on React? Is manually implementing a disable property the best practice? Can use some sort of mixin functionality to avoid repeating ourselves?

lib/gui/components/svg-icon/svg-icon.js
jviotti: BTW, do we have a use case for inlining SVG content? I like the idea, but I wonder if we're building something that is not immediately needed.

lurch: > Also, should we trim the string, and add a test case that passes an SVG string with leading whitespace?

I guess that depends if we're likely to get SVGs with leading whitespace?

lib/gui/components/svg-icon/svg-icon.js
lurch: Out of curiosity, what determines this 40px 'default value'?

lib/gui/components/svg-icon/svg-icon.js
jviotti: Sounds like an easy thing to cover, so I guess no reason not do, just in case :)

lurch: ...but if an SVG with leading whitespace is technically malformed (I dunno, I've never looked at the standard), then we should probably reject it, rather than trying to "massage" it back into a valid SVG?

jviotti: To the best of my knowledge, leading/trailing white-space doesn't affect the validity of SVG elements.

lib/gui/components/svg-icon.js
```diff
@@ -106,7 +106,12 @@ SVGIcon.propTypes = {
   /**
    * @summary SVG image height unit
    */
-  height: propTypes.string
+  height: propTypes.string,
+
+  /**
+   * @summary Should the element be in a disabled state?
+   */
+  disabled: propTypes.bool
```
lurch: Out of curiosity, what does it mean for an SVG to be 'disabled'?

Shou: Here it means the SVG icon will be grayed out.

lurch: Ahh. Perhaps that should be mentioned in the docstring? ;)

Shou: Added that now

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -15,18 +15,18 @@ <h3 class="title"><span class="tick tick--success space-right-tiny"></span> Flas
     <div class="box center">
       <div class="fallback-banner">
         <div class="caption caption-big">Thanks for using
-          <svg-icon path="'../../assets/etcher.svg'"
+          <svg-icon path="'../assets/etcher.svg'"
```
lurch: Hmmm, neither this html file nor the svg files it points to have moved, so why has the relative path changed here? Was it incorrect before?

lurch:
The CI tests are saying that `tests/gui/components/svg-icon.spec.js` needs to be updated too.

lib/gui/pages/main/styles/_main.scss
```diff
@@ -18,7 +18,16 @@
   margin-top: 75px;
 }
 
-.svg-icon[disabled] path {
+svg {
```
jviotti: I feel this should be part of `svg-icon`'s internal styles.

jviotti:
Works great on my system! Waiting to hear back from @Shou regarding my little SVG CSS rule comment, and I'll approve afterwards.

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -15,20 +15,20 @@ <h3 class="title"><span class="tick tick--success space-right-tiny"></span> Flas
     <div class="box center">
       <div class="fallback-banner">
         <div class="caption caption-big">Thanks for using
-          <svg-icon path="../../../assets/etcher.svg"
-            width="150px"
-            height="auto"
+          <svg-icon path="'../assets/etcher.svg'"
```
lurch: Duplicate comment, because my earlier comment got hidden:

AFAICT neither this html file nor the svg files it points to have moved, so why have all the relative paths changed here? Were they 'incorrect' before?

Shou: It's relative to the SVGIcon module which did move, it's a bit of a mess but it should be resolved once we start using JSX and transpiling the source into a single bundled file and everything is relative to that, unless there are any more immediate fixes we could apply. It has bit me recently in the notification issue, where there is no error emitted at wrongful paths.

lib/gui/components/svg-icon/svg-icon.js
Shou: It's functionality that was inherited from the Angular svg-icon directive, but we're not making use of inlining SVG contents, so we could remove it. I can't really think of a use-case where it's absolutely necessary—if we want to dynamically generate SVGs it might be better to just allow passing elements.

lib/gui/pages/finish/templates/success.tpl.html
lurch: > It's relative to the SVGIcon module which did move

Euw, that sounds nasty :-(

> it's a bit of a mess but it should be resolved once we start using JSX

Okay, glad to hear that it's going to be fixed in future :-)

Shou: We could check and emit our own error before reaching `new Notification(...)`. Another idea is to expose the path and add a test that checks if it exists.

jviotti: Yeah, let's so a small check inside `svg-icon` to make it throw if the file is not accessible.

lib/gui/components/svg-icon/svg-icon.js
jviotti: Yeah, let's remove it, and we can always put it back later if we find a use case for it.

lib/gui/pages/finish/templates/success.tpl.html
lurch: Does the GUI get "exercised enough" by the unit tests to ensure that these svg-icon errors would be caught by the CI servers, rather than remaining hidden until an actual user first runs the application?

Shou: Actually it doesn't affect this module luckily, so I'll address this in a separate PR for the notification module which is affected.

jviotti: @lurch That's where my integration testing suggestion comes in (https://github.com/electron/spectron) :)

lurch: > Actually it doesn't affect this module luckily

Huh? Surely svg-icon will still fail if the paths to the SVGs in the templates are incorrect? Which is an error we should attempt to detect?

Shou: Oh yeah, but it's easy to detect those errors since they pop up as very unavoidable alert boxes. It's only in the notification module, so far, that I've noticed silent acceptance of wrong paths.

lib/gui/pages/main/styles/_main.scss
jviotti: Ping @Shou

Shou: Done!

jviotti:
It all looks great, but the icons are not getting automatically highlighted anymore:

screenshot 2017-06-07 11 21 15

Shou:
I believe this is due to the fact that it's React within Angular, so the `disabled` property is kept persistent on the directive/component element. I could be wrong, but according to React the property should work as intended, i.e. disappear when it's set to `false`, and that doesn't happen here. For now I'll pass the property forward to the `div` wrapper around the contained `svg`. It's not exactly ideal but it should work properly when we're only using React.

lib/gui/pages/finish/templates/success.tpl.html
```diff
@@ -15,20 +15,20 @@ <h3 class="title"><span class="tick tick--success space-right-tiny"></span> Flas
...
+            width="'150px'"
+            height="'auto'"
             os-open-external="https://etcher.io?ref=etcher_offline_banner"></svg-icon>
```
jviotti: In the other icon instances, you're using `os-open-external` in a `span` that wraps the icon, but here you're using it directly. What is the correct what of doing so? Can we always add it directly to the `svg-icon` element?

jviotti:
@Shou I just noticed that the footer icons are bit off after these changes. Maybe some unwanted CSS change slipped in?

Before:

screenshot 2017-06-08 10 21 34

After:

screenshot 2017-06-08 10 21 46

lib/gui/pages/finish/templates/success.tpl.html
lurch: Well spotted. From a conceptual point of view, I think it might make sense to have the `os-open-external` on an enclosing `span` or `div`; a bit like the way in HTML you have the `href` attribute on the `A` tag and not on the `IMG` tag.
(but then perhaps react / angular has a different model \*shrug\*)

jviotti: Yeah, I'm really fine with either.

Shou: I (wrongly) assumed it didn't work and did some preliminary work on getting them out of the directive, but it turns out only 'native' properties are removed; `disabled` specifically here. But what @lurch said is something I had in mind when doing it as well so I should complete it.

Shou:
Thanks for noticing, I'll look into that!

Shou:
Rebased and fixed.

lib/gui/components/svg-icon/styles/_svg-icon.scss
```diff
@@ -0,0 +1,13 @@
+
+svg {
+  width: 100%;
+  height: 100%;
+}
+
+svg-icon {
+  display: inline-block;
+
+  > div[disabled] path {
```
jviotti: Can we move this particular rule to the `main.scss` file? This component doesn't control what its non-disabled color is, so I don't think it knows enough about himself to pick a proper disabled color.

Shou: Sure!

Issue #1437, opened 2017-05-16
Word-wrap properly in the drive-selection dialog


lurch:
As pointed out in [gitter](https://gitter.im/resin-io/etcher?at=591a1dcad1a7716a0accc62f) it seems that sometimes the line-wrapping breaks in the wrong place :-(
![image](https://cloud.githubusercontent.com/assets/476186/26100944/4b5437c0-3a27-11e7-801b-4c6245be8e96.png)

In the screenshot above, `15.9` shouldn't be split over two lines.

Shou:
So we need `word-break: break-all` in case of long names, but what we could do here is wrap the size in its own span element and apply `word-break: keep-all` there, so only the whole span will wrap over.

PR #1467, opened 2017-05-25
fix(GUI): don't break up size number in drive selector

Shou:
We make the size number in the drive selector stay whole through
the `word-break: keep-all` CSS property, ensuring that it doesn't
partially overflow to the next line.

See: https://github.com/resin-io/etcher/issues/1437
Changelog-Entry: Don't break up size number in drive selector.


lurch:
LGTM, but requested a review from Juanchi as he knows CSS better than me :)

jviotti:
Looks and works great, but it needs to be rebased because of that Ubuntu APT issue.

Shou:
Rebased

PR #1468, opened 2017-05-25
fix(GUI): stop settings overflow into footer

Shou:
By removing the 'advanced' sub-header we stop the settings from
overflowing into the footer.

See: https://github.com/resin-io/etcher/issues/1383
Changelog-Entry: Stop settings from overflowing into the footer.

### DIFF
--- a/lib/gui/pages/settings/templates/settings.tpl.html
+++ b/lib/gui/pages/settings/templates/settings.tpl.html
@@ -50,8 +50,6 @@
     </label>
   </div>
 
-  <h3 class="subtitle">Advanced</h3>
-
   <div class="checkbox">
     <label>
       <input type="checkbox"
--- a/lib/gui/css/main.css
+++ b/lib/gui/css/main.css
@@ -6567,11 +6567,6 @@ body {
 .page-settings .title {
   color: #fff; }
 
-.page-settings .subtitle {
-  color: #fff;
-  margin-top: 20px;
-  margin-bottom: 15px; }
-
 /*
  * Copyright 2016 resin.io
  *
--- a/lib/gui/pages/settings/styles/_settings.scss
+++ b/lib/gui/pages/settings/styles/_settings.scss
@@ -25,9 +25,3 @@
 .page-settings .title {
   color: $palette-theme-dark-foreground;
 }
-
-.page-settings .subtitle {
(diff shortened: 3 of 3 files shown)


lurch:
Could we have a screenshot, and then approval from @konmouz ?

Shou:
Sure! Here's for comparison.

image
image

lurch:
Looks like the Travis x86 builds are currently failing because the apt mirror-server that the Ubuntu 12.04 x86 Docker image uses ( http://ubuntu.stu.edu.tw/ ) is currently down :-/
(with any luck this will only be a temporary glitch!)

konmouz:
@Shou, can we move the back button to the left? The rest looks alright.

Shou:
Yeah that should be an easy change, but then it wouldn't be in the same spot as the settings cog so users would have to travel from right to left to go back. Is that intended?

lurch:
I agree with @Shou here.

konmouz:
One of the reasons to move to a modal window was to avoid this dilemma. Still, having the back button on the left is a standard practice. Currently it is quite weird.

lib/gui/pages/settings/templates/settings.tpl.html
```diff
@@ -50,8 +50,6 @@ <h1 class="title space-bottom-large">Settings</h1>
     </label>
   </div>
 
-  <h3 class="subtitle">Advanced</h3>
```
jviotti: I'm on mobile, but I think this class can now be deleted. Can you double check?

Shou: Done!

lurch:
One of the Appveyor builds failed again with:
```
./scripts/build/electron-create-resources-app.sh -s . -o release/electron-win32-1.0.0+619ac40-x86-app \
		-v 1.0.0+619ac40 \
		-f "lib,assets"
cp -RLf release/electron-win32-x86-dependencies/* release/electron-win32-1.0.0+619ac40-x86-app
cp: cannot copy cyclic symbolic link `release/electron-win32-x86-dependencies/node_modules/es5-ext/test/array/of'
make: *** [release/electron-win32-1.0.0+619ac40-x86-app] Error 1
```

Just restarted it for you @Shou :)

PR #1472, opened 2017-05-25
feat(GUI): change success page fallback message

Shou:
We change the success page's fallback message from "Thank you for using
Etcher" to "We hope you enjoyed using Etcher".

Changelog-Entry: Change success page fallback message.

### DIFF
--- a/lib/gui/pages/finish/templates/success.tpl.html
+++ b/lib/gui/pages/finish/templates/success.tpl.html
@@ -14,7 +14,7 @@
 
     <div class="box center">
       <div class="fallback-banner">
-        <div class="caption caption-big">Thanks for using
+        <div class="caption caption-big">We hope you enjoyed using
           <svg-icon path="../../../assets/etcher.svg"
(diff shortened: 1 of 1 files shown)


Shou:
@konmouz just to check, this is the message we wanted right?

konmouz:
@Shou, I think so. But it would be good to double check with @alexandrosm

jviotti:
Didn't we decide that this new message implied that were thankful more than the other way around?

lurch:
@Shou Could you update the commit-message? According to the diff, the message used to be "Thanks for using..." rather than "Thank you for using..."
Also, this should be `style` rather than `feat`.
And you'll need to rebase onto latest `master` to get the CI tests passing.

jviotti:
Ping @konmouz @Shou @alexandrosm. Is this change something we discussed somewhere?

Shou:
We are no longer using the banner where this is relevant, closing this.

Issue #1465, opened 2017-05-24
Etcher is still sending some usage/tracking data though disabled


schlamar:
- **Etcher version:** 1.0.0
- **Operating system and architecture:** Win 7, 64bit

I have statistic sending disabled in settings but etcher is still sending some usage/tracking data via HTTP (`http://api.mixpanel.com/track/?data=xxx`).

jviotti:
@schlamar That's weird, thanks for reporting. Can you paste the logs from DevTools (Ctrl+Alt+I), so we can see at which part the events are being sent?

schlamar:
Looks like every event is sent, e.g. navigating to settings and back.

    C:\Apps\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 document not ready yet, trying again in 500 milliseconds...
    C:\Apps\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 Download the React DevTools and use an HTTP server (instead of a file: URL) for a better development experience: https://fb.me/react-devtools
    C:\Apps\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51  _____ _       _
    |  ___| |     | |
    | |__ | |_ ___| |__   ___ _ __
    |  __|| __/ __| '_ \ / _ \ '__|
    | |___| || (__| | | |  __/ |
    \____/ \__\___|_| |_|\___|_|

    Interested in joining the Etcher team?
    Drop us a line at join+etcher@resin.io
    C:\Apps\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 Wed May 24 2017 15:29:56 GMT+0200 (Mitteleuropäische Sommerzeit) Application start
    C:\Apps\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 Wed May 24 2017 15:29:56 GMT+0200 (Mitteleuropäische Sommerzeit) Checking for updates ({"currentVersion":"1.0.0","releaseType":"PRODUCTION","updateSemverRange":"<2.0.0","includeUnstableChannel":false})
    C:\Apps\Etcher\resources\app.asar\node_modules\raven-js\dist\raven.js:51 {"cause":{"code":"UNABLE_TO_VERIFY_LEAF_SIGNATURE"},"isOperational":true,"code":"UNABLE_TO_VERIFY_LEAF_SIGNATURE"}
[...]

PR #1473, opened 2017-05-26
fix(GUI): ignore UNABLE_TO_VERIFY_LEAF_SIGNATURE HTTP issues

jviotti:
This error may get thrown when fetching the list of S3 packages when the
user is behind a proxy tht causes the SSL certificate to incorrectly not
be trusted.

See: https://github.com/atom/apm/issues/103
See: https://github.com/resin-io/etcher/issues/1465
Change-Type: patch
Changelog-Entry: Fix `UNABLE_TO_VERIFY_LEAF_SIGNATURE` error at startup.

### DIFF
--- a/lib/shared/s3-packages.js
+++ b/lib/shared/s3-packages.js
@@ -156,6 +156,11 @@ exports.getRemoteVersions = _.memoize((bucketUrl) => {
       code: 'EACCES'
     }, {
       code: 'ETIMEDOUT'
+    }, {
+
+      // May happen when behind a proxy
+      code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
+
     }, () => {
       return [];
     });
--- a/tests/shared/s3-packages.spec.js
+++ b/tests/shared/s3-packages.spec.js
@@ -712,6 +712,29 @@ describe('Shared: s3Packages', function() {
 
     });
 
+    describe('given UNABLE_TO_VERIFY_LEAF_SIGNATURE', function() {
+
+      beforeEach(function() {
+        const error = new Error('UNABLE_TO_VERIFY_LEAF_SIGNATURE');
+        error.code = 'UNABLE_TO_VERIFY_LEAF_SIGNATURE';
+
+        this.requestGetAsyncStub = m.sinon.stub(request, 'getAsync');
+        this.requestGetAsyncStub.returns(Bluebird.reject(error));
+      });
+
+      afterEach(function() {
+        this.requestGetAsyncStub.restore();
+      });
+
+      it('should resolve an empty array', function(done) {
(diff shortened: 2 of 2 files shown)


jviotti:
I'm getting 404 on that page, but would that solve the issue anyways? Correct me if I'm wrong, but the issue is that the user is behind a proxy, which modifies the certificate chain of the remote certificate such that the proxy servers acts like the internal certificate authority.

Even if we do have the common root certs, we'd need the proxy certificate to workaround the issue.

This does mean users behind such an "unconfigured" proxy will be unable to receive update notifications though, but what options do we have?

jhermsmeier:
Oh, messed up a char in the link somehow, here's the correct one: https://git.daplie.com/Daplie/node-ssl-root-cas

> Even if we do have the common root certs, we'd need the proxy certificate to workaround the issue.
> This does mean users behind such an "unconfigured" proxy will be unable to receive update notifications though, but what options do we have?

I think usually the proxies have some cert with a root CA from that package (thinking about it, Electron should actually already have all of those since it's Chromium), but in case it's completely unconfigured, or uses a self-signed one, we're out of luck I believe. Unless we'd do non-strict TLS just for the notifications, which I don't think is a good idea.

lurch:
Like the one described in https://github.com/atom/apm/issues/103 should we maybe have a (manual) config option for those users that are behind such a proxy who want to turn off strict SSL?
Will this problem also affect the success-banner page?

jviotti:
Looks like we can add a setting to disable strict SSL and use that to set the `strictSSL` `request` option. We also can also integrate https://git.daplie.com/Daplie/node-ssl-root-cas. Which one should we do? I'm happy to give this root CAs module a go before falling back to disabling SSL, but should we always use it, or only some cases? If so, in which?

lurch:
Sounds like @jhermsmeier knows more about SSL than me, so I'll defer to his knowledge...

And I was suggesting adding an option for a user to explicitly set (if they're behind one of these proxies), not always automatically disabling strictSSL. (apologies if I misunderstood your comment @jviotti )

jviotti:
Yeah, I meant adding a setting that when enabled, it will set `strictSSL` to false. Let's wait for @jhermsmeier 's opinion.

jhermsmeier:
I think it'd be best to only include the other root CAs; with that we'd at least lower the number of proxies where SSL fails. But I don't see update notifications as important enough to disable `strictSSL` – and considering that you've already been MITMed on the S3 bucket @jviotti, having the option to disable it is probably a really bad idea.

lurch:
> I think it'd be best to only include the other root CAs

Before doing that, we should verify that they aren't automatically included in all Electron apps anyway? Earlier you said `... Electron should actually already have all of those ...`.

> But I don't see update notifications as important enough to disable strictSSL

As I already asked, will this also affect the success-banner page ?

PR #1475, opened 2017-05-26
chore(package): Make shrinkwrap npm@5 compatible

jhermsmeier:
This updates the `npm-shrinkwrap.json` to work with `npm@5`
and removes the `clean-shrinkwrap` script, as it interferes
with how npm works now, and isn't needed anymore.

Change-Type: patch


npm-shrinkwrap.json
```diff
@@ -1,7395 +1,7258 @@
 {
   "name": "etcher",
   "version": "1.0.0",
+  "lockfileVersion": 1,
```
jviotti: Can we start using the new `package-lock.json` convention?

npm-shrinkwrap.json
```diff
@@ -1,7395 +1,7258 @@
...
     "angular-ui-bootstrap": {
       "version": "2.5.0",
-      "from": "angular-ui-bootstrap@>=2.5.0 <2.6.0",
-      "resolved": "https://registry.npmjs.org/angular-ui-bootstrap/-/angular-ui-bootstrap-2.5.0.tgz"
+      "resolved": "https://registry.npmjs.org/angular-ui-bootstrap/-/angular-ui-bootstrap-2.5.0.tgz",
+      "integrity": "sha1-L6zvuRU4ZlXcX0QVAlgDJRcjbQE="
     },
     "angular-ui-router": {
       "version": "0.4.2",
-      "from": "angular-ui-router@>=0.4.2 <0.5.0",
```
jviotti: Nice to see the `from` parameters going away.

.travis.yml
```diff
@@ -41,6 +41,7 @@ before_install:
     fi
 
 install:
+  - npm install --global npm@5
```
jviotti: Can we only perform this step on OS X builds, so we don't do it twice on GNU/Linux?

npm-shrinkwrap.json
jhermsmeier: That's already the new `package-lock.json` – it just gets renamed to `npm-shrinkwrap.json` if there's already one.

.travis.yml
jhermsmeier: Where do we install npm the second time on Linux?

jviotti: On the Dockerfiles

npm-shrinkwrap.json
jviotti: Sure, I was wondering if we can rename `npm-shrinkwrap.json` to `package-lock.json`, so we always interact with that new filename convention instead.

jhermsmeier: Oh, I see – sure, renamed it back :)

.travis.yml
```diff
@@ -39,6 +39,7 @@ before_install:
     else
       export HOST_OS="$TRAVIS_OS_NAME";
     fi
+  - npm cache clean && npm install --global npm@5
```
jviotti: Does this mean we start from scratch each time?

jhermsmeier: This is more of a debug commit where this is coming from – I'm trying to see if there's an issue with the old cache, because there are really weird post-install script errors from some step in `build/ci/install.sh`

jhermsmeier: And it looks like npm@3 isn't installing npm@5 properly (?!)

jhermsmeier:
Closing this, as npm@5 is not ready yet

npm-shrinkwrap.json
lurch: This is what caused the earlier confusion with our ensure-staged-shrinkwrap script telling us that the shrinkwrap file had changed, even though we had the identical module versions installed.

emirotin:
Hey @jhermsmeier what specific npm5 problems did you have?

lurch:
@emirotin see also #1487

PR #1471, opened 2017-05-25
chore(docker): Change APT mirror used by the Ubuntu12.04 32bit docker image

lurch:
Previously it was using http://ubuntu.stu.edu.tw but this host is no longer
active, which was causing all the 32bit Linux TravisCI builds to fail.

PR #1481, opened 2017-05-29
chore: make use of resin.io's Docker base images

jviotti:
Resin.io already publishes multi-arch Docker base images running Jessie,
which is Debian version we wanted before falling back to Ubuntu 12.04.

The main reason behind this change is to be able to use other resin.io
base images (like ARM ones) without requiring a lot of per-base-image
changes.

### DIFF
--- a/scripts/build/docker/Dockerfile.template
+++ b/scripts/build/docker/Dockerfile.template
@@ -1,19 +1,15 @@
 FROM <%= image %>
 
 # Setup APT sources
-<% if (architecture == 'i686') { %>
-RUN sed s,ubuntu\.stu\.edu\.tw,archive.ubuntu.com, /etc/apt/sources.list > /tmp/sources.list \
-  && mv /tmp/sources.list /etc/apt/sources.list
-<% } %>
-<% if (architecture == 'x86_64') { %>
-RUN echo "deb http://archive.ubuntu.com/ubuntu precise-backports main restricted universe multiverse" >> /etc/apt/sources.list
-<% } %>
+RUN echo "deb http://ftp.debian.org/debian jessie-backports main" >> /etc/apt/sources.list
 
 # Install dependencies
 RUN apt-get update \
   && apt-get install -y \
     build-essential \
     curl \
+    dpkg \
+    fakeroot \
     fuse \
     git \
     jq \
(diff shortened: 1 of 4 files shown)


lurch:
Hmm, but doesn't that mean that we're now back to using the Debian Jessie version of glibc (2.19) rather than the Ubuntu 12.04 version of glibc (2.15) that #1364 introduced?

scripts/build/docker/Dockerfile.template
```diff
@@ -1,19 +1,15 @@
...
 
 # Setup APT sources
-<% if (architecture == 'i686') { %>
-RUN sed s,ubuntu\.stu\.edu\.tw,archive.ubuntu.com, /etc/apt/sources.list > /tmp/sources.list \
-  && mv /tmp/sources.list /etc/apt/sources.list
-<% } %>
-<% if (architecture == 'x86_64') { %>
-RUN echo "deb http://archive.ubuntu.com/ubuntu precise-backports main restricted universe multiverse" >> /etc/apt/sources.list
-<% } %>
+RUN echo "deb http://ftp.debian.org/debian jessie-backports main" >> /etc/apt/sources.list
```
lurch: LOL, I guess #1471 was a pretty short-lived fix ;-)

jviotti:
Hm, you're right. I thought Jessie was the one we wanted. I'll send another PR to use wheezy instead.

lurch:
I think wheezy is too old (?). That's why we were using Ubuntu 12.04 base images... as that's the earliest version that Electron supports.

jviotti:
Looks so. I'm really struggling to get it to compile, plus I get weird segmentation faults out of `node-sass`. I'll try for a bit more, and if I can't make any progress, I'll revert this commit and we'll only use resin.io Docker images for ARM targets.

jviotti:
I see. Let's do that then. I think @nghiant2710 is the man for this task. 

So to summarize:

- Let's get custom resin.io Debian/Ubuntu base images with glibc 2.15
- I'll revert this commit since building on the wheezy images is getting very tricky

alexandrosm:
@jviotti -- can you bring this issue to an arch call to discuss? it's a
tricky problem, and I wonder if all our assumptions are correct.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

PR #1484, opened 2017-06-01
chore: list all needed packages on the Dockerfile

jviotti:
At the moment, we only install the packages that we need but that are
missing from the base Docker image that we use.

For safety purposes (and to make it easier to switch to another base
image), we list all the needed packages, even if they usually come
pre-installed, since that would only be a no-op in the worst case.

See: https://github.com/resin-io/etcher/commit/296a55463714194d15c28eb79bb2939908f8cffd#commitcomment-22358869

### DIFF
--- a/scripts/build/docker/Dockerfile-i686
+++ b/scripts/build/docker/Dockerfile-i686
@@ -12,6 +12,8 @@ RUN apt-get update \
   && apt-get install -y \
     build-essential \
     curl \
+    dpkg \
+    fakeroot \
     fuse \
     git \
     jq \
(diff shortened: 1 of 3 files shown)


lurch:
Linux CI builds are failing because `libsass0` isn't available on Ubuntu. I think if you remove that one package then the Linux docker builds should work again.

lurch:
@alexandrosm is suggesting we don't do this, so I'll let you close this @jviotti

jviotti:
Sounds good, I don't have any strong opinions, so let's close it!

PR #1487, opened 2017-06-02
chore(ci): Use npm@5 on CI services

jhermsmeier:
**Changes:**
- Update all CI services to use Node@8 and npm@5
- Remove preshrinkwrap script (now obsolete)
- Move `npm-shrinkwrap.json` -> `package-lock.json`

**Pending issues:**
- [x] Ubuntu 12.04 docker base image on Linux unsupported by Node@8 package
  - Problem solved by using [tj/n](https://github.com/tj/n) to install Node
- [x] npm@5: Installing npm@latest fails in Docker – https://github.com/npm/npm/issues/16807
  - Problem solved by using [tj/n](https://github.com/tj/n) to install Node
- [x] npm@5: Incorrect unmet peer dependency error – https://github.com/npm/npm/issues/16925
  - [ ] Await npm@5.0.4 **OR**
  - [x] Use a patched version of `html-angular-validate` to circumvent this
- [x] Node ABI mismatch for node-sass in sanity-checks – https://github.com/sass/node-sass/pull/1999
  - [ ] Await new version of `node-sass` with the above fix to be published **OR**
  - [x] Point to a known green commit on `node-sass` master

Connects-To: #1528

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 sudo: false
 node_js:
-  - "6.10.3"
+  - "8.1.0"
 
 services:
   - docker
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@ BUILD_OUTPUT_DIRECTORY = $(BUILD_DIRECTORY)/out
 # Application configuration
 # ---------------------------------------------------------------------
 
-ELECTRON_VERSION = $(shell jq -r '.devDependencies["electron"]' package.json)
-NODE_VERSION = 6.1.0
+ELECTRON_VERSION = $(shell $(NPX) electron --version)
+NODE_VERSION = $(shell node --version)
 COMPANY_NAME = Resinio Ltd
 APPLICATION_NAME = $(shell jq -r '.build.productName' package.json)
 APPLICATION_DESCRIPTION = $(shell jq -r '.description' package.json)
@@ -542,7 +542,9 @@ electron-develop:
 		-s "$(TARGET_PLATFORM)"
 
 sass:
-	$(NPX) node-sass lib/gui/scss/main.scss > lib/gui/css/main.css
+	# NOTE: Since `node-sass` is a native addon, and we install all dependencies for the Electron runtime,
+	# we need to run this through Electron, so that the native module's ABI matches properly
+	$(NPX) ELECTRON_RUN_AS_NODE=1 electron $(which node-sass) lib/gui/scss/main.scss > lib/gui/css/main.css
 
 lint-js:
 	$(NPX) eslint lib tests scripts bin versionist.conf.js
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ BUILD_OUTPUT_DIRECTORY = $(BUILD_DIRECTORY)/out
 # Application configuration
 # ---------------------------------------------------------------------
 
-ELECTRON_VERSION = $(shell $(NPX) electron --version)
+ELECTRON_VERSION = $(shell jq -r '.devDependencies["electron"]' package.json)
 NODE_VERSION = $(shell node --version)
 COMPANY_NAME = Resinio Ltd
 APPLICATION_NAME = $(shell jq -r '.build.productName' package.json)
--- a/Makefile
+++ b/Makefile
@@ -544,7 +544,7 @@ electron-develop:
 sass:
 	# NOTE: Since `node-sass` is a native addon, and we install all dependencies for the Electron runtime,
 	# we need to run this through Electron, so that the native module's ABI matches properly
-	$(NPX) ELECTRON_RUN_AS_NODE=1 electron $(which node-sass) lib/gui/scss/main.scss > lib/gui/css/main.css
+	ELECTRON_RUN_AS_NODE=1 $(NPX) electron node_modules/.bin/node-sass lib/gui/scss/main.scss > lib/gui/css/main.css
 
 lint-js:
 	$(NPX) eslint lib tests scripts bin versionist.conf.js
--- a/Makefile
+++ b/Makefile
@@ -542,9 +542,7 @@ electron-develop:
 		-s "$(TARGET_PLATFORM)"
 
 sass:
-	# NOTE: Since `node-sass` is a native addon, and we install all dependencies for the Electron runtime,
-	# we need to run this through Electron, so that the native module's ABI matches properly
-	ELECTRON_RUN_AS_NODE=1 $(NPX) electron node_modules/.bin/node-sass lib/gui/scss/main.scss > lib/gui/css/main.css
+	node-sass lib/gui/scss/main.scss > lib/gui/css/main.css
 
 lint-js:
 	$(NPX) eslint lib tests scripts bin versionist.conf.js
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4550,9 +4550,7 @@
       "integrity": "sha1-IjvUMa1/suQMVDT4peRVi+JWj70="
     },
     "node-sass": {
-      "version": "4.5.3",
-      "resolved": "https://registry.npmjs.org/node-sass/-/node-sass-4.5.3.tgz",
-      "integrity": "sha1-0JydEXlkEjnRuX/8YjH9zsU+FWg=",
+      "version": "github:sass/node-sass#61d7e1c1abf762aad22bb2a6b3fdac884ca369bd",
       "dev": true,
       "dependencies": {
         "cross-spawn": {
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3500,258 +3500,320 @@
       "integrity": "sha512-cM+1OnKRacBh98mw5RJ3PEmFxhl70lDFAL50+x/a+fxzSZryBF1asADf57zliAtGhCNVHYzrJAZX+8R77AGYGg==",
       "dependencies": {
         "abbrev": {
-          "version": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
+          "version": "1.0.9",
+          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
           "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU="
         },
         "ajv": {
-          "version": "https://registry.npmjs.org/ajv/-/ajv-4.11.5.tgz",
+          "version": "4.11.5",
+          "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.5.tgz",
           "integrity": "sha1-tu50ZXuZOgHc5Et5RNVvSFgo1b0=",
           "dependencies": {
             "co": {
-              "version": "https://registry.npmjs.org/co/-/co-4.6.0.tgz",
+              "version": "4.6.0",
+              "resolved": "https://registry.npmjs.org/co/-/co-4.6.0.tgz",
               "integrity": "sha1-bqa989hTrlTMuOR7+gvz+QMfsYQ="
             }
           }
         },
         "ansi-regex": {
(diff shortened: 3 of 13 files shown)


lurch:
Regarding the `Invalid tag name` CI errors, I found https://github.com/npm/npm/issues/17049

 * Would the Appveyor issue be fixed by changing
```
$(NPX) ELECTRON_RUN_AS_NODE=1 electron $(which node-sass) lib/gui/scss/main.scss > lib/gui/css/main.css
```
to
```
ELECTRON_RUN_AS_NODE=1 $(NPX) electron $(which node-sass) lib/gui/scss/main.scss > lib/gui/css/main.css
```
(in `Makefile`) ?
Although perhaps it would make more sense to install `node-sass` in the same way as we install `uglify-es` ?  IMHO relying on `$(which node-sass)` looks a bit ugly.

 * Looks like the Travis issue is somehow coming from `scripts/build/dependencies-npm.sh` ?

package.json
```diff
@@ -90,7 +90,7 @@
     "flexboxgrid": "^6.3.0",
     "immutable": "^3.8.1",
     "lodash": "^4.5.1",
-    "lzma-native": "^1.5.2",
+    "lzma-native": "^2.0.2",
```
lurch: IIRC this updated version of `lzma-native` also allows us to use a newer version of `node-gyp` (or whatever the relevant package is called - hopefully you know what I mean), rather than locking ourselves to an older version?

Perhaps this change is worth doing anyway in a separate PR against the `master` branch?

npm-shrinkwrap.json
```diff
@@ -3500,258 +3500,320 @@
       "integrity": "sha512-cM+1OnKRacBh98mw5RJ3PEmFxhl70lDFAL50+x/a+fxzSZryBF1asADf57zliAtGhCNVHYzrJAZX+8R77AGYGg==",
       "dependencies": {
         "abbrev": {
-          "version": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
+          "version": "1.0.9",
+          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
```
lurch: Just out of curiosity, how did these "faulty version entries" get in here, and what was required to "fix" them?

jhermsmeier: They got in there through running `npm shrinkwrap`, and I mended them by hand. Yup. Sublime Text multiline editing for the win. Running `npm shrinkwrap` again screws them up again. Haven't found out yet if this is `npm`s fault, or if we're doing something in the pre/post-shrinkwrap scripts that causes this.

package.json
jhermsmeier: > also allows us to use a newer version of `node-gyp`

We don't need to depend on `node-gyp` at all – it ships with npm. Maybe you mean somthing else (bindings, nan)?

> Perhaps this change is worth doing anyway in a separate PR against the master branch?

Yeah, I'm just sorting out all the things we'll need to get this npm@5 thing greeeeeeeen.

npm-shrinkwrap.json
lurch: > I mended them by hand. Yup.

Ouch! :confused: 

> Haven't found out yet if this is npms fault,

If it is, I guess a small reproducible example would be useful.

> or if we're doing something in the pre/post-shrinkwrap scripts that causes this.

Would including the changes from #1514 help with this, or is that something else entirely?

scripts/build/docker/Dockerfile.template
```diff
@@ -31,24 +31,32 @@ RUN apt-get update \
...
   && apt-get update && apt-get install -y gcc-4.8 g++-4.8 \
   && update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 50
 
-# NodeJS
-RUN curl -sL https://deb.nodesource.com/setup_6.x | bash - \
-  && apt-get install -y nodejs
+# Install Node@8.x.x
+RUN git clone https://github.com/tj/n && \
+  cd n && git checkout v2.1.4 && make install && \
+  cd .. && n 8.1.0
```
jhermsmeier: It's real speedy. I'd even argue that it's the fastest way to auto-install node. Also a most splendid node version manager.

package.json
lurch: Nope, looks like I was right - https://github.com/addaleax/lzma-native/issues/30#issuecomment-287521987 and https://github.com/resin-io/etcher/pull/1191

scripts/build/docker/Dockerfile.template
lurch: ...but with a name that's impossible to search for? ;-)

Makefile
```diff
@@ -21,7 +21,7 @@ BUILD_OUTPUT_DIRECTORY = $(BUILD_DIRECTORY)/out
 # ---------------------------------------------------------------------
 
 ELECTRON_VERSION = $(shell jq -r '.devDependencies["electron"]' package.json)
-NODE_VERSION = 6.1.0
+NODE_VERSION = $(shell node --version)
```
jhermsmeier: The node version the builds target is the `ELECTRON_VERSION` (otherwise the native modules wouldn't work in Electron), unless the CLI is the target, in which case it's `NODE_VERSION`, and when the CLI is bundled and this version differs from the node that it's being bundled with then the native modules also won't work – hence this gets the actually installed node version now, to reduce places we have to manually change the node version.
I might just be totally wrong about this though – @jviotti will know for sure though.

scripts/build/docker/Dockerfile.template
lurch: Do you happen to know if this supports NodeJS on each of Linux/x86, Linux/x64, Linux/armv7hf, Darwin/x64, Windows/x86 and Windows/x64?

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 sudo: false
 node_js:
-  - "6.10.3"
+  - "8.1.0"
```
jhermsmeier: Nope, it should work with at least node v4 and up – but in our CI builds node is mostly an auxiliary for running the builds, scripts & installing packages etc. Doesn't make much sense to also run the CI with v6 & v8 to me, especially as that would also amplify our build times (which are already at 60+min) by a few times that.

scripts/build/docker/Dockerfile.template
jhermsmeier: No need to search for it, the URL is right up in there: https://github.com/tj/n
Check out the code, it's pretty readable & straight forward!
PS: It's also the second result after `nvm` when I search for "node version manager" ;P

Makefile
lurch: Does that mean that the node-static-entry-point thingy that @jviotti made would also need to be updated to use node8?  (disclaimer: I've not looked at how that tool works)

package.json
jhermsmeier: Oh right, that one – well, we should still be safe & sound as that made it into the node / npm dists.

Makefile
jhermsmeier: Good point. Haven't looked at that either. Needs checking.

.travis.yml
lurch: Fair enough. All the more reason to attempt to optimise our build times then ;)

jhermsmeier: I concur. Most of it is the wait time for the Mac OS builds though – not much we can do about that (on Travis at least).

package.json
lurch: ...so we should stop shrinkwrapping the `node-gyp` version now? (which I guess also means scripts/ci/ensure-npm-shrinkwrap-versions.sh becomes defunct)

jhermsmeier: We should stop explicitly depending on it altogether, in my opinion. If there's a node version with a buggy node-gyp, we should either move to a different version, or install a non-buggy one globally, I think. But both cases shouldn't be necessary anymore when we're on node 8.

npm-shrinkwrap.json
jhermsmeier: > Ouch!

Haha, not really – with Sublime it's just press & hold <kbd>Cmd</kbd>+<kbd>D</kbd> on a selection that matches them all, edit, save, done ;)

> If it is, I guess a small reproducible example would be useful.

Yeah, I'll fiddle with it tomorrow

scripts/build/docker/Dockerfile.template
lurch: Comparing https://nodejs.org/dist/v6.11.0/ and https://nodejs.org/dist/v7.10.0/ and https://nodejs.org/dist/v8.1.0/ and https://nodejs.org/dist/v8.1.1/ , it seems like v8.1.1 is missing some of the ARM builds?

Ah... I was about to create an [issue](https://github.com/nodejs/nodejs.org/issues), but then spotted that https://github.com/nodejs/nodejs.org/blob/master/locale/en/blog/release/v8.1.1.md says the ARM builds are "_Coming soon_" ;-)

jhermsmeier:
> Although perhaps it would make more sense to install node-sass in the same way as we install uglify-es ? IMHO relying on $(which node-sass) looks a bit ugly.

@lurch I agree – we'd probably run into issue with not all of us using the same globally installed `node-sass` version, and that tripping up the sanity checks...

npm-shrinkwrap.json
jhermsmeier: As it turns out, it is indeed `npm` itself screwing it up... guess I'll create another issue if there isn't one already.

lurch:
Should this PR also include some of the changes from #1514 ? Or are they each addressing orthogonal problems?

scripts/fix-shrinkwrap.js
```diff
@@ -1,46 +1,56 @@
...
   if (!dependency.resolved && /^http(s):/i.test(dependency.version)) {
 
     // Extract the version number from the registry URL
-    var match = dependency.version.match(/(\d+\.\d+\.\d+(\-.+)?)\.tgz$/);
+    const match = dependency.version.match(/(\d+\.\d+\.\d+(-.+)?)\.tgz$/);
 
     // If something goes wrong, we do not want
     // this to continue in any way; so we crash
-    if (!match || !match[1]) {
+    if (!match || !match[ONE]) {
```
lurch: Alternatively, you could change the regex to use a non-capturing group `(?: )`, and then use `_.first` ?

scripts/fix-shrinkwrap.js
```diff
@@ -52,36 +62,38 @@ function fixup(dependency) {
...
+};
 
 const fixes = walk(shrinkwrap);
 
 // Save the shrinkwrap back
-fs.writeFileSync(SHRINKWRAP_FILE, JSON.stringify(shrinkwrap, null, 2) + '\n');
+fs.writeFileSync(SHRINKWRAP_FILE, `${JSON.stringify(shrinkwrap, null, TWO)}\n`);
 
-time = (Date.now() - time) / 1000;
+time = (Date.now() - time) / THOUSAND;
```
lurch: In keeping with the style of the rest of the codebase, `THOUSAND` here should probably be something like `MILLISECONDS_IN_A_SECOND` ?

scripts/fix-shrinkwrap.js
```diff
@@ -52,36 +62,38 @@ function fixup(dependency) {
...
   }
 
-  var keys = Object.keys(tree.dependencies);
-  var fixes = 0;
-  var dep = null;
+  let fixes = ZERO;
 
-  for (var i = 0; i < keys.length; i++) {
-    dep = tree.dependencies[keys[i]];
-    fixes += fixup(dep) ? 1 : 0;
```
lurch: I guess `fixes += fixup(dep) ? 1 : 0;` could alternatively be replaced by `fixes += Number(fixup(dep))` ?

jviotti:
Should we kill this?

jhermsmeier:
Yep, let's wait for npm v5 to stabilise, and then retry

lurch:
Hopefully it'll be "3rd time lucky", LOL.

PR #1488, opened 2017-06-02
chore(package): Update mountutils to 1.2.0

jhermsmeier:
This updates `mountutils` from 1.0.6 to 1.2.0, which includes
various fixes and adds AsyncWorkers:

- fix(windows): Replace use of `wsprintf()`
- fix(darwin): Add local context to avoid global state
- feat(src): Use Nan::AsyncWorker

Change-Type: upgrade


jviotti:
Can we change the commit type to `upgrade`?

jhermsmeier:
@jviotti mended the change-type

PR #1489, opened 2017-06-02
upgrade(package): Update node-sass to 4.5.3

jhermsmeier:
This updates `node-sass` from v3.x to v4.x in anticipation
of addition of Electron ABI versions in an upcoming version.

Change-Type: patch

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3610,6 +3610,12 @@
       "resolved": "https://registry.npmjs.org/jodid25519/-/jodid25519-1.0.2.tgz",
       "optional": true
     },
+    "js-base64": {
+      "version": "2.1.9",
+      "from": "js-base64@>=2.1.8 <3.0.0",
+      "resolved": "https://registry.npmjs.org/js-base64/-/js-base64-2.1.9.tgz",
+      "dev": true
+    },
     "js-message": {
       "version": "1.0.5",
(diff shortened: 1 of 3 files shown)


npm-shrinkwrap.json
```diff
@@ -4962,14 +4974,14 @@
       "resolved": "https://registry.npmjs.org/moment-duration-format/-/moment-duration-format-1.3.0.tgz"
     },
     "mountutils": {
-      "version": "1.0.6",
-      "from": "mountutils@1.0.6",
-      "resolved": "https://registry.npmjs.org/mountutils/-/mountutils-1.0.6.tgz",
+      "version": "1.2.0",
```
lurch: I suspect this PR shouldn't be also bumping `mountutils` ?

jhermsmeier: Indeed, that's not right – fixed it!

jhermsmeier:
@jviotti I wasn't aware we also had an `upgrade` type – fixed!

lurch:
@jhermsmeier You'v confused commit-type and change-type ;-P
https://github.com/resin-io/etcher/blob/master/docs/COMMIT-GUIDELINES.md

jhermsmeier:
Ohhhh boy, clearly a case of scatterbrain – mended it (again) now

PR #1491, opened 2017-06-05
doc(README): Update & normalize badges

jhermsmeier:
This updates & normalizes the badges in the README to all have
a consistent style and adds a release-badge pointing to the website.

Change-Type: patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,13 @@ flashing an SDCard or USB drive is a pleasant and safe experience. It protects
 you from accidentally writing to your hard-drives, ensures every byte of data
 was written correctly and much more.
 
-[![dependencies](https://david-dm.org/resin-io/etcher.svg)](https://david-dm.org/resin-io/etcher.svg)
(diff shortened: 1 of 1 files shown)


lurch:
The number on the waffle icon seems to be white text on a light-gray background, and therefore totally unreadable?

jhermsmeier:
@lurch yeah – that'd be due to https://github.com/badges/shields/issues/620 – and will prob. be fixed sooner or later. Not sure if we'd want to fall back to the previous (which also displayed an incorrect count), or just use this and it'll resolve itself somehow at some point.

jviotti:
Let's keep the new one. The badges provider will definitely fix it (can we actually let them know that its unreadable btw?).

jhermsmeier:
Created a new issue for this particular case: https://github.com/badges/shields/issues/1008

lurch:
@jhermsmeier Fancy also updating https://github.com/resin-io-modules/drivelist ? :)

jhermsmeier:
@lurch haha, yeah, can do :)

PR #1425, opened 2017-05-14
Add support for rpm packages

Dark-Passenger:
This is mostly copy paste job of debian packaging with minor changes for redhat systems. Most of the heavy lifting is being done by electron-installer-redhat.

PR #1501, opened 2017-06-08
refactor(GUI): move ETCHER_DISABLE_UPDATES into package.json

jviotti:
Etcher supports disabling the update notification dialog by setting the
`ETCHER_DISABLE_UPDATES` environment variable.

In order to simplify disabling updates for when these are managed by a
package manager (e.g. in a debian package), this removes support for the
`ETCHER_DISABLE_UPDATES` environment variable, and instead requires
packagers to tweak the `updates.enabled` property of the package.json
file, which is set to `true` by default.

We don't want to encourage end users to disable the update mechanism, so
the documention was removed from `USER-DOCUMENTATION.md`. This option
will remain as something only packagers should tweak.

Change-Type: minor
Changelog-Entry: Remove support for the `ETCHER_DISABLE_UPDATES` environment variable.


scripts/build/debian/config.json
```diff
@@ -4,7 +4,7 @@
   "genericName": "Etcher",
   "homepage": "https://etcher.io/",
   "icon": "assets/icon.png",
-  "bin": "etcher-electron.sh",
+  "bin": "etcher",
```
lurch: (AFAICT) This now means that both the GUI and CLI versions of Etcher are named `etcher`? :-/

lurch:
Ooooh, interesting. To make things easier for packagers, could there maybe be e.g. a Makefile variable that you could set during building, which would automatically turn off updates for (just) the produced build?

lib/gui/models/store.js
```diff
@@ -49,6 +49,7 @@ const DEFAULT_STATE = Immutable.fromJS({
     errorReporting: true,
     unmountOnSuccess: true,
     validateWriteOnSuccess: true,
+    disableUpdates: !packageJSON.updates.enabled,
```
lurch: Is it maybe a bit 'contradictory' that "internally" (i.e. in code) we refer to it as `disableUpdates` but "externally" (i.e. in `package.json`) we refer to it as `updates.enabled`?
Although I guess as long as we're consistent it should be okay.

jviotti: Good point. It looks like the "enabled" form is better.

jviotti:
> Ooooh, interesting. To make things easier for packagers, could there maybe be e.g. a Makefile variable that you could set during building, which would automatically turn off updates for (just) the produced build?

I'm addressing this on the `electron-builder` branch, which I hope to send a proper PR for very soon.

scripts/build/debian/config.json
jviotti: Good point. I added a workaround to rename the binary to `etcher-electron` and reference it as such. It will be fixed the "right" way on the `electron-builder` PR.

jviotti: Actually, I'm reverting the changes. This file has a `name` property at the top which is set to `etcher-electron`, so it doesn't really matter what the name of the binary entry point it, as it will always be aliases to the `name`.

lurch:
> I'm addressing this on the `electron-builder` branch

Ahh, cool. Does that work the same with the `electron-installer-debian` and `electron-installer-redhat` ?
Just when I think I've got to grips with the build system, it changes again... :laughing: 
(don't worry, I know you're trying to switch to a more "standardised" build system and use less of our "custom" build system)

lib/gui/app.js
```diff
@@ -100,14 +99,12 @@ app.run((ErrorService, SelectionStateModel) => {
   });
 
   const currentReleaseType = release.getReleaseType(currentVersion);
+  const enableUpdates = settings.get('enableUpdates');
```
lurch: Perhaps `updatesEnabled` would be better, as `enableUpdates` could be misinterpreted as a verb (command) ?

jviotti: That's a great idea. Done!

jviotti:
lol yeah, most of the options from those packages can be adapted to `electron-builder`.

lurch:
If this gets merged standalone, it'll result in `.deb` files being created with updates enabled (which is undesirable). So I'd prefer to see this merged at the same time as (or at least very close to) the `electron-builder` PR.

jviotti:
Fair enough. Check my latest changes. I added a workaround to deal with this, so debian packages are not broken.

scripts/build/jq-insert.sh
```diff
@@ -56,5 +56,12 @@ if [ -z "$ARGV_PROPERTY" ] ||
 fi
 
 TEMPORARY_FILE="$ARGV_TEMPORARY_DIRECTORY/$(basename "$ARGV_FILE").TMP"
-cat "$ARGV_FILE" | jq ".$ARGV_PROPERTY=\"$ARGV_VALUE\"" > "$TEMPORARY_FILE"
+
+if [ "$ARGV_VALUE" == "true" ] || [ "$ARGV_VALUE" == "false" ]; then
```
lurch: I guess this is basically testing value-is-not-a-string, i.e. if we ever wanted to set a numeric value, we'd need that to also follow this first `if` branch. I wonder if a better solution might be to require string values passed to `jq-insert.sh` to already be double-quoted? e.g. change https://github.com/resin-io/etcher/blob/master/Makefile#L220 to
```
-v "\"$(ANALYTICS_SENTRY_TOKEN)\"" \
```
?

Makefile
```diff
@@ -148,6 +148,14 @@ $(warning No Mixpanel token found (ANALYTICS_MIXPANEL_TOKEN is not set))
 endif
 
 # ---------------------------------------------------------------------
+# Updates
+# ---------------------------------------------------------------------
+
+ifdef DISABLE_UPDATES
+$(warning Update notification dialog was disabled (DISABLE_UPDATES is set))
```
lurch: Minor grammar nit: strange mix here of `was` and `is` - perhaps `was` could be changed to `has been` ?

Makefile
```diff
@@ -225,6 +233,14 @@ ifdef ANALYTICS_MIXPANEL_TOKEN
 		-t $(BUILD_TEMPORARY_DIRECTORY)
 endif
 
+ifdef DISABLE_UPDATES
+	./scripts/build/jq-insert.sh \
+		-p "updates.enabled" \
+		-v "false" \
+		-f $@/package.json \
+		-t $(BUILD_TEMPORARY_DIRECTORY)
+endif
```
lurch: Very nice :+1: 
Should this `DISABLE_UPDATES` variable be documented somewhere?

Makefile

Makefile

scripts/build/jq-insert.sh
jviotti: Yeah, that's a much better approach. Check my latest changes!

docs/PUBLISHING.md
```diff
@@ -96,6 +96,12 @@ make electron-installer-zip
 make electron-installer-nsis
 ```
 
+Disable updates
```
lurch: For consistency, this ought to be `Disabling updates`

jviotti: Done! :)

lurch:
Given the removal of `etcher-electron.sh` first (which is also used in the RPM PR), do you want to 'delay' merging this PR until after #1425 has been merged?

docs/PUBLISHING.md
lurch: Quick!

jviotti:
@lurch I tested the PR at some point and seems to be working great, so I think we can just merge that other one to unblock this one today. I'm going to move the rpm generation logic to `electron-builder` anyway very soon, so I'll be testing all that stuff again anyway :)

lurch:
#1425 is merged! So now you get to add the rpm files to this PR too ;-)

jviotti:
Oh no! :P I'll update the PR in a second.

jviotti:
Done!

PR #1502, opened 2017-06-08
chore: remove unused `electron-packager` dependency

jviotti:
We're not using this development dependency anymore. Furthermore, we're
also not using the `packageignore.js` script, which was meant to be used
with `electron-packager`.


lurch:
I wondered why my magic CI bash scripts didn't pick this up, but then I remembered that my scripts only check that the modules `require`d by the source code are available in `package.json` ;-)
I _started_ writing a script to do the inverse check (check that all packages in `package.json` are actually used in the source code) but I ran into a slight problem and never got round to finishing it off. I'll have to see if I can find it again...

lurch:
Ahhh, in fact... https://github.com/resin-io/etcher/pull/1237#issuecomment-290767373
:grinning:

PR #1503, opened 2017-06-08
chore: remove ampersand from package description

jviotti:
The ampersand confuses nupkg when generating Windows installers from
`electron-builder`.

The referenced issue talks about an issue where the ampersand is present
on the application name, but anything that gets into the `.nuspec` XML
file, including the description, triggers the issue.

See: https://github.com/electron-userland/electron-builder/issues/517

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "semverRange": "<2.0.0"
   },
   "main": "lib/start.js",
-  "description": "Flash OS images to SD cards & USB drives, safely and easily.",
+  "description": "Flash OS images to SD cards and USB drives, safely and easily.",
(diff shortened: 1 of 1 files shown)


lurch:
The latter comments in that thread imply that there's a workaround?

jviotti:
@lurch The workaround seems to be:

> @initPRAGMA just use nsis target (default target for Windows now).

Which we don't want to do, since we want to generate Squirrel installers to implement auto updates in the future.

jviotti:
Ping @lurch

lurch:
The pedant in me says that `electron-builder` ought to fix their scripts to escape things properly, but the pragmatist in me says all the CI tests are passing so I'll approve ;-)

jviotti:
Haha, I'll see if I can escape it somehow after the electron-builder transition.

PR #1505, opened 2017-06-08
chore: move npm targets to Makefile

jviotti:
We currently have various npm script target, and some of them are
getting complex enough that making sense out of them in package.json is
not a trivial task.

This commit moves all npm targets that are not directly recognisable by
npm (like `start`, `test`, `preshrinkwrap`, etc) into the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 # Build configuration
 # ---------------------------------------------------------------------
 
+NODE_MODULES_BIN=./node_modules/.bin
+
 # This directory will be completely deleted by the `clean` rule
 BUILD_DIRECTORY ?= release
 
@@ -170,6 +172,9 @@ define execute-command
 	$(1)
 
 endef
+
+CHANGELOG.md:
+	$(NODE_MODULES_BIN)/versionist
 
 $(BUILD_DIRECTORY):
 	mkdir $@
@@ -426,6 +431,15 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGE
 TARGETS = \
 	help \
 	info \
+	lint \
+	lint-js \
+	lint-sass \
+	lint-cpp \
+	lint-html \
+	lint-spell \
+	test-gui \
+	test-sdk \
+	test \
 	sanity-checks \
 	clean \
 	distclean \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -171,6 +171,9 @@ define execute-command
 
 endef
 
+CHANGELOG.md:
+	./node_modules/.bin/versionist
```
lurch: Could we make `./node_modules/.bin` a variable please?

lurch:
Coool! :-D
I once considered suggesting exactly this quite a while ago, but I didn't because I thought you were too much of an NPM fan ;)  (and at the time, I wasn't aware of the `node_modules/.bin/` 'trick').

package.json
```diff
@@ -16,15 +16,7 @@
...
-    "test": "npm run lint && electron-mocha --recursive --renderer tests/gui -R spec && electron-mocha --recursive tests/shared tests/child-writer tests/image-stream -R spec",
-    "sass": "node-sass ./lib/gui/scss/main.scss > ./lib/gui/css/main.css",
-    "cpplint": "cpplint --recursive src",
-    "jslint": "eslint lib tests scripts bin versionist.conf.js",
-    "sasslint": "sass-lint lib/gui/scss",
-    "htmllint": "node scripts/html-lint.js",
-    "codespell": "codespell.py --skip *.gz,*.bz2,*.xz,*.zip,*.img,*.dmg,*.iso,*.rpi-sdcard,.DS_Store lib tests docs scripts Makefile *.md LICENSE",
-    "lint": "npm run jslint && npm run sasslint && npm run cpplint && npm run codespell && npm run htmllint",
-    "changelog": "versionist",
+    "test": "make lint test",
```
lurch: Is it confusing to have `make test` do one thing, but then have `npm run test` do extra stuff?

jviotti: Fair enough, I'll unify both.

Makefile

lurch: ...except you forgot to change it here, doh! ;P

lurch:
...and on the topic of Makefiles, something I've been wondering recently is if there should also be a rule which deletes the `node_modules` and `build` folders? Something like a `distclean` (which would also depend on the `clean` rule) ?

Makefile
jviotti: Good catch, fixed!

jviotti:
@lurch I think you forgot you already took care of that yourself :P https://github.com/resin-io/etcher/blob/master/Makefile#L559

jviotti:
I'm merging this since I'd like to do some `electron-builder` work during the weekend, and I believe all comments were addressed. Please let me know if there was something else you wanted me to address and I'll send a future PR with it :)

Makefile
```diff
@@ -532,6 +546,41 @@ electron-develop:
 		-t electron \
 		-s "$(TARGET_PLATFORM)"
 
+sass:
+	$(NODE_MODULES_BIN)/node-sass ./lib/gui/scss/main.scss > ./lib/gui/css/main.css
```
lurch: Do we really need the `./` on the front of the paths here?

jviotti: Good catch. I'm sending a new PR for this.

PR #1509, opened 2017-06-09
refactor(GUI): remove angular dependency from selection-state

Shou:
We remove the dependency on Angular from SelectionStateModel and rename
it to selectionState.

Changelog-Entry: Remove Angular dependency from selection-state

### DIFF
--- a/lib/gui/models/selection-state.js
+++ b/lib/gui/models/selection-state.js
@@ -16,338 +16,325 @@
 
 'use strict';
 
-/**
- * @module Etcher.Models.SelectionState
- */
-
 const _ = require('lodash');
-const angular = require('angular');
 const Store = require('./store');
 const availableDrives = require('./available-drives');
-const MODULE_NAME = 'Etcher.Models.SelectionState';
-const SelectionStateModel = angular.module(MODULE_NAME, []);
 
-SelectionStateModel.service('SelectionStateModel', function() {
+/**
+ * @summary Set a drive
+ * @function
+ * @public
+ *
+ * @param {String} drive - drive device
+ *
+ * @example
+ * selectionState.setDrive('/dev/disk2');
+ */
+module.exports.setDrive = (drive) => {
+  Store.dispatch({
+    type: Store.Actions.SELECT_DRIVE,
+    data: drive
+  });
+};
 
-  /**
-   * @summary Set a drive
-   * @function
-   * @public
-   *
-   * @param {String} drive - drive device
-   *
-   * @example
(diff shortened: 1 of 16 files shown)


lib/gui/models/selection-state.js
```diff
@@ -16,338 +16,325 @@
...
+ * }
+ */
+module.exports.isCurrentDrive = (drive) => {
+  if (!drive) {
+    return false;
+  }
 
-module.exports = MODULE_NAME;
+  return drive === _.get(this.getDrive(), [ 'device' ]);
+};
```
lurch: Kind of annoying that the GH diff-display can't do a better job with this file. Ho hum :-/

lib/gui/models/selection-state.js
```diff
@@ -16,338 +16,325 @@
...
+ * @summary Set a drive
+ * @function
+ * @public
+ *
+ * @param {String} drive - drive device
+ *
+ * @example
+ * selectionState.setDrive('/dev/disk2');
+ */
+module.exports.setDrive = (drive) => {
```
jviotti: Let's use just `exports.<function>`, since the initial `module` is redundant (CommonJS automatically aliases `exports` to `module.exports`).

lib/gui/models/selection-state.js
```diff
@@ -16,338 +16,325 @@
...
+ * @function
+ * @public
+ *
+ * @param {String} drive - drive device
+ *
+ * @example
+ * selectionState.toggleSetDrive('/dev/disk2');
+ */
+module.exports.toggleSetDrive = (drive) => {
+  if (this.isCurrentDrive(drive)) {
```
jviotti: I don't think `this` is valid now. We should use `exports.<function>` instead.

lib/gui/models/selection-state.js
jviotti: Indeed, its a PITA to review.

tests/gui/models/selection-state.spec.js
```diff
@@ -3,74 +3,63 @@
...
-  ));
-
-  describe('SelectionStateModel', function() {
-
-    let SelectionStateModel;
-
-    beforeEach(angular.mock.inject(function(_SelectionStateModel_) {
-      SelectionStateModel = _SelectionStateModel_;
-    }));
+  describe('selectionState', function() {
```
jviotti: I think this `describe` block is redundant given that we're already saying `Browser: selectionState` above. Maybe we can remove it and get rid of one indentation level? :)

Shou:
Done

Issue #1030, opened 2017-01-17
Allow installation to non-default location


WasabiFan:
- **Etcher version:** Beta 18
- **Operating system and architecture:** Windows 10

The Etcher installer should allow the user to choose where they want the app to be installed instead of always using the default location.

jviotti:
This is a great suggestion. We're currently using a third party module that generates the installer for us, but we've been looking to move away from it, and this change can be done while doing so.

Issue #877, opened 2016-11-18
Windows installer directory "error"


lurch:
- **Etcher version:** 1.0.0-beta16
- **Operating system and architecture:** Windows7 x64
- **Do you see any meaningful error information on DevTools?** n/a

I installed the x64 version of Etcher onto my 64-bit Windows 7 laptop tonight, and was very surprised to see the Etcher files ending up in `C:\Program Files (x86)\Etcher\`! According to [this](http://answers.microsoft.com/en-us/windows/forum/windows_7-files/programs-files-vs-program-files-x86/7d631676-e688-472a-a1bb-eefccfeca6d7), `Program Files (x86)` is only supposed to be used for 32-bit programs. Using [this](http://superuser.com/questions/358434/how-to-check-if-a-binary-is-32-or-64-bit-on-windows#889267) I verified that `C:\Program Files (x86)\Etcher\Etcher.exe` is indeed a 64-bit application.

So then as a test I uninstalled the x64 version of Etcher, and installed the x86 version. It again installed into `C:\Program Files (x86)\Etcher\`, but this time using the info from the earlier superuser.com link I was able to verify that `C:\Program Files (x86)\Etcher\Etcher.exe` was now a 32-bit application.

I.e. it's not a packaging problem, as `Etcher-1.0.0-beta.16-win32-x64.exe` contains the x64 application and `Etcher-1.0.0-beta.16-win32-x86.exe` contains the x86 application - it's merely an installation-directory problem.

jviotti:
Hm, there must be something going on with the current installer. At the moment, we just use a third-party module to take care of the installation directory, which means we don't have any insights/control about that its actually doing. 

I suggest to solve this once we transition to making the installer in our own build scripts instead of using third party stuff.

lurch:
> once we transition to making the installer in our own build scripts instead of using third party stuff.

Hmmm, is there an issue for that?

jviotti:
Not really, but is a requirement for the issue about optimizing the final package size, so I guess it's covered by that one.

PR #1510, opened 2017-06-10
chore: use the new `electron-builder` version to create NSIS installer

jviotti:
We've been using `electron-builder` v2 all this time to create the NSIS
installer. This commit upgrade `electron-builder` to v18.6.2, and keeps
using it just to create the NSIS installer (for now).

The final package behaves exactly like the one we have before, just that
we needed various tweaks to upgrade to the latest `electron-builder`
version.

In more detail:

- Inject data to package.json using the new `--extraMetadata` option
- Remove old `.builder` package.json property
- Change the author of the project to Resin Inc. (the company name used
  in our code-signing certificate)

As an extra, the new NSIS installer allows the user to install the
application to any location, and fixes the fact that the previous
installer copied the application to C:\Program Files (x86) even on x64
systems.

Change-Type: patch
Fixes: https://github.com/resin-io/etcher/issues/1030
Fixes: https://github.com/resin-io/etcher/issues/877


jviotti:
The plan is to slowly transition to `electron-builder` so we don't need to create and merge a huge un-reviewable PR.

Makefile
```diff
@@ -483,7 +492,7 @@ endif
 
 ifeq ($(TARGET_PLATFORM),win32)
 electron-installer-zip: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
-electron-installer-nsis: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe
+electron-installer-nsis: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe
```
jviotti: `electron-builder` outputs final packages to `dist/` instead of `dist/out` (like we do at the moment), but let's live with this little inconsistency until we completely move to `electron-builder`.

lurch: Couldn't we just `mv` the file from `dist/` to `dist/out/` after it's been created?

package.json
```diff
@@ -18,8 +18,36 @@
...
+    },
+    "nsis": {
+      "oneClick": false,
+      "allowToChangeInstallationDirectory": true,
+      "runAfterFinish": true,
+      "installerIcon": "assets/icon.ico",
+      "uninstallerIcon": "assets/icon.ico",
+      "deleteAppDataOnUninstall": true,
+      "license": "LICENSE",
+      "artifactName": "${productName}-${version}-win32-${env.TARGET_ARCH}.${ext}"
```
lurch: Interesting - does NSIS fill in these placeholders automatically then?

Makefile
```diff
@@ -400,10 +405,19 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TAR
...
-	| $(BUILD_OUTPUT_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
-	./scripts/build/electron-installer-nsis-win32.sh -n $(APPLICATION_NAME) -a $< -t $(BUILD_TEMPORARY_DIRECTORY) -o $@
+ELECTRON_BUILDER_OPTIONS=--$(ELECTRON_BUILDER_TARGET_ARCH) --extraMetadata.version=$(APPLICATION_VERSION)
+
+ifdef ANALYTICS_SENTRY_TOKEN
+ELECTRON_BUILDER_OPTIONS+= --extraMetadata.analytics.sentry.token=$(ANALYTICS_SENTRY_TOKEN)
+endif
+
+ifdef ANALYTICS_MIXPANEL_TOKEN
+ELECTRON_BUILDER_OPTIONS+= --extraMetadata.analytics.mixpanel.token=$(ANALYTICS_MIXPANEL_TOKEN)
```
lurch: Wow, I've never seen command-line flags "used" (abused?) in this way before ;)

Makefile
```diff
@@ -162,6 +162,11 @@ APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 # Fix hard link Appveyor issues
 CPRF = cp -RLf
 
+ELECTRON_BUILDER_TARGET_ARCH=$(TARGET_ARCH)
+ifeq ($(TARGET_ARCH),x86)
+ELECTRON_BUILDER_TARGET_ARCH=ia32
+endif
```
jviotti: Sounds good, done!

scripts/clean-shrinkwrap.js
```diff
@@ -28,7 +29,8 @@ console.log('Removing:', shrinkwrapIgnore.join(', '));
  * @returns {ChildProcess}
  */
 const npm = (command) => {
-  return spawn('npm', command, {
+  const binary = os.platform() === 'win32' ? 'npm.cmd' : 'npm';
```
jviotti: Done!

scripts/ci/build-installers.sh
```diff
@@ -48,8 +48,8 @@ if [ "$ARGV_OPERATING_SYSTEM" == "linux" ]; then
   ./scripts/build/docker/run-command.sh \
     -r "$TARGET_ARCH" \
     -s "$(pwd)" \
-    -c 'make installers-all'
+    -c 'make electron-develop installers-all'
```
jviotti: Yes, the problem is that the sanity check that ensures the shrinkwrap file doesn't contain optional dependencies actually removes the optional dependencies from `node_modules`, which means that when we run the installers, some optional dependencies are not needed anymore.

We need find a good way to fix that, but as a quick fix, we can re-run `electron-develop`, which will only install the dependencies that are missing, before creating the installers.

npm-shrinkwrap.json
```diff
@@ -7156,12 +7306,6 @@
       "version": "2.6.0",
       "from": "yauzl@2.6.0",
       "resolved": "https://registry.npmjs.org/yauzl/-/yauzl-2.6.0.tgz"
-    },
-    "zip-stream": {
-      "version": "1.1.1",
-      "from": "zip-stream@>=1.1.0 <2.0.0",
-      "resolved": "https://registry.npmjs.org/zip-stream/-/zip-stream-1.1.1.tgz",
-      "dev": true
```
jviotti: Yeah, the shrinkwrap diff is huge.

package.json
jviotti: Yep, it supports a wide range of placeholders.

Makefile
```diff
@@ -400,10 +405,19 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TAR
 	| $(BUILD_OUTPUT_DIRECTORY)
 	./scripts/build/zip-file.sh -f $< -s $(TARGET_PLATFORM) -o $@
 
-$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH) \
-	| $(BUILD_OUTPUT_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
-	./scripts/build/electron-installer-nsis-win32.sh -n $(APPLICATION_NAME) -a $< -t $(BUILD_TEMPORARY_DIRECTORY) -o $@
+ELECTRON_BUILDER_OPTIONS=--$(ELECTRON_BUILDER_TARGET_ARCH) --extraMetadata.version=$(APPLICATION_VERSION)
+
+ifdef ANALYTICS_SENTRY_TOKEN
```
jviotti: I moved things around a bit to unify as most as we can right now. Everything is grouped except the code that injects analytics on the CLI, which I think we can finish grouping once we move to `pkg`.

scripts/ci/build-installers.sh
lurch: Well for me, `make electron develop` always seems to take ages. And IIRC you said in one of the meetings that you want to optimise the time taken by the CI tests? So wouldn't it be better to just "fix" the optional dependencies sanity check in some way?
(when you say "not needed" above, did you actually mean "not installed"?)

BTW, if this really is a problem, how come we never noticed it before???

Makefile
```diff
@@ -400,10 +412,9 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TAR
 	| $(BUILD_OUTPUT_DIRECTORY)
 	./scripts/build/zip-file.sh -f $< -s $(TARGET_PLATFORM) -o $@
 
-$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH) \
-	| $(BUILD_OUTPUT_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
-	./scripts/build/electron-installer-nsis-win32.sh -n $(APPLICATION_NAME) -a $< -t $(BUILD_TEMPORARY_DIRECTORY) -o $@
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe: \
+	| $(BUILD_DIRECTORY)
+	PATH=$(NODE_MODULES_BIN):$$PATH TARGET_ARCH=$(TARGET_ARCH) build --win nsis $(ELECTRON_BUILDER_OPTIONS)
```
lurch: Hmmm, so no way to run electron-builder without adding things to the `$PATH`? :-/
Even so, it might be better to still explicitly call `$(NODE_MODULES_BIN)/build` rather than just `build`?

Makefile
lurch: Hmmm, and does leaving the file in `dist/` instead of `dist/out/` mean that the snapshot-build publishing step will be broken?

Makefile
lurch: I wonder if it might be worth removing the blank lines at https://github.com/resin-io/etcher/blob/0b067495c636c19bf934ddda28b07683e9c367db/Makefile#L228 and https://github.com/resin-io/etcher/blob/0b067495c636c19bf934ddda28b07683e9c367db/Makefile#L236 to make it clearer that they're all part of the same make-rule, and not actually separate rules? I had to read it a few times before I realised what was going on...

Makefile
jviotti: I tried that, but `electron-builder` calls other scripts that are inside `node_modules/.bin` (like `node-gyp`), so we must ensure the whole thing is in the PATH.

scripts/ci/build-installers.sh
jviotti: > So wouldn't it be better to just "fix" the optional dependencies sanity check in some way?

Trust me I tried, but couldn't find a way to do so without some non-trivial effort. I think we can avoid the whole thing by making a program that parses the shrinkwrap files instead of simply taking a list of optional modules to operate, but that'll take time, and I want to get this transition done ASAP so we can focus on other stuff.

I'll try to get something up and running while these PRs get merged. If I finish fast enough, I'll send a PR and we can rebase this one. What do you think?

> (when you say "not needed" above, did you actually mean "not installed"?)

So let's say there's an optional dependency for GNU/Linux called `foo`. When you run `electron-develop`, `foo` gets installed, however when attempting to check that the shrinkwrap is up to date, `foo` will get removed in OSes other that GNU/Linux. If the tasks that follow need to make use of this module, they will fail with ENOENT errors on GNU/Linux.

> BTW, if this really is a problem, how come we never noticed it before???

Looks like none of the optional dependencies that were getting removed were needed to generate the installers.

Makefile
jviotti: Oh, I missed this comment. The snapshot build publishing step still works correctly, since it will upload the file pointed by the new path. I'll move it to `dist/out` anyway!

Makefile
jviotti: Oh, the blank line suggestions were actually by @jhermsmeier, who commented that it was hard to read without blank lines :P I'm happy with both to be honest.

scripts/ci/build-installers.sh
lurch: Hmm, so do we know _which_ optional dependencies are needed by the installers? Rather than re-running the whole of `make electron-develop`, is there maybe some way to only install the optional dependencies needed by the installers?

Hmmmmm, thinking about it I guess the "problem" is that we have the dependencies (with some of them being optional) required by Etcher itself, and we also have the dependencies (with some of them being optional) required to create the Etcher installers; and from a conceptual point of view these are two entirely separate sets of dependencies, and yet from a practical point of view they're both stored in the same `package.json` / `npm-shrinkwrap.json` file? (have I understood that correctly?)
Hopefully this doesn't mean that we end up with modules only needed to create the Etcher installers, included inside Etcher's `.asar`?

Makefile
lurch: Even so, it might be better to still explicitly call `$(NODE_MODULES_BIN)/build` rather than just `build`?
;-)

Makefile
```diff
@@ -401,9 +413,9 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TAR
 	./scripts/build/zip-file.sh -f $< -s $(TARGET_PLATFORM) -o $@
 
 $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH) \
-	| $(BUILD_OUTPUT_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
-	./scripts/build/electron-installer-nsis-win32.sh -n $(APPLICATION_NAME) -a $< -t $(BUILD_TEMPORARY_DIRECTORY) -o $@
+	| $(BUILD_OUTPUT_DIRECTORY)
+	PATH=$(NODE_MODULES_BIN):$$PATH TARGET_ARCH=$(TARGET_ARCH) build --win nsis $(ELECTRON_BUILDER_OPTIONS)
+	mv $(BUILD_DIRECTORY)/$(notdir $@) $@
```
lurch: :+1:

Makefile

scripts/ci/build-installers.sh
jviotti: That's right. And the fact that electron-builder doesn't support prefixed npm installs makes things worse. The npm version that we're using adds `optional: true` flags to shrinkwrap entries, so I think we can use that instead. I'll give it a go and see how it goes.

lurch: Thanks for the explanations Juanchi, I think I understand what's going on now, so I'll approve :)

It's super-hacky, but could another solution be to have a `etcher-shrinkwrap.json` and a `electron-builder-shrinkwrap.json` and then copy each file to `npm-shrinkwrap.json` as appropriate? (and I guess you could rename the `node_modules` directory back and forth too). But of course that may well be worse than the current situation ;-)

Makefile
```diff
@@ -138,15 +138,25 @@ endif
 endif
 
 # ---------------------------------------------------------------------
+# Electron Builder
+# ---------------------------------------------------------------------
+
+ELECTRON_BUILDER_OPTIONS=--$(TARGET_ARCH_ELECTRON_BUILDER) --extraMetadata.version=$(APPLICATION_VERSION)
```
lurch: Do these variable declarations work with a space either side of the `=` (or `+=`) too?

scripts/ci/build-installers.sh
jviotti: You're welcome! Yeah, I think there are some ways to fix this. I'll see what I can do and hopefully send a PR soon.

Makefile
jviotti: Yep, in fact quoting the value means that the quotes will be present on the final command. For example:

```sh
foo "--bar baz"
```

lurch: Huh? Maybe we're talking at cross purposes. What I was asking is whether changing:
```
ELECTRON_BUILDER_OPTIONS=--$(TARGET_ARCH_ELECTRON_BUILDER) --extraMetadata.version=$(APPLICATION_VERSION)
```
to
```
ELECTRON_BUILDER_OPTIONS = --$(TARGET_ARCH_ELECTRON_BUILDER) --extraMetadata.version=$(APPLICATION_VERSION)
```
would work?  (and then the same for the `ELECTRON_BUILDER_OPTIONS+=` lines further down)

jviotti: Ah, I totally misunderstood the comment. That works fine, I'm updating the PR :)

scripts/ci/build-installers.sh
jviotti: And the solution is https://github.com/resin-io/etcher/pull/1514. We can manually remove the entries without ever running npm.

Makefile
jhermsmeier: If run through an `npm run` script then `node_modules/.bin` is already in the path (details: `$ npm help run`) and all tools should resolve to binaries in there. I'd suggest adding 
```json
"electron-builder": "electron-builder"
```
to the scripts in the `package.json`, so you can run it as `ENV_VAR=value npm run electron-build -- build --win nsis` etc.

Alternatively, we could use [npx](https://github.com/zkat/npx), which is built to solve this particular problem of executing npm package binaries. This would also avoid cluttering the `package.json`.

Makefile
lurch: Well perhaps a small comment above the rule then, explaining that it actually extends over multiple lines, including the blank ones? :)

lurch:
Seems like the 'related' #1511 removes a lot more darwin-specific code than this PR removes win32-specific code?

Makefile
jviotti: Oh, I didn't know about `npx`, such an awesome addition. Regarding npm scripts, I'd like to keep most of the tasks in the Makefile, since it seems npm scripts can easily become a mess to maintain.

jviotti:
The other PR moves all the macOS builds to electron-builder, but this one just moves the NSIS generation, which means there are a lot of Windows scripts that we can't get rid of yet.

lurch:
Ah, so I guess there'll be further Windows build PRs in future? Cool :+1:

Makefile

Makefile
```diff
@@ -2,7 +2,7 @@
 # Build configuration
 # ---------------------------------------------------------------------
 
-NODE_MODULES_BIN=./node_modules/.bin
+NPX=./node_modules/.bin/npx
```
lurch: Could put a space either side of the `=` for readability / consistency.

Makefile
```diff
@@ -426,9 +438,9 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TAR
 	./scripts/build/zip-file.sh -f $< -s $(TARGET_PLATFORM) -o $@
 
 $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH) \
-	| $(BUILD_OUTPUT_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
-	./scripts/build/electron-installer-nsis-win32.sh -n $(APPLICATION_NAME) -a $< -t $(BUILD_TEMPORARY_DIRECTORY) -o $@
+	| $(BUILD_OUTPUT_DIRECTORY)
+	TARGET_ARCH=$(TARGET_ARCH) $(NPX) build --win nsis $(ELECTRON_BUILDER_OPTIONS)
```
lurch: Much nicer, thanks @jhermsmeier :)

scripts/ci/build-installers.sh
lurch: So as this PR has now been merged, and #1514 apparently fixes this problem, after #1514 has been rebased to `master`, should it also revert `-c 'make electron-develop installers-all'` back to `-c 'make installers-all'` ?

jviotti: Yeah, let's hope we can get that fix ready soon. Turns out there are some edge cases if one of these nasty platform specific optional dependencies has its own dependencies :(

lurch: So dependencies of an optional dependency aren't automatically optional?

jviotti: They are optional, however we don't want to simply remove every optional dependency, but platform specific optional dependencies, and optional dependencies of platform specific optional dependencies.

PR #1511, opened 2017-06-10
chore: use `electron-builder` to generate macOS builds

jviotti:
This commit makes use of electron-builder to replace what our scripts
were already doing.

Change-Type: patch


Makefile
```diff
@@ -423,6 +385,17 @@ ifdef CODE_SIGN_CERTIFICATE_PASSWORD
 endif
 endif
 
+assets/osx/installer.tiff: assets/osx/installer.png assets/osx/installer@2x.png
+	tiffutil -cathidpicheck $^ -out $@
```
jviotti: That dependency comes with macOS (and its only used there). Do you think we should still document it?

Makefile
```diff
@@ -423,6 +385,17 @@ ifdef CODE_SIGN_CERTIFICATE_PASSWORD
...
+
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION).dmg: assets/osx/installer.tiff
+	PATH=$(NODE_MODULES_BIN):$$PATH CSC_NAME="$(CODE_SIGN_IDENTITY)" CSC_IDENTITY_AUTO_DISCOVERY=false \
+		build --mac dmg $(ELECTRON_BUILDER_OPTIONS)
```
jviotti: I guess this will be fixed with the `npx` program

Makefile
```diff
@@ -423,6 +385,17 @@ ifdef CODE_SIGN_CERTIFICATE_PASSWORD
...
+
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-mac.zip: assets/osx/installer.tiff
+	PATH=$(NODE_MODULES_BIN):$$PATH CSC_NAME="$(CODE_SIGN_IDENTITY)" CSC_IDENTITY_AUTO_DISCOVERY=false \
+		build --mac zip $(ELECTRON_BUILDER_OPTIONS)
```
jviotti: The ZIP is required to auto-updating purposes. We've been looking to do that since a long time ago, which is the reason why we're publishing it. Moving to `electron-builder` should make this feature a bit easier to implement, and then ZIPs will be very widely used.

Makefile
```diff
@@ -462,16 +435,16 @@ package-electron: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-
 package-cli: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)
 
 ifeq ($(TARGET_PLATFORM),darwin)
-electron-installer-app-zip: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
-electron-installer-dmg: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg
+electron-installer-app-zip: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-mac.zip
+electron-installer-dmg: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION).dmg
```
jviotti: The renaming is because `electron-builder` doesn't allow us to change the final artifact in OSes other than Windows (just I guess we can just stick with what they recommend).

I updated the PR to move artifacts to `$(BUILD_OUTPUT_DIRECTORY)`

Makefile
lurch: If it's _always_ built-in, then it's probably fine (much like we don't document the need for `cp` or `mv`).

Makefile
lurch: Hmmm, seems weird that they allow changing the build artefact name in Windows but not in other OSes? \*shrug\*

If we're going to be switching to `electron-builder`-recommended names, then it probably makes sense to switch to the 'recommended' filename on Windows too? Although I guess these changed filenames will also impact @craig-mulligan 's code on the https://github.com/resin-io/etcher-homepage (the next time we do a 'proper' release). And I guess changing filenames might make https://github.com/resin-io/etcher-homepage/issues/57 more important too?

Alternatively, we could stick with the download filenames we're already using, and then simply rename the file at the same time we move it to `$(BUILD_OUTPUT_DIRECTORY)`.

I have no strong opinions on which option we go with.

jviotti: This is totally undocumented, but looks like `artifactName` works for every electron-builder target. I'll make use of it here, plus open an issue to improve the docs.

jviotti: Check the latest changes

Makefile
```diff
@@ -371,32 +358,20 @@ endif
...
+$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).zip: assets/osx/installer.tiff \
 	| $(BUILD_OUTPUT_DIRECTORY)
-ifdef CODE_SIGN_IDENTITY
-	./scripts/build/electron-sign-dmg-darwin.sh \
-		-n "$(APPLICATION_NAME)" \
-		-d $< \
-		-i "$(CODE_SIGN_IDENTITY)"
-endif
-	./scripts/build/electron-create-readonly-dmg-darwin.sh -d $< -o $@
+	CSC_NAME="$(CODE_SIGN_IDENTITY)" CSC_IDENTITY_AUTO_DISCOVERY=false TARGET_ARCH=$(TARGET_ARCH) \
```
lurch: I'm assuming this also works if `$(CODE_SIGN_IDENTITY)` isn't set?

PR #1512, opened 2017-06-12
[WIP] refactor(GUI): remove angular dependency from drive scanner

Shou:
Remove the Angular dependency from DriveScanner and with it the service,
exposing it through the module directly.

Changelog-Entry: Remove Angular dependency from DriveScanner.


jviotti:
The changes look great. I think we can merge as soon as the tests pass.

Shou:
Since it's kind of hard to read the diffs here with the indentation changes, the only important changes should be using `driveScanner.once` instead of `driveScanner.on` in the tests, because there were some issues with `done()` being called several times on the first test using `driveScanner.on` and that killed the subsequent tests as well.

PR #1513, opened 2017-06-12
chore: don't use ./ when generating sass files

jviotti:
The `./` prefix is unnecessary.

See: https://github.com/resin-io/etcher/pull/1505#pullrequestreview-43444274

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -547,7 +547,7 @@ electron-develop:
 		-s "$(TARGET_PLATFORM)"
 
 sass:
-	$(NODE_MODULES_BIN)/node-sass ./lib/gui/scss/main.scss > ./lib/gui/css/main.css
+	$(NODE_MODULES_BIN)/node-sass lib/gui/scss/main.scss > lib/gui/css/main.css
 
 lint-js:
 	$(NODE_MODULES_BIN)/eslint lib tests scripts bin versionist.conf.js


lurch:
Oooh, eagle-eyed-Andrew spots a different `Signed-off-by:` address :wink:

jviotti:
@lurch Yeah, some guy eventually bought jviotti.com :P

lurch:
```
This site can’t be reached

jviotti.com refused to connect.
```
:crying_cat_face:

lurch:
@lekkas Just out of curiosity, is there any kind of "process standard" for these signed-off addresses, i.e. using your private email address vs. using your resin.io email address, etc. ?

jviotti:
lol, the size is not up yet, but the e-mail is :P

jviotti:
@lurch If we're using GPG signatures, we need to sign using the e-mail from our signature, so I guess its fine to allow any e-mail the signer is comfortable with, unless we want to enforce employees to add custom user ids to their keys :/

lurch:
Ahh, I've never used signed-off-by myself, so I didn't realise it links to GPG.

lurch:
Is that something that a ProcBot could fix? :laughing:

jviotti:
The issue is that any rebase operation will destroy the signature, since stuff like dates change. The only way to preserve the signature would be to require the commiter to make any necessary changes.

PR #1514, opened 2017-06-12
chore: don't remove optional dependencies in clean-shrinkwrap.js

jviotti:
If we include a platform specific optional dependency in the shrinkwrap
file, then npm will insist in installing it even if the platform doesn't
match. As a solution, we figured out we can avoid putting this platform
specific optional dependencies in the npm-shrinkwrap.json file.

In order to do this, we currently have a script called
`clean-shrinkwrap.js` that runs *before* any `npm shrinkwrap` file (its
a `preshrinkwrap` npm script) that deletes all the platform specific
modules we know about using `npm rm`.

The problem with this approach is that `npm rm` will remove the module's
code from `node_modules`, which means that if we run `npm shrinkwrap`,
we will lose certain optional dependencies, that may be needed at a
later stage.

The solution is to modify the `clean-shrinkwrap.js` script to parse
`npm-shrinkwrap.json`, and manually delete the entries that we want to
omit. Also, the script needs to be run *after* `npm shrinkwrap`, so we
change the npm script name to `postshrinkwrap`.

### DIFF
--- a/scripts/clean-shrinkwrap.js
+++ b/scripts/clean-shrinkwrap.js
@@ -1,43 +1,159 @@
+/*
+ * Copyright 2017 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 /**
- * This script is in charge of generating the `shrinkwrap` file.
- *
- * `npm shrinkwrap` has a bug where it will add optional dependencies
- * to `npm-shrinkwrap.json`, therefore causing errors if these optional
- * dependendencies are platform dependent and you then try to build
- * the project in another platform.
- *
- * As a workaround, we keep a list of platform dependent dependencies in
- * the `shrinkwrapIgnore` property of `package.json`, and manually remove
- * them from `npm-shrinkwrap.json` if they exists.
- *
- * See: https://github.com/npm/npm/issues/2679
+ * This script is in charge of removing platform specific optional
+ * dependencies, and their dependencies from the `shrinkwrap` file.
  */
 
 'use strict';
 
+const _ = require('lodash');
+const fs = require('fs');
 const path = require('path');
-const os = require('os');
-const packageJSON = require('../package.json');
-const spawn = require('child_process').spawn;
-const shrinkwrapIgnore = packageJSON.shrinkwrapIgnore;
+const childProcess = require('child_process');
 
-console.log('Removing:', shrinkwrapIgnore.join(', '));
+const NPM_SHRINKWRAP_FILE_PATH = path.join(__dirname, '..', 'npm-shrinkwrap.json');
+const shrinkwrapFile = require(NPM_SHRINKWRAP_FILE_PATH);
+const platformSpecificDependencies = require('../package.json').platformSpecificDependencies;
 
 /**
- * Run an npm command
- * @param {Array} command - list of arguments
- * @returns {ChildProcess}
+ * @summary Normalize npm qualifiers
+ * @function
+ * @private
+ *
+ * @param {String[]} qualifiers - npm qualifiers
+ * @returns {String[]} normalized qualifiers
+ *
+ * @example
+ * const qualifiers = normalizeNpmQualifiers([ 'drivelist@^5.0.0', 'electron@1.6.6' ]);
+ * console.log(qualifiers);
+ * > [ 'drivelist', 'electron' ]
  */
-const npm = (command) => {
-  const npmBinary = os.platform() === 'win32' ? 'npm.cmd' : 'npm';
-  return spawn(npmBinary, command, {
-    cwd: path.join(__dirname, '..'),
-    env: process.env,
-    stdio: [ process.stdin, process.stdout, process.stderr ]
+const normalizeNpmQualifiers = (qualifiers) => {
+  return _.map(qualifiers, (qualifier) => {
+    return _.first(_.split(qualifier, '@'));
   });
 };
 
-npm([ 'rm', '--ignore-scripts' ].concat(shrinkwrapIgnore))
-  .once('close', () => {
-    console.log('Done.');
+/**
+ * @summary Get the top level dependencies of a certain npm specifier
+ * @function
+ * @private
+ *
+ * @param {String} specifier - npm specifier
+ * @returns {String[]} top level dependencies
+ *
+ * @example
+ * const dependencies = getTopLevelDependencies('drivelist@^5.0.0');
+ */
+const getTopLevelDependencies = _.memoize((specifier) => {
+  const json = childProcess.execSync(`npm view --json ${specifier} dependencies`, {
+    encoding: 'utf8'
   });
+
+  if (!json) {
+    return [];
+  }
+
+  const data = JSON.parse(json);
+  const dependencies = _.isPlainObject(data) ? _.toPairs(data) : _.flatMap(data, _.toPairs);
+  return _.uniq(_.map(dependencies, ([ name, version ]) => {
+    return `${name}@${version}`;
+  }));
+});
+
+/**
+ * @summary Recursively fetch the dependency tree of an npm specifier
+ * @function
+ * @private
+ *
+ * @param {String} specifier - npm specifier
+ * @returns {String[]} all dependencies
+ *
+ * @example
+ * const dependencies = getNormalizedDependencyTree('drivelist@^5.0.0');
+ */
+const getNormalizedDependencyTree = (specifier) => {
+  const topLevelDependencies = getTopLevelDependencies(specifier);
+  const nestedDependencies = _.flatMapDeep(topLevelDependencies, getNormalizedDependencyTree);
+  return _.uniq(_.concat(normalizeNpmQualifiers(topLevelDependencies), nestedDependencies));
+};
+
+/**
+ * @summary Remove certain dependencies from a shrinkwrap object
+ * @function
+ * @private
+ *
+ * @description
+ * A shrinkwrap object is a recursive data structure, that apart
+ * from some extra metadata, has the following structure:
+ *
+ * {
+ *   ...
+ *   "dependencies": {
+ *     "<dependency_name>": <recursive definition>,
+ *     "<dependency_name>": <recursive definition>,
+ *     "<dependency_name>": <recursive definition>,
+ *     ...
+ *   }
+ * }
+ *
+ * The purpose of this function is to remove certain dependencies
+ * that match a blacklist. In order to do so, we start from the top
+ * level object, remove the blacklisted dependencies, and recurse
+ * if possible.
+ *
+ * @param {Object} object - shrinkwrap object
+ * @param {String[]} blacklist - list of blacklist
+ * @param {Object} qualifiers - qualifiers
+ * @param {Boolean} qualifiers.dev - development dependency
+ * @param {Boolean} qualifiers.optional - optional dependency
+ * @returns {Object} filtered shrinkwrap object
+ */
+const removeDependencies = (object, blacklist, qualifiers) => {
+  if (!_.isEmpty(object.dependencies)) {
+    object.dependencies = _.chain(object.dependencies)
+      .omitBy((dependency, name) => {
+        return _.every([
+          _.includes(blacklist, name),
+          dependency.dev === qualifiers.dev,
+          dependency.optional === qualifiers.optional
+        ]);
+      })
+      .mapValues((dependency) => {
+        return removeDependencies(dependency, blacklist, qualifiers);
+      })
+      .value();
+  }
+
+  return object;
+};
+
+console.log(`Fetching dependency tree for ${_.join(platformSpecificDependencies, ', ')} (this may take a couple of minutes)`);
+const blacklist = _.uniq(normalizeNpmQualifiers(
+  _.concat(platformSpecificDependencies,
+  _.flatMap(platformSpecificDependencies, getNormalizedDependencyTree))));
+console.log(`Attempting to remove: ${_.join(blacklist, ', ')}`);
+
+const JSON_INDENTATION_SPACES = 2;
+const result = JSON.stringify(removeDependencies(shrinkwrapFile, blacklist, {
+  dev: true,
+  optional: true
+}), null, JSON_INDENTATION_SPACES);
+
+fs.writeFileSync(NPM_SHRINKWRAP_FILE_PATH, `${result}\n`);
+console.log('Done');
--- a/package.json
+++ b/package.json
@@ -77,21 +77,18 @@
   "scripts": {
     "test": "make test",
     "start": "electron lib/start.js",
-    "preshrinkwrap": "node ./scripts/clean-shrinkwrap.js",
+    "postshrinkwrap": "node ./scripts/clean-shrinkwrap.js",
     "configure": "node-gyp configure",
     "build": "node-gyp build",
     "install": "node-gyp rebuild"
   },
   "author": "Resin Inc. <hello@etcher.io>",
   "license": "Apache-2.0",
-  "shrinkwrapIgnore": [
-    "macos-alias",
-    "fs-xattr",
-    "ds-store",
-    "appdmg",
-    "7zip-bin-mac",
-    "7zip-bin-win",
-    "7zip-bin-linux"
+  "platformSpecificDependencies": [
+    "appdmg@0.3.10",
+    "7zip-bin-mac@1.0.1",
+    "7zip-bin-win@2.1.0",
+    "7zip-bin-linux@1.1.0"
   ],
   "dependencies": {
     "angular": "1.6.3",
--- a/scripts/ci/build-installers.sh
+++ b/scripts/ci/build-installers.sh
@@ -48,8 +48,8 @@ if [ "$ARGV_OPERATING_SYSTEM" == "linux" ]; then
   ./scripts/build/docker/run-command.sh \
     -r "$TARGET_ARCH" \
     -s "$(pwd)" \
-    -c 'make electron-develop installers-all'
+    -c 'make installers-all'
 else
   ./scripts/build/check-dependency.sh make
-  make electron-develop installers-all
+  make installers-all
 fi


scripts/clean-shrinkwrap.js
```diff
@@ -15,27 +15,42 @@
...
+  object.dependencies = _.chain(object.dependencies)
+    .omitBy((dependency, name) => {
+      return _.includes(dependencies, name);
+    })
+    .mapValues((dependency) => {
+      return removeDependencies(dependency, dependencies);
+    })
+    .value();
+
+  return object;
```
lurch: So I'm guessing (I find this function hard to read, as I'm less of a JS expert than you and @jhermsmeier ) that this recursively trawls through the dependencies tree in `object`, filtering out each dependency in `dependencies`?

Perhaps the code (especially the line `removeDependencies(dependency, dependencies)`) would be easier to read if you renamed the `dependencies` parameter to `dependency_list`, so that it's harder to confuse with `object.dependencies`, and more obvious what's just a list of strings and what's a nested object-tree? \*shrug\*

lurch:
So, for those of us that aren't NPM packaging experts, would you mind explaining in more detail what this does, how it's different to the previous approach, and how it solves the problem that we had previously?
Doesn't it still mean that optional dependencies will be missing from `npm-shrinkwrap.json`, and thus won't be installed on 'fresh' systems?

lurch:
For some reason both the Linux 32-bit and 64-bit Travis tests are passing, but the OSX Travis PR build is [failing](https://travis-ci.org/resin-io/etcher/jobs/242194837) the sanity-check with:
```
node version: v6.10.3
npm version: 3.10.10
wrote npm-shrinkwrap.json
> etcher@1.0.0 postshrinkwrap /Users/travis/build/resin-io/etcher
Removing: macos-alias, fs-xattr, ds-store, appdmg
Done
    npm shrinkwrap --dev
index 4f697c5..718936d 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -260,6 +260,13 @@
       "resolved": "https://registry.npmjs.org/array-back/-/array-back-1.0.4.tgz",
       "dev": true
     },
+    "array-buffer-from-string": {
+      "version": "0.1.0",
+      "from": "array-buffer-from-string@>=0.1.0 <0.2.0",
+      "resolved": "https://registry.npmjs.org/array-buffer-from-string/-/array-buffer-from-string-0.1.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "array-filter": {
       "version": "0.0.1",
       "from": "array-filter@>=0.0.0 <0.1.0",
@@ -465,6 +472,13 @@
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-0.4.2.tgz",
       "dev": true
     },
+    "base32-encode": {
+      "version": "0.1.0",
+      "from": "base32-encode@>=0.1.0 <0.2.0",
+      "resolved": "https://registry.npmjs.org/base32-encode/-/base32-encode-0.1.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "base64-js": {
       "version": "0.0.8",
       "from": "base64-js@0.0.8",
@@ -565,6 +579,13 @@
       "from": "bootstrap-sass@>=3.3.5 <4.0.0",
       "resolved": "https://registry.npmjs.org/bootstrap-sass/-/bootstrap-sass-3.3.6.tgz"
     },
+    "bplist-creator": {
+      "version": "0.0.7",
+      "from": "bplist-creator@>=0.0.3 <0.1.0",
+      "resolved": "https://registry.npmjs.org/bplist-creator/-/bplist-creator-0.0.7.tgz",
+      "dev": true,
+      "optional": true
+    },
     "brace-expansion": {
       "version": "1.1.6",
       "from": "brace-expansion@>=1.0.0 <2.0.0",
@@ -1100,6 +1121,13 @@
       "from": "core-util-is@>=1.0.0 <1.1.0",
       "resolved": "https://registry.npmjs.org/core-util-is/-/core-util-is-1.0.2.tgz"
     },
+    "cp-file": {
+      "version": "3.2.0",
+      "from": "cp-file@>=3.1.0 <4.0.0",
+      "resolved": "https://registry.npmjs.org/cp-file/-/cp-file-3.2.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "crc": {
       "version": "3.4.4",
       "from": "crc@>=3.4.4 <4.0.0",
@@ -2405,6 +2433,13 @@
       "from": "flexboxgrid@>=6.3.0 <7.0.0",
       "resolved": "https://registry.npmjs.org/flexboxgrid/-/flexboxgrid-6.3.0.tgz"
     },
+    "fmix": {
+      "version": "0.1.0",
+      "from": "fmix@>=0.1.0 <0.2.0",
+      "resolved": "https://registry.npmjs.org/fmix/-/fmix-0.1.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "for-in": {
       "version": "0.1.8",
       "from": "for-in@>=0.1.5 <0.2.0",
@@ -2499,6 +2534,13 @@
       "resolved": "https://registry.npmjs.org/fs-extra-tf/-/fs-extra-tf-0.30.4.tgz",
       "dev": true
     },
+    "fs-temp": {
+      "version": "1.1.2",
+      "from": "fs-temp@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/fs-temp/-/fs-temp-1.1.2.tgz",
+      "dev": true,
+      "optional": true
+    },
     "fs.realpath": {
       "version": "1.0.0",
       "from": "fs.realpath@>=1.0.0 <2.0.0",
@@ -3013,11 +3055,24 @@
       "resolved": "https://registry.npmjs.org/ignore/-/ignore-3.2.6.tgz",
       "dev": true
     },
+    "image-size": {
+      "version": "0.5.5",
+      "from": "image-size@>=0.5.0 <0.6.0",
+      "resolved": "https://registry.npmjs.org/image-size/-/image-size-0.5.5.tgz",
+      "dev": true,
+      "optional": true
+    },
     "immutable": {
       "version": "3.8.1",
       "from": "immutable@>=3.8.1 <4.0.0",
       "resolved": "https://registry.npmjs.org/immutable/-/immutable-3.8.1.tgz"
     },
+    "imul": {
+      "version": "1.0.1",
+      "from": "imul@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/imul/-/imul-1.0.1.tgz",
+      "dev": true
+    },
     "imurmurhash": {
       "version": "0.1.4",
       "from": "imurmurhash@>=0.1.4 <0.2.0",
@@ -4800,6 +4855,13 @@
       "from": "ms@0.7.2",
       "resolved": "https://registry.npmjs.org/ms/-/ms-0.7.2.tgz"
     },
+    "murmur-32": {
+      "version": "0.1.0",
+      "from": "murmur-32@>=0.1.0 <0.2.0",
+      "resolved": "https://registry.npmjs.org/murmur-32/-/murmur-32-0.1.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "mute-stream": {
       "version": "0.0.5",
       "from": "mute-stream@0.0.5",
@@ -4848,6 +4910,13 @@
       "resolved": "https://registry.npmjs.org/ncp/-/ncp-2.0.0.tgz",
       "dev": true
     },
+    "nested-error-stacks": {
+      "version": "1.0.2",
+      "from": "nested-error-stacks@>=1.0.1 <2.0.0",
+      "resolved": "https://registry.npmjs.org/nested-error-stacks/-/nested-error-stacks-1.0.2.tgz",
+      "dev": true,
+      "optional": true
+    },
     "netmask": {
       "version": "1.0.6",
       "from": "netmask@>=1.0.4 <1.1.0",
@@ -5443,6 +5512,13 @@
       "resolved": "https://registry.npmjs.org/querystring-es3/-/querystring-es3-0.2.1.tgz",
       "dev": true
     },
+    "random-path": {
+      "version": "0.1.1",
+      "from": "random-path@>=0.1.0 <0.2.0",
+      "resolved": "https://registry.npmjs.org/random-path/-/random-path-0.1.1.tgz",
+      "dev": true,
+      "optional": true
+    },
     "randomatic": {
       "version": "1.1.6",
       "from": "randomatic@>=1.1.3 <2.0.0",
@@ -6266,6 +6342,13 @@
       "resolved": "https://registry.npmjs.org/stream-browserify/-/stream-browserify-2.0.1.tgz",
       "dev": true
     },
+    "stream-buffers": {
+      "version": "2.2.0",
+      "from": "stream-buffers@>=2.2.0 <2.3.0",
+      "resolved": "https://registry.npmjs.org/stream-buffers/-/stream-buffers-2.2.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "stream-chunker": {
       "version": "1.2.8",
       "from": "stream-chunker@>=1.2.8 <2.0.0",
@@ -6617,6 +6700,13 @@
       "from": "tmp@0.0.31",
       "resolved": "https://registry.npmjs.org/tmp/-/tmp-0.0.31.tgz"
     },
+    "tn1150": {
+      "version": "0.1.0",
+      "from": "tn1150@>=0.1.0 <0.2.0",
+      "resolved": "https://registry.npmjs.org/tn1150/-/tn1150-0.1.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "to-arraybuffer": {
       "version": "1.0.1",
       "from": "to-arraybuffer@>=1.0.0 <2.0.0",
@@ -6862,6 +6952,13 @@
       "from": "underscore.string@>=3.2.2 <4.0.0",
       "resolved": "https://registry.npmjs.org/underscore.string/-/underscore.string-3.3.4.tgz"
     },
+    "unorm": {
+      "version": "1.4.1",
+      "from": "unorm@>=1.4.1 <2.0.0",
+      "resolved": "https://registry.npmjs.org/unorm/-/unorm-1.4.1.tgz",
+      "dev": true,
+      "optional": true
+    },
     "update-json": {
       "version": "1.0.0",
       "from": "update-json@>=1.0.0 <2.0.0",
The command "./scripts/ci/test.sh -o $HOST_OS -r $TARGET_ARCH" exited with 2.
```
:-S

scripts/clean-shrinkwrap.js
```diff
@@ -15,27 +15,42 @@
...
+const fs = require('fs');
 const path = require('path');
-const packageJSON = require('../package.json');
-const spawn = require('child_process').spawn;
-const shrinkwrapIgnore = packageJSON.shrinkwrapIgnore;
 
-console.log('Removing:', shrinkwrapIgnore.join(', '));
+const NPM_SHRINKWRAP_FILE_PATH = path.join(__dirname, '..', 'npm-shrinkwrap.json');
+const shrinkwrapFile = require(NPM_SHRINKWRAP_FILE_PATH);
+const shrinkwrapIgnore = require('../package.json').shrinkwrapIgnore;
```
jviotti: We need the path to the shrinkwrap file at the bottom of the script (in order to write to it), so by extracting it to a constant, we don't have to duplicate it.

scripts/clean-shrinkwrap.js
jviotti: Yeah, let me clarify the whole thing. I'll do some variable renaming, plus extend the docstring to explain what's going on.

jviotti:
> So, for those of us that aren't NPM packaging experts, would you mind explaining in more detail what this does, how it's different to the previous approach, and how it solves the problem that we had previously?
Doesn't it still mean that optional dependencies will be missing from npm-shrinkwrap.json, and thus won't be installed on 'fresh' systems?

This script parses the shrinkwrap files and removes entries that match a certain blacklist that we pass. The reason why we need to do this is that we have to avoid putting platform dependent optional dependencies in the shrinkwrap file, since npm will insist in installing them even if the desired OS doesn't match.

So optional dependencies are happily shrinkwrapped, UNLESS they are platform specific (e.g. the package.json declares that it only works on linux, win32, or darwin).

The previous approach deleted the dependencies by running `npm rm`, which removes them from the shrinkwrap file AND from `node_modules`, while now we manually remove them only from the shrinkwrap file.

jviotti:
> For some reason both the Linux 32-bit and 64-bit Travis tests are passing, but the OSX Travis PR build is failing the sanity-check with:

Looks like these are dependencies of one of the platform specific optional dependency. I'll take care of it in a minute.

scripts/clean-shrinkwrap.js
lurch: Hmmm, but we could still use just `const NPM_SHRINKWRAP_FILE_PATH = '../npm-shrinkwrap.json';` ?
Not that it really matters of course...

lurch:
Ahh, so platform-specific optional dependencies _will_ still get installed on the appropriate platform, even if they're not shrinkwrapped?

Does this PR address the issue I alluded to on #1510 about ensuring that packages only needed to create the Etcher installers don't get included in Etcher's `.asar`?

scripts/clean-shrinkwrap.js
jviotti: Its a best practice to make use of `path.join` for any path concatenation task, given it will use the correct delimiters for each OS.

jviotti:
OK, I went hardcore on this. Turns out all the dependencies we were listing in `shrinkwrapIgnore` are all dependencies of a single platform specific dependency: `appdmg`. I reduced the list to just `appdmg`, and implemented logic to recursively discover the dependency tree of a certain dependency, using the npm registry as a reference.

Based in all this information, we remove the entries that match the dependency tree and that are optional (which means they are only used by an optional dependency), and for development purposes (since we don't allow production optional dependencies).

jviotti:
I'm thinking that we could make use of this little script to investigate the dependency tree of all our top level dependencies, as a way to see what dependencies are the ones that bring most dependencies, etc

scripts/clean-shrinkwrap.js
lurch: ...in which case (for consistency) should you do the same thing for `'../package.json'` too?

lurch:
> /Users/travis/build/resin-io/etcher/scripts/clean-shrinkwrap.js
  27:7  error  'packageJSON' is assigned a value but never used  no-unused-vars
  28:7  error  'spawn' is assigned a value but never used        no-unused-vars

Tut tut @jviotti ;-)

scripts/clean-shrinkwrap.js
```diff
@@ -1,43 +1,166 @@
...
+    })
+    .map(([ name, version ]) => {
+      return `${name}@${version}`;
+    })
+    .uniq()
+    .value();
+});
+
+/**
+ * @summary Recursively fetch the dependency tree of an npm specified
```
lurch: specified -> specifier

scripts/clean-shrinkwrap.js
```diff
@@ -1,43 +1,166 @@
...
+/**
+ * @summary Get the top level dependencies of a certain npm specifier
+ * @function
+ * @private
+ *
+ * @param {String} specifier - npm specifier
+ * @returns {String[]} top level dependencies
+ *
+ * @example
+ * const dependencies = getTopLevelDependencies('drivelist@^5.0.0');
```
lurch: Does this need to be modified to fetch details about a specific version, rather than a 'version range' (or whatever the correct terminology is) ?
Compare:
```
$ npm view drivelist@^5.0.0 dependencies
drivelist@5.0.0 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.1 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.2 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.3 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.4 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.5 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.6 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.7 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.8 { 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.9 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.10 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.11 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.12 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.13 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.14 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.15 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.16 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.17 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.18 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.19 { debug: '^2.6.0', 'js-yaml': '^3.4.1', lodash: '^4.16.4' }
drivelist@5.0.20  
{ bluebird: '^3.5.0',
  debug: '^2.6.0',
  'js-yaml': '^3.4.1',
  lodash: '^4.16.4',
  tmp: '0.0.31' }
drivelist@5.0.21  
{ bluebird: '^3.5.0',
  debug: '^2.6.0',
  'js-yaml': '^3.4.1',
  lodash: '^4.16.4',
  tmp: '0.0.31' }
drivelist@5.0.22  
{ bluebird: '^3.5.0',
  debug: '^2.6.0',
  'js-yaml': '^3.4.1',
  lodash: '^4.16.4',
  tmp: '0.0.31' }
```
with:
```
$ npm view drivelist@5.0.22 dependencies
{ bluebird: '^3.5.0',
  debug: '^2.6.0',
  'js-yaml': '^3.4.1',
  lodash: '^4.16.4',
  tmp: '0.0.31' }
```

Obviously it's possible that different versions of a module might have distinct dependencies, and we only want the dependency-list for the _actual_ version of the module we're using?
E.g. `somemodule@1.0.2` might depend on `['othermoduleA', 'othermoduleB']`, `somemodule@1.0.3` might depend on `['othermoduleB', 'othermoduleC']` and `somemodule@1.0.4` might depend on `['othermoduleC', 'othermoduleD']`; and if we happen to be using `somemodule@1.0.3`, then when we ask for `somemodule`'s dependencies, we want to be told `['othermoduleB', 'othermoduleC']` and not  `['othermoduleA', 'othermoduleB', 'othermoduleC', 'othermoduleD']`. (if that makes sense?)

But for modules that aren't shrinkwrapped, I've no idea how difficult it is to find out the actual version we're using, so feel free to ignore me if I'm just overcomplicating things ;-)

lurch:
Once this is all working properly, should this PR also include [this](https://github.com/resin-io/etcher/pull/1510#discussion_r121718624) change?

scripts/clean-shrinkwrap.js
jviotti: Fixed!

scripts/clean-shrinkwrap.js
```diff
@@ -1,43 +1,166 @@
...
+ * @function
+ * @private
+ *
+ * @param {String} specifier - npm specifier
+ * @returns {String[]} all dependencies
+ *
+ * @example
+ * const dependencies = getDependencyTree('drivelist@^5.0.0');
+ */
+const getDependencyTree = (specifier) => {
```
jviotti: Fixed!

scripts/clean-shrinkwrap.js
jviotti: Determining which versions are used did get very tricky, given the shrinkwrap file doesn't preserve information about the packages relationship. The current approach is to get the dependencies for all versions of the packages that match the range, and join them together, however I think that the ideal approach would be to not use ranges in the package.json, given that since we're using npm-shrinkwrap.json, we don't use ranges at all (I'll push another PR with that change)

jviotti:
> Once this is all working properly, should this PR also include this change?

Done!

lurch:
> ./node_modules/.bin/npx eslint lib tests scripts bin versionist.conf.js
C:\projects\etcher\scripts\clean-shrinkwrap.js
  73:9  error  'dependencies' is assigned a value but never used  no-unused-vars


Also: having just read the commit-message again, I wonder if this PR would also be a good place to delete the `from:` lines from the `npm-shrinkwarap.json`? They're not actually needed (I think) and they caused us all kinds of conflict headaches previously...

scripts/clean-shrinkwrap.js
lurch: > I think that the ideal approach would be to not use ranges in the package.json

Hmm, I'd naturally be a bit hesitant about removing information (in this case version ranges from `package.json`) that _might_ be useful. If you want to ignore any version ranges from package.json inside this `clean-shrinkwrap.js` code, couldn't you just filter them out inside `clean-shrinkwrap.js, and leave them in place for use anywhere else?

jhermsmeier:
> I wonder if this PR would also be a good place to delete the from: lines from the npm-shrinkwarap.json? They're not actually needed (I think) and they caused us all kinds of conflict headaches previously...

Those will just be put back the next time you run `npm shrinkwrap`... I also don't think we should touch auto-generated files, unless necessary.

lurch:
> Those will just be put back the next time you run `npm shrinkwrap`... 

...but then automatically removed again by this `postshrinkwrap` script? ;-)

> I also don't think we should touch auto-generated files, unless necessary.

The problem with the `from:` lines is that if Etcher developer "Bob" does `make electron-develop` the `from:` lines also get stored somewhere (I dunno the full details) in his `node_modules` directory. So if Etcher developer "Charlie" later installs a later version of a module, that might change that module's `from:` line in `npm-shrinkwrap.json`, which Charlie will then also commit to the repo. And so when Bob does a `git pull` and another `make electron-develop`, Bob will get the later version of the module, but the `from:` line in `npm-shrinkwrap.json` will then be reset back to what Bob has in his `node_modules` directory, and git will shout at him that there's now a change in `npm-shrinkwrap.json`, even though that's not what Bob intended. After _**much**_ confusion and head-scratching, and attempts at git-reverting things, Bob will _eventually_ realise (been there, done that :wink: ) that the only way around this is to `rm -rf node_modules` and **then** `make electron-develop`.  Whereas if the `from:` lines weren't in `npm-shrinkwrap.json` then there'd be no need to (sometimes) do `rm -rf node_modules` before doing `make electron-develop`.

ISTR reading that the removal of the `from:` lines was one of the improvements in NPM@5 ? (or perhaps I'm mis-remembering) But I dunno if they're actually still important in NPM@3 / NPM@4 ?

scripts/clean-shrinkwrap.js
jviotti: The issue is that if you have ranges, you can't know what precise version is being installed by the application. For example, if we have `foo@^2.0.1` in `package.json`, then the application code could be using `2.0.1`, `2.1.0`, `2.9.0`, etc which can have a completely different dependency tree.

By using a shrinkwrap file, we're basically negating the whole ranges functionality, and I can't think of any scenario where the ranges could actually be useful.

lurch: Okay, so I guess we could get rid of ranges in `package.json`. But won't the sub-packages themselves likely to be using ranges for their dependencies, re-creating the problem all over again?
As I've probably mentioned before, NPM is all a bit of a mystery to me, so please take all my comments with a large pinch of salt, and ignore them if appropriate ;-)

jviotti: Damn, you're right. I have other ideas to solve this problem. Let's see what comes out of that :)

jviotti:
Closing since I have better implementation here: https://github.com/resin-io/etcher/pull/1551. I didn't re-use this PR since the approach is completely different.

lurch:
Completely different approach, and yet an identical commit message? ;-)

jviotti:
Haha, yeah, the problem remains the same :)

PR #1515, opened 2017-06-12
refactor(GUI): replace SET_SETTING with an atomic SET_SETTINGS action

jviotti:
This commit is the first on a series of commit to incrementally
implement support for configuration files (so we avoid a huge PR like we
have at the moment).

Once of the first things we can do is replace the `SET_SETTING` redux
action with an atomic `SET_SETTINGS` action that sets all the settings
for the application at once.

The purpose of this change is that later the `SET_SETTINGS` action can
be modified to stringify all the settings and store them in a
configuration file, without having to deal with merges, conflicts, etc
(since the client application if forced to resolve those problems before
calling the `SET_SETTINGS` action.)

The behaviour of the code remains almost the same, with the exception
that the user can now set settings that we don't know about, so the user
can switch between Etcher versions without getting weird errors if one
of the configuration keys he has doesn't exist in the other version.

See: https://github.com/resin-io/etcher/pull/1382
Change-Type: patch

### DIFF
--- a/lib/gui/models/settings.js
+++ b/lib/gui/models/settings.js
@@ -20,7 +20,9 @@
  * @module Etcher.Models.Settings
  */
 
+const _ = require('lodash');
 const Store = require('./store');
+const errors = require('../../shared/errors');
 
 /**
  * @summary Set a setting value
@@ -34,12 +36,25 @@ const Store = require('./store');
  * settings.set('unmountOnSuccess', true);
  */
 exports.set = (key, value) => {
+  if (_.isNil(key)) {
+    throw errors.createError({
+      title: 'Missing setting key'
+    });
+  }
+
+  if (!_.isString(key)) {
+    throw errors.createError({
+      title: `Invalid setting key: ${key}`
+    });
+  }
+
+  const newSettings = _.assign(exports.getAll(), {
+    [key]: value
+  });
+
   Store.dispatch({
-    type: Store.Actions.SET_SETTING,
-    data: {
-      key,
-      value
-    }
+    type: Store.Actions.SET_SETTINGS,
+    data: newSettings
   });
 };
 
--- a/lib/gui/models/store.js
+++ b/lib/gui/models/store.js
@@ -78,7 +78,7 @@ const ACTIONS = _.fromPairs(_.map([
   'SELECT_IMAGE',
   'REMOVE_DRIVE',
   'REMOVE_IMAGE',
-  'SET_SETTING'
+  'SET_SETTINGS'
 ], (message) => {
   return [ message, message ];
 }));
@@ -451,35 +451,40 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
       return state.deleteIn([ 'selection', 'image' ]);
     }
 
-    case ACTIONS.SET_SETTING: {
-      const key = action.data.key;
-      const value = action.data.value;
-
-      if (!key) {
+    case ACTIONS.SET_SETTINGS: {
+      if (!action.data) {
         throw errors.createError({
-          title: 'Missing setting key'
+          title: 'Missing settings'
         });
       }
 
-      if (!_.isString(key)) {
+      if (!_.isPlainObject(action.data)) {
         throw errors.createError({
-          title: `Invalid setting key: ${key}`
+          title: `Invalid settings: ${action.data}`
         });
       }
 
-      if (!DEFAULT_STATE.get('settings').has(key)) {
+      const invalidKey = _.find(_.keys(action.data), (key) => {
+        return !_.isString(key);
+      });
+
+      if (!_.isNil(invalidKey)) {
         throw errors.createError({
-          title: `Unsupported setting: ${key}`
+          title: `Invalid setting key: ${invalidKey}`
         });
       }
 
-      if (_.isObject(value)) {
+      const invalidPair = _.find(_.toPairs(action.data), (pair) => {
+        return _.isObject(_.last(pair));
+      });
+
+      if (!_.isNil(invalidPair)) {
         throw errors.createError({
-          title: `Invalid setting value: ${value}`
+          title: `Invalid setting value: ${_.last(invalidPair)} for ${_.first(invalidPair)}`
         });
       }
 
-      return state.setIn([ 'settings', key ], value);
+      return state.setIn([ 'settings' ], Immutable.fromJS(action.data));
     }
 
     default: {
--- a/tests/gui/models/settings.spec.js
+++ b/tests/gui/models/settings.spec.js
@@ -9,20 +9,20 @@ describe('Browser: settings', function() {
 
   describe('settings', function() {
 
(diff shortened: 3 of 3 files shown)


lib/gui/models/store.js
```diff
@@ -451,35 +451,37 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
-      if (!_.isString(key)) {
+      if (!_.isPlainObject(action.data)) {
         throw errors.createError({
-          title: `Invalid setting key: ${key}`
+          title: `Invalid settings: ${action.data}`
         });
       }
 
-      if (!DEFAULT_STATE.get('settings').has(key)) {
+      const invalidKey = _.find(_.keys(action.data), (key) => {
```
lurch: Should `invalidKey` be named `invalidKeys` ?

EDIT: Ah, I've just read up and now see that `_.find` only finds the first match, not all matches :)

lib/gui/models/store.js
```diff
@@ -451,35 +451,37 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
+
+      if (!_.isNil(invalidKey)) {
         throw errors.createError({
-          title: `Unsupported setting: ${key}`
+          title: `Invalid setting key: ${invalidKey}`
         });
       }
 
-      if (_.isObject(value)) {
+      const invalidValue = _.find(_.values(action.data), _.overSome([ _.isArray, _.isObject ]));
```
lurch: Why did it previously only check `_.isObject` but now it's checking both `_.isArray` and `_.isObject` ?

lib/gui/models/settings.js
```diff
@@ -34,12 +36,25 @@ const Store = require('./store');
  * settings.set('unmountOnSuccess', true);
  */
 exports.set = (key, value) => {
```
lurch: Should this module grow a function to set multiple settings at once? Or is that something for a later PR?

tests/gui/models/settings.spec.js
```diff
@@ -33,10 +33,10 @@ describe('Browser: settings', function() {
 
     describe('.set()', function() {
 
-      it('should throw if the key is not supported', function() {
```
lurch: Given that 'not supported keys' are no longer a thing, maybe it makes sense for `SUPPORTED_KEYS` to be renamed to something like `DEFAULT_KEYS`?

lib/gui/models/settings.js
jviotti: I guess we can add that function once we need it.

lib/gui/models/store.js
jviotti: Amazing catch. This is redundant, since an array is always considered an object. Fixed!

tests/gui/models/settings.spec.js
jviotti: Good point. Done!

jviotti:
@lurch How do you feel about this patch now?

lib/gui/models/store.js
```diff
@@ -451,35 +451,37 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
+          title: `Invalid setting key: ${invalidKey}`
         });
       }
 
-      if (_.isObject(value)) {
+      const invalidValue = _.find(_.values(action.data), _.isObject);
+      if (!_.isNil(invalidValue)) {
         throw errors.createError({
-          title: `Invalid setting value: ${value}`
+          title: `Invalid setting value: ${invalidValue}`
```
lurch: It's something that the code didn't do previously, but to aid with debugging (should the problem ever occur), I wonder if it'd be useful to also include _which_ `key` we were attempting to set to `invalidValue` in this error message?

tests/gui/models/settings.spec.js
```diff
@@ -9,20 +9,20 @@ describe('Browser: settings', function() {
...
     afterEach(function() {
-      _.each(SUPPORTED_KEYS, (supportedKey) => {
+      _.each(DEFAULT_KEYS, (supportedKey) => {
         settings.set(supportedKey, this.settings[supportedKey]);
       });
     });
 
     it('should be able to set and read values', function() {
-      const keyUnderTest = _.first(SUPPORTED_KEYS);
+      const keyUnderTest = _.first(DEFAULT_KEYS);
```
lurch: Given the number of occurrences of `const keyUnderTest = _.first(DEFAULT_KEYS);`, would it be worth moving it to e.g. just under `const DEFAULT_KEYS = ...` ?

lurch:
With the talk about "the user can switch between Etcher versions without getting weird errors if one of the configuration keys he has doesn't exist in the other version.", does that mean that this PR might be a good place to address https://github.com/resin-io/etcher/pull/1382#issuecomment-299179917

jviotti:
Yeah, I'll try to reproduce here and see if I can include a fix.

tests/gui/models/settings.spec.js
jviotti: Even better: instead of always using the first available setting for testing purposes, lets pick one of them at random. What do you think?

lib/gui/models/store.js
jviotti: Sounds good, check my latest changes!

lurch:
Is the mention of `reconfix` in the commit-message a typo?

lib/gui/models/settings.js
```diff
@@ -34,12 +36,25 @@ const Store = require('./store');
...
+  }
+
+  if (!_.isString(key)) {
+    throw errors.createError({
+      title: `Invalid setting key: ${key}`
+    });
+  }
+
+  const newSettings = _.assign(exports.getAll(), _.fromPairs([
+    [ key, value ]
```
Shou: I don't know which is preferable but you can also do this, and remove the `fromPairs` usage.

```js
const newSettings = _.assign(exports.getAll(), {
  [key]: value
});
```

lurch: I was looking at this line, and suspected there must be a simpler way :+1: 

Ah no wait, it was https://github.com/resin-io/etcher/pull/1515/files#diff-74bcbd72e100aecd2f68e833e7a169f4R71 that I was suspecting could be simplified.

jviotti: That's a cool suggestion. I didn't know such syntax sugar existed. Let me check if it works here...

jviotti:
> Is the mention of reconfix in the commit-message a typo?

Damn, I always confuse Reconfix and Redux :P I'll fix it.

lurch:
>>  does that mean that this PR might be a good place to address https://github.com/resin-io/etcher/pull/1382#issuecomment-299179917

> Yeah, I'll try to reproduce here and see if I can include a fix.

So was there no (easy) fix for this then?

jviotti:
Let's address it when moving to a configuration file.

lurch:
> Let's address it when moving to a configuration file.

Okay. I guess with a configuration file this becomes even more important, since the users can edit the config file, so we need to be able to assume sensible defaults for any values the user deliberately / accidentally removes ;-)

PR #1132, opened 2017-02-23
chore: automatic CLI packaging

jviotti:
This commit introduces the phony `package-cli` target, used to package
the Electron CLI in a directory.

Other related changes:

- The `package` target has been renamed to `package-electron`

Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

PR #1517, opened 2017-06-13
chore: fix version/platform order inconsistencies in package names

jviotti:
See: https://github.com/resin-io/etcher/pull/1132#discussion_r121654527


Makefile
```diff
@@ -275,7 +275,7 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERS
 	| $(BUILD_DIRECTORY)
 	./scripts/build/concatenate-javascript.sh -e lib/cli/etcher.js -b $< -o $@ -m
 
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH): \
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(TARGET_PLATFORM)-$(APPLICATION_VERSION)-$(TARGET_ARCH): \
```
jhermsmeier: Isn't the convention usually to have it in the order `name-version-platform-arch`?

lurch: I've no opinion on which is "correct", but https://github.com/electron/electron/releases uses `name-version-platform-arch` and https://etcher.io uses `name-version-platform-arch` and https://etcher.io/cli/ uses `name-version-platform-arch`, so it probably makes sense to stick to that convention here in the `Makefile`? ;)

jviotti: I made sure that convention is honored in the whole Makefile. Check the latest changes.

PR #1518, opened 2017-06-13
chore: generate single-binary portable installers on Windows

jviotti:
We currently support portable builds that are basically ZIPs containing
the main Etcher executable and all its related libraries.

Turns out `electron-builder` supports NSIS-based portable builds that
can create a single executable that has everything it needs to run,
including any external assets.

This commit makes use of this new portable Windows installer
functionality, replacing the old ZIP approach.

Change-Type: patch
Changelog-Entry: Generate single-binary portable installers on Windows.


Makefile
```diff
@@ -432,25 +411,18 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_V
...
 		-c scripts/build/redhat/config.json
 
-$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).zip: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH) \
+$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH)-portable.exe: \
 	| $(BUILD_OUTPUT_DIRECTORY)
-	./scripts/build/zip-file.sh -f $< -s $(TARGET_PLATFORM) -o $@
+	CSC_LINK=$(CODE_SIGN_CERTIFICATE) CSC_KEY_PASSWORD=$(CODE_SIGN_CERTIFICATE_PASSWORD) TARGET_ARCH=$(TARGET_ARCH) \
+		$(NPX) build --win portable $(ELECTRON_BUILDER_OPTIONS)
+	mv $(BUILD_DIRECTORY)/$(notdir $@) $@
```
jhermsmeier: Why is this being moved out of the `build/` dir? Could we maybe add a comment explaining that?

Makefile
```diff
@@ -432,25 +411,18 @@ $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-electron_$(APPLICATION_V
 	./scripts/build/electron-installer-redhat-linux.sh -p $< -r "$(TARGET_ARCH)" -o $| \
 		-c scripts/build/redhat/config.json
 
-$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).zip: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH) \
+$(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH)-portable.exe: \
 	| $(BUILD_OUTPUT_DIRECTORY)
-	./scripts/build/zip-file.sh -f $< -s $(TARGET_PLATFORM) -o $@
+	CSC_LINK=$(CODE_SIGN_CERTIFICATE) CSC_KEY_PASSWORD=$(CODE_SIGN_CERTIFICATE_PASSWORD) TARGET_ARCH=$(TARGET_ARCH) \
```
lurch: Does this still work properly if `CODE_SIGN_CERTIFICATE` and `CODE_SIGN_CERTIFICATE_PASSWORD` haven't been set?

Makefile
```diff
@@ -523,15 +495,15 @@ PUBLISH_BINTRAY_REDHAT += \
 endif
 
 ifeq ($(TARGET_PLATFORM),win32)
-electron-installer-zip: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
+electron-installer-portable: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH)-portable.exe
```
lurch: The change from `$(TARGET_PLATFORM)` to `win32` might be redundant, given that no other platform uses `.exe` files?
Or perhaps this change does make sense for consistency reasons? \*shrug\*

Makefile
```diff
@@ -523,15 +495,15 @@ PUBLISH_BINTRAY_REDHAT += \
...
 electron-installer-nsis: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-win32-$(TARGET_ARCH).exe
 cli-installer-zip: $(BUILD_OUTPUT_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
 TARGETS += \
-	electron-installer-zip \
+	electron-installer-portable \
```
lurch: Does this change mean that the build system will stop creating Linux zip files? Perhaps we should add a `electron-installer-portable` target and not remove `electron-installer-zip` ?

lurch:
> Turns out electron-builder supports NSIS-based portable builds that
can create a single executable that has everything it needs to run,
including any external assets.

Cool! :+1:

Makefile
jviotti: `electron-builder` outputs packages in `dist` (e.g. `BUILD_DIRECTORY`). The final path of his Make target is under `BUILD_OUT_DIRECTORY`. This little line simply moves the file from `BUILD_DIRECTORY` to `BUILD_OUT_DIRECTORY`, so it matches our current directory structure. We will eventually get rid of this plus `BUILD_OUT_DIRECTORY` once the transition process is over.

Makefile
jviotti: Yep, I confirm it works fine.

Makefile
jviotti: Oh, good catch. I'll unify things in this PR.

Makefile
jviotti: We don't generate zip Linux archives, so its removal makes no difference.

lurch: We do, but I've now realised that they have a separate `electron-installer-appimage` Makefile target, so the removal of `electron-installer-zip` is fine :)

lurch:
Will this approach still work properly with e.g. #1382 ?

jviotti:
Yep, the application executable is standalone, but it can still write to any location.

lurch:
Interesting. So does that mean that there'll be a 'default' `.etcherrc` included inside the single-binary executable? Or will the `.etcherrc` simply get written to a "pre-defined location" the first time that Etcher starts up?
I think people might expect the `.etcherrc` for the portable version of Etcher to e.g. be written to the same directory as the Etcher executable, rather than to some location on their hard-disk (that the 'installable' version of Etcher would use)?

jviotti:
Only the assets needed to run the application are embedded with the binary, which excludes any external location like `.etcherrc`. 

> I think people might expect the .etcherrc for the portable version of Etcher to e.g. be written to the same directory as the Etcher executable, rather than to some location on their hard-disk (that the 'installable' version of Etcher would use)?

Yeah, that's a good point. Let's detect if a version is portable, and change the location of the configuration file (once that is merged).

PR #1520, opened 2017-06-15
chore: make use of electron-builder to build GNU/Linux packages

jviotti:
Change-Type: patch


Makefile
```diff
@@ -142,6 +145,17 @@ endif
 ELECTRON_BUILDER_OPTIONS = --$(TARGET_ARCH_ELECTRON_BUILDER) --extraMetadata.version=$(APPLICATION_VERSION)
 
 # ---------------------------------------------------------------------
+# Updates
+# ---------------------------------------------------------------------
+
+DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS = --extraMetadata.analytics.updates.enabled=false
```
lurch: Should something similar be done for the other `--extraMetadata.` flags too?

docs/PUBLISHING.md
```diff
@@ -96,12 +96,6 @@ make electron-installer-zip
 make electron-installer-nsis
 ```
 
-Disabling updates
------------------
-
-You can disable updates in the final artifact by passing `DISABLE_UPDATES=1` to
-the `make` targets described above.
```
lurch: Why was this removed? `Makefile` is still checking for `DISABLE_UPDATES` ?

lib/gui/models/store.js
```diff
@@ -539,7 +545,7 @@ module.exports = _.merge(redux.createStore(
       // is not very elegant, nor performant, so we decide
       // to intentionally ignore this little flaw since
       // adding extra properties makes no damage at all.
-      return state.set(PERSISTED_PATH, subset);
+      return state.set(PERSISTED_PATH, state.get(PERSISTED_PATH).merge(subset));
```
lurch: When https://github.com/electron-userland/electron-builder/issues/1674 gets fixed, should both changes to this file be reverted?

lurch:
Wow, amazing to see the removal of such a huge amount of custom code! :tada: :100: 

It's a bit of an overly-broad question, but what's the relationships (if any) between electron-installer-debian, electron-installer-redhat, AppImageKit, AppImageAssistant, electron-builder, electron-packager, electron-forge, electron-prebuilt, etc. ?

scripts/build/docker/Dockerfile.template
```diff
@@ -16,6 +16,8 @@ RUN apt-get update \
     curl \
     fuse \
     git \
+    graphicsmagick \
```
lurch: Hmmm, how come?

develar:
> what's the relationships (if any) between electron-installer-debian, electron-installer-redhat, AppImageKit, AppImageAssistant, electron-builder, electron-packager, electron-forge, electron-prebuilt, etc. 

electron-builder decided to not reinvent the wheel and rely on proven and tested existing solution — [FPM](https://github.com/jordansissel/fpm).

> AppImageKit, AppImageAssistant

electron-builder solution was developed in cooperation with AppImage author. AppImageAssistant it is only UI.

> electron-forge

https://github.com/electron-userland/electron-builder/issues/1193 Duplicates electron-builder and webpack :)

scripts/build/docker/Dockerfile.template
develar: I suggest to extend electron-builder Docker image. `electronuserland/electron-builder:latest` https://github.com/electron-userland/electron-builder/wiki/Docker

lurch: Does that support both x64 and x86? (and ARMv7 support too would be fantastic :smile: )

lurch:
Thanks for the explanations @develar :+1:

scripts/build/docker/Dockerfile.template
develar: @lurch You can build ia32 on x64 Linux. The same, I guess, applies for ARMv7.

`gcc-multilib g++-multilib` is preinstalled in the electron-builder docker images.

develar:
@lurch BTW, I don't recommend electron-compile because of https://github.com/electron/electron-compile/issues/221 and because in the wild things a "little" bit complicated and no need to yet another tool if we have so awesome tool as webpack (e.g. fork-ts-checker-webpack-plugin solution exists only for webpack and so on) It is another reason why electron-forge is a little bit odd for me (regardless of this electron-compile is supported by electron-builder, but because of these reasons, we do not promote electron-compile/forge).

scripts/build/docker/Dockerfile.template
lurch: We use our CI Docker images for both building and testing... (and IIRC we had some problem previously when we tried building the ia32 image on a x64 host? I think some part of the build process mis-identified the ARCH and built a x64 native-module for the x86 Etcher)

lurch:
Yeah, I agree that it would seem to make sense to use electron-compile purely for development (which we don't, because we don't need any of it's features) and then transpile / bundle just to pure JS as part of the build process for release, rather than doing it at runtime.

scripts/build/docker/Dockerfile.template
develar: > mis-identified the ARCH and built a x64 native-module for the x86 Etcher

Solution `gcc-multilib g++-multilib` is tested and proven to work. I am not 100%, but you can check :) (using electron-builder, of course).

> and testing

Our docker image contains Chrome to test ;)

develar:
> to make sense to use electron-compile purely for development (which we don't, because we don't need any of its features) 

Electorn-compile is not suitable for development as well, because when you see webpack features and existing plugins... development with webpack is much more convenient.

scripts/build/docker/Dockerfile.template
lurch: Even `gcc-multilib` can't stop a not-cross-compilation-aware `configure` script setting things up wrong ;-)

WRT testing, I meant that we test that a 32-bit image runs on a pure 32-bit host. I think if we compiled on a 64-bit host, and ended up with a '32-bit' image which had a mix of 32-bit and 64-bit parts, then trying to test it on a 64-bit host would still "work"?

lurch:
Fair enough, I'm still a bit of a newbie to the whole JS "ecosystem", so most of this stuff goes over my head ;-)  And that's why I sometimes ask "stupid questions".

scripts/build/docker/Dockerfile.template
develar: > can't stop a not-cross-compilation-aware configure script setting things up wrong ;-)

Yeah, if you have some native deps that you build outside of node-gyp...

Ok, I stopped promoting electron-builder docker image :) (currently, I don't have time to test and add ia32 version of image).

PR #1521, opened 2017-06-15
chore(travis): remove unused afsctool macOS dependency

jviotti:


jviotti:
Good catch. Fixed!

Issue #1499, opened 2017-06-08
unsupported compression method: 9


lurch:
- **Etcher version:** 1.0.0
- **Operating system and architecture:** Any
- **Image flashed:** http://www.protoneer.com/files/RPI-CNC-V400.zip
- **Do you see any meaningful error information in the DevTools?**
```
document not ready yet, trying again in 500 milliseconds...
/usr/lib/etcher-electron/resources/app.asar/node_modules/raven-js/dist/raven.js:51 Download the React DevTools and use an HTTP server (instead of a file: URL) for a better development experience: https://fb.me/react-devtools
/usr/lib/etcher-electron/resources/app.asar/node_modules/raven-js/dist/raven.js:51  _____ _       _
|  ___| |     | |
| |__ | |_ ___| |__   ___ _ __
|  __|| __/ __| '_ \ / _ \ '__|
| |___| || (__| | | |  __/ |
\____/ \__\___|_| |_|\___|_|

Interested in joining the Etcher team?
Drop us a line at join+etcher@resin.io
/usr/lib/etcher-electron/resources/app.asar/node_modules/raven-js/dist/raven.js:51 Thu Jun 08 2017 12:11:11 GMT+0100 (BST) Application start
/usr/lib/etcher-electron/resources/app.asar/node_modules/raven-js/dist/raven.js:51 Thu Jun 08 2017 12:11:11 GMT+0100 (BST) Not checking for updates ({"shouldCheckForUpdates":false,"disableUpdatesEnvironmentVariable":"1","releaseType":"PRODUCTION"})
/usr/lib/etcher-electron/resources/app.asar/node_modules/raven-js/dist/raven.js:51 Thu Jun 08 2017 12:11:13 GMT+0100 (BST) Open image selector
/usr/lib/etcher-electron/resources/app.asar/node_modules/raven-js/dist/raven.js:51 {}
console.(anonymous function) @ /usr/lib/etcher-electron/resources/app.asar/node_modules/raven-js/dist/raven.js:51
```
As reported by @htcnish on [gitter](https://gitter.im/resin-io/etcher?at=59386a34e531dbc905bf0e59), when selecting the `RPI-CNC-V400.zip` file, Etcher displays:
![image](https://user-images.githubusercontent.com/476186/26925905-6ca6c36a-4c43-11e7-9061-c87ee22143b2.png)

[...]

PR #1535, opened 2017-06-20
test(image-stream): Fix test image data

jhermsmeier:
This updates the base test image to be a complete, but small
disk image, as thre previous image was quite large for running tests (32MB),
and was chopped off at an arbitrary position, causing other tools to fail
on it as an input.

Change-Type: patch


tests/image-stream/metadata/zip.spec.js
```diff
@@ -127,19 +126,18 @@ describe('ImageStream: Metadata ZIP', function() {
...
-      '    <BlocksCount> 9 </BlocksCount>',
-      '    <MappedBlocksCount> 4     </MappedBlocksCount>',
-      '    <BmapFileSHA1> d90f372215cbbef8801caca7b1dd7e587b2142cc </BmapFileSHA1>',
+      '    <BlocksCount> 1280 </BlocksCount>',
+      '    <MappedBlocksCount> 1280 </MappedBlocksCount>',
+      '    <BmapFileSHA1> cc6f077565c73a46198777b259c231875df4e709 </BmapFileSHA1>',
       '    <BlockMap>',
-      '        <Range sha1="193edb53bde599f58369f4e83a6c5d54b96819ce"> 0-1 </Range>',
-      '        <Range sha1="193edb53bde599f58369f4e83a6c5d54b96819ce"> 7-8 </Range>',
+      '        <Range sha1="7b7d6e1fc44ef224a8c57d3ec6ffc3717c428a14"> 0-1280 </Range>',
```
lurch: Hmmm, so I guess there's not actually any unmapped blocks, which wouldn't make this a good test of bmap-writing?

lurch:
For the compressed and archive files, should we have multiple files testing different compression methods and ratios? See also #1499 

And (just for reference purposes) perhaps it'd be useful to have a text file listing the compressed and uncompressed size of each of the test-images, and for the archive files (i.e. `.zip`) you could list each archive's directory tree?

tests/image-stream/metadata/zip.spec.js
jhermsmeier: Yeah, that's because I couldn't get Tizen's bmap-tool to function, it kept complaining about `ioctl` not returning something it wanted. Handcrafted it instead. The test is about the discovering & reading the metadata, not blockmapping – and since this is only a 5MB disk image, this wouldn't be a good test for blockmaps either way – so this doesn't really matter imo.

jhermsmeier:
> For the compressed and archive files, should we have multiple files testing different compression methods and ratios? See also #1499

I'm not sure that makes sense to me – at that point we're starting to test the compression algorithms... Regarding #1499, yeah, I think that'd be good to create a user error in that case and test that. Better suited to be done in a separate PR.

lurch:
> I'm not sure that makes sense to me – at that point we're starting to test the compression algorithms... 

But if we ever switch decompression libraries, we'll be sure that the new library (and therefore Etcher) is able to decompress all the compression types and ratios that the old library supported? \*shrug\*

> Regarding #1499, yeah, I think that'd be good to create a user error in that case and test that. Better suited to be done in a separate PR.

Oh yeah, I agree that that should be done in a separate PR, I was just using it as an example of a decompression library not supporting every single possible compression type ;-)

jviotti:
> But if we ever switch decompression libraries, we'll be sure that the new library (and therefore Etcher) is able to decompress all the compression types and ratios that the old library supported? *shrug*

@jhermsmeier We should definitely at least give the compression algorithms a pass, and ensure the result matches the original image. We've caught various issues this way, and is usually what helps us choose a compression library.

jhermsmeier:
> But if we ever switch decompression libraries, we'll be sure that the new library (and therefore Etcher) is able to decompress all the compression types and ratios that the old library supported?

Oh yes, sure that makes sense, that didn't cross my mind.

PR #1539, opened 2017-06-22
feat(GUI): make size units closest relative

Shou:
We make the size units used the closest relative unit through a new
filter `closestUnit` replacing the old `gigabyte` filter.

Changelog-Entry: Make the size units the closest relative.

### DIFF
--- a/lib/gui/utils/byte-size/byte-size.js
+++ b/lib/gui/utils/byte-size/byte-size.js
@@ -29,7 +29,7 @@ const ByteSize = angular.module(MODULE_NAME, []);
 
 /* eslint-disable lodash/prefer-lodash-method */
 
-ByteSize.filter('gigabyte', require('./filters/gigabyte'));
+ByteSize.filter('closestUnit', require('./filters/closest-unit'));
 
 /* eslint-enable lodash/prefer-lodash-method */
 
--- a/lib/gui/utils/byte-size/byte-size.js
+++ b/lib/gui/utils/byte-size/byte-size.js
@@ -29,7 +29,7 @@ const ByteSize = angular.module(MODULE_NAME, []);
 
 /* eslint-disable lodash/prefer-lodash-method */
 
-ByteSize.filter('closestUnit', require('./filters/closest-unit'));
+ByteSize.filter('closestUnit', require('./filter.js'));
 
 /* eslint-enable lodash/prefer-lodash-method */
 
--- a/lib/shared/units.js
+++ b/lib/shared/units.js
@@ -17,6 +17,25 @@
 'use strict';
 
 /**
+ * @summary A number precision of 4, allowing at most 4 digits.
+ * @constant
+ * @private
+ * @type {Number}
+ */
+const PRECISION_FOUR = 4;
+
+/**
+ * @summary Terabyte to byte ratio
+ * @constant
+ * @private
+ * @type {Number}
+ *
+ * @description
+ * 1 TB = 1e+3 B
+ */
+const TERABYTE_TO_BYTE_RATIO = 1e+12;
+
+/**
  * @summary Gigabyte to byte ratio
  * @constant
  * @private
@@ -39,6 +58,17 @@ const GIGABYTE_TO_BYTE_RATIO = 1e+9;
 const MEGABYTE_TO_BYTE_RATIO = 1e+6;
 
 /**
+ * @summary Kilobyte to byte ratio
+ * @constant
+ * @private
+ * @type {Number}
+ *
+ * @description
+ * 1 KB = 1e+3 B
+ */
+const KILOBYTE_TO_BYTE_RATIO = 1e+3;
+
+/**
  * @summary Milliseconds in a day
  * @constant
  * @private
@@ -48,21 +78,6 @@ const MEGABYTE_TO_BYTE_RATIO = 1e+6;
  * From 24 * 60 * 60 * 1000
  */
 const MILLISECONDS_IN_A_DAY = 86400000;
-
-/**
- * @summary Convert bytes to gigabytes
- * @function
- * @public
- *
- * @param {Number} bytes - bytes
- * @returns {Number} gigabytes
- *
- * @example
- * const result = units.bytesToGigabytes(7801405440);
- */
-exports.bytesToGigabytes = (bytes) => {
-  return bytes / GIGABYTE_TO_BYTE_RATIO;
-};
 
 /**
  * @summary Convert bytes to megabytes
@@ -77,6 +92,37 @@ exports.bytesToGigabytes = (bytes) => {
  */
 exports.bytesToMegabytes = (bytes) => {
   return bytes / MEGABYTE_TO_BYTE_RATIO;
+};
+
+/**
+ * @summary Convert bytes to most appropriate unit string
+ * @function
+ * @public
+ *
+ * @param {Number} bytes - bytes
+ * @param {Number} precision - number precision
+ * @returns {String} size and unit string
+ *
+ * @example
+ * const humanReadable = units.bytesToClosestUnit(7801405440);
+ */
+exports.bytesToClosestUnit = (bytes, precision = PRECISION_FOUR) => {
+  const units = [
+    [ KILOBYTE_TO_BYTE_RATIO, 'KB' ],
+    [ MEGABYTE_TO_BYTE_RATIO, 'MB' ],
+    [ GIGABYTE_TO_BYTE_RATIO, 'GB' ],
+    [ TERABYTE_TO_BYTE_RATIO, 'TB' ]
+  ];
+
+  let result = `${bytes} B`;
+
+  for (const [ size, unit ] of units) {
+    if (size <= bytes) {
+      result = `${(bytes / size).toPrecision(precision)} ${unit}`;
+    }
+  }
+
+  return result;
 };
 
 /**
--- a/lib/shared/units.js
+++ b/lib/shared/units.js
@@ -16,35 +16,9 @@
 
 'use strict';
 
-/**
- * @summary A number precision of 4, allowing at most 4 digits.
- * @constant
- * @private
- * @type {Number}
- */
-const PRECISION_FOUR = 4;
+const _ = require('lodash');
 
-/**
- * @summary Terabyte to byte ratio
- * @constant
- * @private
- * @type {Number}
- *
- * @description
- * 1 TB = 1e+3 B
- */
-const TERABYTE_TO_BYTE_RATIO = 1e+12;
-
-/**
- * @summary Gigabyte to byte ratio
- * @constant
- * @private
- * @type {Number}
- *
- * @description
- * 1 GB = 1e+9 B
- */
-const GIGABYTE_TO_BYTE_RATIO = 1e+9;
+const prettyBytes = require('pretty-bytes');
 
 /**
  * @summary Megabyte to byte ratio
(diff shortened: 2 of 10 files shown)


Shou:
What do you think of this one?

jhermsmeier:
What do you think about just using [pretty-bytes](https://github.com/sindresorhus/pretty-bytes)?

Shou:
I think we were trying to reduce the amount of packages we use, right? If not, I'd be completely down with using that instead.

jhermsmeier:
Yeah, true – I just spotted that it has hardcoded 3 decimal precision, and I think it'd be better to reduce the precision to only one decimal `7.948 GB` looks more like 7900 GB at first glance

jviotti:
This PR is great. I do think the `units.js` logic is starting to get too hairy, so let's try to find a module that does this for us, and that has ideally no dependencies.

Shou:
`pretty-bytes` doesn't seem to have any dependencies, so I'm guessing you're good with that?

lurch:
I dunno how it compares to `pretty-bytes`, but way back in #1021 I suggested using https://www.npmjs.com/package/byte-size

npm-shrinkwrap.json
```diff
@@ -1208,8 +1205,7 @@
     "currently-unhandled": {
       "version": "0.4.1",
       "from": "currently-unhandled@>=0.4.1 <0.5.0",
-      "resolved": "https://registry.npmjs.org/currently-unhandled/-/currently-unhandled-0.4.1.tgz",
-      "dev": true
+      "resolved": "https://registry.npmjs.org/currently-unhandled/-/currently-unhandled-0.4.1.tgz"
```
lurch: Seems to be a lot of unwanted changes here? :-/

lib/shared/units.js
```diff
@@ -50,33 +43,38 @@ const MEGABYTE_TO_BYTE_RATIO = 1e+6;
...
- * @returns {Number} gigabytes
+ * @returns {Number} megabytes
  *
  * @example
- * const result = units.bytesToGigabytes(7801405440);
+ * const result = units.bytesToMegabytes(7801405440);
  */
-exports.bytesToGigabytes = (bytes) => {
-  return bytes / GIGABYTE_TO_BYTE_RATIO;
+exports.bytesToMegabytes = (bytes) => {
```
lurch: Do we need to keep this?

lib/gui/utils/byte-size/byte-size.js
```diff
@@ -29,7 +29,7 @@ const ByteSize = angular.module(MODULE_NAME, []);
 
 /* eslint-disable lodash/prefer-lodash-method */
 
-ByteSize.filter('gigabyte', require('./filters/gigabyte'));
+ByteSize.filter('closestUnit', require('./filters/closest-unit'));
```
lurch: As there'll (presumably) now only be ever one filter, could/should `lib/gui/utils/byte-size/filters/` be removed?

npm-shrinkwrap.json
jviotti: Indeed. `pretty-bytes` has no dependencies, so there must be something going on here. Can you double check @Shou ?

lib/shared/units.js
Shou: Yup, it's used in `lib/gui/models/flash-state.js`, and since it deals directly with numbers to numbers it can't be replaced by the other, as it formats to a string.

jviotti: We could parse it as a number at the end. I agree its ugly, but I guess that if we can get rid of this function, then that's cool.

Shou: We can't really control what unit it becomes so we'd have to parse and do some ugly stuff for this to work, I can't really think of a good, clean way to use the other function exclusively.

jviotti: Fair enough :)

jviotti:
npm shrinkwrap is going crazy. Can you rebase this PR? I think that everything will work fine after my latest shrinkwrap related changes get here.

jviotti:
Hm, looks like its up to date. Damn npm ranges, I guess we'll have to update the ranges as npm suggests.

lib/shared/units.js
lurch: AFAICT (but I could be wrong), that speed only needs to be a number, because of https://github.com/resin-io/etcher/blob/master/lib/gui/pages/main/templates/main.tpl.html#L98
So could we instead change it to just be a string (created by pretty-bytes) instead, and get rid of the formatting and 'MB' in `main.tpl.html` ? This would also mean that if somebody had a really slow drive, Etcher might display "123 KB/s" instead of "0.12 MB/s", which might be nicer?

lurch:
See, @jhermsmeier ? This is why I suggested we could / should filter the `from:` lines out of `npm-shrinkwrap.json` ;-)

jhermsmeier:
> See, @jhermsmeier ? This is why I suggested we could / should filter the from: lines out of npm-shrinkwrap.json ;-)

Afaik the `from:` field is purely informational, and isn't used as the selector for the package, unless the dependency is a git repo or something else, iirc – and even then, the `resolved:` field is used for fetching. But who knows what *actually* happens inside `npm`...

lib/shared/units.js
jviotti: Yeah, that sounds like a great way to go!

lurch:
Yeah, thinking about it, I think this *is* related to @jviotti 's removal of version-ranges of `package.json`, and I still believe the commands in my previous comment (nuking `node_modules`) will fix this for @Shou 

AFAICT, what's happened is this:
1. `package.json` used to store version-ranges, e.g. `"yargs": "^4.6.0",`
2. This means that when @Shou did `make electron-develop` using the old `package.json`, `npm-shrinkwrap.json` then stored:
```json
"yargs": {
  "version": "4.7.1",
  "from": "yargs@>=4.6.0 <5.0.0",
  "resolved": "https://registry.npmjs.org/yargs/-/yargs-4.7.1.tgz"
},
```
(and this info is also cached somewhere inside `node_modules`)
3. In #1549 @jviotti changed `package.json` to read `"yargs": "4.7.1",` and `npm-shrinkwrap.json` then stored:
```json
"yargs": {
  "version": "4.7.1",
  "from": "yargs@4.7.1",
  "resolved": "https://registry.npmjs.org/yargs/-/yargs-4.7.1.tgz"
},
```
(note that the `version` is identical, it's only the `from` that has changed)
And the `ensure-staged-shrinkwrap.sh` script on the CI-servers is happy because the local `npm-shrinkwrap.json` which has been freshly-updated from the newly-created `node_modules` directory, matches up with the `npm-shrinkwrap.json` in the repo. So far so good.
4. @Shou then did a `git pull` to update his repo (which got him the newer `npm-shrinkwrap.json` but didn't touch his `node_modules` directory), and when he next did a `make electron-develop`, npm spotted that all of the module versions in his local `node_modules` directory already matched the `version`s specified in `npm-shrinkwrap.json` and so nothing changed. So far so good.
5. However as part of this PR @Shou presumably did `npm install --save pretty-bytes@1.0.4`. npm then added `"pretty-bytes": "1.0.4",` to `package.json`, and then recreated `npm-shrinkwrap.json`. However, when recreating `npm-shrinkwrap.json` it used the `from` lines cached inside `node_modules` (ignoring the `from` lines in the old `npm-shrinkwrap.json`) which meant that e.g. yargs got reverted to
```json
"yargs": {
  "version": "4.7.1",
  "from": "yargs@>=4.6.0 <5.0.0",
  "resolved": "https://registry.npmjs.org/yargs/-/yargs-4.7.1.tgz"
},
```
as can be seen in this PR's file-changes.
6. When the CI-tests run, the `ensure-staged-shrinkwrap.sh` script on the CI-servers is now sad because the local `npm-shrinkwrap.json` which has been freshly-updated from the newly-created `node_modules` directory, no longer matches up with the `npm-shrinkwrap.json` in the repo, as can be seen in the Appveyor logfiles
```
     "yargs": {
       "version": "4.7.1",
-      "from": "yargs@>=4.6.0 <5.0.0",
+      "from": "yargs@4.7.1",
       "resolved": "https://registry.npmjs.org/yargs/-/yargs-4.7.1.tgz"
     },
```
7. This is because the old `node_modules` folder on @Shou 's PC has cached the old `"from": "yargs@>=4.6.0 <5.0.0",` line, and so when @Shou updates `npm-shrinkwrap.json` this is what gets inserted (leading to the crazy number of diffs in `npm-shrinkwrap.json` in this PR). However on Appveyor the newly-installed `node_modules` folder has cached the new `"from": "yargs@4.7.1",` line, and so this is what gets inserted when Appveyor tries updating `npm-shrinkwrap.json` (leading to the huge number of complaints from `ensure-staged-shrinkwrap.sh` in the Appveyor logs)
8. So I believe if @Shou deletes his old `node_modules` folder, then that'll get rid of all the old cached `from` lines, and so when he updates `npm-shrinkwrap.json` the _only_ change will be to `pretty-bytes`, and then the `npm-shrinkwrap.json` updated by @Shou and the `npm-shrinkwrap.json` updated by Appveyor should match, and so everything should be all good again.

> But who knows what actually happens inside `npm`...

Haha, reminds me of https://en.wikipedia.org/wiki/God%27s_algorithm ;-)

AFAICT my theory above makes sense, but it doesn't explain why `ensure-staged-shrinkwrap.sh` complains on Appveyor, but passes happily on Travis :-S

jviotti:
What about trying to install the dependency with `--save-exact`, so ranges are never passed to npm? 

> AFAICT my theory above makes sense, but it doesn't explain why ensure-staged-shrinkwrap.sh complains on Appveyor, but passes happily on Travis :-S

I suspect there is a big `if (os.platform() === 'win32') { ... } else { ... }` block in npm's source code :P

jviotti:
@lurch @Shou I got to the bottom of it, but still have no clear solution. See https://github.com/resin-io/etcher/pull/1547

lurch:
Or, as I discussed in #1547 the difference between Travis and Appveyor might be due to each CI service caching things differently :-/

lurch:
@Shou Hopefully #1580 fixes this, so do you want to try rebasing again? ;)

jviotti:
Ping @Shou. We have conflicts here, so I guess we need to rebase anyway.

Shou:
Yup, working on it right now actually!

lurch:
Looks like the `from:` lines in `npm-shrinkwrap.json` are now all okay (hooray! :tada: ) but it's now getting confused about which dependencies are `dev` or not, e.g.
```
-      "resolved": "https://registry.npmjs.org/array-find-index/-/array-find-index-1.0.2.tgz",
-      "dev": true
+      "resolved": "https://registry.npmjs.org/array-find-index/-/array-find-index-1.0.2.tgz"
```

@Shou I see in the list of changed files that this PR updates `package.json`, but there's no corresponding update to `npm-shrinkwrap.json`, which seems a bit suspicious?

jviotti:
I confirm the "dev" changes the CI services are suggesting are valid. The dependency that was being installed here was now already installed as development dependency on `master`, so now npm wants to simply remove the "dev" flags.

Shou:
So this is good now?

lib/shared/units.js
lurch: Hmmm, so I guess you decided to not go for this change then? :confused:

Shou: My bad! I'll address this in a separate PR.

PR #1540, opened 2017-06-22
chore: make use of electron-builder for the package-electron target 

jviotti:
This target builds a package for the current operating system, but
doesn't create an installer out of it, which is useful for debugging
purposes.

Change-Type: patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,7 @@ endif
 
 ifdef DISABLE_UPDATES
 $(warning Update notification dialog has been disabled (DISABLE_UPDATES is set))
+ELECTRON_BUILDER_OPTIONS += --extraMetadata.updates.enabled=false
 endif
 
 # ---------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -433,7 +434,14 @@ TARGETS = \
 	installers-all \
 	electron-develop
 
-package-electron: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)
```
lurch: So should the `$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)` rule be removed too, if there's no target referencing it?

jviotti: The Linux rules still use that target. We will be able to remove it after merging the Linux electron-builder PR.

lurch: Oh yeah, I hadn't spotted that `$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-linux-$(TARGET_ARCH)` _is_ still being used :)

PR #1543, opened 2017-06-24
chore: use various electron-builder defaults

jviotti:
Now that all GUI builds are done through electron-builder, we'll
simplify our Makefile by adopting certain electron-builder defaults:

- Output final packages in `dist/`
- Adopt electron-builder code-signing environment variables

Change-Type: patch


lurch:
Looks like the Etcher CLI builds are the only thing still using `$(BUILD_OUTPUT_DIRECTORY)`. Should they be moved to `$(BUILD_DIRECTORY)` too, and then `$(BUILD_OUTPUT_DIRECTORY)` can be removed?

jviotti:
Sounds cool. Can you review https://github.com/resin-io/etcher/pull/1547? We could address these changes there.

PR #1544, opened 2017-06-24
chore: move all electron-builder configuration to electron-builder.yml

jviotti:
The electron-builder configuration we're putting in package.json is
growing. To make it easier to read and edit, we'll move the whole
`build` property into a configuration file called
`electron-builder.yml`.

Change-Type: patch

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -1,0 +1,93 @@
+appId: io.resin.etcher
+copyright: Copyright 2016 Resinio Ltd
+productName: Etcher
+npmRebuild: true
+nodeGypRebuild: true
+publish: null
+files:
+  - lib
+  - assets/icon.png
+  - node_modules/**/*
+mac:
+  icon: assets/icon.icns
+  artifactName: "${productName}-${version}-darwin-${env.TARGET_ARCH}.${ext}"
+  category: public.app-category.developer-tools
+dmg:
+  background: assets/osx/installer.tiff
+  icon: assets/icon.icns
+  iconSize: 110
+  contents:
+    - x: 140
+      y: 225
+    - x: 415
+      y: 225
+      type: link
+      path: /Applications
+  window:
+    width: 540
+    height: 405
+win:
+  icon: assets/icon.ico
+nsis:
+  oneClick: false
+  allowToChangeInstallationDirectory: true
+  runAfterFinish: true
+  installerIcon: assets/icon.ico
+  uninstallerIcon: assets/icon.ico
+  deleteAppDataOnUninstall: true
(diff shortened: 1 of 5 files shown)


electron-builder.yml
```diff
@@ -0,0 +1,93 @@
...
+      y: 225
+    - x: 415
+      y: 225
+      type: link
+      path: /Applications
+  window:
+    width: 540
+    height: 405
+win:
+  icon: asseets/icon.ico
```
jhermsmeier: Typo here: asseets -> assets

jviotti: Good catch, fixed!

electron-builder.yml
```diff
@@ -0,0 +1,93 @@
...
+  icon: assets/icon.ico
+nsis:
+  oneClick: false
+  allowToChangeInstallationDirectory: true
+  runAfterFinish: true
+  installerIcon: assets/icon.ico
+  uninstallerIcon: assets/icon.ico
+  deleteAppDataOnUninstall: true
+  license: LICENSE
+  artifactName: ${productName}-${version}-win32-${env.TARGET_ARCH}.${ext}
```
jhermsmeier: Also looks like the `productName` is null here for some reason:
```
Building NSIS installer
  Packaging NSIS installer for arch x64
mv dist/null-1.0.0+2866085-win32-x64-portable.exe dist/out/null-1.0.0+2866085-win32-x64-portable.exe
mv: cannot stat `dist/null-1.0.0+2866085-win32-x64-portable.exe': No such file or directory
make: *** [dist/out/null-1.0.0+2866085-win32-x64-portable.exe] Error 1
```

jviotti: Oops, fixed!

jviotti:
@jhermsmeier All green! Is this good to merge?

lib/child-writer/writer-proxy.js
```diff
@@ -89,7 +89,7 @@ return permissions.isElevated().then((elevated) => {
     console.log(`Running: ${commandArguments.join(' ')}`);
 
     return permissions.elevateCommand(commandArguments, {
-      applicationName: packageJSON.build.productName,
+      applicationName: packageJSON.displayName,
```
lurch: Like one of your comments in #1438 , should this be picking up the `initialTitle` rather than a value from `package.json` ?

jviotti: The issue is that this script doesn't run in the renderer process, so we don't have access to the DOM. I guess we could make the writer proxy take the title as a command line argument. I'll try to put a PR for it soon.

Issue #1531, opened 2017-06-20
Build the CLI using pkg


jviotti:
See https://github.com/zeit/pkg

lurch:
So does that (effectively) do the same thing as your node-static-entry-point thing?

jviotti:
Yep, but we'll try to re-use as most as we can in order to avoid having to maintain custom code.

lurch:
Should node-static-entry-point now be marked as a 'deprecated' project?

jviotti:
Done!

Issue #1450, opened 2017-05-18
Etcher CLI looks for index.js in pwd instead of relative to path of etcher binary


tkfu:
- **Etcher version:** CLI 1.0.0
- **Operating system and architecture:** Ubuntu 16.04 x64
- **Image flashed:** n/a
- **Do you see any meaningful error information in the DevTools?** n/a

Like the title says. Some terminal output illustrating the problem:

```
jon@jon-lenovo ~ $ etcher -v
module.js:457
    throw err;
    ^

Error: Cannot find module '/home/jon/index.js'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.runMain (module.js:590:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
jon@jon-lenovo ~ $ mkdir -p some/random/dir
jon@jon-lenovo ~ $ cd !!:2
cd some/random/dir
jon@jon-lenovo ~/some/random/dir $ etcher -v
module.js:457
    throw err;
    ^

Error: Cannot find module '/home/jon/some/random/dir/index.js'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.runMain (module.js:590:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
jon@jon-lenovo ~/some/random/dir $ cd ~/bin
jon@jon-lenovo ~/bin $ etcher -v
1.0.0
```

Note: I put the contents of the tarball in `~/bin` instead of `/opt/etcher-cli` as suggested, but I also tested it with the contents in `/opt/etcher-cli` just in case there were some hard-coded path searching shenanigans going on. Same problem.

tkfu:
It looks like it works if I explicitly specify the path:
```sh
jon@jon-lenovo ~ $ ~/bin/etcher -v
1.0.0
```
This is kinda crappy behaviour, though, and it's undocumented to boot.

lurch:
Yep, I can confirm this behaviour.

> This is kinda crappy behaviour

To be fair, we did state that this was the first experimental release of the CLI ;-)
I'm afraid that the CLI packaging took us quite a while to sort out, and it was only completed shortly before the release of Etcher 1.0.0, so it didn't get as much testing as we'd ideally have liked.
[...]

PR #1547, opened 2017-06-26
chore: make use of pkg to package the Etcher CLI

jviotti:
This commit replaces our home-grown CLI packaging mechanism based on
browserify + node-static-entry-point with pkg, an open source tool to
package Node.js applications for distribution.

Some highlights:

- Removing browserify got rid of a lot of dependencies from
  npm-shrinkwrap.json

- pkg currently has an issue where macOS binaries can't be code-signed
  (https://github.com/zeit/pkg/issues/128), therefore this commit
  comments-out the binary signing section for that operating system

- pkg currently has an issue where Windows binaries can't be branded
  (zeit/pkg#149), therefore this commit
  comments-out the branding section for that operating system

See: https://github.com/zeit/pkg
Fixes: https://github.com/resin-io/etcher/issues/1531
Fixes: https://github.com/resin-io/etcher/issues/1450
Change-Type: patch


Makefile
```diff
@@ -215,64 +215,73 @@ $(BUILD_OUTPUT_DIRECTORY): | $(BUILD_DIRECTORY)
...
-	$(BUILD_DIRECTORY)/node-$(TARGET_PLATFORM)-$(TARGET_ARCH)-dependencies \
-	| $(BUILD_DIRECTORY)
-	mkdir $@
-	cp $(word 1,$^) $@
-	$(CPRF) $(word 2,$^) $@
-	$(CPRF) $(word 3,$^)/* $@
-
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).js: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app \
+	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/node_modules \
```
lurch: Barf! Having `$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app` depend on `$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/node_modules` looks **so** wrong :-/

It's effectively saying that you need to make `dir1/dir2/` before you can make `dir1/`

Makefile
```diff
@@ -215,64 +215,73 @@ $(BUILD_OUTPUT_DIRECTORY): | $(BUILD_DIRECTORY)
...
+# See https://github.com/zeit/pkg/issues/128
+# ifeq ($(TARGET_PLATFORM),darwin)
+# ifdef CSC_NAME
+#		./scripts/build/electron-sign-file-darwin.sh -f $@/$(ETCHER_CLI_BINARY) -i "$(CSC_NAME)"
+# endif
+# endif
+
+# pkg currently has a bug where Windows executables
+# can't be branded
+# See https://github.com/zeit/pkg/issues/149
```
lurch: Should this be mentioned in the PR description too?

Makefile
```diff
@@ -215,64 +215,73 @@ $(BUILD_OUTPUT_DIRECTORY): | $(BUILD_DIRECTORY)
...
-		-e $(word 2,$^) \
-		-r $(TARGET_ARCH) \
-		-s $(TARGET_PLATFORM)
+TARGET_PLATFORM_PKG = $(TARGET_PLATFORM)
+ifeq ($(TARGET_PLATFORM_PKG),darwin)
+TARGET_PLATFORM_PKG = macos
+endif
+ifeq ($(TARGET_PLATFORM_PKG),win32)
+TARGET_PLATFORM_PKG = win
+endif
```
lurch: "smells like" there ought to be a `target-convert.sh` script, similar to the `architecture-convert.sh` script? ;)

Makefile
```diff
@@ -215,64 +215,73 @@ $(BUILD_OUTPUT_DIRECTORY): | $(BUILD_DIRECTORY)
...
-ifeq ($(TARGET_PLATFORM),darwin)
-ifdef CSC_NAME
-	./scripts/build/electron-sign-file-darwin.sh -f $@/etcher -i "$(CSC_NAME)"
-endif
-endif
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH): \
+	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app \
+	| $(BUILD_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
+	mkdir $@
+	$(NPX) pkg --output $@/$(ETCHER_CLI_BINARY) -t node6-$(TARGET_PLATFORM_PKG)-$(TARGET_ARCH) $</lib/cli/etcher.js
```
lurch: I wonder if there should be a Makefile 'constant' (variable) for this `lib/cli/etcher.js` path?

scripts/build/concatenate-javascript.sh
```diff
@@ -1,67 +0,0 @@
...
-fi
-
-"$BROWSERIFY" "$ARGV_ENTRY_POINT" \
-  --node \
-  --basedir "$ARGV_BASE_DIRECTORY" \
-  --outfile "$ARGV_OUTPUT"
-
-if [ "$ARGV_MINIFY" == "true" ]; then
-  ./scripts/build/check-dependency.sh uglifyjs
-  uglifyjs --compress --output "$ARGV_OUTPUT.MIN" -- "$ARGV_OUTPUT"
```
lurch: Looks like this removes the only usage of `uglifyjs`, so the CI scripts can be updated too? (to not install it)

scripts/build/node-package-cli.sh
```diff
@@ -1,81 +0,0 @@
...
-  || [ -z "$ARGV_NODE_MODULES" ] \
-  || [ -z "$ARGV_ARCHITECTURE" ] \
-  || [ -z "$ARGV_OPERATING_SYSTEM" ] \
-  || [ -z "$ARGV_OUTPUT" ]; then
-  usage
-fi
-
-cp "$ARGV_ENTRY_POINT" "$ARGV_OUTPUT/index.js"
-
-./scripts/build/dependencies-npm-extract-addons.sh \
```
lurch: Should this PR delete this script too?

scripts/build/node-package-cli.sh
```diff
@@ -1,81 +0,0 @@
...
-  -d "$ARGV_NODE_MODULES" \
-  -o "$ARGV_OUTPUT/node_modules"
-
-APPLICATION_NAME_LOWERCASE="$(echo "$ARGV_APPLICATION_NAME" | tr '[:upper:]' '[:lower:]')"
-BINARY_LOCATION="$ARGV_OUTPUT/$APPLICATION_NAME_LOWERCASE"
-if [ "$ARGV_OPERATING_SYSTEM" == "win32" ]; then
-  BINARY_LOCATION="$BINARY_LOCATION.exe"
-fi
-
-./scripts/build/node-static-entry-point-download.sh \
```
lurch: Should this PR delete this script too?

Makefile
jviotti: Sounds like a good idea, will do.

Makefile

scripts/build/node-package-cli.sh
jviotti: This is still being used to extract the native dependencies and ship them along with the pkg generated binary.

scripts/build/node-package-cli.sh
jviotti: Yep, done

Makefile

Makefile
jviotti: What would be the recommended approach? We could install `node_modules` somewhere else, but we would have to copy it, which brings us to the weird symbolic link issues in Appveyor CI.

We basically need `node_modules`, `lib`, and `package.json` all in the same place to run `pkg`, but I can't think of a way to get this setup without having to copy `node_modules`

scripts/build/concatenate-javascript.sh
jviotti: Done!

jviotti:
> Should this PR totally remove the usage of $(BUILD_OUTPUT_DIRECTORY) ?

Done!

jviotti:
Ping @lurch. Lets try to get this merged soon, we can get rid of those "cyclic links" Appveyor issues when using MinGW.

Makefile
```diff
@@ -205,89 +204,82 @@ CHANGELOG.md:
...
-	./scripts/build/electron-brand-exe.sh \
-		-f $@/etcher.exe \
-		-n $(APPLICATION_NAME) \
-		-d "$(APPLICATION_DESCRIPTION)" \
-		-v "$(APPLICATION_VERSION)" \
-		-c "$(APPLICATION_COPYRIGHT)" \
-		-m "$(COMPANY_NAME)" \
-		-i assets/icon.ico \
-		-w $(BUILD_TEMPORARY_DIRECTORY)
+ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE).exe
```
lurch: Given that this is only defining a variable name, perhaps it should be moved to earlier in the file along with the other variable-definitions?

Makefile
lurch: Given that nothing else depends on `$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/node_modules`, I guess one slightly ugly way might be to move the rules for `$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app/node_modules` inside the rules for `$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH)-app` ?

>  without having to copy `node_modules`

So it doesn't matter that we _are_ copying the `lib` directory? And you probably don't need the `-r` flag when copying `package.json` ;-)

scripts/build/node-package-cli.sh
lurch: Ahh, I hadn't noticed that it'd been moved into the `Makefile` ;)
Can you remind me what this does? Why isn't it something that pkg can do itself?

Makefile
```diff
@@ -205,89 +204,82 @@ CHANGELOG.md:
...
+# ifeq ($(TARGET_PLATFORM),win32)
+#		./scripts/build/electron-brand-exe.sh \
+#			-f $@/$(ETCHER_CLI_BINARY) \
+#			-n $(APPLICATION_NAME) \
+#			-d "$(APPLICATION_DESCRIPTION)" \
+#			-v "$(APPLICATION_VERSION)" \
+#			-c "$(APPLICATION_COPYRIGHT)" \
+#			-m "$(COMPANY_NAME)" \
+#			-i assets/icon.ico \
+#			-w $(BUILD_DIRECTORY)
```
lurch: (when this works again) does it make sense to just download `rcedit.exe` into `$(BUILD_DIRECTORY)`, given that it's not a target that we're building?

Makefile

Makefile
jviotti: Fair enough. I brought back the temporary directory.

jviotti:
Ping @lurch

jviotti:
We're getting some shrinkwrap issues, exactly like in https://github.com/resin-io/etcher/pull/1539, and I think I got to the bottom if it, and its related to `node_modules` caches in the CI servers. Here's what's going on on this particular case:

- I installed `pkg` using `npm install --save pkg` (notice I didn't specify a version)
- That resulted in `package.json` containing:

```
    "pkg": "^4.1.1"
```

And in `npm-shrinkwrap.json` containing:

```
    "pkg": {
      "version": "4.1.1",
      "from": "pkg@latest",
```

(notice the `from` property).

- I then manually edited the shrinkwrap file and package.json to eliminate `latest` and any ranges (which is what npm would have done for me)
- See what's inside `node_modules/pkg/package.json`:

```
  "_from": "pkg@latest",
  "_id": "pkg@4.1.1",
``` 

So my shrinkwrap file says that `pkg` came `from: pkg@4.1.1`, but the actual `package.json` of the `pkg` installation says it came from `pkg@latest`.

- If I run `npm shrinkwrap --dev`, this is what I get:

```diff
diff --git a/npm-shrinkwrap.json b/npm-shrinkwrap.json
index d32af581..fe0a9c25 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -8659,7 +8659,7 @@
     },
     "pkg": {
       "version": "4.1.1",
-      "from": "pkg@4.1.1",
+      "from": "pkg@latest",
       "resolved": "https://registry.npmjs.org/pkg/-/pkg-4.1.1.tgz",
       "dependencies": {
         "acorn": {
```

So npm is obeying what's on `node_modules`, and updating `npm-shrinkwrap.json` to reflect that.
 
- Of course, the fix is to also update `_from` from `node_modules/pkg/package.json`
- The problem is gone if I wipe out `node_modules` and do everything from scratch, since `pkg` will be correctly installed as what the shrinkwrap says, and there will be no inconsistency
- The issue is when the CI services use cached node_modules :)

lurch:
So.... given that these annoying `from:` lines seem to be the cause of 90% of our problems with `npm-shrinkwrap.json`, would it be worth trying to remove them all during `clean-shrinkwrap.js`, as I've already suggested a couple of times? I **think** that these lines are only informational (and they're gone in NPM 5), but obviously @jviotti knows NPM much better than I do, so he can better test whether removing these lines actually breaks NPM 3 or NPM 4 ;-)

appveyor.yml
```diff
@@ -6,7 +6,7 @@ image: Visual Studio 2015
 cache:
   - C:\Users\appveyor\.node-gyp
   - '%AppData%\npm-cache'
-  - node_modules
+  - node_modules -> package.json
```
lurch: Would `npm-shrinkwrap.json` here be "more correct" than `package.json` ?

lurch:
@jviotti Looks like you might have missed one of my previous comments? https://github.com/resin-io/etcher/pull/1547#discussion_r125349252

lurch:
Depending on what you do with the `Makefile`, it looks like you might be able to get rid of the `CPRF` variable too :)

appveyor.yml
jviotti: Yeah, I think so!

Makefile
```diff
@@ -205,89 +210,77 @@ CHANGELOG.md:
 $(BUILD_DIRECTORY):
 	mkdir $@
 
-$(BUILD_TEMPORARY_DIRECTORY): | $(BUILD_DIRECTORY)
```
jviotti: Good catch. Fixed!

npm-shrinkwrap.json
```diff
@@ -8698,7 +8308,7 @@
     },
     "pretty-bytes": {
       "version": "1.0.4",
-      "from": "pretty-bytes@>=1.0.2 <2.0.0",
+      "from": "pretty-bytes@1.0.4",
```
jviotti: Seems to be legit. Maybe some of the new dependencies are directly referencing this particular version?

.travis.yml
```diff
@@ -5,6 +5,8 @@ node_js:
 
 cache:
   ccache: true
+  directories:
+    - node_modules
```
jviotti: Docker mirrors every file that it creates on the current directory, so `node_modules` (as well as `dist`) end up in the right place regardless.

> Hopefully re-enabling this won't uncover the same bug that @jhermsmeier hit ;-)

Haha, lets hope for the best :D

jviotti:
> So.... given that these annoying from: lines seem to be the cause of 90% of our problems with npm-shrinkwrap.json, would it be worth trying to remove them all during clean-shrinkwrap.js, as I've already suggested a couple of times? I think that these lines are only informational (and they're gone in NPM 5), but obviously @jviotti knows NPM much better than I do, so he can better test whether removing these lines actually breaks NPM 3 or NPM 4 ;-)

I think we should try to push it as far as we can before doing non-standard change to this file. We don't know for sure if some npm operations rely on this little tag (even for comparison purposes). If they are removed in later npm versions, then we rather put effort in upgrading our npm version.

Makefile
jviotti: Yeah, I think that's a fair approach. I'll update the PR!

jviotti:
> Depending on what you do with the Makefile, it looks like you might be able to get rid of the CPRF variable too :)

Done! :tada:

jviotti:
I'm testing that the CLI build pipeline didn't break, and I'll push my latest changes afterwards.

.travis.yml
lurch: Ahhh, apologies for the mix-up, it gets a bit confusing sometimes ;)

Makefile
```diff
@@ -179,15 +178,18 @@ TARGET_ARCH_DEBIAN = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET
 TARGET_ARCH_REDHAT = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t redhat)
 TARGET_ARCH_APPIMAGE = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t appimage)
 TARGET_ARCH_ELECTRON_BUILDER = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t electron-builder)
+TARGET_PLATFORM_PKG = $(shell ./scripts/build/target-convert.sh -r $(TARGET_PLATFORM) -t pkg)
```
lurch: WTF?!?!  AFAICT `./scripts/build/target-convert.sh` doesn't actually exist, so how did this PR **ever** pass the CI tests? :confounded:

jviotti: Are you in the wrong branch? I can see it from my side:

```sh
$ ls -l scripts/build/target-convert.sh
-rwxr-xr-x  1 jviotti  staff  1390 Jun 30 12:34 scripts/build/target-convert.sh*
```

lurch: Well apparently this PR has been merged to `master`, but the file doesn't show up at https://github.com/resin-io/etcher/tree/master/scripts/build

Did you forget to check it in? Or am I just being really stupid? :-/

lurch: Ahhh, I've just spotted that Appveyor says:
```
make: ./scripts/build/target-convert.sh: Command not found
mkdir dist/Etcher-cli-1.0.0+b67afbe-win32-x64
./node_modules/.bin/npx pkg --output dist/Etcher-cli-1.0.0+b67afbe-win32-x64/etcher.exe -t node6--x64 dist/Etcher-cli-1.0.0+b67afbe-win32-x64-app/lib/cli/etcher.js
```
and Travis says:
```
make: ./scripts/build/target-convert.sh: Command not found
mkdir dist/Etcher-cli-1.0.0+b67afbe-linux-x64
./node_modules/.bin/npx pkg --output dist/Etcher-cli-1.0.0+b67afbe-linux-x64/etcher -t node6--x64 dist/Etcher-cli-1.0.0+b67afbe-linux-x64-app/lib/cli/etcher.js
```
so it's **not** me being stupid, and this indeed isn't quite working as expected - note the `node6--x64` strings with the missing platform specifier.

jviotti: What :| That is really weird. I triple checked, and the script is there on my local tree, for the latest master:

```
$ git log -1
commit e52ef501bbcefe7d90bc760ef829211aefeb9fc3 (HEAD -> master, origin/master)
Author: Juan Cruz Viotti <jv@jviotti.com>
Date:   Tue Jul 11 17:12:44 2017 -0400

    chore: avoid brew taking a long time on macOS Travis builds (#1583)

    - Homebrew now auto-updates before performing any installation, which
      can take quite some time. We can disable this by enabling
      HOMEBREW_NO_AUTO_UPDATE

    - We can pass the --force-bottle option to `brew install` to force it to
      download pre-compiled binaries, and don't attempt any compilations

    See: https://github.com/resin-io/etcher/pull/1582
    Change-Type: patch

$ ls -l scripts/build
total 96
-rwxr-xr-x  1 jviotti  staff  2278 Jul  6 15:39 architecture-convert.sh*
-rwxr-xr-x  1 jviotti  staff  1276 Jun  6 12:59 check-dependency.sh*
-rwxr-xr-x  1 jviotti  staff  1476 Jun 12 15:22 dependencies-npm-extract-addons.sh*
-rwxr-xr-x  1 jviotti  staff  4021 Jul 10 11:21 dependencies-npm.sh*
drwxr-xr-x  8 jviotti  staff   272 Jul 11 15:07 docker/
-rwxr-xr-x  1 jviotti  staff  2061 Mar 30 18:05 download-tool.sh*
-rw-r--r--  1 jviotti  staff  2670 Mar 31 13:17 electron-brand-exe-win32.sh
-rwxr-xr-x  1 jviotti  staff  2670 Mar 31 13:17 electron-brand-exe.sh*
-rwxr-xr-x  1 jviotti  staff  1789 Mar  9 16:32 electron-sign-exe-win32.sh*
-rwxr-xr-x  1 jviotti  staff  1247 Mar 31 13:17 electron-sign-file-darwin.sh*
-rwxr-xr-x  1 jviotti  staff  1125 Apr 24 10:57 tar-gz-file.sh*
-rwxr-xr-x  1 jviotti  staff  1390 Jun 30 12:34 target-convert.sh*
-rwxr-xr-x  1 jviotti  staff  1629 Jun 27 12:52 zip-file.sh*
```

I'll investigate...

jviotti: OK, there's something messed up. Looks like git is not tracking changes for THAT particular file...

jviotti: Ah, fuck. Check `.gitignore` :/

```
# Compiled binary addons (http://nodejs.org/api/addons.html)
build
```

jviotti: See https://github.com/resin-io/etcher/pull/1588

PR #1548, opened 2017-06-26
upgrade: `electron-builder` to v19.9.1

jviotti:
- Exclude \*.dll/\*.exe files from the asar in non-Windows operating
  systems (from 19.8.0)

- Correctly parse boolean flags in `--extraMetadata` (in v19.9.0)

See: https://github.com/electron-userland/electron-builder/releases/tag/v19.9.1
Change-Type: patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ endif
 # Electron Builder
 # ---------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


lurch:
Failing on Travis with `TypeError: s.trim is not a function` (but works on Appveyor, which is strange!)

jviotti:
Trying to debug the issue here: https://github.com/electron-userland/electron-builder/issues/1784

jviotti:
OK, it was totally our fault. I updated the Makefile, and now everything should work.

PR #1549, opened 2017-06-26
chore: remove npm ranges from package.json

jviotti:
We don't make use of npm ranges at all (every dependency and its
subdependencies are shrinkwrapped). Removing npm ranges allows us to
simplify our clean-shrinkwrap.js script to get rid of platform specific
optional dependencies, and their subdependencies.

See: https://github.com/resin-io/etcher/pull/1514


package.json
```diff
@@ -149,70 +149,70 @@
...
+    "html-angular-validate": "0.1.9",
+    "mochainon": "1.0.0",
+    "nock": "9.0.9",
+    "node-gyp": "3.5.0",
+    "node-sass": "4.5.3",
+    "npx": "5.2.0",
+    "sass-lint": "1.10.2",
     "tmp": "0.0.31",
-    "versionist": "^2.1.0"
+    "versionist": "^2.8.1"
```
jhermsmeier: Aaaand here

package.json
```diff
@@ -149,70 +149,70 @@
...
-    "mochainon": "^1.0.0",
-    "nock": "^9.0.9",
-    "node-gyp": "^3.5.0",
-    "node-sass": "^4.5.3",
-    "npx": "^5.2.0",
-    "sass-lint": "^1.10.2",
+    "electron-builder": "18.6.2",
+    "electron-mocha": "3.3.0",
+    "eslint": "^3.18.0",
+    "eslint-plugin-lodash": "^2.3.6",
```
jhermsmeier: Same here

package.json
```diff
@@ -149,70 +149,70 @@
...
+    "redux": "3.5.2",
+    "redux-localstorage": "0.4.1",
+    "request": "2.81.0",
+    "resin-cli-form": "1.4.1",
+    "resin-cli-visuals": "1.3.1",
+    "resin-corvus": "1.0.0-beta.26",
+    "rx": "4.1.0",
+    "semver": "5.1.1",
+    "sudo-prompt": "6.1.0",
+    "trackjs": "^2.3.1",
```
jhermsmeier: Missed a range here, or is this intended?

package.json
jviotti: Damn, I'm still stuck at this imperfect human being state :P

PR #1551, opened 2017-06-27
chore: don't remove optional dependencies in clean-shrinkwrap.js

jviotti:
If we include a platform specific optional dependency in the shrinkwrap
file, then npm will insist in installing it even if the platform doesn't
match. As a solution, we figured out we can avoid putting this platform
specific optional dependencies in the npm-shrinkwrap.json file.

In order to do this, we currently have a script called
`clean-shrinkwrap.js` that runs *before* any `npm shrinkwrap` file (its
a `preshrinkwrap` npm script) that deletes all the platform specific
modules we know about using `npm rm`.

The problem with this approach is that `npm rm` will remove the module's
code from `node_modules`, which means that if we run `npm shrinkwrap`,
we will lose certain optional dependencies, that may be needed at a
later stage.

The solution is to modify the `clean-shrinkwrap.js` script to parse
`npm-shrinkwrap.json`, and manually delete the entries that we want to
omit. Also, the script needs to be run *after* `npm shrinkwrap`, so we
change the npm script name to `postshrinkwrap`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,21 +19,17 @@
   "scripts": {
     "test": "make test",
     "start": "electron lib/start.js",
-    "preshrinkwrap": "node ./scripts/clean-shrinkwrap.js",
+    "postshrinkwrap": "node ./scripts/clean-shrinkwrap.js",
     "configure": "node-gyp configure",
     "build": "node-gyp build",
     "install": "node-gyp rebuild"
   },
   "author": "Resin Inc. <hello@etcher.io>",
   "license": "Apache-2.0",
-  "shrinkwrapIgnore": [
-    "macos-alias",
-    "fs-xattr",
-    "ds-store",
-    "appdmg",
-    "7zip-bin-mac",
-    "7zip-bin-win",
-    "7zip-bin-linux"
+  "platformSpecificDependencies": [
+    [ "7zip-bin-mac" ],
+    [ "7zip-bin-win" ],
+    [ "7zip-bin-linux" ]
   ],
   "dependencies": {
     "angular": "1.6.3",
--- a/scripts/clean-shrinkwrap.js
+++ b/scripts/clean-shrinkwrap.js
@@ -15,29 +15,258 @@
 
 'use strict';
 
+const _ = require('lodash');
+const fs = require('fs');
 const path = require('path');
-const os = require('os');
 const packageJSON = require('../package.json');
-const spawn = require('child_process').spawn;
-const shrinkwrapIgnore = packageJSON.shrinkwrapIgnore;
+const NPM_SHRINKWRAP_FILE_PATH = path.join(__dirname, '..', 'npm-shrinkwrap.json');
+const shrinkwrapFile = require(NPM_SHRINKWRAP_FILE_PATH);
+const platformSpecificDependencies = packageJSON.platformSpecificDependencies;
+const JSON_INDENTATION_SPACES = 2;
 
-console.log('Removing:', shrinkwrapIgnore.join(', '));
+console.log('Removing:', platformSpecificDependencies.join(', '));
 
 /**
- * Run an npm command
- * @param {Array} command - list of arguments
- * @returns {ChildProcess}
+ * @summary Get a shrinkwrap dependency object
+ * @function
+ * @private
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} shrinkwrap object
+ *
+ * @example
+ * const object = getShrinkwrapDependencyObject(require('./npm-shrinkwrap.json'), [
+ *   'drivelist',
+ *   'lodash'
+ * ]);
+ *
+ * console.log(object.version);
+ * console.log(object.dependencies);
  */
-const npm = (command) => {
-  const npmBinary = os.platform() === 'win32' ? 'npm.cmd' : 'npm';
-  return spawn(npmBinary, command, {
-    cwd: path.join(__dirname, '..'),
-    env: process.env,
-    stdio: [ process.stdin, process.stdout, process.stderr ]
-  });
+const getShrinkwrapDependencyObject = (shrinkwrap, shrinkwrapPath) => {
+  return _.reduce(shrinkwrapPath, (accumulator, dependency) => {
+    return _.get(accumulator, [ 'dependencies', dependency ], {});
+  }, shrinkwrap);
 };
 
-npm([ 'rm', '--ignore-scripts' ].concat(shrinkwrapIgnore))
-  .once('close', () => {
-    console.log('Done.');
+/**
+ * @summary Get a cleaned shrinkwrap dependency object
+ * @function
+ * @private
+ *
+ * @description
+ * This function wraps `getShrinkwrapDependencyObject()` to
+ * omit unnecessary properties such as `from`, or `dependencies`.
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} pretty shrinkwrap object
+ *
+ * @example
+ * const object = getPrettyShrinkwrapDependencyObject(require('./npm-shrinkwrap.json'), [
+ *   'drivelist',
+ *   'lodash'
+ * ]);
+ *
+ * console.log(object.name);
+ * console.log(object.path);
+ * console.log(object.version);
+ */
+const getPrettyShrinkwrapDependencyObject = (shrinkwrap, shrinkwrapPath) => {
+  const object = getShrinkwrapDependencyObject(shrinkwrap, shrinkwrapPath);
+
+  if (_.isEmpty(object)) {
+    return null;
+  }
+
+  return {
+    name: _.last(shrinkwrapPath),
+    path: shrinkwrapPath,
+    version: object.version,
+    development: Boolean(object.dev),
+    optional: Boolean(object.optional)
+  };
+};
+
+/**
+ * @summary Get the manifest (package.json) of a shrinkwrap dependency
+ * @function
+ * @private
+ *
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} dependency manifest
+ *
+ * @example
+ * const manifest = getShrinkwrapDependencyManifest([ 'bluebird' ]);
+ * console.log(manifest.devDependencies);
+ */
+const getShrinkwrapDependencyManifest = (shrinkwrapPath) => {
+  const manifestPath = _.chain(shrinkwrapPath)
+    .flatMap((dependency) => {
+      return [ 'node_modules', dependency ];
+    })
+    .concat([ 'package.json' ])
+    .reduce((accumulator, file) => {
+      return path.join(accumulator, file);
+    }, '.')
+    .value();
+
+  try {
+
+    // For example
+    // ./node_modules/drivelist/node_modules/lodash/package.json
+    return require(`.${path.sep}${manifestPath}`);
+
+  } catch (error) {
+    if (error.code === 'MODULE_NOT_FOUND') {
+      return null;
+    }
+
+    throw error;
+  }
+};
+
+/**
+ * @summary Get the top level dependencies of a shrinkwrap object
+ * @function
+ * @private
+ *
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} top level dependencies
+ *
+ * @example
+ * const dependencies = getTopLevelDependenciesForShrinkwrapPath([ 'debug' ]);
+ * console.log(dependencies);
+ * > {
+ * >   "lodash": "^4.0.0"
+ * > }
+ */
+const getTopLevelDependenciesForShrinkwrapPath = (shrinkwrapPath) => {
+  return _.get(getShrinkwrapDependencyManifest(shrinkwrapPath), [ 'dependencies' ], {});
+};
+
+/**
+ * @summary Get the dependency tree of a shrinkwrap dependency
+ * @function
+ * @private
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object[]} dependency tree
+ *
+ * @example
+ * const dependencyTree = getDependencyTree(require('./npm-shrinkwrap.json'), [ 'drivelist' ]);
+ *
+ * _.each(dependencyTree, (dependency) => {
+ *   console.log(dependency.name);
+ *   console.log(dependency.path);
+ *   console.log(dependency.version);
+ * })
+ */
+const getDependencyTree = (shrinkwrap, shrinkwrapPath) => {
+  const dependencies = getTopLevelDependenciesForShrinkwrapPath(shrinkwrapPath);
+
+  if (_.isEmpty(dependencies)) {
+    return [];
+  }
+
+  const object = getShrinkwrapDependencyObject(shrinkwrap, shrinkwrapPath);
+  const result = _.map(dependencies, (version, name) => {
+    const dependencyPath = _.has(object.dependencies, name) ? _.concat(shrinkwrapPath, [ name ]) : [ name ];
+    return getPrettyShrinkwrapDependencyObject(shrinkwrap, dependencyPath);
   });
+
+  return _.concat(result, _.flatMapDeep(result, (dependency) => {
+    return getDependencyTree(shrinkwrap, dependency.path);
+  }));
+};
+
+/**
+ * @summary Remove certain development optional dependencies from a shrinkwrap file
+ * @function
+ * @private
+ *
+ * @description
+ * A shrinkwrap object is a recursive data structure, that apart
+ * from some extra metadata, has the following structure:
+ *
+ * {
+ *   ...
+ *   "dependencies": {
+ *     "<dependency_name>": <recursive definition>,
+ *     "<dependency_name>": <recursive definition>,
+ *     "<dependency_name>": <recursive definition>,
+ *     ...
+ *   }
+ * }
+ *
+ * The purpose of this function is to remove certain dependencies
+ * that match a blacklist. In order to do so, we start from the top
+ * level object, remove the blacklisted dependencies, and recurse
+ * if possible.
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap object
+ * @param {Object[]} blacklist - dependency blacklist
+ * @returns {Object} filtered shrinkwrap object
+ *
+ * @example
+ * const shrinkwrapFile = require('./npm-shrinkwrap.json');
+ * const dependencyTree = getDependencyTree(shrinkwrapFile, [ 'drivelist' ]);
+ * const filteredShrinkwrap = removeOptionalDevelopmentDependencies(shrinkwrapFile, dependencyTree);
+ */
+const removeOptionalDevelopmentDependencies = (shrinkwrap, blacklist) => {
+  if (!_.isEmpty(shrinkwrap.dependencies)) {
+    shrinkwrap.dependencies = _.chain(shrinkwrap.dependencies)
+      .omitBy((dependency, name) => {
+        return _.every([
+          _.find(blacklist, {
+            name,
+            version: dependency.version
+          }),
+          dependency.dev,
+          dependency.optional
+        ]);
+      })
+      .mapValues((dependency) => {
+        return removeOptionalDevelopmentDependencies(dependency, blacklist);
+      })
+      .value();
+  }
+
+  return shrinkwrap;
+};
+
+/**
+ * @summary Get the dependency tree of a dependency plus the dependency itself
+ * @function
+ * @private
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object[]} tree
+ *
+ * @example
+ * const tree = getTree(require('./npm-shrinkwrap.json'), [ 'drivelist' ]);
+ *
+ * _.each(tree, (dependency) => {
+ *   console.log(dependency.name);
+ *   console.log(dependency.path);
+ *   console.log(dependency.version);
+ * });
+ */
+const getTree = (shrinkwrap, shrinkwrapPath) => {
+  return _.compact(_.concat([
+    getPrettyShrinkwrapDependencyObject(shrinkwrap, shrinkwrapPath)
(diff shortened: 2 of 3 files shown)


jviotti:
Let's see what the CI servers say. Fingers-crossed!

jviotti:
YESS, finally everything works \o/

jviotti:
What do you think @jhermsmeier ?

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * @summary Get a shrinkwrap dependency object
+ * @function
+ * @private
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} shrinkwrap object
+ *
+ * @example
+ * const object = getShrinkwrapDependencyObject(require('npm-shrinkwrap.json'), [
```
jhermsmeier: The required path needs a `./` to understand that it's a local path afaik – might be good to have that in the example as well?

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * @function
+ * @private
+ *
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} top level dependencies
+ *
+ * @example
+ * const dependencies = getTopLevelDependenciesFromShrinkwrapObject([ 'debug' ]);
+ */
+const getTopLevelDependenciesFromShrinkwrapObject = (shrinkwrapPath) => {
```
jhermsmeier: This is starting to feel like I'm reading Java 😝

package.json
```diff
@@ -19,21 +19,17 @@
...
-    "fs-xattr",
-    "ds-store",
-    "appdmg",
-    "7zip-bin-mac",
-    "7zip-bin-win",
-    "7zip-bin-linux"
+  "platformSpecificDependencies": [
+    [ "7zip-bin-mac" ],
+    [ "7zip-bin-win" ],
+    [ "7zip-bin-linux" ]
```
jhermsmeier: What's the reasoning behind the nested arrays here?

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+  return shrinkwrap;
+};
+
+const blacklist = _.reduce(platformSpecificDependencies, (accumulator, dependencyPath) => {
+  return _.compact(_.concat(accumulator, [
+    getPrettyShrinkwrapDependencyObject(shrinkwrapFile, dependencyPath)
+  ], getDependencyTree(shrinkwrapFile, dependencyPath)));
+}, []);
+
+const JSON_INDENTATION_SPACES = 2;
```
jviotti: Done

scripts/clean-shrinkwrap.js
jviotti: Well, its **Java**Script after all :P

scripts/clean-shrinkwrap.js
jviotti: Fixed!

package.json
jviotti: The new approach works based on shrinkwrap "paths", which we set using array syntax. Basically this means that we want to omit `7zip-bin-mac` from the top level of the shrinkwrap file for example. If `7zip-bin-mac` was present inside `foo` in the shrinkwrap file, we should add `[ 'foo', '7zip-bin-mac' ]`.

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * @description
+ * This function transforms the output of `getShrinkwrapDependencyObject()` to
+ * omit unnecessary properties such as `from`, or `dependencies`.
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} shrinkwrap object
+ *
+ * @example
+ * const object = getPrettyShrinkwrapDependencyObject(require('npm-shrinkwrap.json'), [
```
jhermsmeier: Another relative path missing in the example here

jhermsmeier:
There's also a build failure on Windows:

```
[...]
mkdir dist/Etcher-cli-1.0.0+2a000ae-win32-x86-app
cp package.json dist/Etcher-cli-1.0.0+2a000ae-win32-x86-app
cp -RLf lib dist/Etcher-cli-1.0.0+2a000ae-win32-x86-app
cp -RLf dist/node-win32-x86-dependencies/* dist/Etcher-cli-1.0.0+2a000ae-win32-x86-app
cp: cannot copy cyclic symbolic link `dist/node-win32-x86-dependencies/node_modules/ajv/node_modules'
make: *** [dist/Etcher-cli-1.0.0+2a000ae-win32-x86-app] Error 1
Command exited with code 2
```

Is this a transient error, that only occasionally occurs?

lurch:
> `cp: cannot copy cyclic symbolic link`
Is this a transient error, that only occasionally occurs?

Yeah, it keeps popping up now and then, but only on Appveyor, and we've no idea why :-(  When you restart the build (which I've just done), it usually passes on the second attempt.
(and annoyingly, unlike on Travis CI where you can restart just one arch-specific build, on Appveyor you have to restart all arch builds)

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
-npm([ 'rm', '--ignore-scripts' ].concat(shrinkwrapIgnore))
-  .once('close', () => {
-    console.log('Done.');
+/**
+ * @summary Get the manifest of a shrinkwrap dependency
+ * @function
+ * @private
+ *
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} dependency manifest
```
lurch: I've got no idea what a `dependency manifest` is - an example might be great ;)

EDIT: Ah, now I see that it's the `package.json` for that dependency :)

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+      return [ 'node_modules', dependency ];
+    })
+    .concat([ 'package.json' ])
+    .reduce((accumulator, file) => {
+      return path.join(accumulator, file);
+    }, '.')
+    .value();
+
+  try {
+    return require(`.${path.sep}${manifestPath}`);
```
lurch: And an inline example comment of what this `manifestPath` might look like would be useful too.

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+  } catch (error) {
+    if (error.code === 'MODULE_NOT_FOUND') {
```
lurch: Just out of curiosity, what triggers this error? Is there some other attribute we could check, to stop us `require`-ing non-existent modules? (perhaps something as simple as checking the file exists?)

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * @private
+ *
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} top level dependencies
+ *
+ * @example
+ * const dependencies = getTopLevelDependenciesFromShrinkwrapObject([ 'debug' ]);
+ */
+const getTopLevelDependenciesFromShrinkwrapObject = (shrinkwrapPath) => {
+  return _.get(getShrinkwrapDependencyManifest(shrinkwrapPath), [ 'dependencies' ], {});
```
lurch: Seems that this is the only place that `getShrinkwrapDependencyManifest` is used so perhaps it should just be inlined?

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * ]);
  */
-const npm = (command) => {
-  const npmBinary = os.platform() === 'win32' ? 'npm.cmd' : 'npm';
-  return spawn(npmBinary, command, {
-    cwd: path.join(__dirname, '..'),
-    env: process.env,
-    stdio: [ process.stdin, process.stdout, process.stderr ]
-  });
+const getShrinkwrapDependencyObject = (shrinkwrap, shrinkwrapPath) => {
```
lurch: Would there be any benefit to memoizing this function?

scripts/clean-shrinkwrap.js
lurch: From what I can tell, would `getTopLevelDependenciesForShrinkwrapPath` be a better name? \*shrug\*

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * @param {String[]} blacklist - dependency blacklist
+ * @param {Object} qualifiers - qualifiers
+ * @param {Boolean} qualifiers.development - development dependency
+ * @param {Boolean} qualifiers.optional - optional dependency
+ * @returns {Object} filtered shrinkwrap object
+ *
+ * @example
+ * const shrinkwrapFile = require('./npm-shrinkwrap.json');
+ * const dependencyTree = getDependencyTree(shrinkwrapFile, [ 'drivelist' ]);
+ * const filteredShrinkwrap = removeDependencies(shrinkwrapFile, dependencyTree, {
```
lurch: I think this example is wrong, because `getDependencyTree` returns `Object[]` but the `blacklist` parameter here is `String[]` ?

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+  }
+
+  return shrinkwrap;
+};
+
+const blacklist = _.reduce(platformSpecificDependencies, (accumulator, dependencyPath) => {
+  return _.compact(_.concat(accumulator, [
+    getPrettyShrinkwrapDependencyObject(shrinkwrapFile, dependencyPath)
+  ], getDependencyTree(shrinkwrapFile, dependencyPath)));
+}, []);
```
lurch: Should this chunk of code be moved into e.g. a `buildBlacklistFromDependencyTree` function or something?
I'm afraid it looks pretty impenetrable to a non-Lodash-expert ;)

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * }
+ *
+ * The purpose of this function is to remove certain dependencies
+ * that match a blacklist. In order to do so, we start from the top
+ * level object, remove the blacklisted dependencies, and recurse
+ * if possible.
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} blacklist - dependency blacklist
+ * @param {Object} qualifiers - qualifiers
```
lurch: Given the single-purpose nature of this script, I wonder if it'd be worth renaming `removeDependencies` to `removeOptionalDevDependencies` and then getting rid of the `qualifiers` parameter? \*shrug\*

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+  ], getDependencyTree(shrinkwrapFile, dependencyPath)));
+}, []);
+
+const result = JSON.stringify(removeDependencies(shrinkwrapFile, blacklist, {
+  development: true,
+  optional: true
+}), null, JSON_INDENTATION_SPACES);
+
+fs.writeFileSync(NPM_SHRINKWRAP_FILE_PATH, `${result}\n`);
+console.log('Done');
```
lurch: `console.log('Done, thank f***!');` :smile:

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ *     ...
+ *   }
+ * }
+ *
+ * The purpose of this function is to remove certain dependencies
+ * that match a blacklist. In order to do so, we start from the top
+ * level object, remove the blacklisted dependencies, and recurse
+ * if possible.
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
```
lurch: Does this parameter description need to be updated to reflect the recursive nature of this function? (or is that an internal-only detail?)

scripts/clean-shrinkwrap.js
jviotti: Fixed!

jviotti:
> Yeah, it keeps popping up now and then, but only on Appveyor, and we've no idea why :-( When you restart the build (which I've just done), it usually passes on the second attempt.
(and annoyingly, unlike on Travis CI where you can restart just one arch-specific build, on Appveyor you have to restart all arch builds)

It looks like a MinGW bug. Moving stuff using Node.js file-system works fine. We only use this for the CLI generation, which we will move to `pkg`, so it should be gone soon.

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} shrinkwrap object
+ *
+ * @example
+ * const object = getShrinkwrapDependencyObject(require('./npm-shrinkwrap.json'), [
+ *   'drivelist',
+ *   'lodash'
+ * ]);
```
jviotti: Sure, I'll add some more example code.

scripts/clean-shrinkwrap.js
jviotti: Indeed, that's a great idea.

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+  }, shrinkwrap);
+};
+
+/**
+ * @summary Get a cleaned shrinkwrap dependency object
+ * @function
+ * @private
+ *
+ * @description
+ * This function transforms the output of `getShrinkwrapDependencyObject()` to
```
jviotti: Fixed!

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,222 @@
...
+ * omit unnecessary properties such as `from`, or `dependencies`.
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap file contents
+ * @param {String[]} shrinkwrapPath - path to shrinkwrap dependency
+ * @returns {Object} shrinkwrap object
```
jviotti: Fixed

scripts/clean-shrinkwrap.js
jviotti: Haha, I clarified it anyways.

scripts/clean-shrinkwrap.js
jviotti: Done!

scripts/clean-shrinkwrap.js
jviotti: The `require()` function does. We could check if the file exists before doing so, which will mean using `fs.access`, however the code would remain almost the same, with the difference that the `try`/`catch` block will be checking for `ENOENT` rather than `MODULE_NOT_FOUND`

scripts/clean-shrinkwrap.js
jviotti: Done!

scripts/clean-shrinkwrap.js
jviotti: Yeah, it was inlined at some point, but I extracted out since `getShrinkwrapDependencyManifest` was quite big already.

scripts/clean-shrinkwrap.js
jviotti: Done!

scripts/clean-shrinkwrap.js
jviotti: Makes sense, done!

scripts/clean-shrinkwrap.js
jviotti: Oops, the type declaration was indeed wrong. Thanks for the heads up!

scripts/clean-shrinkwrap.js
jviotti: Haha, what this does is basically taking every dependency in the black list, and for each of them, get an array of its dependencies + the dependency itself (hence `_.concat`). We then flatten the whole thing as we go by using a fold.

I guess we can rename the internal bit to something like "get dependency tree AND the dependency itself". Maybe that could be just `getTree()`?

scripts/clean-shrinkwrap.js
lurch: Hmmm, I've just seen in the NodeJS docs that `Using fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.`
Fair enough then ;-)

jviotti:
Fixed!

lurch:
```
Done
    npm shrinkwrap --dev
index 2c2240b..44d9863 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -53,6 +53,13 @@
       "resolved": "https://registry.npmjs.org/7zip-bin/-/7zip-bin-2.1.0.tgz",
       "dev": true
     },
+    "7zip-bin-win": {
+      "version": "2.1.0",
+      "from": "7zip-bin-win@>=2.1.0 <3.0.0",
+      "resolved": "https://registry.npmjs.org/7zip-bin-win/-/7zip-bin-win-2.1.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "abbrev": {
       "version": "1.1.0",
       "from": "abbrev@>=1.0.0 <2.0.0",
warning: LF will be replaced by CRLF in npm-shrinkwrap.json.
The file will have its original line endings in your working directory.
Command exited with code 2
```

```
Done
    npm shrinkwrap --dev
index 2c2240b..dc1282c 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -53,6 +53,13 @@
       "resolved": "https://registry.npmjs.org/7zip-bin/-/7zip-bin-2.1.0.tgz",
       "dev": true
     },
+    "7zip-bin-linux": {
+      "version": "1.1.0",
+      "from": "7zip-bin-linux@>=1.1.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/7zip-bin-linux/-/7zip-bin-linux-1.1.0.tgz",
+      "dev": true,
+      "optional": true
+    },
     "abbrev": {
       "version": "1.1.0",
       "from": "abbrev@>=1.0.0 <2.0.0",
The command "./scripts/ci/test.sh -o $HOST_OS -r $TARGET_ARCH" exited with 2.
```

:cry:

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,258 @@
...
+ * if possible.
+ *
+ * @param {Object} shrinkwrap - the shrinkwrap object
+ * @param {Object[]} blacklist - dependency blacklist
+ * @returns {Object} filtered shrinkwrap object
+ *
+ * @example
+ * const shrinkwrapFile = require('./npm-shrinkwrap.json');
+ * const dependencyTree = getDependencyTree(shrinkwrapFile, [ 'drivelist' ]);
+ * const filteredShrinkwrap = removeDependencies(shrinkwrapFile, dependencyTree);
```
lurch: function-name in the example needs updating ;)

scripts/clean-shrinkwrap.js
lurch: But `getTopLevelDependenciesForShrinkwrapPath` is such a small function, it doesn't seem to warrant being standalone? \*shrug\*

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,258 @@
...
+    .concat([ 'package.json' ])
+    .reduce((accumulator, file) => {
+      return path.join(accumulator, file);
+    }, '.')
+    .value();
+
+  try {
+
+    // For example
+    // ./node_modules/drivelist/node_modules/package.json
```
lurch: Umm, should this be `./node_modules/drivelist/package.json` ? Or maybe `./node_modules/drivelist/node_modules/lodash/package.json` ?

jviotti: Right, fixed!

jviotti:
Looks like I made a couple of subtle issues while addressing the review comments. It should work now.

scripts/clean-shrinkwrap.js
jviotti: Done!

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,258 @@
 
 'use strict';
 
+const _ = require('lodash');
+const fs = require('fs');
 const path = require('path');
-const os = require('os');
 const packageJSON = require('../package.json');
```
jviotti: We only need to `require` the package.json, however we need to shrinkwrap path in two places: a `require` call and the output of `fs.writeFile`, so better to extract it as a constant to avoid path issues (`require()` has special logic to understand `./` on Windows, but its not safe to assume that for `fs.writeFile` as well)

lurch: > but its not safe to assume that for fs.writeFile as well)

Ahh, that's the bit I was missing, thanks. So `require` definitely works with both slash-types on Windows?

scripts/clean-shrinkwrap.js
```diff
@@ -15,29 +15,258 @@
...
+    .reduce((accumulator, file) => {
+      return path.join(accumulator, file);
+    }, '.')
+    .value();
+
+  try {
+
+    // For example
+    // ./node_modules/drivelist/node_modules/lodash/package.json
+    return require(`.${path.sep}${manifestPath}`);
```
lurch: Given your comment about `require` in a comment just now, does this _need_ to be `.${path.sep}` instead of just `./` ?

scripts/clean-shrinkwrap.js
lurch: Really? The example still says `removeDependencies` whereas the function is named `removeOptionalDevelopmentDependencies`

jviotti: I forgot to push 😆

scripts/clean-shrinkwrap.js
jviotti: It would work. Since `manifestPath` is generated using `path.join`, it may include backslashes, so I decided to use `path.sep` just for consistency purposes (so we don't have paths like `./node_modules\package.json`).

lurch: ...so this line could also use `path.join` for consistency purposes? ;)

jviotti: Yeah, I really wanted to, but:

```js
> path.join('.', 'foo')
'foo'
```

(`path.join` omits the period, which `require` needs to work properly, so we need to preppend `path.sep` anyway)

jviotti:
Let's merge this to check if the logs are reduced enough to make Travis CI builds complete.

scripts/clean-shrinkwrap.js
lurch: Cool, thanks for the explanation! :+1:

PR #1553, opened 2017-06-27
chore(ci): Enable node_modules & ccache caching

jhermsmeier:
This enables caching of `node_modules`, and `ccache` directories,
as well as installation of ccache on Mac OS.

Change-Type: patch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ sudo: false
 node_js:
   - "6.10.3"
 
+cache:
+  ccache: true
+  directories:
+    - node_modules
+
 services:
   - docker
 
@@ -34,6 +39,7 @@ os:
   - osx
 
 before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then PATH=/usr/local/opt/ccache/libexec:$PATH; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       export HOST_OS="darwin";
     else
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,7 @@ image: Visual Studio 2015
 cache:
   - C:\Users\appveyor\.node-gyp
   - '%AppData%\npm-cache'
+  - node_modules
   - C:\ProgramData\chocolatey\bin -> appveyor.yml
   - C:\ProgramData\chocolatey\lib -> appveyor.yml
   - C:\Users\appveyor\AppData\Local\Temp\chocolatey -> appveyor.yml
--- a/scripts/ci/install.sh
+++ b/scripts/ci/install.sh
@@ -52,7 +52,7 @@ if [ "$ARGV_OPERATING_SYSTEM" == "linux" ]; then
 else
   if [ "$ARGV_OPERATING_SYSTEM" == "darwin" ]; then
     ./scripts/build/check-dependency.sh brew
-    brew install jq
+    brew install ccache jq
   elif [ "$ARGV_OPERATING_SYSTEM" == "win32" ]; then
     ./scripts/build/check-dependency.sh choco
     choco install nsis -version 2.51


.travis.yml
```diff
@@ -39,6 +44,8 @@ before_install:
     else
       export HOST_OS="$TRAVIS_OS_NAME";
     fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install ccache; fi
```
jviotti: Could we move this to `scripts/ci/install.sh`?

.travis.yml
```diff
@@ -39,6 +44,8 @@ before_install:
...
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then PATH=/usr/local/opt/ccache/libexec:$PATH; fi
```
jviotti: Would it do any harm if we add this in the `env` section?

.travis.yml
jhermsmeier: Sure, done

.travis.yml
jhermsmeier: I moved it up now, not sure which env section exactly you're referring to?

jviotti: I mean this: https://github.com/resin-io/etcher/blob/master/.travis.yml#L17

jhermsmeier: Oh, well, this is Mac OS only though...

jviotti: Fair enough. Why is this needed though? Shouldn't `brew` take care of symlinking any binary to `/usr/local/bin`?

jhermsmeier: I was thinking the same, but
```
$ brew info ccache
[...]
==> Caveats
To install symlinks for compilers that will automatically use
ccache, prepend this directory to your PATH:
  /usr/local/opt/ccache/libexec
```

jviotti: Bah, fair enough, I'll approve :)

lurch:
Does this mean that the Travis Linux images come with ccache pre-installed, but the Travis OSX images don't? Do we need to do anything special to get ccache and node_modules caching working inside the Docker images too?

.travis.yml
lurch: I've also just found https://docs.travis-ci.com/user/caching/#ccache-cache

jhermsmeier:
I guess we could map the `node_modules` into the docker containers, but that'd probably mess up the cache every time... and the caches are meant for the travis box itself, so I'm not sure we should do that.

lurch:
@jhermsmeier Any idea why the Travis builds are failing in #1558 ? They're printing messages like:
```
single-line-log@1.1.2 /etcher/node_modules/electron/node_modules/single-line-log
semver@5.3.0 /etcher/nochange detected (content changed, file is created, or file is deleted):
/home/travis/build/resin-io/etcher/node_modules/lzma-native/build/liblzma/build/lib/liblzma.a
/home/travis/build/resin-io/etcher/node_modules/lzma-native/build/liblzma/config.log
/home/travis/build/resin-io/etcher/node_modules/lzma-native/build/liblzma/libtool
/home/travis/build/resin-io/etcher/node_modules/lzma-native/build/liblzma/src/liblzma/.libs/liblzma.a
changes detected, packing new archive
.
.
.
.
FAILED: tar -Pzcf /home/travis/.casher/push.tgz /home/travis/build/resin-io/etcher/node_modules
tar: /home/travis/build/resin-io/etcher/node_modules: file changed as we read it
uploading archive
Done. Your build exited with 1.
```

jhermsmeier:
@lurch why did the one where it got merged go green then? https://travis-ci.org/resin-io/etcher/builds/247705868 Only the PR that followed that into master went red... Something doesn't add up :/

jhermsmeier:
Also not sure why the `node_modules` would still be modified at that point in time

jhermsmeier:
Looks like the cache builder starts while `$ ./scripts/ci/build-installers.sh -o $HOST_OS -r $TARGET_ARCH` is still running, from what I can tell

PR #1554, opened 2017-06-27
chore(scripts): Make npm output less verbose

jhermsmeier:
This adds the `--silent` flag to several npm commands,
ideally making the output less verbose, for better overview in CI logs.

Change-Type: patch

### DIFF
--- a/scripts/build/dependencies-npm.sh
+++ b/scripts/build/dependencies-npm.sh
@@ -99,9 +99,9 @@ function run_install() {
 
   # When changing between target architectures, rebuild all dependencies,
   # since compiled add-ons will not work otherwise.
-  npm rebuild
+  npm rebuild --silent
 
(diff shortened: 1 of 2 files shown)


lurch:
Cool, as long as this doesn't hide any error messages :)

lurch:
A thought I just had in the meeting - maybe it'd be useful to have a `NO_SILENT` env-var, which when set would stop the `--silent` flag being added, so that you could temporarily set `$NO_SILENT` in case you want to see the full output?
(but then again, I dunno what this flag actually turns off, so perhaps that _wouldn't_ be useful?)

jhermsmeier:
Wouldn't be useful – we're just trying to shut up the dep tree noise it puts out. On CI, loglevel warn should be enough, and `--silent` doesn't make that go away.

Issue #336, opened 2016-04-18
Implement auto-update functionality


jviotti:

probonopd:
Check [AppImageUpdate](https://github.com/probonopd/AppImageKit/blob/master/AppImageUpdate.AppDir/README.md) - it can do binary delta updates which means that only the changed parts of the binary need to be downloaded during an update.

jviotti:
Hi @probonopd ,

Thanks a lot for the suggestion. 

> We are using the ISO 9660 Volume Descriptor #1 field "Application Used" to store this information.

Do you have more documentation, or maybe an example, on how to add the update information this way?

***

What's the recommended way to ask end users to perform the upgrade? I wouldn't want to ask them to download an application `AppImageUpdate.AppImage` to perform this task. Maybe I can embed distribute this command line tool along with the `AppImage`, but renamed to something like `EtcherUpdate`?

probonopd:
> Do you have more documentation, or maybe an example, on how to add the update information this way?

Yes, [here](https://github.com/probonopd/AppImageKit/blob/master/AppImageUpdate.AppDir/usr/bin/appimageupdate#L121-L130).

>  Maybe I can embed distribute this command line tool along with the AppImage, but renamed to something like EtcherUpdate?

Absolutely, consider the `appimageupdate` script and the AppImageUpdate.AppImage GUI around it a sample implementation. You could instead have your app perform the same kind of thing natively. You could do a GUI inside your app similar to https://sparkle-project.org/.

jsejcksn:
Is this the last issue before 1.0 is released? I was looking around the site and the issues and didn't see any indicators about release date or progress toward 1.0. Thanks!

jviotti:
Hi @jsejcksn ,

The only things delaying us for v1.0 are currently:

- Auto updating functionality in all operating systems.
- Only elevating the process that performs the flashing, which I'm working on at the moment.

jsejcksn:
@jviotti Thanks for the quick reply. Is there a page other than [milestones](https://github.com/resin-io/etcher/milestones) where you post info about release progress? A twitter or blog, perhaps?

jviotti:
Hi @jsejcksn ,

Not sadly, I'm afraid, but its definitely something we should be dealing with better. As far as I've seen, most OSS projects make use of Milestones for that, so probably cleaning it up a bit should work.

Do you have another examples/suggestions?

jsejcksn:
[...]

PR #1555, opened 2017-06-27
feat(GUI): publish Electron packages to GitHub Releases

jviotti:
electron-builder provides a mechanism to deploy packages to GitHub
Releases, and then to implement auto-updates for Windows and macOS based
on them.

This commits does the following changes to the build system:

- Configure GitHub Releases deployment for DMG, ZIP, AppImage, NSIS, and
  Windows Portable (deb and rpm packages need to be published to
  Bintray, which we will do soon)

- Introduce a PUBLISH Make environment variable which when set, will
  trigger electron-builder deployments

- Cleanup Makefile now that we don't need to keep references to most of
  the packages

And the following changes to the application code:

- Re-work s3-packages.js to work with GitHub Releases

- Show the update notification dialog if a new GitHub Release was
  published

See: https://github.com/resin-io/etcher/issues/336


jviotti:
Please take a look a this PR again. The update notification dialog now works with GitHub releases.

Makefile
```diff
@@ -155,6 +155,24 @@ ELECTRON_BUILDER_OPTIONS += $(DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS)
...
+ifdef PUBLISH
+ifndef GH_TOKEN
+$(error No GitHub token found (GH_TOKEN is not set))
+endif
+ELECTRON_BUILDER_OPTIONS += --publish always
+ifeq ($(RELEASE_TYPE),production)
+# We want a human to push the final red button
+ELECTRON_BUILDER_OPTIONS += --draft
+endif
+ifeq ($(RELEASE_TYPE),snapshot)
```
lurch: Given that `RELEASE_TYPE` can't be both `production` and `snapshot` at the same time, you could put this within an else clause https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html

Makefile
```diff
@@ -286,15 +304,13 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATF
 assets/osx/installer.tiff: assets/osx/installer.png assets/osx/installer@2x.png
 	tiffutil -cathidpicheck $^ -out $@
 
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).dmg: assets/osx/installer.tiff \
-	| $(BUILD_DIRECTORY)
+electron-installer-dmg: assets/osx/installer.tiff
```
lurch: I'm :-1: on this change (and the similar changes) - I think it's useful for the `Makefile` to still refer to the actual filenames that get generated, rather than only using phoney targets everywhere.

Makefile
```diff
@@ -363,21 +375,16 @@ package-electron:
...
-electron-installer-dmg: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg
 cli-installer-tar-gz: $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).tar.gz
 TARGETS += \
 	electron-installer-dmg \
 	electron-installer-app-zip \
 	cli-installer-tar-gz
 PUBLISH_AWS_S3 += \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg \
 	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).tar.gz
```
lurch: Do we (will we?) want to shift the CLI publishing to GH too?

lurch:
Hmmm, don't the CI servers still publish snapshot builds to S3? With this change, those snapshot builds will no longer prompt to update to a newer snapshot build? :-(

lurch:
And just out of curiosity, where does the "auto-update" part fit into all of this? Or will that be a separate later PR?

Makefile

Makefile
jviotti: Yeah, we will. Currently the GUI packages are uploaded by electron-builder. I'll create a new GitHub Publish script and use it for the CLI in a following PR.

jviotti:
> Hmmm, don't the CI servers still publish snapshot builds to S3? With this change, those snapshot builds will no longer prompt to update to a newer snapshot build? :-(

Snapshot releases will now be published to GitHub Releases as well, by using the "pre-release" tag, so everything will continue working as usual, just that we will be also able to find snapshot builds by filtering for pre-release releases on GitHub.

> And just out of curiosity, where does the "auto-update" part fit into all of this? Or will that be a separate later PR?

This PR is only concerned with moving assets to GitHub Releases, and making the update notification work with that. I'll then send a new PR that only uses the update notification dialog for AppImages, but attempt to auto-update otherwise.

Makefile
jviotti: Yeah, the main motivation behind this change is that with electron-builder, is the tool what determines the final file name. Trying to keep the electron-builder configuration in sync with the target name involved a lot of work (e.g. to match how the injected extra metadata affects the final package, etc), and we don't need to reference this files anywhere else.

What would be the benefits of using the actual paths as Makefile targets? Should we go for it once we do have a Makefile rule that needs to reference it?

lurch: > Trying to keep the electron-builder configuration in sync with the target name involved a lot of work

But presumably that hard work has been done now, and we won't be changing filenames?

> What would be the benefits of using the actual paths as Makefile targets?

One of the reasons for using a Makefile instead of just a shell script is that `make` can use the dependency-graph to work out which targets need to be rebuilt when other files change, which it obviously can't do if everything is just phoney targets.

Makefile
```diff
@@ -322,19 +337,15 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET
 		--extraMetadata.version=$(APPLICATION_VERSION) \
 		--extraMetadata.packageType=AppImage
 
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-linux-$(TARGET_ARCH).zip: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_ARCH_APPIMAGE).AppImage \
-	| $(BUILD_DIRECTORY)
+electron-installer-appimage: $(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_ARCH_APPIMAGE).AppImage
```
lurch: Not related to this PR specifically, but why are the AppImages named with `$(APPLICATION_NAME_LOWERCASE)` i.e. `etcher-...` when previously they were named like `Etcher-...` ?

scripts/ci/build-installers.sh
```diff
@@ -44,6 +44,10 @@ if [ -z "$ARGV_OPERATING_SYSTEM" ] || [ -z "$ARGV_ARCHITECTURE" ]; then
   usage
 fi
 
+if [ "${GH_TOKEN:-false}" != "false" ]; then
```
lurch: So if `$GH_TOKEN` is set, we _automatically_ set `$PUBLISH` to 1? Why do we want this behaviour, instead of setting both `$GH_TOKEN` and `$PUBLISH` separately? (I guess it's easy to imagine `$GH_TOKEN` being stored as a global env-var, and then `$PUBLISH` only being selectively set on the command-line where we invoke `build-installers.sh` ?)

Makefile
```diff
@@ -155,6 +155,23 @@ ELECTRON_BUILDER_OPTIONS += $(DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS)
...
+ifdef PUBLISH
+ifndef GH_TOKEN
+$(error No GitHub token found (GH_TOKEN is not set))
+endif
+ELECTRON_BUILDER_OPTIONS += --publish always
+ifeq ($(RELEASE_TYPE),production)
+# We want a human to push the final red button
+ELECTRON_BUILDER_OPTIONS += --draft
+else
+ELECTRON_BUILDER_OPTIONS += --prerelease
```
jviotti: I see, done! Notice that we're already guarding against invalid versions in the line 40 case.

Makefile
jviotti: Is what `electron-builder` does, so we have to get along :)

docs/USER-DOCUMENTATION.md
```diff
@@ -133,15 +133,11 @@ run Etcher on a GNU/Linux system.
...
-
-Note that the value of the variable will be ignored if it doesn't match the
-release type of the current application version. For example, setting the
-variable to a production version (e.g. `ETCHER_FAKE_S3_LATEST_VERSION=2.0.0`)
-will be ignored if you're running a snapshot build, and vice-versa.
+You can set the `ETCHER_FAKE_GITHUB_LATEST_VERSION` environment variable to a
+valid semver version (greater than the current version) to trick the
+application into thinking that what you put there is the latest available
+version, therefore causing the update notification dialog to be presented at
+startup.
```
jviotti: I got rid of the fact that if the value of the variable doesn't match the current release type it gets ignored, since `electron-builder`, once its auto updates mode is enabled, allows clients to switch between production and snapshot channels at will.

lib/gui/app.js
```diff
@@ -118,7 +118,7 @@ app.run((ErrorService) => {
     includeUnstableChannel
   });
 
-  s3Packages.getLatestVersion(currentReleaseType, {
+  githubReleases.getLatestVersion({
```
jviotti: I changed the semantics a bit to accommodate to `electron-builder`'s auto update functionality, which allows the user to move between unstable and production channels. Now, when determining the available versions, we'll query ALL of them, but filter the pre-release ones if `includeUnstableChannel` is not enabled.

jviotti: I fix the `currentReleaseType` case :)

scripts/ci/build-installers.sh
jviotti: The issue is that secure environment variables, like `GH_TOKEN`, don't get applied during pull request builds, while non secure variable do (like `PUBLISH`). Setting `GH_TOKEN` and `PUBLISH` in the Etcher project meant that during pull requests, `PUBLISH` would be present while `GH_TOKEN` would be missing, causing `electron-builder` to complain about not being able to publish because there was no token.

I guess we can fix this by checking the opposite case: If `PUBLISH` is enabled, but there is no `GH_TOKEN`, we show a warning and unset `PUBLISH`

Makefile
jviotti: Fair enough. I'll revert these set of changes...

Makefile
```diff
@@ -370,21 +386,19 @@ package-electron:
...
 	cli-installer-tar-gz
 PUBLISH_AWS_S3 += \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).dmg \
 	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).tar.gz
 endif
 
 ifeq ($(TARGET_PLATFORM),linux)
-electron-installer-appimage: $(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
+electron-installer-appimage: $(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-linux-$(TARGET_ARCH).zip
```
lurch: I guess the change here (and also similar changes on lines 389, 390, 419 and 420) isn't strictly necessary, given the `ifeq ($(TARGET_PLATFORM),linux)` immediately above? ;)

Makefile
lurch: Thanks :)  It also allows some of my clever Makefile tricks (which I'll hopefully be PRing this week) to work ;-)

Makefile
lurch: Hmm, and there's no option to override that, to keep consistency with our original filenames (and other OSes) ? :-/

scripts/ci/build-installers.sh
lurch: I guess an 'easy' way to fix that, would be to make `PUBLISH` a secure variable too, so that it's only set when `GH_TOKEN` is? ;-)

> I guess we can fix this by checking the opposite case: If PUBLISH is enabled, but there is no GH_TOKEN, we show a warning and unset PUBLISH

That also works :)

lurch:
For some reason you didn't reply to my comment:
> So, previously (when we were using AWS S3) we had both Stable/Unstable versions, and Production/Snapshot releases, with these two distinctions being orthogonal (so a particular release of Etcher was either Stable Production (e.g. 1.0.0), Unstable Production (e.g. 1.0.0-beta.19), Stable Snapshot (e.g. 1.0.0+2201e5f) or Unstable Snapshot (e.g. 1.0.0-beta.19+531ab82) ). See https://github.com/resin-io/etcher/blob/master/docs/PUBLISHING.md for more info.
But it looks like with this new scheme (using GitHub Releases), the only distinction is between release and prerelease? How do the old categories map onto the new categories? And how does this affect the update logic? Especially with regards to the includeUnstableChannel option?
(I'm afraid I've not studied the code closely enough to be able to answer this myself).

Looking again at the code myself, it looks like `Stable Production` and `Unstable Production` now both map to `release`, and `Stable Snapshot` and `Unstable Snapshot` now both map to `prerelease` ?
AFAICT there's no additional filtering done in `github-releases.js` based on whether we've got `1.0.0` or `1.0.0-beta.19`, so I guess `includeUnstableChannel` now _effectively_ means `includePrereleaseChannel` (i.e. `includeSnapshotChannel`), as there's no distinction between stable and unstable versions? If this **is** the approach we're going for, then I guess this means some of the unit-tests could be simplified, as I guess combinations like
```
{
    tag_name: 'v1.0.0-beta.16',
    prerelease: true
},
```
are now invalid, as that'd actually be a `release` (production) version rather than a `prerelease` (snapshot) version. And the info on https://github.com/resin-io/etcher/blob/master/docs/PUBLISHING.md about the different version types and update channels probably needs updating too. (in fact, there are several places where that document needs updating, e.g. we no longer build to `build/out`, and after this PR we won't be publishing to S3)

When we _do_ start publishing snapshot builds to https://github.com/resin-io/etcher/releases (from the CI services) I assume that GH will hide them from that page by default? Because otherwise our 'real' production releases would get swamped in the noise?

> since electron-builder, once its auto updates mode is enabled, allows clients to switch between production and snapshot channels at will.

Are you just referring to the `includeUnstableChannel` setting here, or did you mean some other feature that `electron-builder` offers?
Thinking about it, I wonder if it **is** worth renaming that setting, since a "Snapshot channel" is something _very_ different to a "Unstable Production channel". (because (at least in theory) an Unstable Production version would have more testing than a Snapshot version, and an Unstable Production channel would have less frequent updates than a Snapshot channel)


Apologies for all the awkward questions, but this is quite a big change so I want to make sure that we're doing it right ;-)

Makefile
jviotti: Right, fixed!

Makefile
jviotti: We could fix it the manual way, but I guess its not worth it. In the future, it'd be nice to move to electron-builder defaults (which is kinda consistent with itself and also consistent with other Electron apps out there).

scripts/ci/build-installers.sh
jviotti: Haha, yeah, let's go for it. After all, we don't want everyone to see that such variable is set to `1` :P

Makefile
lurch: Fair enough.

> it'd be nice to move to electron-builder defaults

Given all the other stuff we're currently overhauling, would it be worth making that change now, before the release of v1.1 ?

Makefile
```diff
@@ -322,19 +341,16 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET
 		--extraMetadata.version=$(APPLICATION_VERSION) \
 		--extraMetadata.packageType=AppImage
 
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-linux-$(TARGET_ARCH).zip: \
-	$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_ARCH_APPIMAGE).AppImage \
-	| $(BUILD_DIRECTORY)
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip: \
```
lurch: In contrast to the change you just made, these lines are _outside_ of any `ifeq ($(TARGET_PLATFORM),` lines (although I can't remember now if there's a good reason for that), which means the _only_ difference between the Linux zipfile target and the OSX zipfile target is that the Linux zipfile starts with a lowercase letter!
darwin:
```
$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
```
linux:
```
$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_PLATFORM)-$(TARGET_ARCH).zip
```

jviotti:
No worries, I'm glad you're reviewing this :)

So here's more or less the plan: S3 allows us to impose our own directory structure, and have more control over what things represent (e.g. stable snapshots, etc). The GitHub model is way, way simpler, and as you noticed, only allows two states: stable and pre-release.

electron-builder is heavily inspired by GitHub Releases, and mostly works with the same stable and pre-release states. electron-builder will provide us with the whole auto updating mechanism, so we have to play along (a side effect of ditching our custom build system, etc).

This PR then tries to simplify our more complex system of releases to fit that. electron-builder doesn't care about what a version is (e.g. it doesn't even care if its semver or any random string), and **only** looks at the pre-release tag.

The current approach then becomes:

- 1.0.0 -> stable
- 1.0.0-beta.19 -> pre-release
- 1.0.0+aaaaaaa -> pre-release
- 1.0.0-beta.19+aaaaaaa -> pre-release

As you can see, everything is pre-release except for the "main" production versions.

Of course, this means that if a user enables the unstable update channel, then he will get all of the pre-release ones, but I guess we can live with that (not much we can do without going custom in the auto update logic, which is a lot of work).

> AFAICT there's no additional filtering done in github-releases.js based on whether we've got 1.0.0 or 1.0.0-beta.19, so I guess includeUnstableChannel now effectively means includePrereleaseChannel (i.e. includeSnapshotChannel), as there's no distinction between stable and unstable versions? If this is the approach we're going for, then I guess this means some of the unit-tests could be simplified, as I guess combinations like

Right. Renaming the variable to `includePrereleaseChannel` sounds like a great idea.

> as I guess combinations like ... are now invalid, as that'd actually be a release (production) version rather than a prerelease (snapshot) version.

electron-builder doesn't care about the actual version, so we could in theory publish 1.1.0 and call it a pre-release, and electron-builder will think that's simply an unstable version.

> And the info on https://github.com/resin-io/etcher/blob/master/docs/PUBLISHING.md about the different version types and update channels probably needs updating too. (in fact, there are several places where that document needs updating, e.g. we no longer build to build/out, and after this PR we won't be publishing to S3)

Right, good point. I'll go through the whole docs and update them accordingly.

> When we do start publishing snapshot builds to https://github.com/resin-io/etcher/releases (from the CI services) I assume that GH will hide them from that page by default? Because otherwise our 'real' production releases would get swamped in the noise?

I don't think we can do much about that unless we want to give up on snapshot releases. My thought was that we should point users to the etcher website, which will abstract the mess and provide filtering features, plus always keep the stable one at the top.

> Are you just referring to the includeUnstableChannel setting here, or did you mean some other feature that electron-builder offers?

Yeah, the electron-builder auto-updates technology allows you to pass a boolean option to check the pre-release channel or not, so we can simply plug in our existing `includeUnstableChannel` setting.

> Thinking about it, I wonder if it is worth renaming that setting, since a "Snapshot channel" is something very different to a "Unstable Production channel". (because (at least in theory) an Unstable Production version would have more testing than a Snapshot version, and an Unstable Production channel would have less frequent updates than a Snapshot channel)

I think we'll have to live with the fact we won't be able to make that distinction. Being able to re-use so many auto updating and packaging code probably outweighs that disadvantage.

Makefile
jviotti: Sounds good, I'll be updating this PR with those changes.

Makefile
jviotti: Lets see how this looks after making use of electron-builder's defaults. I think the distinction will be better then.

lurch:
Thanks for answering my questions :)

> The current approach then becomes:
1.0.0 -> stable
1.0.0-beta.19 -> pre-release
1.0.0+aaaaaaa -> pre-release
1.0.0-beta.19+aaaaaaa -> pre-release

Ahhh, in that case you might want to tweak the logic in the `Makefile` to reflect that then (if possible), as currently `1.0.0-beta.19` will get published as a stable release.

> Right. Renaming the variable to `includePrereleaseChannel` sounds like a great idea.

With the corresponding changes in the settings and GUI code please ;)

> Of course, this means that if a user enables the unstable update channel, then he will get all of the pre-release ones, but I guess we can live with that (not much we can do without going custom in the auto update logic, which is a lot of work).

Hmmm, what if we instead removed the `includePrereleaseChannel` option from the GUI (and settings) and instead tweaked the logic so that stable releases only auto-update to stable releases, and pre-releases only auto-update to pre-releases? (I've not looked into the electron-builder auto-updates at all, but is that something they allow?) Would possibly eliminate some of this confusion.

> I don't think we can do much about that unless we want to give up on snapshot releases.

Something I forgot to ask earlier - given the amount of snapshot releases we generate (across all combinations of OSes, architectures and package-types), are GitHub "happy" for their releases mechanism to also host the vast number of snapshot builds we'll be publishing, or is it "supposed" to only be used for 'real' releases?

> My thought was that we should point users to the etcher website, which will abstract the mess and provide filtering features, plus always keep the stable one at the top.

Cool. Maybe we could also have an 'advanced' or 'beta' page on the etcher website, with a link to the latest pre-release version?

> > unit-tests...

> so we could in theory publish 1.1.0 and call it a pre-release, and electron-builder will think that's simply an unstable version.

Still, probably doesn't hurt to update the unit-tests to reflect what will actually be happening in reality, to prevent any future confusion? ;)

> Being able to re-use so many auto updating and packaging code probably outweighs that disadvantage.

Sounds good :+1:

jviotti:
This PR is stale. Closing...

PR #1557, opened 2017-06-28
feat(gui): Show friendlier error dialog when opening images

jhermsmeier:
This displays a friendlier error dialog if opening an image fails
due any reason (like i.e. an unsupported compression method)

screen shot 2017-06-23 at 17 21 10

Connects To: #1499 
Change-Type: minor


jhermsmeier:
I'll add more compression level / type tests for the other compression algorithms in a separate PR, to keep the PRs more concise and reviewable.

PR #1558, opened 2017-06-28
feat(GUI): add webview api version parameter

Shou:
We add the API version sent to the banner via a GET search parameter,
allowing for more intricate control over what data needs or can be sent.

Changelog-Entry: Add Webview API version parameter.

### DIFF
--- a/lib/gui/components/safe-webview.js
+++ b/lib/gui/components/safe-webview.js
@@ -46,6 +46,22 @@ const VERSION_PARAM = 'etcher-version';
 const ELECTRON_SESSION = 'persist:success-banner';
 
 /**
+ * @summary API version search parameter key
+ * @constant
+ * @private
+ * @type {String}
+ */
+const API_VERSION_PARAM = 'api-version';
+
+/**
+ * @summary Webview API version
+ * @constant
+ * @private
+ * @type {String}
+ */
+const API_VERSION = 1;
+
+/**
  * @summary Webviews that hide/show depending on the HTTP status returned
  * @type {Object}
  * @public
@@ -69,6 +85,7 @@ class SafeWebview extends react.PureComponent {
 
     // We set the version GET parameter here.
     url.searchParams.set(VERSION_PARAM, packageJSON.version);
+    url.searchParams.set(API_VERSION_PARAM, API_VERSION);
 
     this.entryHref = url.href;
 
--- a/lib/gui/components/safe-webview.js
+++ b/lib/gui/components/safe-webview.js
@@ -30,14 +30,6 @@ const MODULE_NAME = 'Etcher.Components.SafeWebview';
 const angularSafeWebview = angular.module(MODULE_NAME, []);
 
 /**
- * @summary GET parameter sent to the initial webview source URL
- * @constant
- * @private
- * @type {String}
- */
-const ETCHER_VERSION_PARAM = 'etcher-version';
-
-/**
  * @summary Electron session identifier
  * @constant
  * @private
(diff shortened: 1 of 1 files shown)


lib/gui/components/safe-webview.js
```diff
@@ -46,6 +46,22 @@ const VERSION_PARAM = 'etcher-version';
 const ELECTRON_SESSION = 'persist:success-banner';
 
 /**
+ * @summary API version search parameter key
+ * @constant
+ * @private
+ * @type {String}
+ */
+const API_VERSION_PARAM = 'api-version';
```
jviotti: Can we add a big comment in the `@description` section stating that changing this will break compatibility with the HTML banner and therefore we should only change it if absolutely necessary?

lib/gui/components/safe-webview.js
```diff
@@ -46,6 +46,22 @@ const VERSION_PARAM = 'etcher-version';
...
+
+/**
+ * @summary Webview API version
+ * @constant
+ * @private
+ * @type {String}
+ */
+const API_VERSION = 1;
```
lurch: Should we start off at 2, and treat `API_VERSION` 1 as the version from before the `api-version` key was added?

lib/gui/components/safe-webview.js
```diff
@@ -69,6 +85,7 @@ class SafeWebview extends react.PureComponent {
 
     // We set the version GET parameter here.
     url.searchParams.set(VERSION_PARAM, packageJSON.version);
```
lurch: Perhaps the `VERSION_PARAM` constant should now be renamed to `ETCHER_VERSION_PARAM` ?
And I guess the comment above could be changed to `// We set the version GET parameters here.` ;)

lib/gui/components/safe-webview.js
lurch: I guess you could also explain that the HTML banner can use this version-number to determine which features the safe-webview supports (independent of the Etcher version-number).

(perhaps it's just me, but 'API' feels like the wrong term for this)

lib/gui/components/safe-webview.js
Shou: @jviotti mentioned that we should just treat no given `API_VERSION` as such, which we can call the the "zeroth" API version I guess.

Shou:
Once the tests pass we should be good to go.

lib/gui/components/safe-webview.js
```diff
@@ -46,6 +46,28 @@ const VERSION_PARAM = 'etcher-version';
 const ELECTRON_SESSION = 'persist:success-banner';
 
 /**
+ * @summary API version search parameter key
```
lurch: Minor nit, but could we hyphenate this as `search-parameter` ? To stop people thinking that "version search" is a thing ;)
And could we update the `@summary` for `ETCHER_VERSION_PARAM` so that the two read similarly? (and given their similarity, perhaps `ETCHER_VERSION_PARAM` and `API_VERSION_PARAM` should be moved next to each other in the source-code?)

lurch:
Looks like the Travis builds have been broken by #1553 ?

lib/gui/components/safe-webview.js
Shou: Done

lib/gui/components/safe-webview.js
jviotti: Ping @Shou

jviotti: Ah, you addressed it right below, sorry!

lurch:
@Shou Hopefully if you rebase this to master, the Travis tests should work now? :)

Shou:
@lurch doesn't look that way, unfortunately.

lurch:
The TravisCI logfiles (even in 'raw' mode) are still dying in a seemingly random place :cry: 
Ping @jhermsmeier Did you manage to get anywhere with this?

jhermsmeier:
@lurch didn't get it last friday, but have just kicked off https://travis-ci.org/resin-io/etcher/builds/249631499, with silenced output – which will hopefully get us some insight as to what's happening there

lurch:
@Shou Could you try rebasing again? Looks like Jonas and Juanchi (the J-brothers?) have fixed it :)

Shou:
Awesome, it's all pristine green now.

Issue #1459, opened 2017-05-19
Error: File path or name contains unicode character


jiejiecn:
- **Etcher version: 1.0.0**
- **Operating system and architecture: WIN10 x64 Chinese**

If the image file is located in a unicode path or file name, Etcher rasied an error "Unable to access file"
The image xxxxxxxxxx is not accessible

![snip_20170519170553](https://cloud.githubusercontent.com/assets/11746870/26241026/cff9fd76-3cb5-11e7-9905-b4f714330088.png)

lurch:
Ahhh, thanks for the bug report @jiejiecn , looks like I also need to add Unicode characters to the testing I'm doing in #1376

lurch:
I've done some testing, and this appears to be limited to Windows (hands up all those who find that surprising!).

I created a test directory named `test丠ir` and placed an ISO inside it, and on Linux Etcher 1.0.0 was able to flash the iso okay, but on Windows I got a similar error to @jiejiecn above...
![image](https://cloud.githubusercontent.com/assets/476186/26252750/048c253a-3ca9-11e7-8ec7-78d8a651ad4b.png)

The interesting thing is that the DevTools console displays the file-path correctly:
```
WRITER: Running: C:\Program Files\Etcher\Etcher.exe C:\Program Files\Etcher\resources\app.asar\lib\child-writer\writer-proxy.js C:\Program Files\Etcher\resources\app.asar C:\Users\IEUser\Downloads\test丠ir\alpine-standard-3.5.2-x86_64.iso --drive \\.\PHYSICALDRIVE1 --unmount --check

[...]

PR #1559, opened 2017-06-28
fix(elevator): make use of wide characters everywhere

jviotti:
Switching the Windows language to a language encoded using UTF-16 (like
Japanese) revealed that our native elevator module garbles language
specific strings, making the child writer process unable to find an
image inside a language specific directory, and thus resulting in a
"File is not accessible" error.

As a solution, we make use of `std::wstring` and wide character Windows
API functions in the elevator module.

Change-Type: patch
Changelog-Entry: Fix "file is not accessible" error when flashing an image that lives inside a directory whose name is UTF-16 encoded on Windows.
Fixes: https://github.com/resin-io/etcher/issues/1459

### DIFF
--- a/src/os/win32/elevate.cpp
+++ b/src/os/win32/elevate.cpp
@@ -16,11 +16,11 @@
 
 #include "os/elevate.h"
 
-static std::string JoinArguments(std::vector<std::string> arguments) {
-  std::ostringstream result;
+static std::wstring JoinArguments(std::vector<std::wstring> arguments) {
+  std::wostringstream result;
 
   std::copy(arguments.begin(), arguments.end(),
-            std::ostream_iterator<std::string>(result, " "));
+            std::ostream_iterator<std::wstring, wchar_t>(result, L" "));
 
   return result.str();
 }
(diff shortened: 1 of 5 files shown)


src/os/win32/elevate.cpp
```diff
@@ -16,34 +16,36 @@
...
 #include "os/elevate.h"
 
-static std::string JoinArguments(std::vector<std::string> arguments) {
-  std::ostringstream result;
+static std::wstring JoinArguments(std::vector<std::wstring> arguments) {
+  std::wstring result = L"";
 
-  std::copy(arguments.begin(), arguments.end(),
-            std::ostream_iterator<std::string>(result, " "));
+  for (const std::wstring argument : arguments) {
```
jviotti: I resorted to this loop given making `stringstream` work with `wstring` was not very straightforward. I'll keep trying to crack that nut in the mean time.

jviotti: Done!

src/utils/v8utils.h
```diff
@@ -20,18 +20,22 @@
...
+std::vector<std::wstring> GetArguments(v8::Local<v8::Array> arguments);
 }  // namespace v8utils
 }  // namespace etcher
 
 #define YIELD_ERROR(CALLBACK, ERROR)                                           \
   {                                                                            \
-    v8::Local<v8::Value> argv[1] = {Nan::Error((ERROR))};                      \
-    Nan::MakeCallback(Nan::GetCurrentContext()->Global(), (CALLBACK), 1,       \
-                      argv);                                                   \
+    const uint16_t *message = (const uint16_t *)(ERROR).c_str();               \
```
lurch: Why does this use `uint16_t` when you used `wchar_t` earlier?

lurch:
I guess this fixes #1459 ?

src/utils/v8utils.h
jviotti: This is already a `wchar_t`. The `NewFromTwoByte` takes an array of `uint16_t` (https://v8.paulfryzel.com/docs/master/classv8_1_1_string.html#a6b3efe5df1016e4f9f485e79b0acdf5b), so I'm basically casting it for that purpose.

jviotti:
Yep, that's exactly the issue I encountered. I didn't know there was an open ticket for it already! :)

src/utils/v8utils.h
lurch: I wonder if it'd be better to only cast it inside the call to `NewFromTwoByte`, thus keeping it as `wchar_t` for as long as possible?
i.e. `Nan::Error(v8::String::NewFromTwoByte(isolate, (const uint16_t *)message))`

lurch:
So perhaps you could reference the open issue number in the PR and commit message? ;-)

src/utils/v8utils.h
jviotti: Makes sense, done!

jviotti:
Done!

PR #1560, opened 2017-06-28
chore:(windows): make use of NSIS one-click installers

jviotti:
One-click installers are a "modern" version of usual wizard installers,
and somewhat resemble the Squirrel.Windows installers you can see in
apps like Atom.

As the name implies, the user needs to agree the license, and that's
all, the remaining installer is non-interactive, leading to a very
smooth user experience.

Change-Type: patch

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -29,8 +29,7 @@ dmg:
 win:
   icon: assets/icon.ico
 nsis:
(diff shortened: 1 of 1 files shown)


jviotti:
Looks like we're getting cache related issues here. Can you take a look @jhermsmeier ?

PR #1561, opened 2017-06-29
chore(package): Disable node_modules caching on Travis CI

jhermsmeier:
This disabled caching of the `node_modules` directory again,
as this is causing failures on Travis CI while building the installers.

Change-Type: patch
Connects To: #1553

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ node_js:
 
 cache:
   ccache: true
-  directories:
-    - node_modules
 
 services:
   - docker


jhermsmeier:
Lets disable this for now, so we can continue working, until we get to the bottom of this.

lurch:
Maybe it'd be worth getting in touch with Travis support, to see if there's a way round this caching problem?

Travis builds are still failing on Linux, but looking at the logs, I can't see any reason why - they just seem to stop at a random place? :sob:

PR #1565, opened 2017-06-30
test(image-stream): Test more compression methods

jhermsmeier:
This adds more compressed test images with various compression methods used for the respective archive / compression types (bz2, dmg, and zip). Nothing is added for gzip, as there is only one compression method specified for gzip (CM=8, 0-7 are reserved), which is deflate, and already being tested.

Change-Type: patch
Connects To: #1499


lurch:
Fantastic :)

PR #1567, opened 2017-07-03
chore(scripts): Silence stdout for npm install

jhermsmeier:
This silences `npm install`'s stdout to omit noisy output, and only show warnings & errors

Change-Type: patch


jviotti:
The logs are so, so beautifully small. I love it.

lurch:
It's nice not to have the logs spammed with all the lzma-building commands :)

PR #1569, opened 2017-07-04
refactor(GUI): decouple localStorage from the redux store

jviotti:
We're currently persisting the user settings in localSettings by using a
redux plugin called redux-localstorage. As a way to decouple the redux
store from a technology that is browser specific, this commit makes the
following changes:

- Create local-settings.js, which is concerned with managing settings in
  a persisting location

- Decouple the redux store from the persisting storage method

- Extend the settings model to persist settings, cache reads, etc

Change-Type: patch

### DIFF
--- a/lib/gui/models/local-settings.js
+++ b/lib/gui/models/local-settings.js
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2017 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+/**
+ * @summary Local storage settings key
+ * @constant
+ * @type {String}
+ */
+const LOCAL_STORAGE_SETTINGS_KEY = 'etcher-settings';
+
+/**
+ * @summary Read all local settings
+ * @function
+ * @public
+ *
+ * @returns {Object} local settings
+ *
+ * @example
+ * const settings = localSettings.readAll();
+ */
+exports.readAll = () => {
+  return JSON.parse(localStorage.getItem(LOCAL_STORAGE_SETTINGS_KEY)) || {};
+};
+
+/**
+ * @summary Write local settings
+ * @function
+ * @public
+ *
+ * @param {Object} settings - settings
+ *
+ * @example
+ * localSettings.writeAll({
+ *   foo: 'bar'
+ * });
+ */
+exports.writeAll = (settings) => {
+  const INDENTATION_SPACES = 2;
+  localStorage.setItem(LOCAL_STORAGE_SETTINGS_KEY, JSON.stringify(settings, null, INDENTATION_SPACES));
+};
+
+/**
+ * @summary Clear the local settings
+ * @function
+ * @private
+ *
+ * @description
+ * Exported for testing purposes
+ *
+ * @example
+ * localSettings.clear();
+ */
+exports.clear = () => {
+  localStorage.removeItem(LOCAL_STORAGE_SETTINGS_KEY);
+};
--- a/lib/gui/models/settings.js
+++ b/lib/gui/models/settings.js
@@ -22,7 +22,99 @@
 
 const _ = require('lodash');
 const Store = require('./store');
+const localSettings = require('./local-settings');
 const errors = require('../../shared/errors');
+
+/**
+ * @summary Set a settings object
+ * @function
+ * @private
+ *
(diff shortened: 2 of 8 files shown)


jviotti:
@lurch Check this PR. It fixes https://github.com/resin-io/etcher/pull/1382#issuecomment-299179917 as far as I can tell (the bug report is a bit complex, so not sure I covered all cases.)

jviotti:
After this PR is merged, we should be able to move the redux store and all models except "settings", to the SDK side 👯

lib/gui/models/local-settings.js
```diff
@@ -0,0 +1,70 @@
...
+ *   foo: 'bar'
+ * });
+ */
+exports.writeAll = (settings) => {
+  const INDENTATION_SPACES = 2;
+  localStorage.setItem(LOCAL_STORAGE_SETTINGS_KEY, JSON.stringify(settings, null, INDENTATION_SPACES));
+};
+
+/**
+ * @summary Clear the locl settings
```
Shou: Small typo here, `locl` -> `local`

lib/gui/models/settings.js
```diff
@@ -70,7 +157,7 @@ exports.set = (key, value) => {
  * const value = settings.get('unmountOnSuccess');
  */
 exports.get = (key) => {
-  return this.getAll()[key];
+  return _.get(exports.getAll(), key);
```
Shou: Should we do `[ key ]` here for consistency? Either works though.

lib/gui/models/local-settings.js
```diff
@@ -0,0 +1,70 @@
...
+ * @function
+ * @public
+ *
+ * @returns {Object} local settings
+ *
+ * @example
+ * const settings = localSettings.readAll();
+ */
+exports.readAll = () => {
+  return JSON.parse(localStorage.getItem(LOCAL_STORAGE_SETTINGS_KEY)) || {};
```
Shou: They're "implicitly imported", they just exist in the browser environment basically.

lib/gui/models/local-settings.js
jviotti: Fixed!

lib/gui/models/settings.js
jviotti: Sounds good, done!

lib/gui/models/local-settings.js
jviotti: Yeah, browser-related global stuff :)

lib/gui/models/local-settings.js
jviotti: Fixed!

lurch:
I can't quite tell if this bug fixes the situation I described in https://github.com/resin-io/etcher/pull/1382#issuecomment-299179917 or not (I tried testing, but with all the different versions (and with some versions of Etcher seeming to store their settings in `~/.config/etcher/Local Storage/file__0.localstorage` and others using `~/.config/Electron/Local Storage/file__0.localstorage`) I got in a bit of a muddle), so let me try explaining the problem a bit clearer:

In version1 of the application imagine there are settings configA (defaults to false) and configB (defaults to true).
In version2 of the application, imagine there's still settings configA and configB, and there's also a new setting configC (defaults to true).

I think the problem I was describing in #1382 was that if (on a never-before touched PC) you install and run version1, then you'll get `{configA: false, configB: true}`.
And if on a different never-before-touched PC you install and run version2, then you'll correctly get `{configA: false, configB: true, configC: true}`.
But then if on a 3rd PC you installed and ran version1, and then closed it and installed and ran version2, you'd get `{configA: false, configB: true, configC: false}`. (Note that configC is false, when you expect it to default to true)

Perhaps you can figure out @jviotti if that is indeed fixed by this PR? (i.e. on the PC where you installed and ran version1 and then installed and ran version2, will configC now default to true?)  Are there any unit-tests that cover this?

jviotti:
@lurch Thanks for the explanations. I think we can write a unit test for that particular case. Let me check...

jviotti:
I can confirm this issue is fixed, and there is already a test case the checks that scenario:

```js
it('should extend the application state with the local settings content', function() {
  const object = {
    foo: 'bar'
  };

  m.chai.expect(settings.getAll()).to.deep.equal(DEFAULT_SETTINGS);
  localSettings.writeAll(object);
  m.chai.expect(settings.getAll()).to.deep.equal(DEFAULT_SETTINGS);
  settings.load();
  m.chai.expect(settings.getAll()).to.deep.equal(_.assign({}, DEFAULT_SETTINGS, object));
});
```

Here is what's going on:

- The redux application settings are the default ones
- The local settings contain just `foo: bar`
- The `.load()` function is called, which reads and loads the local settings
- The resulting application settings are the default ones + `foo: bar`

This covers the issue you discovered, because even if the local file lacks settings that are on the default ones (e.g. because new ones were introduced), they are both merged before been assigned to the redux store.

lurch:
Okay, cool. So it definitely worked differently before this PR then? I'm afraid with `settings`, `localSettings` and `DEFAULT_SETTINGS` I find it kind of hard to follow what the unit-tests are doing ;-)  (but that's my problem, AFAIK there's nothing wrong with the unit-tests)

jviotti:
Yeah, there were some issues with data merging before, and we didn't have a lot of tests that covered how data was synced in local storage.

Issue #1328, opened 2017-04-20
Inject a tag to builds to determine where they came from


jviotti:
For example, we can have an "AppImage" tag, or an "installer" or "standalone" tag for Windows. We can then send that to Mixpanel.

lurch:
We should probably break down "installer" into "downloaded from the Etcher website" and "installed by chocolately"? Or is that level of distinction not possible? Does "standalone" mean the "portable" version?

Similarly on Linux, it might be nice if we could tell if the user had installed the .deb file by downloading it from S3, or if they'd installed it from bintray.

PR #1570, opened 2017-07-04
chore: send analytics about package types

jviotti:
This commit adds a `packageType` property to package.json, which is set
according to the package type of the Etcher target (e.g: dmg, nsis, deb,
etc).

This information is then sent to Mixpanel as the `packageType` property
of the "Application start" event.

Change-Type: patch
Changelog-Entry: Send anonymous analytics about package types.
Fixes: https://github.com/resin-io/etcher/issues/1328

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,12 +289,14 @@ assets/osx/installer.tiff: assets/osx/installer.png assets/osx/installer@2x.png
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).dmg: assets/osx/installer.tiff \
 	| $(BUILD_DIRECTORY)
 	TARGET_ARCH=$(TARGET_ARCH) $(NPX) build --mac dmg $(ELECTRON_BUILDER_OPTIONS) \
-		 --extraMetadata.version=$(APPLICATION_VERSION)
+		--extraMetadata.version=$(APPLICATION_VERSION) \
+		--extraMetadata.packageType=dmg
 
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).zip: assets/osx/installer.tiff \
 	| $(BUILD_DIRECTORY)
 	TARGET_ARCH=$(TARGET_ARCH) $(NPX) build --mac zip $(ELECTRON_BUILDER_OPTIONS) \
-		 --extraMetadata.version=$(APPLICATION_VERSION)
+		--extraMetadata.version=$(APPLICATION_VERSION) \
+		--extraMetadata.packageType=zip
 
 APPLICATION_NAME_ELECTRON = $(APPLICATION_NAME_LOWERCASE)-electron
 
@@ -303,6 +305,7 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)-$(APPLICATION_VERSION_REDHAT).$(
 	$(NPX) build --linux rpm $(ELECTRON_BUILDER_OPTIONS) \
 		--extraMetadata.name=$(APPLICATION_NAME_ELECTRON) \
 		--extraMetadata.version=$(APPLICATION_VERSION_REDHAT) \
+		--extraMetadata.packageType=rpm \
 		$(DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS)
 
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)_$(APPLICATION_VERSION_DEBIAN)_$(TARGET_ARCH_DEBIAN).deb: \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -299,11 +300,13 @@ assets/osx/installer.tiff: assets/osx/installer.png assets/osx/installer@2x.png
...
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).dmg: assets/osx/installer.tiff \
 	| $(BUILD_DIRECTORY)
-	TARGET_ARCH=$(TARGET_ARCH) $(NPX) build --mac dmg $(ELECTRON_BUILDER_OPTIONS)
+	TARGET_ARCH=$(TARGET_ARCH) $(NPX) build --mac dmg $(ELECTRON_BUILDER_OPTIONS) \
+		--extraMetadata.$(PACKAGE_TYPE_PROPERTY)=dmg
 
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-darwin-$(TARGET_ARCH).zip: assets/osx/installer.tiff \
 	| $(BUILD_DIRECTORY)
-	TARGET_ARCH=$(TARGET_ARCH) $(NPX) build --mac zip $(ELECTRON_BUILDER_OPTIONS)
+	TARGET_ARCH=$(TARGET_ARCH) $(NPX) build --mac zip $(ELECTRON_BUILDER_OPTIONS) \
```
lurch: Now that we're using electron-builder's `build` command, why are we still setting `TARGET_ARCH` here?

Makefile
```diff
@@ -312,18 +315,21 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)-$(APPLICATION_VERSION_REDHAT).$(
...
 		--extraMetadata.name=$(APPLICATION_NAME_ELECTRON) \
 		--extraMetadata.version=$(APPLICATION_VERSION_DEBIAN) \
+		--extraMetadata.$(PACKAGE_TYPE_PROPERTY)=deb \
 		$(DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS)
 
 $(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_ARCH_APPIMAGE).AppImage: \
 	| $(BUILD_DIRECTORY)
-	$(NPX) build --linux AppImage $(ELECTRON_BUILDER_OPTIONS)
+	$(NPX) build --linux AppImage $(ELECTRON_BUILDER_OPTIONS) \
+		--extraMetadata.$(PACKAGE_TYPE_PROPERTY)=appimage
```
lurch: I wonder if it'd be worth casing this as `AppImage` instead of `appimage`, so that in every Makefle rule the packageType exactly matches the target string given to electron-builder?

lib/gui/app.js
```diff
@@ -87,7 +87,9 @@ app.run(() => {
 });
 
 app.run((ErrorService) => {
-  analytics.logEvent('Application start');
+  analytics.logEvent('Application start', {
+    packageType: packageJSON.packageType
```
lurch: Should this also log the application version? AFAICT from reading through the code, the version is only logged by the 'Checking for updates' event, so if we instead hit the 'Not checking for updates' event (i.e. `(!shouldCheckForUpdates || !updatesEnabled)`) then the `currentVersion` doesn't get logged?

Also, something I've been wondering for a while is whether the `console.log` on line 76 should also report the `currentVersion`, so that when people copy'n'paste us their DevTools logs, we can see from the logs what version they're running. And if you think that's a good idea, then it's probably worth adding this `packageType` info in there too?

Makefile
jviotti: We make use of the variable to tweak the file names in the configuration file:

```yml
artifactName: "${productName}-${version}-darwin-${env.TARGET_ARCH}.${ext}"
```

Makefile
jviotti: Sounds reasonable!

Makefile
lurch: Ahhh. So how come it's only on the `--mac` and `--win` but not the `--linux` lines?

lib/gui/app.js
jviotti: Sounds like a great idea, done!

jviotti:
Damn Appveyor cache. I'll rebase once https://github.com/resin-io/etcher/pull/1580 is merged.

jviotti:
Ping @lurch. What do you think of this?

Makefile
```diff
@@ -189,6 +189,7 @@ PRODUCT_NAME = etcher
 APPLICATION_NAME_LOWERCASE = $(shell echo $(APPLICATION_NAME) | tr A-Z a-z)
 APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 APPLICATION_VERSION_REDHAT = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
+PACKAGE_TYPE_PROPERTY = packageType
```
lurch: I'm in two minds whether it's worth having a separate variable for this - it looks a bit weird having both:
```
--extraMetadata.$(PACKAGE_TYPE_PROPERTY)=dmg
```
and I wonder if it'd look better / more consistent as:
```
--extraMetadata.packageType=dmg
```
instead :question:

lib/gui/app.js
```diff
@@ -79,17 +79,24 @@ app.run(() => {
     '\\____/ \\__\\___|_| |_|\\___|_|',
     '',
     'Interested in joining the Etcher team?',
-    'Drop us a line at join+etcher@resin.io'
+    'Drop us a line at join+etcher@resin.io',
+    '',
+    `Version = ${packageJSON.version}, Type = ${packageJSON.packageType}`
```
lurch: Is it worth including the detected OS here too? \*shrug\*
(at the moment each unique `packageType` only maps to a single OS, but I guess that's not guaranteed)

lib/gui/app.js
```diff
@@ -79,17 +79,24 @@ app.run(() => {
...
   ].join('\n'));
 });
 
 app.run(() => {
-  analytics.logEvent('Application start');
+  const currentVersion = packageJSON.version;
+
+  analytics.logEvent('Application start', {
```
lurch: Does the OS somehow get reported "automatically" to our analytics services?

lurch:
Few small questions, but on the whole this looks great :+1:

lib/gui/app.js
lurch: But then again if we include the OS, do we also include the CPU architecture, etc. ? So rather than diving down the rabbit hole perhaps just the `Version` and `Type` are sufficient for now.

lib/gui/app.js
jviotti: Yeah, corvus does it under the hood :)

Makefile

lib/gui/app.js
jviotti: Yeah, lets see how it goes, and if its useful when troubleshooting users, then we can add more info.

PR #1573, opened 2017-07-06
fix(gui): Fix image checks in image selections

jhermsmeier:
I must have screwed something up during a rebase,
this mends the partition table an windows image checks.

Change-Type: patch

### DIFF
--- a/lib/gui/pages/main/controllers/image-selection.js
+++ b/lib/gui/pages/main/controllers/image-selection.js
@@ -83,10 +83,10 @@ module.exports = function(
     Bluebird.try(() => {
       let message = null;
 
-      if (supportedFormats.looksLikeWindowsImage(image)) {
+      if (supportedFormats.looksLikeWindowsImage(image.path)) {
         analytics.logEvent('Possibly Windows image', image);
         message = messages.warning.looksLikeWindowsImage();
-      } else if (supportedFormats.missingPartitionTable(image)) {
+      } else if (!image.hasMBR) {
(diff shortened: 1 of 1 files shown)


lurch:
Could / should we add extra unit-tests to prevent similar mistakes getting through the CI checks in the future?

jviotti:
The only thing I can think of to catch these types of issues, is to write GUI integration tests.

lurch:
Oh? From what I can tell from the changes in this PR, it looks like the error was in the `selectImage` function, which currently doesn't seem to have any [unit-tests](https://github.com/resin-io/etcher/search?utf8=%E2%9C%93&q=selectImage&type=code) ?  Or is this code so heavily tied into the GUI that it's impossible to unit-test standalone (with mocking etc.)?

lurch:
And on a related note, perhaps it'd be worth looking at adding some code-coverage checks to Etcher?

jviotti:
> Or is this code so heavily tied into the GUI that it's impossible to unit-test standalone (with mocking etc.)?

Yeah, exactly. The code is too GUI centric to property unit-test. Spectron integration tests would cover these types of things.

> And on a related note, perhaps it'd be worth looking at adding some code-coverage checks to Etcher?

I'm a big fan of https://istanbul.js.org/, however I'm not sure how to configure it given that we use `electron-mocha` (a very custom wrapper around mocha that runs test cases on Electron itself), and that Electron applications have independent main/renderer processes. 

Let's definitely research what's needed. If its doable, then lets go for it!

PR #1575, opened 2017-07-06
refactor(GUI): remove flash of white electron webContents workaround

jviotti:
The lib/gui/etcher.js file contained a workaround to an Electron issue
that have since then been solved. In summary:

- Replace the `did-finish-load` event of `webContents` with
  `ready-to-show`, which is only emitted when the whole webview has
  finished loading

Change-Type: patch

### DIFF
--- a/lib/gui/etcher.js
+++ b/lib/gui/etcher.js
@@ -51,17 +51,7 @@ electron.app.on('ready', () => {
   });
 
   // Prevent flash of white when starting the application
-  // https://github.com/atom/electron/issues/2172
-  mainWindow.webContents.on('did-finish-load', () => {
-    const WEBVIEW_LOAD_TIMEOUT_MS = 100;
-
-    // The flash of white is still present for a very short
-    // while after the WebView reports it finished loading
-    setTimeout(() => {
-      mainWindow.show();
-    }, WEBVIEW_LOAD_TIMEOUT_MS);
-
-  });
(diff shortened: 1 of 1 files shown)


lurch:
"not enabled" or "now enabled" ? ;)

lib/gui/etcher.js
```diff
@@ -76,11 +66,6 @@ electron.app.on('ready', () => {
   // See http://electron.atom.io/docs/api/global-shortcut/
 
   mainWindow.on('focus', () => {
-    electron.globalShortcut.register('CmdOrCtrl+Alt+I', () => {
-      mainWindow.webContents.openDevTools({
-        mode: 'detach'
```
lurch: Does it now default to `detach` mode?

jviotti: Good catch. It doesn't :( I'll revert this change.

lurch:
nit: should the commit-title change from `remove some now-unnecessary electron hacks` to `remove a now-unnecessary electron hack` ? ;)

lib/gui/etcher.js
```diff
@@ -51,17 +51,7 @@ electron.app.on('ready', () => {
   });
 
   // Prevent flash of white when starting the application
```
lurch: Is this comment still relevant?

jviotti: Yep, the method is different (and doesn't involve that hacky timeout we had before), but the intention is the same.

PR #1576, opened 2017-07-06
refactor(GUI): move redux store to lib/shared

jviotti:
After https://github.com/resin-io/etcher/pull/1569, the redux store can
run outside of a browser context. This commit moves it to `lib/shared`,
where we will likely move all the other models as well. As an extra, I
renamed `Store` to `store`, since there was no reason for that variable
to be capitalized.

Change-Type: patch

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -33,6 +33,7 @@ const EXIT_CODES = require('../shared/exit-codes');
 const messages = require('../shared/messages');
 const s3Packages = require('../shared/s3-packages');
 const release = require('../shared/release');
+const store = require('../shared/store');
 const packageJSON = require('../../package.json');
 const flashState = require('./models/flash-state');
 const settings = require('./models/settings');
@@ -42,8 +43,6 @@ const updateNotifier = require('./components/update-notifier');
 const availableDrives = require('./models/available-drives');
 const selectionState = require('./models/selection-state');
 const driveScanner = require('./modules/drive-scanner');
-
-const Store = require('./models/store');
 
 const app = angular.module('Etcher', [
   require('angular-ui-router'),
@@ -153,7 +152,7 @@ app.run((ErrorService) => {
 });
 
 app.run(() => {
-  Store.subscribe(() => {
+  store.subscribe(() => {
     const currentFlashState = flashState.getFlashState();
 
     // There is usually a short time period between the `isFlashing()`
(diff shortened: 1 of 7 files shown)


lurch:
CI error:
```
/etcher/lib/gui/app.js
  46:1  error  More than 1 blank line not allowed  no-multiple-empty-lines
```

Apart from that, LGTM!

jviotti:
Fixed!

lurch:
Hmm, all the CI tests passed in this PR, but then Appveyor failed on the merge to `master` :-/
https://ci.appveyor.com/project/resin-io/etcher/build/1.0.5090
```
/scripts/ci/ensure-staged-shrinkwrap.sh
node version: v6.10.3
npm version: 3.10.10
wrote npm-shrinkwrap.json
> etcher@1.0.0 postshrinkwrap C:\projects\etcher
Done
    npm shrinkwrap --dev
index c42b7d8..4cd828c 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -8698,7 +8698,7 @@
     },
     "pretty-bytes": {
       "version": "1.0.4",
       "resolved": "https://registry.npmjs.org/pretty-bytes/-/pretty-bytes-1.0.4.tgz",
       "dev": true
     },
```

Which is really weird, because the previous commits all [built okay](https://github.com/resin-io/etcher/commits/master), and this PR doesn't touch `package.json` or `npm-shrinkwrap.json`. Bloody NPM! :cry:  (perhaps this is also related to https://github.com/resin-io/etcher/pull/1547#discussion_r126389887 )

Ahhh, or perhaps this is related to caching `node_modules`, where a PR-build will produce one set of `node_modules`, which will then get cached and re-used by the next master-build of Etcher. So perhaps [this change](https://github.com/resin-io/etcher/pull/1547#discussion_r126391223) will indeed fix things for Appveyor, but I dunno how we'd resolve things on Travis, which doesn't seem to have a similar cache-dependency feature.
Unless, as I suggested elsewhere, we're able to remove these pesky `from:` lines from `npm-shrinkwrap.json`... (which I believe would fix a ton of problems, but I dunno if it'd introduce new problems)

lurch:
Hmmmm, @jhermsmeier does https://github.com/npm/npm/pull/17076 mean that **keeping** the `from` field will make our upgrade to NPM 5 easier? (or is that irrelevant, given that we're using locked versions rather than ranges?)

jviotti:
Hm, let me extract those changes to a new PR, so we can get that merged sooner.

PR #1577, opened 2017-07-07
docs: initial draft of the Etcher SDK spec

jviotti:
This document describes the design and interface of the Etcher SDK.

### DIFF
--- a/docs/sdk-spec.markdown
+++ b/docs/sdk-spec.markdown
@@ -1,0 +1,263 @@
+Etcher SDK
+==========
+
+This document describes the design and interface of the Etcher SDK.
+
+Preface
+-------
+
+Etcher does a traditional type of image flashing, which is to simply copy the
+image byte-by-byte (with some exceptions, like omitting certain blocks when
+using block maps) to a destination device (like an SD Card).
+
+We discovered deviations of this flashing "mode," like to flash to a board's
+internal memory, and noticed a pattern: they all need two "actions":
+
+- Scanning for a set of destinations that are meaningful to the flashing mode
+- Flashing an image to one or more destinations found in the scanning phase
+
+In the future, there may be extra actions such as "configure" and "manage",
+which modes may decide to give or not.
+
+This specification describes a design for an Etcher SDK that supports different
+types of flashing modes.
+
+API
+---
+
+The Etcher SDK will expose this API:
+
+### `String[] etcher.SUPPORTED_MODES`
+
+An array of the modes this particular version of the Etcher SDK supports. For
+example:
+
+- The "normal" flashing mode we have now
+- dfu
+- CHIP
+- FEL/fastboot
+
+### `EventEmitter etcher.image.fromURL(String url)`
+
+This function creates an *image* object out of an image URL. The implementation
+will at least support HTTP, HTTPS, and `file://`.
+
+- `String url`: The url to the image file
+
+The returned `EventEmitter` emits these events:
+
+- `done`: The operation was successful, and the handler argument is the image
+  object
+- `progress` (optional): The progress details, useful when constructing an
+  image object for a certain image will take more time
+- `error`: An error occurred
+
+### `EventEmitter etcher.image.fromFile(String path)`
+
+This function is like calling `etcher.image.fromURL()` with a `file://`
+protocol.
+
+### `String[] etcher.getSupportedModesForImage(Image image)`
+
+This function returns a list of supported modes for a particular image object.
+
+### `Promise<Destination[]> etcher.scan(Mode mode, Object options)`
+
+This function scans for destinations, passing the `options` object to the
+scanner's `.getAll()` function as-is.
+
+If the scanner singleton is not running, this function will yield back an empty
+set of destinations.
+
+### `EventEmitter etcher.flash(Mode mode, Image image, Destination[] destinations, Object options)`
+
+This function flashes an image to a set of destinations using a certain mode.
+
+The options object may contain these properties, although there is no guarantee
+that every mode will support them all.
+
+- `Boolean check`: Check the result of a flashing process, in whatever way
+  makes sense on that mode
+- `Object extra`: Extra options specific to the chosen mode
+
+The returned `EventEmitter` emits these events:
+
+- `done`: The operation was successful, and the handler argument is a mode
+  specific results object
+- `progress`: Flashing progress details
+- `error`: An error occurred
+
+The flash implementation may emit custom events, as long as they include an
+'extra:' prefix.
+
+Interfaces
+----------
+
+The SDK deals with various "entities." This is a formal definition of what
+these entities look like:
+
+### Progress
+
+An in-progress action. A progress object may include these properties:
+
+- `String id`: A unique identifier for what the progress is about
+- `Number percentage`: The progress percentage
+- `Number transferred`: The number of transferred parts (whatever that means in
+  the progress context)
+- `Number length`: The total number of transferred parts
+- `Number remaining`: The number of parts that are pending
+- `Number eta`: The number of estimated remaining seconds
+- `Number runtime`: The number of seconds the progress was active
+- `Number speed`: The average number of parts processed in each tick
+
+### Image
+
+An operating system image file, like `2016-11-25-raspbian-jessie.img`, or
+`flintos_rpi_v0.3.img.xz`. An image object should contain these properties:
+
+#### `ReadStream stream`
+
+A readable stream that consumes the image in its original form. For a
+compressed image, this stream emits the compressed data.
+
+#### `String type.original`
+
+THe MIME type of the original image size. For a compressed image, this value
+should be the compression/archive MIME type.
+
+#### `String type.final`
+
+The MIME type of the final uncompressed image file. For example,
+`application/octet-stream`, or `application/x-apple-diskimage`.
+
+#### `String filename`
+
+The name of the original filename of the image. Like `foo.img`, or `bar.zip`.
+For a compressed image, the value should be the file name of the compressed
+image.
+
+#### `TransformStream stream`
+
+The `stream` uses this transform stream to get the final image data. For a
+compressed image, this transform stream should decompress the image. If the
+image doesn't need any transformation, then the transform stream should be a
+`PassThrough` stream.
+
+#### `Number size.final.value`
+
+The final size of the image. For a compressed image, this is the size of the
+uncompressed data. If the image is already uncompressed, then this property
+should equal `size.original`.
+
+#### `Boolean size.final.estimation`
+
+This value conveys whether `size.final.value` is an estimated value. There are
+some compression formats, like Bzip2, for which its hard to find the final
+uncompressed size without uncompressing the image. If this property is `true`,
+then `size.final.value` can be a rough estimation.
+
+#### `Number size.original.value`
+
+The original size of the image. For a compressed image, this is the size of the
+compressed data.
+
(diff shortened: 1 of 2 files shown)


docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+`.getAll()` function as-is.
+
+If the scanner singleton is not running, this function will yield back an empty
+set of targets.
+
+### `EventEmitter etcher.flash(Mode mode, Image image, Target[] targets, Object options)`
+
+This function flashes an image to a set of targets using a certain mode.
+
+The options object may contain these properties, although its there is no
```
Shou: `its there` -> `there`?

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+- `error`: An error occurred
+
+The flash implementation may emit custom events, as long as they include an
+'extra:' prefix.
+
+Interfaces
+----------
+
+The SDK deals with various "entities." This is a formal definition of how these
+entities look like:
```
Shou: "How" in the phrase "how these entities look like" feels off, I think it'd be better to use "what" there, "what these entities look like".

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+
+#### `String filename`
+
+The name of the original filename of the image. Like `foo.img`, or `bar.zip`.
+For a compressed image, the value should be the file name of the compressed
+image.
+
+#### `TransformStream stream`
+
+The `stream` uses the this transform stream to get the final image data. For a
```
Shou: `the this` -> `this`?

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+#### `Number size.final.value`
+
+The final size of the image. For a compressed image, this is the size of the
+uncompressed data. If the image is already uncompressed, then this property
+should equal `size.original`.
+
+#### `Boolean size.final.estimation`
+
+This value conveys whether `size.final.value` is an estimated value. There are
+some compression formats, like Bzip2, for which is hard to find the final
```
Shou: `for which is` -> `for which it is`?

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+- `Object options.extra`: Scanner-specific extra options
+
+The returned `EventEmitter` should emit these events:
+
+- `start`: A scan phase has just started. The handler argument is the number of
+  times the scanner will start
+- `stop`: The scanner will stop scanning
+- `done`: The scan phase ended. The handler argument is an array of *targets*
+- `found`: Emitted potentially various times during a single scan phase,
+  passing a single *target* at a time. This event is useful for scanners than
```
Shou: `than` -> `that`?

docs/sdk-spec.markdown
jviotti: Fixed!

docs/sdk-spec.markdown
jviotti: Fixed!

docs/sdk-spec.markdown
jviotti: Fixed :P

docs/sdk-spec.markdown
jviotti: Fixed!

docs/sdk-spec.markdown
jviotti: Fixed!

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+- `done`: The stop action finished successfully
+- `error`: An error occurred, and the scanner might be still running
+
+The scanner implementation may emit custom events, as long as they include an
+`extra:` prefix.
+
+#### `Boolean .isScanning()`
+
+This function checks if a scanner is active. This function should return `true`
+even if the scanner is on a waiting phase.
```
Shou: `on` -> `in`

jviotti: Done!

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+
+### `Promise<Target[]> etcher.scan(Mode mode, Object options)`
+
+This function scans for targets, passing the `options` object to the scanner's
+`.getAll()` function as-is.
+
+If the scanner singleton is not running, this function will yield back an empty
+set of targets.
+
+### `EventEmitter etcher.flash(Mode mode, Image image, Target[] targets, Object options)`
```
jviotti: lol, sounds good!

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,253 @@
...
+The name of the original filename of the image. Like `foo.img`, or `bar.zip`.
+For a compressed image, the value should be the file name of the compressed
+image.
+
+#### `TransformStream stream`
+
+The `stream` uses this transform stream to get the final image data. For a
+compressed image, this transform stream should decompress the image. If the
+image doesn't need any transformation, then the transform stream should be a
+`PassThrough` stream.
```
jviotti: The issues are:

- We might want to re-use the same "transformation" on multiple sources. If each source is responsible for its own transformation, then we will have duplication, which is not the case if the transformations are central

We can of course mitigate this if the sources implementations are inside the SDK source tree (e.g. we can re-use in a source code level), however for many that's not the case (particularly third party implementations on which we don't have control over).

Take the UDF module, for example. Lets pretend some people want to compress the DMGs (I know, non-sense, but lets pretend that's the case), so we need to flash things like `.dmg.xz`. If each source is responsible for all its transformations, then the UDF module would need to add support for all the decompression formats, instead of re-using a single transformation phase that all other sources have access to. 

- In some cases (bzip2, gz, etc), we don't know the size of the final "data," which could cause sources implementations to provide readable streams with no size information. By centralizing transformations, we can exploit the fact that we always know the size of the origin, and apply any transformation on top

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,256 @@
...
+fixed and system drives.
+
+#### `EventEmitter .get(String id)`
+
+This function retrieves a single destination by its id.
+
+- If the scanner is active, and the id exists, then it will return the
+  destination
+- If the scanner is active, and the id exists, then it will return `null`
+- If the scanner is not running, then it will return `null`
```
josephroberts: Does this mean you expect the scanner to be running continuously?

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,256 @@
...
+option could be a `safe` boolean that would cause `.getAll()` to filter out
+fixed and system drives.
+
+#### `EventEmitter .get(String id)`
+
+This function retrieves a single destination by its id.
+
+- If the scanner is active, and the id exists, then it will return the
+  destination
+- If the scanner is active, and the id exists, then it will return `null`
```
josephroberts: typo: id does not exist

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,256 @@
...
+  the progress context)
+- `Number length`: The total number of transferred parts
+- `Number remaining`: The number of parts that are pending
+- `Number eta`: The number of estimated remaining seconds
+- `Number runtime`: The number of seconds the progress was active
+- `Number speed`: The average number of parts processed in each tick
+
+### Image
+
+An operating system image file, like `2016-11-25-raspbian-jessie.img`, or
```
josephroberts: this image could be a zip, elf or bin file. how would we encode the extra properties into these img types?

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,256 @@
...
+types of flashing modes.
+
+API
+---
+
+The Etcher SDK will expose this API:
+
+### `String[] etcher.SUPPORTED_MODES`
+
+An array of the modes this particular version of the Etcher SDK supports.
```
josephroberts: Can you give some example modes?

docs/sdk-spec.markdown
```diff
@@ -0,0 +1,256 @@
...
+
+This function scans for destinations, passing the `options` object to the
+scanner's `.getAll()` function as-is.
+
+If the scanner singleton is not running, this function will yield back an empty
+set of destinations.
+
+### `EventEmitter etcher.flash(Mode mode, Image image, Destination[] destinations, Object options)`
+
+This function flashes an image to a set of destinations using a certain mode.
```
josephroberts: is mode how you differentiate between different flasher mechanisms? in terms of micros pretty each device will have a different flashing mechanism

docs/sdk-spec.markdown
jviotti: Yeah, in the context of Etcher, all scanners will be continuously running, and the client, which is the current dialog to select the destination, will query it as needed.

We could add a way to allow the scanner to run once and stop if that's needed.

docs/sdk-spec.markdown
jviotti: Fixed!

docs/sdk-spec.markdown
jviotti: Not sure what you mean. This is an internal data structure created by the SDK when it reads the image, so nothing needs to be embedded into the image types themselves.

docs/sdk-spec.markdown
josephroberts: we will need this for certain types of radios, for example, with Bluetooth you can only do one operation which means you need to scan first and then update

docs/sdk-spec.markdown
jviotti: Exactly. Some modes we can have are `dfu`, the "normal" flashing mode we have now, the CHIP custom flashing mode, etc

docs/sdk-spec.markdown
jviotti: Done!

docs/sdk-spec.markdown
jviotti: Could the bluetooth scanner scan for devices, save the information, and then scan again after X amount of time?

docs/sdk-spec.markdown
josephroberts: so we could have `ble_microbit`, `ble_nrf51822`, `ble_whatever` etc?

docs/sdk-spec.markdown
josephroberts: ok, so the OS image file is a property of the image object?

docs/sdk-spec.markdown
jviotti: Yep, we can have as many modes as needed to support special requirements. What does `ble` stand for, though? Could we unify all these into a single `ble` mode, or are they very different?

docs/sdk-spec.markdown
jviotti: Yeah, exactly.

docs/sdk-spec.markdown
josephroberts: `ble` stands for `bluetooth`  - they all use the bluetooth radio, but they all have very different update protocols. for micros `mode` would basically mean `device type`

docs/sdk-spec.markdown
josephroberts: you wouldn't want to do that as scanning might start during another operation e.g. an update. the workflow would be something like:
* Start scanning and wait for it to finish
* Loop through each device returned and update them one by one
* Start scanning again

jviotti:
This PR is stale. Closing...

PR #1582, opened 2017-07-10
chore: cache as much as possible from electron-builder

jviotti:
These directories contain files downloaded by electron-builder (like
electron itself), that we can cache in Travis CI and Appveyor CI to
speed up builds.

Change-Type: patch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,16 @@ sudo: false
 node_js:
   - "6.10.3"
 
+# Remove wine from cache
+before_cache:
+  - rm -rf $HOME/.cache/electron-builder/wine
+
 cache:
   ccache: true
+  directories:
+    - $HOME/.cache/electron
+    - $HOME/.cache/electron-builder
+    - $HOME/.npm/_prebuilds
 
 services:
   - docker
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,6 +5,8 @@ image: Visual Studio 2015
 
 cache:
   - C:\Users\appveyor\.node-gyp
+  - '%LOCALAPPDATA%\electron\Cache'
+  - '%LOCALAPPDATA%\electron-builder\cache'
   - '%AppData%\npm-cache'
   - node_modules
   - C:\ProgramData\chocolatey\bin -> appveyor.yml


lurch:
Are you sure that this won't break (like in #1580 ) if one PR updates the electron-builder version, while other PRs are still using the 'older' electron-builder? ;-)

"as most as possible" should be "as much as possible"

jviotti:
It should work fine. The types of things electron-builder caches are independent from the actual electron-builder version. The electron-builder demo project contains these exact lines, and they don't use a shrinkwrap file, so it should work just fine.

develar:
Must be not merge in this state — wine must be removed from cache. Please see recommended config — https://github.com/develar/onshape-desktop-shell/blob/master/.travis.yml

develar:
Cache items are versioned and concurrent access / downloading supported. It is safe to use the same cache across electron-builder versions

lurch:
Arghh, seems like the Travis OSX builds are timing out every single flipping time! :-(

develar:
@lurch Rule is simple – consider to never use brew on travis. It is way to hell. Feel free to ask me as electron-builder maintainer to prepare some prebuild (as we do for wine and so on) for travis os agent for you. Or see https://github.com/develar/build-travis-mac-env/blob/master/.travis.yml how to prebuild deps.

jviotti:
Thanks a lot for reaching out @develar ! 

> @lurch Rule is simple – consider to never use brew on travis. It is way to hell. Feel free to ask me as electron-builder maintainer to prepare some prebuild (as we do for wine and so on) for travis os agent for you. Or see https://github.com/develar/build-travis-mac-env/blob/master/.travis.yml how to prebuild deps.

This is a very interesting approach. I think the only dependency related to electron-builder we use at the moment is `nsis` from Chocolatey. Can this be embedded in electro-builder itself? In the case of brew, we're using `ccache` and `jq`, but I guess it wouldn't make much sense to embed those.

By the way, isn't there a way to cache homebrew's Cellar or something like that instead of separately building and uploading the binaries?

jviotti:
BTW, I added the recommended `before_cache` phase. Thanks a lot @develar !

develar:
> only dependency related to electron-builder we use at the moment is nsis from Chocolatey

NSIS is prebundled for all platforms. No need to install it.

> In the case of brew, we're using ccache and jq, but I guess it wouldn't make much sense to embed those

>  isn't there a way to cache homebrew's Cellar or something like that instead of separately building and uploading the binaries

No way to cache. No easy way to create portable brew. And since last year every innocent call to brew leads "ouch..." because now brew automatically call `brew update` and... and brew state equals to time when OS X image was created by Travis. So, even, even if you explicitly pass `--force-bottle`, `brew update` will occur and take time. I tired to fight and decided to get rid of brew and simply prepare prebuilds for.

develar:
@jviotti If don't want to prepare prebuilds, please set env `HOMEBREW_NO_AUTO_UPDATE=1` and pass `--force-bottle` when do you use brew. I hope it will help you.

lurch:
`appveyor/pr` is passing, but `appveyor/branch` is failing with:
```
node version: v6.10.3
npm version: 3.10.10
wrote npm-shrinkwrap.json
> etcher@1.0.0 postshrinkwrap C:\projects\etcher
Done
    npm shrinkwrap --dev
index c42b7d8..4cd828c 100644
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -8698,7 +8698,7 @@
     },
     "pretty-bytes": {
       "version": "1.0.4",
       "resolved": "https://registry.npmjs.org/pretty-bytes/-/pretty-bytes-1.0.4.tgz",
       "dev": true
     },
```
WTF? :cry: :cry: 
(or will this 'simply' be fixed by manually deleting the Appveyor caches?)

jviotti:
> NSIS is prebundled for all platforms. No need to install it.

Oh, I didn't realize that. Thanks for the tip, I'll get rid of it :D

> No way to cache. No easy way to create portable brew. And since last year every innocent call to brew leads "ouch..." because now brew automatically call brew update and... and brew state equals to time when OS X image was created by Travis. So, even, even if you explicitly pass --force-bottle, brew update will occur and take time. I tired to fight and decided to get rid of brew and simply prepare prebuilds for.
> @jviotti If don't want to prepare prebuilds, please set env HOMEBREW_NO_AUTO_UPDATE=1 and pass --force-bottle when do you use brew. I hope it will help you.

Hm, that sucks. Lets give `HOMEBREW_NO_AUTO_UPDATE` and `--force-bottle` a go, and see how far that takes us, before going for the more complex approach of manually caching binaries.

BTW, is there a way to manually download brew bottles and their dependencies, and manually installing them? Could that be an easier approach?

develar:
> is there a way to manually download brew bottles and their dependencies, and manually installing them? Could that be an easier approach

Yes, this approach was used in the electron-budder.

```
- curl -L https://dl.bintray.com/develar/bin/wine.7z -o /tmp/wine.7z
- /tmp/7za x -o/usr/local/Cellar -y /tmp/wine.7z
- brew link --overwrite fontconfig freetype
 ```

replace `https://dl.bintray.com/develar/bin/wine.7z` to brew bottle url (if bottle exists ;)).

jviotti:
I sent 2 PRs based on @develar 's feedback:

- https://github.com/resin-io/etcher/pull/1584
- https://github.com/resin-io/etcher/pull/1583

jviotti:
I see. That's actually a cool approach. We're moving to a self-hosted CI setup for Etcher, and given we will have full control over the machines, we can install the dependencies in advance, so we'll see how that goes.

PR #1583, opened 2017-07-11
chore: avoid brew taking a long time on Travis CI

jviotti:
- Homebrew now auto-updates before performing any installation, which
  can take quite some time. We can disable this by enabling
  HOMEBREW_NO_AUTO_UPDATE

- We can pass the --force-bottle option to `brew install` to force it to
  download pre-compiled binaries, and don't attempt any compilations

See: https://github.com/resin-io/etcher/pull/1582
Change-Type: patch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ env:
     - CCACHE_COMPRESS=1
     - CC="clang"
     - CXX="clang++"
+    - HOMEBREW_NO_AUTO_UPDATE=1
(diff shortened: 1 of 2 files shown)


lurch:
Is it worth mentioning in the commit message that this only affects the OSX builds?

jviotti:
Makes sense. Given how funny the CI servers have been, I'll edit the commit messages when merging :)

PR #1585, opened 2017-07-11
chore: add eslint-jsdoc-plugin to eslint

jviotti:
This plugin helps us detect some things the built-in jsdoc rules can't,
like whether there is an example or not.

As expected, the addition of this plugin helped detect some minor JSDoc
issues.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -6,7 +6,13 @@ env:
   mocha: true
 plugins:
   - lodash
+  - jsdoc
 extends: 'eslint:recommended'
+settings:
+  jsdoc:
+    additionalTagNames:
+      customTags:
+        - fulfil
 rules:
 
   # Possible Errors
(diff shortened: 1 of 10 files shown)


.eslintrc.yml
```diff
@@ -645,3 +651,24 @@ rules:
...
+
+  # JSDoc
+
+  jsdoc/check-param-names:
+    - error
+  jsdoc/check-tag-names:
+    - error
+  jsdoc/newline-after-description:
+    - error
+  jsdoc/require-example:
```
jhermsmeier: Do we really want to force an example for *every* function & method? Seems to me that'd be a bit of overkill, and introduce a lot of noise in the code.

lurch: There's a few places in this PR where it adds `/* eslint-disable jsdoc/require-example */` ;-)

jviotti: Yeah, we can enable the exception for things like React components, etc.

Issue #1572, opened 2017-07-06
[Suggestion] macOS Installer Support


KevinAT:
i don't think has been asked before. have you thought about adding the ability to burn macOS Installers with etcher?

jhermsmeier:
Hey @KevinAT, I'm not sure I understand this correctly; are you referring to the setup image that installs the operating system itself?

KevinAT:
@jhermsmeier yes, exactly those - like the windows os installer just the mac one. 

under OSX/macOS they are displayed as "Install OS X %cat_or_mountain_name%.app" and treated like normal executables. it's possible to burn them to dvds/usb drives/etc but from what i know in recent versions apple made it more difficult to burn them, as in requiring the app/image to be modified to be bootable. osx tools like Install Disk Creator and DiskMaker X 6 are able to create install mediums from the OSX Installer app files, but i think they modify them a bit in the background, and i'm not sure if those modifications would be easy to port to other OSs like linux. porting it to windows might not work at all, since ntfs doesn't seem to be able to handle the macOS installer files.

though i don't know much about this and am just speculating from what i heard/read. anyways it would be great if i could create bootable macOS installer drives in the future from my linux distro without having to use macOS myself.

jhermsmeier:
Okay – you can already do that, nothing complicated like preprocessing or modification necessary, as Etcher already supports the Apple Disk Image format.

### Step 1: Download Mac OS Whatever



<p align="center">
<img alt="screen shot 2017-07-06 at 14 59 28" src="https://user-images.githubusercontent.com/244907/27912321-bd45b220-625c-11e7-8315-0216174e7fc2.png">
</p>

### Step 2: Navigate into `.app`'s package contents



<p align="center">
<img alt="screen shot 2017-07-06 at 15 00 43" src="https://user-images.githubusercontent.com/244907/27912386-fb260f5e-625c-11e7-8023-61198c442e8a.png">
</p>

### Step 3: Find the `InstallESD.dmg` disk image

**NOTE:** It might be named differently for different Mac OS versions, but whatever has a `.dmg` extension in there and is ~4-6GB in size is what you're looking for.



<p align="center">
<img alt="screen shot 2017-07-06 at 15 01 39" src="https://user-images.githubusercontent.com/244907/27912445-420ed96e-625d-11e7-828f-637e79957bbe.png">
</p>

### Step 4: Flash it with Etcher

[...]

PR #1586, opened 2017-07-11
upgrade(package): Update udif 0.9.0 -> 0.10.0

jhermsmeier:
This updates `udif` from 0.9.0 to 0.10.0, fixing two bugs;

* [57ddf788] fix(readstream): Avoid read-before-open situations
* [6333a183] fix(image): try/catch parsing the resource fork plist

[57ddf788]: https://github.com/jhermsmeier/node-udif/commit/57ddf788b72cc1ff17317e560532e5e89adf695c
[6333a183]: https://github.com/jhermsmeier/node-udif/commit/6333a183f6863348723f95b3c8183fea16a30703

Change-Type: patch
Connects To: #1572

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -10117,9 +10117,9 @@
       "resolved": "https://registry.npmjs.org/ua-parser-js/-/ua-parser-js-0.7.12.tgz"
     },
     "udif": {
-      "version": "0.9.0",
-      "from": "udif@0.9.0",
-      "resolved": "https://registry.npmjs.org/udif/-/udif-0.9.0.tgz",
+      "version": "0.10.0",
+      "from": "udif@0.10.0",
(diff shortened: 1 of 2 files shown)


lurch:
@jhermsmeier Apologies in advance for sounding like a broken record, but is there any way that the bugs these fix could be added to the unit-tests? (either on `Etcher` and/or `udif`, whichever is most appropriate)

jhermsmeier:
@lurch haha, no worries – the funny thing is that there *are* tests in both, Etcher and udif, that make use of the readStream (all those dmg image tests, i.e.) in the same way – but none of them experience this particular case.

jhermsmeier:
I'll see if I can somehow reproduce this race-condition reliably outside of Etcher, and then add that to `udif`, but my hopes of managing that are quite low (due to how the fs read requests work).

PR #1587, opened 2017-07-11
fix(writer): Use final size if it's not an estimation

jhermsmeier:
This avoids running into the "flashstate percentage above 100%" error again.

Change-Type: patch
See also: #1572

### DIFF
--- a/lib/image-stream/handlers.js
+++ b/lib/image-stream/handlers.js
@@ -157,7 +157,9 @@ module.exports = {
         extension: fileExtensions.getLastFileExtension(imagePath),
         stream: udif.createReadStream(imagePath),
         size: {
-          original: options.size,
+          // FIXME(jhermsmeier): Originally `options.size`,
+          // See discussion in https://github.com/resin-io/etcher/pull/1587
+          original: size || options.size,
           final: {
             estimation: false,
             value: size
--- a/tests/image-stream/dmg.spec.js
+++ b/tests/image-stream/dmg.spec.js
@@ -81,14 +81,13 @@ describe('ImageStream: DMG', function() {
       it('should return the correct metadata', function() {
         const image = path.join(DMG_PATH, 'etcher-test-zlib.dmg');
         const uncompressedSize = fs.statSync(path.join(IMAGES_PATH, 'etcher-test.img')).size;
-        const compressedSize = fs.statSync(image).size;
 
         return imageStream.getImageMetadata(image).then((metadata) => {
           m.chai.expect(metadata).to.deep.equal({
             path: image,
             extension: 'dmg',
             size: {
-              original: compressedSize,
+              original: uncompressedSize,
               final: {
                 estimation: false,
                 value: uncompressedSize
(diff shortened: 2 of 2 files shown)


jhermsmeier:
Not sure how this particular error popped up again – I thought we'd dealt with this before?

jviotti:
Are you sure this isn't breaking other image formats? As a recap: `etcher-image-write` potentially takes an image stream that represents the *compressed* version of an image. The module also expects the size you pass to it to be the size of the *compressed* stream, since that's what its going to consume.

The writer then pipes the compressed stream through a progress transform stream, and **then** applies the image transform stream that takes care of the decompression. This design decision is to workaround the fact that for some image types, we have no way to determine the final uncompressed size with certainty. Despite this, we always know the compressed size, so it means we can always show an accurate progress bar.

So again, I'm pretty sure that the value we need to pass to etcher-image-write is the size of the compressed package, and then pass a transform stream that is able to decompress it.

As far as I can see on the code, the UDIF image stream handler passes a stream that extracts the file directly, which kinda breaks our design (I didn't notice this before). We have two solutions:

 - Make the UDIF image stream handler return a stream of the compressed package, set the original and final sizes accordingly, and also pass a transform stream that is able to decompress it

- Keep everything as it is, but set both the original and final value to the final value, since decompression is already being done on the UDIF module (of course this only applies if we know that the uncompressed size we got is absolutely correct, which I think its the case)

jviotti:
To be honest, I'm surprised our test cases didn't catch this. Can we fix it somehow? I believe the error would have been detected as incorrect progress percentages while flashing the DMGs. If so, can we extend the DMGs test to ensure we get sane progress percentages?

lurch:
I wonder if the test DMGs are so small, that the difference between the compressed and uncompressed file sizes is too small to register as > 100% ? \*shrug\*

@jviotti Rather than *always* calculating the progress based on the size of the compressed image, and the decompression-stream-progress; if we **know** for definite what the final uncompressed image size is, could we have an alternative progress-calculation scheme based on the size of the uncompressed image, and the bytes-actually-written-to-disk stream? This would prevent inaccurate progress readings when the decompression speed is much faster than the disk-write speed, and it would also prevent the "write speed" slowing down massively towards the end for those images that have a lot of low-compressible data at the start, followed by a huge amount of empty space (high-compressible data) at the end, which I believe is what's occurring in #1171 and #638 , regardless of the actual compression method being used.

That's kinda hard to explain in words, so maybe this picture helps?
decompressionprogress

jviotti:
@lurch Yeah, I'm happy to make that change, which should be now easier to do with pipage.

jhermsmeier:
@jviotti @lurch I have made this a DMG specific workaround for now, so that we can get rid of the errors, and test with it – and I'd do the proper fix as a follow up to this; what do you think?

jviotti:
@jhermsmeier Let's do the following instead, which is localized in the DMG handler:

```diff
diff --git a/lib/image-stream/handlers.js b/lib/image-stream/handlers.js
index 62069f89..095a24ee 100644
--- a/lib/image-stream/handlers.js
+++ b/lib/image-stream/handlers.js
@@ -157,7 +157,7 @@ module.exports = {
         extension: fileExtensions.getLastFileExtension(imagePath),
         stream: udif.createReadStream(imagePath),
         size: {
-          original: options.size,
+          original: size,
           final: {
             estimation: false,
             value: size
```

So basically since the stream `node-udif` gives us is already the one with uncompressed data, we can tell Etcher that the source size is the uncompressed size.

jhermsmeier:
@jviotti alright, done.

PR #1588, opened 2017-07-11
chore: only ignore top level directories in gitignore for some cases

jviotti:
We accidentally added build/ to gitignore, causing any new files from
`scripts/build` to be ignored.

See: https://github.com/resin-io/etcher/pull/1547#discussion_r126790010


scripts/build/target-convert.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+###
+# Copyright 2016 resin.io
```
lurch: Should be 2017 ;)

lurch:
Doh! Looks like this bug has been there since 2nd May https://github.com/resin-io/etcher/commit/b75dfd3ecedf5f1218d00c8988085e562f2bce19 :(

Are there any other file-paths in `.gitignore` that _ought_ to use an 'absolute' path?

scripts/build/target-convert.sh
jviotti: Done!

jviotti:
Ugh, that sucks. I'll revise `gitignore` and update this PR accordingly.

.gitignore
```diff
@@ -21,15 +21,15 @@ coverage
 .lock-wscript
 
 # Compiled binary addons (http://nodejs.org/api/addons.html)
-build
+/build
 
 # Dependency directory
 # https://docs.npmjs.com/misc/faq#should-i-check-my-node-modules-folder-into-git
 node_modules
 bower_components
```
lurch: Do we still need this line, if we're not using bower any more?

.gitignore
```diff
@@ -21,15 +21,15 @@ coverage
 .lock-wscript
 
 # Compiled binary addons (http://nodejs.org/api/addons.html)
-build
+/build
 
 # Dependency directory
 # https://docs.npmjs.com/misc/faq#should-i-check-my-node-modules-folder-into-git
 node_modules
```
lurch: Should this have a leading slash too?

.gitignore
```diff
@@ -21,15 +21,15 @@ coverage
...
 bower_components
 
 # Compiled Electron releases
```
lurch: Electron -> Etcher ?

.gitignore
jviotti: I thought about that, but I don't think a directory name such as `node_modules` could ever be the source of confusion, plus this rule would cover us from accidental nested `node_modules` in other directories.

.gitignore
jviotti: Good catch. I'll get rid of it!

.gitignore
jviotti: Fixed!

PR #1590, opened 2017-07-12
refactor(image-stream): parse xz and gzip metadata using a custom read function

jviotti:
This commit refactors the xz and gzip image handlers to pass/use a
custom read function to be able to determine the uncompressed size, and
other needed metadata.

By using this function (which currently only uses the `fs` module), we
can implement support for getting the uncompressed size of compressed
files using HTTP Ranges.

Change-Type: patch

### DIFF
--- a/lib/image-stream/gzip.js
+++ b/lib/image-stream/gzip.js
@@ -16,9 +16,6 @@
 
 'use strict';
 
-const Bluebird = require('bluebird');
-const fs = Bluebird.promisifyAll(require('fs'));
-
 /**
  * @summary The byte length of ISIZE
  * @type {Number}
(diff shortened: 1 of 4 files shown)


lib/image-stream/gzip.js
```diff
@@ -38,37 +35,34 @@ const ISIZE_LENGTH = 4;
...
-        }
+exports.getUncompressedSize = (options) => {
+  const ISIZE_BUFFER_START = 0;
 
-        return isizeBuffer.readUInt32LE(ISIZE_BUFFER_START);
-      });
-    });
+  // Determine the uncompressed size of the gzip file by reading
+  // its `ISIZE`. The specification clarifies that this value is
+  // just an estimation.
```
lurch: This comment seems unnecessary, as it just duplicates the function's `@description` ?

lib/image-stream/gzip.js
```diff
@@ -38,37 +35,34 @@ const ISIZE_LENGTH = 4;
  * by reading its `ISIZE`. The specification clarifies that this
  * value is just an estimation.
```
lurch: I wonder if it's worth adding a comment here that this involves reading from the very end of the file? If we're streaming, we wouldn't want to download the whole file twice ;-)

lib/image-stream/gzip.js
```diff
@@ -38,37 +35,34 @@ const ISIZE_LENGTH = 4;
...
+exports.getUncompressedSize = (options) => {
+  const ISIZE_BUFFER_START = 0;
 
-        return isizeBuffer.readUInt32LE(ISIZE_BUFFER_START);
-      });
-    });
+  // Determine the uncompressed size of the gzip file by reading
+  // its `ISIZE`. The specification clarifies that this value is
+  // just an estimation.
+  return options.read(ISIZE_LENGTH, options.size - ISIZE_LENGTH).then((buffer) => {
```
lurch: Only a minor nit, but I wonder if adding a `const ISIZE_OFFSET = options.size - ISIZE_LENGTH;` line here would help to clarify this code? (or perhaps such a small change isn't worth it)

lib/image-stream/handlers.js
```diff
@@ -115,9 +125,14 @@ module.exports = {
    * @returns {Promise}
    */
   'application/x-xz': (imagePath, options) => {
-    return fs.openAsync(imagePath, 'r').then((fileDescriptor) => {
-      return lzma.parseFileIndexFDAsync(fileDescriptor).tap(() => {
-        return fs.closeAsync(fileDescriptor);
+    return Bluebird.using(fs.openAsync(imagePath, 'r').disposer((fileDescriptor) => {
+      return fs.closeAsync(fileDescriptor);
+    }), (fileDescriptor) => {
```
lurch: Nice to see such uniformity between the gzip and xz code :)

lib/image-stream/gzip.js
jviotti: Indeed, I accidentally duplicated it during one of my refactorings.

lib/image-stream/gzip.js
jviotti: Sounds good, done!

lib/image-stream/gzip.js
jviotti: Done! In any case, passing a `read()` function that uses HTTP Ranges, for example, allows you to avoid downloading the whole file

cedricve:
overall looks good, no remarks! good job.

jviotti:
@lurch 

> Could / should any of these changes be unit-tested?

All these changes are covered by the integration tests that use real images.

lurch:
Given that we actually have a bunch of different-format test images in `tests/image-stream/data/` (thanks to @jhermsmeier ) I guess I was suggesting that we could actually test that we get the 'expected' uncompressed sizes (whether that's the real size or just an estimate) for some of these test files?
But perhaps that belongs in a separate PR, so I'll approve **this** PR after the release of Etcher v1.1.1 :)

lib/image-stream/gzip.js
```diff
@@ -29,46 +26,40 @@ const fs = Bluebird.promisifyAll(require('fs'));
...
  * @example
- * gzip.getUncompressedSize('path/to/file.gz').then((uncompressedSize) => {
+ * const fd = fs.openSync('path/to/image', 'r');
+ *
+ * gzip.getUncompressedSize({
+ *   size: fs.statSync('path/to/image.gz').size,
+ *   read: (count, offset) => {
+ *     const buffer = Buffer.alloc(count);
+ *     fs.readSync(fd, buffer, 0, count, offset);
+ *     return buffer;
```
jhermsmeier: Looks like this should be returning a `Promise`, given `options.read().then()` on line 62

lib/image-stream/utils.js
```diff
@@ -17,6 +17,41 @@
...
+ * @returns {Promise}
+ *
+ * @example
+ * fs.openAsync('path/to/image.img', 'r').then((fileDescriptor) => {
+ *   return utils.readBufferFromImageFileDescriptor(fileDescriptor, 512, 0);
+ * }).then((buffer) => {
+ *   console.log(buffer);
+ * });
+ */
+exports.readBufferFromImageFileDescriptor = (fileDescriptor, count, offset) => {
```
jhermsmeier: Can we swap offset & count here to `(fd, position, length)`, to adhere to the function signatures in the rest of Node's API?

lib/image-stream/gzip.js
jviotti: Indeed, fixed!

lib/image-stream/utils.js
jviotti: Done!

cedricve:
nice @jviotti one step closer!

PR #1591, opened 2017-07-13
chore: cache electron-builder temporary files in macOS

jviotti:
Our current Travis CI configuration caches temporary files created by
electron-builder, but only for GNU/Linux builds. In macOS, electron and
electron-builder store their cache at ~/Library/Caches, so we must cache
that as well.

This commit was motivated by the fact that most macOS builds are failing
for Electron package download timeouts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ cache:
     - $HOME/.cache/electron
     - $HOME/.cache/electron-builder
     - $HOME/.npm/_prebuilds
+    - $HOME/Library/Caches/electron
+    - $HOME/Library/Caches/electron-builder
     - node_modules
 
 services:


lurch:
@develar is this something worth adding to your examples?

lurch:
I don't use OSX myself, but does `wine` need to be removed from the `electron-builder` cache there too?

develar:
It is not proper solution. We officially recommend:

```
env:
  global:
    - ELECTRON_CACHE=$HOME/.cache/electron
    - ELECTRON_BUILDER_CACHE=$HOME/.cache/electron-builder
```

develar:
If you use `electron-download` — do not use it. Please use `electron-download-tf`. Vote for https://github.com/electron-userland/electron-download/pull/60

lurch:
https://www.npmjs.com/package/electron-download-tf and https://www.npmjs.com/package/electron-download-tf look almost identical - and they have the same 'home page' link? ( https://github.com/electron-userland/electron-download )

develar:
@lurch electron-download maintainers doesn't accept PR quickly, so, we have to use fork. You can see repo in https://github.com/electron-userland/electron-download/pull/60 PR.

PR #1594, opened 2017-07-13
refactor(image-stream): extract part of utils.js into mime.js

jviotti:
This is a small refactoring commit that extracts the MIME related
function from utils.js into a new file called mime.js, and stores the
default image MIME type there as a constant, to avoid duplicating it in
multiple parts of the code.

Change-Type: patch


lib/image-stream/mime.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {String} filename - file path
+ * @fulfil {String} - mime type
+ * @returns {Promise}
+ *
+ * @example
+ * mime.getMimeTypeFromFile('path/to/raspberrypi.img.gz').then((mimeType) => {
+ *   console.log(mimeType);
+ * });
+ */
+exports.getMimeTypeFromFile = (filename) => {
```
jhermsmeier: What do you think about naming this `getMimeTypeFromFilename()` instead, to avoid any confusion (i.e. not passing a buffer)?

lib/image-stream/mime.js
```diff
@@ -0,0 +1,65 @@
...
+const fileType = require('file-type');
+const mime = require('mime-types');
+const fs = require('fs');
+
+/**
+ * @summary The default image MIME type
+ * @type {String}
+ * @constant
+ */
+exports.DEFAULT_IMAGE_MIME_TYPE = 'application/octet-stream';
```
jhermsmeier: What do you think about just `DEFAULT_MIME_TYPE`, as that's the default MIME type for everything unknown, and not specific to disk images?

jviotti: Sounds great, done!

lib/image-stream/mime.js
jviotti: Makes sense, done!

jhermsmeier:
@jviotti looks like we might want to add credentials to not have the API rate limit kick in (from the failed Mac OS build):

```
./node_modules/.bin/npx pkg --output dist/Etcher-cli-1.0.0+8d2475e-darwin-x64/etcher -t node6--x64 dist/Etcher-cli-1.0.0+8d2475e-darwin-x64-app/lib/cli/etcher.js
> pkg@4.1.1
> Fetching base Node.js binaries to: ~/.pkg-cache
> Error! read ETIMEDOUT
> Asset 'v2.3/uploaded-v2.3-node-v6.10.3-macos-x64' not found by direct link
> Error! API rate limit exceeded for 208.78.110.199. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Command failed: /Users/travis/build/resin-io/etcher/node_modules/.bin/pkg --output dist/Etcher-cli-1.0.0+8d2475e-darwin-x64/etcher -t node6--x64 dist/Etcher-cli-1.0.0+8d2475e-darwin-x64-app/lib/cli/etcher.js
make: *** [dist/Etcher-cli-1.0.0+8d2475e-darwin-x64] Error 2
```

lurch:
Sounds like a `pkg` bug to me - why is it trying to download the Node.js binaries from a service that has rate-limits? Do we know _which_ service this is that's being rate-limited?

lib/image-stream/mime.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * @summary The default image MIME type
+ * @type {String}
+ * @constant
+ */
+exports.DEFAULT_MIME_TYPE = 'application/octet-stream';
+
+/**
+ * @summary Get archive mime type
```
lurch: Should `archive` here be replaced with `a file's` ?

EDIT: And perhaps the `@summary` should explain that this method _might_ involve actually opening and reading the file itself, i.e. it might not _just_ look at the filename.

tests/image-stream/mime.spec.js
```diff
@@ -0,0 +1,93 @@
...
+'use strict';
+
+const m = require('mochainon');
+const path = require('path');
+const DATA_PATH = path.join(__dirname, 'data');
+const mime = require('../../lib/image-stream/mime');
+
+describe('ImageStream: MIME', function() {
+
+  describe('.getArchiveMimeType()', function() {
```
lurch: Does this line need updating to `.getMimeTypeFromFileName` ?

tests/image-stream/mime.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        m.chai.expect(type).to.equal('application/x-apple-diskimage');
+      });
+    });
+
+    it('should resolve application/octet-stream for an unrecognized file type', function() {
+      const file = path.join(DATA_PATH, 'unrecognized', 'random.rpi-sdcard');
+      return mime.getMimeTypeFromFileName(file).then((type) => {
+        m.chai.expect(type).to.equal('application/octet-stream');
+      });
+    });
```
lurch: Maybe there should also be some tests to exercise the `file-type` functionality, i.e. a file with an unknown extension, but with a header-format that `file-type` is able to identify?

lib/image-stream/mime.js
```diff
@@ -0,0 +1,65 @@
...
+'use strict';
+
+const _ = require('lodash');
+const Bluebird = require('bluebird');
+const fileType = require('file-type');
+const mime = require('mime-types');
+const fs = require('fs');
+
+/**
+ * @summary The default image MIME type
```
jviotti: Done!

lib/image-stream/mime.js
jviotti: Done!

tests/image-stream/mime.spec.js
jviotti: Done!

jviotti:
> The PR comment mentions utils.js, but that file doesn't actually got modified in this PR?? Was this PR supposed to also remove utils.getArchiveMimeType ?

Oops, fixed!

tests/image-stream/mime.spec.js
jviotti: Done, I added a test case for that!

jviotti:
> @jviotti looks like we might want to add credentials to not have the API rate limit kick in (from the failed Mac OS build):

Hm, comes from GitHub. I'll add the pkg cache directory to the CI services cache.

jviotti:
See https://github.com/resin-io/etcher/pull/1598

tests/image-stream/mime.spec.js
```diff
@@ -0,0 +1,100 @@
...
+
+    it('should resolve the actual MIME type, without caring about the extension', function() {
```
lurch: nit: it _does_ care about the extension, because it only checks the first 262 bytes if it doesn't recognise the extension. Perhaps something like `should resolve the actual MIME type for an unrecognized extension` ?
Is there any danger that a future version of the `mime-types` module might recognise `*.md` as `text/markdown` ?

lurch: Haha, looking at the Travis logs it already does! So you'll need to choose a different unrecognized extension ;-)

tests/image-stream/mime.spec.js
```diff
@@ -0,0 +1,100 @@
...
+
+    it('should resolve application/octet-stream for an unrecognized file type', function() {
+      const file = path.join(DATA_PATH, 'unrecognized', 'random.rpi-sdcard');
+      return mime.getMimeTypeFromFileName(file).then((type) => {
+        m.chai.expect(type).to.equal('application/octet-stream');
+      });
+    });
+
+    it('should resolve the correct MIME type given an invalid extension', function() {
+      const file = path.join(DATA_PATH, 'unrecognized', 'xz-with-invalid-extension.foo');
```
lurch: Cool. Just for the sake of completion, could we also have a file named `xz-with-no-extension` ?

jviotti: Done!

.gitattributes
```diff
@@ -39,3 +39,4 @@ Makefile text
 *.zip binary diff=hex
 *.dmg binary diff=hex
 *.rpi-sdcard binary diff=hex
+*.foo binary diff=hex
```
lurch: Looks like you'll need to add
```
xz-without-extension binary diff=hex
```
here too.

jviotti: Fixed!

PR #1595, opened 2017-07-13
test(image-stream): increase DMG tests timeouts

jviotti:
I caught an sporadic issue a couple of times, where the DMG tests would
exceed the default 2000ms mocha timeout. All the other image
decompression tests have a much higher timeout already, so this commit
adds the same timeout to the DMG tests file.

### DIFF
--- a/tests/image-stream/dmg.spec.js
+++ b/tests/image-stream/dmg.spec.js
@@ -27,6 +27,8 @@ const tester = require('./tester');
 
 describe('ImageStream: DMG', function() {
 
+  this.timeout(20000);
+
   describe('compression method', function() {
 
     describe('NONE', function() {


jhermsmeier:
Although 20 seconds is a bit high, I think. Anyways, out of curiosity – where did those timeouts happen? On CI or locally? And can you remember with which one exactly (I suspect bzip2, or adc)?

jviotti:
We used to have lower timeouts, but at some point I increased them to 20s because of Appveyor I think. I'm pretty sure it was bzip2.

lurch:
> All the other image
decompression tests have a much higher timeout already, so this commit
adds the same timeout to the DMG tests file.

Could this somehow be configured as a shared 'disk tests timeout' constant, so that it doesn't need to be duplicated in several different places? (which will be useful if we ever need to increase it again in future, for some reason)

jviotti:
Done, see https://github.com/resin-io/etcher/pull/1599

PR #1596, opened 2017-07-13
refactor: move most models to lib/shared/

jviotti:
Now that the Redux store is no longer front-end dependent, we can move
most of the models to lib/shared. Currently, lib/shared is a mess, and
contains all sorts of functionality, however moving things out of the
GUI is the first step. Once we have everything decoupled, we can
organise all the code we have there.


lurch:
LGTM, but @Shou should probably review

PR #1597, opened 2017-07-13
refactor(GUI): make settings model setter asynchronous

jviotti:
This is part of the process of implementing support for a configuration
file. We previously decoupled the Redux store from localStorage by
moving the logic that actually persists the data to localStorage to a
local-settings.js file, however the localStorage API is synchronous, so
it follows that at the moment, all functions that interact with are also
synchronous.

Moving to storing the settings to a file means turning all these
functions to promises, which we do in this commit, in order to not mix
the addition of the configuration file feature with the huge amount of
refactoring it requires.

See: https://github.com/resin-io/etcher/issues/1356
Change-Type: patch

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -93,65 +93,66 @@ app.run(() => {
     version: currentVersion
   });
 
-  settings.load();
-
-  const shouldCheckForUpdates = updateNotifier.shouldCheckForUpdates({
-    currentVersion,
-    lastSleptUpdateNotifier: settings.get('lastSleptUpdateNotifier'),
-    lastSleptUpdateNotifierVersion: settings.get('lastSleptUpdateNotifierVersion')
-  });
-
-  const currentReleaseType = release.getReleaseType(currentVersion);
-  const updatesEnabled = settings.get('updatesEnabled');
-
-  if (!shouldCheckForUpdates || !updatesEnabled) {
-    analytics.logEvent('Not checking for updates', {
-      shouldCheckForUpdates,
-      updatesEnabled,
-      releaseType: currentReleaseType
+  settings.load().then(() => {
+    const shouldCheckForUpdates = updateNotifier.shouldCheckForUpdates({
+      currentVersion,
+      lastSleptUpdateNotifier: settings.get('lastSleptUpdateNotifier'),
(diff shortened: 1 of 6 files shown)


lurch:
This might be a stupid question :wink: but with the settings file being so small, it'll probably load and save really quickly, so what problem(s) would there be if this code was left as synchronous?

jhermsmeier:
@lurch regardless of file size, doing I/O would block the current tick, while doing this asynchronously allows the I/O to happen in libuv's thread without blocking the JS one.
For example, imagine your storage device is under high load (even SSDs can reach response times in the 100s of milliseconds); then you'd be freezing up the whole thing for that duration if done synchronously.
Not something you'd want to happen.

jviotti:
I guess that could work, but better to be prepared in case the file size increases, I/O is slow, etc. By using promises, the user can actually decide to wait or not, which increases flexibility.

lurch:
Thanks for confirming that it was a stupid question :laughing: 

"doing this asynchronously allows the I/O to happen in libuv's thread without blocking the JS one" is interesting :+1: 

So (apologies for another stupid question) why didn't we have an async wrapper around the synchronous localStorage API before now?

jviotti:
The localStorage API is synchronous, and we never thought of a configuration file before, so I guess no particular reason apart from not anticipating we would switch to something else.

jhermsmeier:
> So (apologies for another stupid question) why didn't we have an async wrapper around the synchronous localStorage API before now?

Because `localStorage` is a [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and access is always synchronous and implementation browser dependent (in this case, with Chromium, it's backed by LevelDB if my memory serves me right), while this whole libuv and doing I/O outside the main JS thread is a Node thing.

lurch:
When I looked into it [previously](https://github.com/resin-io/etcher/pull/1183) it appeared to be using SQLite.

jhermsmeier:
Oh, gone are those days then

lurch:
> if my memory serves me right

...

> gone are those days then

Happens to everyone, as you get older :wink: :rofl:

jviotti:
Is this good to merge?

jviotti:
I'm assuming there are no further comments here, but let me know if that's not the case!

PR #1598, opened 2017-07-14
chore: cache ~/.pkg-cache in Travis CI and Appveyor CI

jviotti:
Looks like pkg downloads node binaries from GitHub Releases, and GitHub
may impose a rate limiting if we download too many things from them in a
short period of time (likely to happen when there are many concurrent
PRs).

In order to mitigate this issue, this commit adds `$HOME/.pkg-cache`,
the directory where pkg stores node binaries, to the CI services cache.

See: https://github.com/resin-io/etcher/pull/1594

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ cache:
     - $HOME/.npm/_prebuilds
     - $HOME/Library/Caches/electron
     - $HOME/Library/Caches/electron-builder
+    - $HOME/.pkg-cache
     - node_modules
 
 services:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ cache:
   - '%LOCALAPPDATA%\electron\Cache'
   - '%LOCALAPPDATA%\electron-builder\cache'
   - '%AppData%\npm-cache'
+  - '%USERPROFILE%\.pkg-cache'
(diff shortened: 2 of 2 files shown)


lurch:
> likely to happen when there are many concurrent PRs

... or when builds on Travis or Appveyor from entirely unrelated projects also happen to be using `pkg` in their CI tests ;-)

Will this rate-limiting ever cause problems for us now that we're switching to GH Releases for Etcher?

jviotti:
> ... or when builds on Travis or Appveyor from entirely unrelated projects also happen to be using pkg in their CI tests ;-)

Oh, right, I didn't think about that.

> Will this rate-limiting ever cause problems for us now that we're switching to GH Releases for Etcher?

Nope, pkg is using the GitHub API, which has rate limiting, but that shouldn't be a problem if you're downloading the assets directly.

lurch:
> that shouldn't be a problem if you're downloading the assets directly.

I guess I was referring more to https://github.com/resin-io/etcher/pull/1555/files#diff-9cf255022386c90c5a7e721284b1901cR63 but I see that you already handle the case of the ratelimit being reached, by pretending there are no versions available. All good then ;)
(BTW I'm still in the middle of reviewing #1555 )

PR #1599, opened 2017-07-14
test(image-stream): move extraction timeout to a shared constant

jviotti:
The image stream tests that extract images all have a 20s timeout, which
is manually set in all the suites related to extraction.

This commit extracts that number as a constant called DEFAULT_TIMEOUT.

See: https://github.com/resin-io/etcher/pull/1595

### DIFF
--- a/tests/image-stream/tester.js
+++ b/tests/image-stream/tester.js
@@ -43,6 +43,8 @@ const deleteIfExists = (file) => {
   });
 };
 
+exports.DEFAULT_IMAGE_TESTS_TIMEOUT = 20000;
+
 exports.expectError = function(file, errorMessage, errorDetail) {
   it('should be rejected with an error', function() {
     return imageStream.getFromFilePath(file).catch((error) => {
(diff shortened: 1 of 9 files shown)


tests/image-stream/tester.js
```diff
@@ -43,6 +43,8 @@ const deleteIfExists = (file) => {
   });
 };
 
+exports.DEFAULT_TIMEOUT = 20000;
```
lurch: I think this should have a name something more like `DEFAULT_IMAGE_TESTS_TIMEOUT` ?

lurch:
@jhermsmeier Feel free to go through a bunch of recent Appveyor and Travis build logs (for all arches & OSes) to work out what a more reasonable default is, that won't cause the CI tests to fail ;-)

jhermsmeier:
@lurch yeah, it's not that important to me – and I can also totally imagine the fs on a CI box being real slow at times, so 20s is okay, where 10s would do as well, I reckon.

tests/image-stream/tester.js
jviotti: Done!

lurch: Thanks :heart:

PR #1600, opened 2017-07-14
refactor(gui): Only enable full debug output on demand

jhermsmeier:
This disables full wildcard debug output by default now,
leaving the possibility to manually enable debug output
via the `DEBUG` environment variable.

Change-Type: patch

### DIFF
--- a/lib/gui/etcher.js
+++ b/lib/gui/etcher.js
@@ -22,10 +22,6 @@ const path = require('path')
 const EXIT_CODES = require('../shared/exit-codes')
 let mainWindow = null
 
-// Enable drivelist debugging information
-// See https://github.com/resin-io-modules/drivelist
-process.env.DRIVELIST_DEBUG = 1
-
 electron.app.on('window-all-closed', electron.app.quit)
 
 // Sending a `SIGINT` (e.g: Ctrl-C) to an Electron app that registers
--- a/lib/gui/index.html
+++ b/lib/gui/index.html
@@ -9,7 +9,12 @@
 
     <!-- Enable debug information from all modules that use `debug` -->
     <!-- See https://github.com/visionmedia/debug#browser-support -->
-    <script>window.localStorage.debug = '*';</script>
+    <script>
+      // Enable drivelist debugging information
+      // See https://github.com/resin-io-modules/drivelist
+      process.env['DRIVELIST_DEBUG'] = /drivelist|^\*$/i.test(process.env['DEBUG']) ? '1' : '';
+      window.localStorage.debug = process.env['DEBUG'];
+    </script>
 
     <script src="./app.js"></script>
   </head>
--- a/lib/start.js
+++ b/lib/start.js
@@ -16,10 +16,6 @@
 
 'use strict'
 
-// Enable debug information from all modules that use `debug`
-// See https://github.com/visionmedia/debug#browser-support
-process.env.DEBUG = '*'
-
 // See http://electron.atom.io/docs/v0.37.7/api/environment-variables/#electronrunasnode
 //
 // Notice that if running electron with `ELECTRON_RUN_AS_NODE`, the binary


Shou:
Looks good to me. I'm curious though, what prompted the change?

jhermsmeier:
A noisy neighbour called drivelist :)

screen shot 2017-07-14 at 16 50 47

lurch:
Hmmm, so @jviotti spends lots of time adding debug code so that he can see what's going wrong, and then @jhermsmeier turns it all off again? ;-)

Are we really expecting end-users to be able to set the `DEBUG` environment variable when we ask them for help with tracking down what's causing them problems? Or have I grabbed the wrong end of the stick yet again?

jviotti:
Maybe a better approach is to blacklist the ones we're not interested in? Does `debug` support logging "levels"? If so, maybe we can use that to still see valuable information, but ignore unnecessary things?

jhermsmeier:
> Are we really expecting end-users to be able to set the DEBUG environment variable when we ask them for help with tracking down what's causing them problems? Or have I grabbed the wrong end of the stick yet again?

As far as I can see the only `debug` user so far is `drivelist` (and some stuff in pipage will be using it as well) – and that's generally really internal state debug output, not the kind we've got going for `mountuils` and such. So, I don't think we'll need users to enable that.

> Maybe a better approach is to blacklist the ones we're not interested in? Does debug support logging "levels"? If so, maybe we can use that to still see valuable information, but ignore unnecessary things?

It doesn't have logging levels as such, but the `DEBUG` value is a selector, so you can do stuff like
```fish
# Debug everything expect drivelist
DEBUG="*,-drivelist"
# Debug only drivelist, and blockmap readstream
DEBUG="drivelist*,blockmap:readstream*"
```
Details [here](https://github.com/visionmedia/debug#wildcards)

jhermsmeier:
Also, I didn't turn it all off – I just made it come on selectively ;P

lurch:
> As far as I can see the only debug user so far is drivelist

Oh, so is there some other kind of debugging used in e.g. `mountutils`? Should the different approaches be harmonised?
Apologies if I'm misunderstanding again, but is it worth adding any of this 'debug support' stuff to the Robot interface?

> Also, I didn't turn it all off – I just made it come on selectively ;P

I was only pulling your leg, hence the smiley ;-)

jhermsmeier:
> Oh, so is there some other kind of debugging used in e.g. mountutils? Should the different approaches be harmonised?

Yeah, the mountutils debug output is coming from C++ land, hence the `debug` module wasn't used. And yup, I think having a standard approach to debug output would be sweet, preferably with `debug` (because being able to selectively enable only what's needed is splendid).

jviotti:
Maybe we should create a `node-debug-cpp` add-on :)

jviotti:
@jhermsmeier I've been taking a look at this, and I have an alternative proposal:

- We can not log "OK" if the temporary script `unlink` phase was successful, and instead only log the error if an error happens (the OK acknowledge is not necessary and by removing it we cut 50% of the logs)

- Instead of trying to unlink the file and not check its callback, maybe we can only proceed once the unlink function ended (successfully or not), to be able to correlate any unlink errors with a specific execution

- Not disable global debugging, since the amount of logs will now probably be fine

What do you guys think?

lurch:
From the example output Jonas posted above, I guess you'd also need to modify drivelist's `write` to not output anything if it succeeded?
BTW, does the `+1s` in the screenshot above mean that it took a whole second to write out the drivelist `.sh` file? That sounds awfully slow for such a small file? :-/  (or does it actually indicate something else?)

jviotti:
> From the example output Jonas posted above, I guess you'd also need to modify drivelist's write to not output anything if it succeeded?

I guess knowing the final script path is good for debugging purposes.

> BTW, does the +1s in the screenshot above mean that it took a whole second to write out the drivelist .sh file? That sounds awfully slow for such a small file? :-/ (or does it actually indicate something else?)

I think it represents the time between that message and the previous one, and I know for sure that drivelist is slow (which is also a reason why I want to go full native on it). I send a PR that makes things better: https://github.com/resin-io-modules/drivelist/pull/194, but still, running the actual script is the bottleneck

jhermsmeier:
> BTW, does the +1s in the screenshot above mean that it took a whole second to write out the drivelist .sh file?

Those times are times between the debug messages as @jviotti pointed out, meaning `drivelist` is called every second, and it takes ~750ms in total until it's done writing out, executing & deleting the script.

jhermsmeier:
> @jhermsmeier I've been taking a look at this, and I have an alternative proposal:

@jviotti I still don't think that we should have debug output enabled at all in a build, unless manually enabled. It has zero value 95% of the time, and cluttering up the console has a cost that we shouldn't ignore (if all of it would be written to a logfile that would be nuked on successful flash, that'd be a different story). Plus logging to the console is really slow compared to anything else, so this also has a speed impact.

I would counter-propose completely disabling it by default (as this PR does), but adding a global `DEBUG()` function users could call in the console should it become necessary. For example `DEBUG()` could toggle `localStorage.debug = '*'` and `process.env['DEBUG'] = '*'`, and when called again, would disable it again.

What do you think?

lurch:
Just to add my 2 cents: In the recently-released (and then pulled) v1.1.0 release it was _really_ annoying having `drivelist write` spam the DevTools console every 2 seconds, and it makes the logs **much** harder to read. And of course it also results in the 'useful logs' quickly scrolling off the top of the screen :-/
So iMHO this part (at least) should be disabled by default ;-)

@jhermsmeier I'd be :-1: on a `DEBUG()` function that added as a 'toggle' - wouldn't it be better to have `DEBUG('*')` turn on debugging, and `DEBUG('')` (or `DEBUG()`) turn off debugging? Of course those functions in turn could be wrapped by `ENABLE_DEBUG()` and `DISABLE_DEBUG()` functions.

jhermsmeier:
@lurch @jviotti I've added a `DEBUG()` function now, that works like this:

```js
// Enable output for all subsystems:
DEBUG('*')
// Enable output for a specific subsystem:
DEBUG('etcher:drive-scanner*')
// Disable all debug output by passing any false-y value:
DEBUG(); DEBUG(false); DEBUG(''); DEBUG(null); DEBUG(undefined); DEBUG(0)
```

jhermsmeier:
@jviotti is this still good to merge?

lurch:
Some docs to go along with this (i.e. how to enable debugging messages when required) would be useful ;-)

jviotti:
I'll approve, but I agree with @lurch that some docs would be very nice.

jviotti:
Ah, never mind, I had already approved :P

PR #1605, opened 2017-07-17
feat(GUI): enlarge reflash button on success page

Shou:
We change the `Flash Another` button in the top-right corner of the
success screen to be bigger, matching the banner buttons' size.

Changelog-Entry: Enlarge Flash Another button on success page.

### DIFF
--- a/lib/gui/pages/finish/styles/_finish.scss
+++ b/lib/gui/pages/finish/styles/_finish.scss
@@ -26,7 +26,7 @@
 }
 
 .page-finish .button-primary {
-  min-width: 255px;
+  min-width: $btn-large-min-width;
   min-height: 53px;
   font-size: 15px;
 }
@@ -63,7 +63,7 @@
 .page-finish .center {
   display: flex;
   align-items: center;
-  justify-content: center;
+  justify-content: space-around;
 }
 
 .page-finish .box > div {
--- a/lib/gui/pages/finish/styles/_finish.scss
+++ b/lib/gui/pages/finish/styles/_finish.scss
@@ -26,7 +26,9 @@
 }
 
 .page-finish .button-primary {
-  min-width: $btn-min-width;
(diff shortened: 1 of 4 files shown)


Shou:
image

lib/gui/pages/finish/styles/_finish.scss
```diff
@@ -26,7 +26,9 @@
 }
 
 .page-finish .button-primary {
-  min-width: $btn-min-width;
```
jviotti: We shouldn't just stop of using this variable. We either update it, so all buttons have the same min width, or adjust the banner's buttons to match the application.

lurch: I think if we made _all_ the buttons have this same minimum width, then the three buttons of the main UI wouldn't fit into the window?

lurch:
@Shou just for comparison purposes, could you also post a screenshot of how this change looks when the offline banner is displayed?

lib/gui/pages/finish/styles/_finish.scss
jviotti: I'm also happy with a new variable called `$btn-large-min-width` or something like that. Basically my point is to try to not make many different exceptions to something like a button width on different parts of the app.

Shou:
@konmouz so the size is matched exactly now, but it still appears bigger, probably because it's lurid compared to the other buttons. 
image

@lurch here's the offline banner.
image

(the discrepancies of the two are quite clear here, I wish we synced their similarities somehow...)

lurch:
Thanks for the screenshots @Shou !

> the discrepancies of the two are quite clear here, I wish we synced their similarities somehow...

I believe @jviotti is planning to publish Etcher v1.1 soon, so maybe that'd be a good opportunity to update the offline-banner to better match the (current) online-banner?

PR #1608, opened 2017-07-19
chore: fix typos in target-convert.sh

lurch:


lurch:
Not sure what impact this would have had? Maybe `pkg` was just defaulting to the host target OS?

jviotti:
Wow, good catch. So `pkg` takes a target in the following form: `node-<platform>-arch`. This change caused `<platform>` to be blank (because the script exitted with an unbound variable error), so we got something like `node--x64`, which was correctly parsed by `pkg`, and as you pointed out, makes it default to the current platform:

```sh
./scripts/build/target-convert.sh: line 53: ARGV_ARCHITECTURE: unbound variable
mkdir dist/Etcher-cli-1.0.0+227533a7-darwin-x64
./node_modules/.bin/npx pkg --output dist/Etcher-cli-1.0.0+227533a7-darwin-x64/etcher -t node6--x64 dist/Etcher-cli-1.0.0+227533a7-darwin-x64-app/lib/cli/etcher.js
```

So this looks great, however I still want to investigate why the Makefile was not failing if some of its scripts was failing :S

jviotti:
So this is used in the top of the Makefile:

```
TARGET_PLATFORM_PKG = $(shell ./scripts/build/target-convert.sh -r $(TARGET_PLATFORM) -t pkg)
```

I guess that either scripts calling from outside a Make rule don't affect the final exit code? Maybe there is a way to turn that on somehow?

jviotti:
Ah OK, so the final variable gets assigned the error message :P

jviotti:
Yeah, looks like `$(shell ...)` doesn't care about the exit code, and will just assign whatever the command output to stderr or stdout.

jviotti:
From https://www.gnu.org/software/make/manual/html_node/Shell-Function.html:

> After the shell function or ‘!=’ assignment operator is used, its exit status is placed in the .SHELLSTATUS variable. 

I guess we can create a small macro that checks this variable, and delegates the message to `$(error)`

lurch:
Unfortunately even if we were using `shellcheck` on our CI servers, it wouldn't have caught this error :-/ https://github.com/koalaman/shellcheck/issues/351

PR #1609, opened 2017-07-19
chore: use electron-builder default package names

jviotti:
We're currently manually changing the names of the final packages
created by `electron-builder`. This commit makes Etcher use the default
package names that `electron-builder` picks for us.

The Windows final package names contain spaces, so I did keep the
`artifactName` entries for them, which now basically use what
`electron-builder` recommends, but use hyphens instead of spaces.

Change-Type: minor
See: https://github.com/resin-io/etcher/pull/1555


lurch:
Could you show how the v1.0 filenames (e.g. from https://etcher.io ) map to the new `electron-builder` default filenames please? :)
(just as an extra comment on this PR, no need to update the commit message)

jviotti:
Sure, here's how it looks now, and after this change:

- Etcher-1.0.0-darwin-x64.dmg -> Etcher-1.0.0.dmg
- Etcher-1.0.0-win32-x64.exe -> Etcher-Setup-1.0.0-x64.exe
- Etcher-1.0.0-win32-x86.exe -> Etcher-Setup-1.0.0-x86.exe
- Etcher-1.0.0-win32-x64.zip -> Etcher-Portable-1.0.0-x64.exe
- Etcher-1.0.0-win32-x86.zip -> Etcher-Portable-1.0.0-x86.exe
- Etcher-1.0.0-linux-x64.AppImage-> etcher-1.0.0-x86_64.AppImage
- Etcher-1.0.0-linux-x86.AppImage -> etcher-1.0.0-i386.AppImage

lurch:
Hmmm, so does that mean we'd be distributing `etcher-1.0.0-i386.AppImage` inside `etcher-1.0.0-linux-x86.zip` ?

jhermsmeier:
@lurch if I remember it right, that's because of the permission bits on the AppImage?

lurch:
@jhermsmeier I wasn't querying the use of zipfiles, but the fact that the zip is named 'x86' but the file inside it is named 'i386'. I've got no idea what the best approach here would be :-/

jhermsmeier:
Oh, sorry – missed that :/

jviotti:
Good catch. I'll update the Makefile on this PR as well.

jviotti:
Done!

jhermsmeier:
@jviotti is it all using x86, or i386 now?

jviotti:
The ZIPs are now using the same convention from the actual AppImages (e.g. i386)

PR #1610, opened 2017-07-19
chore: remove the concept of target and host platforms

jviotti:
Its very unlikely that we will ever support cross platform builds. For
that reason, let's simplify the whole Makefile by removing the concept
of target and host platforms.


Makefile
```diff
@@ -178,10 +172,10 @@ TARGET_ARCH_DEBIAN = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET
 TARGET_ARCH_REDHAT = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t redhat)
 TARGET_ARCH_APPIMAGE = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t appimage)
 TARGET_ARCH_ELECTRON_BUILDER = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t electron-builder)
-TARGET_PLATFORM_PKG = $(shell ./scripts/build/target-convert.sh -r $(TARGET_PLATFORM) -t pkg)
+PLATFORM_PKG = $(shell ./scripts/build/target-convert.sh -r $(PLATFORM) -t pkg)
```
lurch: Probably makes sense to rename `target-convert.sh` to `platform-convert.sh` ?

lurch:
Oh, is cross-platform builds not a feature that `electron-builder` supports?

jviotti:
@lurch Up to a certain point. See https://github.com/electron-userland/electron-builder/wiki/Multi-Platform-Build

> Don't expect that you can build app for all platforms on one platform
> - If your app has native dependencies, it can be compiled only on the target platform. prebuild is a solution, but most node modules don't provide prebuilt binaries.
> - macOS Code Signing works only on macOS. Cannot be fixed.

Both points affect us, plus I don't think its a battle worth fighting.

lurch:
Fair enough :+1: 
Rename `target-convert.sh` and I'll approve ;)

Makefile

jviotti:
Cool, I just pushed that change :)

jhermsmeier:
> macOS Code Signing works only on macOS. Cannot be fixed.

This sounds like something that can be fixed by some mild reverse engineering :)

lurch:
I think getting the MacOS CLI code-signed via `pkg` would be more important than fixing `electron-builder` to be able to code-sign MacOS binaries when running on Linux ;)

jhermsmeier:
Yeah, but this sure sounds like a fun spare-time project to me :)

lurch:
As long as Apple doesn't come after you under the DMCA ;-)

jhermsmeier:
Reverse engineering is legal, unless stuff has been copied from whatever it is your tearing apart, afaik. Also... I'm in Germany, not the US ;P

lurch:
:de: :tada:

PR #1611, opened 2017-07-19
chore: publish development Bintray packages

jviotti:
This commit includes several changes to adapt the CI configuration files
and Bintray publish script to perform development deployments.

- Move our Bintray details to the Makefile
- Deploy to a new Bintray component if `RELEASE_TYPE` is `snapshot`
- Call `publish-bintray-debian` and `publish-bintray-redhat` in the CI
    deployment script
- Call the Bintray deployment scripts for RPMs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,29 @@ APPLICATION_NAME = $(shell jq -r '.displayName' package.json)
 APPLICATION_DESCRIPTION = $(shell jq -r '.description' package.json)
 APPLICATION_COPYRIGHT = $(shell cat electron-builder.yml | shyaml get-value copyright)
 
+BINTRAY_ORGANIZATION = resin-io
+BINTRAY_REPOSITORY_DEBIAN = debian
+BINTRAY_REPOSITORY_REDHAT = redhat
+
+# ---------------------------------------------------------------------
+# Extra variables
+# ---------------------------------------------------------------------
+
+TARGET_ARCH_DEBIAN = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t debian)
+TARGET_ARCH_REDHAT = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t redhat)
+TARGET_ARCH_APPIMAGE = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t appimage)
+TARGET_ARCH_ELECTRON_BUILDER = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t electron-builder)
+PLATFORM_PKG = $(shell ./scripts/build/platform-convert.sh -r $(PLATFORM) -t pkg)
+ENTRY_POINT_CLI = lib/cli/etcher.js
+ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE)
+ifeq ($(TARGET_PLATFORM),win32)
+ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE).exe
+endif
+
+APPLICATION_NAME_LOWERCASE = $(shell echo $(APPLICATION_NAME) | tr A-Z a-z)
+APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
+APPLICATION_VERSION_REDHAT = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
+
 # ---------------------------------------------------------------------
 # Release type
 # ---------------------------------------------------------------------
@@ -36,11 +59,13 @@ PACKAGE_JSON_VERSION = $(shell jq -r '.version' package.json)
 ifeq ($(RELEASE_TYPE),production)
 APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)
 S3_BUCKET = resin-production-downloads
+BINTRAY_COMPONENT = $(APPLICATION_NAME_LOWERCASE)
 endif
 ifeq ($(RELEASE_TYPE),snapshot)
 CURRENT_COMMIT_HASH = $(shell git log -1 --format="%h")
 APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)+$(CURRENT_COMMIT_HASH)
 S3_BUCKET = resin-nightly-downloads
+BINTRAY_COMPONENT = $(APPLICATION_NAME_LOWERCASE)-devel
 endif
 ifndef APPLICATION_VERSION
 $(error Invalid release type: $(RELEASE_TYPE))
@@ -163,26 +188,6 @@ $(warning No Mixpanel token found (ANALYTICS_MIXPANEL_TOKEN is not set))
 else
 ELECTRON_BUILDER_OPTIONS += --extraMetadata.analytics.mixpanel.token=$(ANALYTICS_MIXPANEL_TOKEN)
 endif
-
-# ---------------------------------------------------------------------
-# Extra variables
-# ---------------------------------------------------------------------
-
-TARGET_ARCH_DEBIAN = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t debian)
-TARGET_ARCH_REDHAT = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t redhat)
-TARGET_ARCH_APPIMAGE = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t appimage)
-TARGET_ARCH_ELECTRON_BUILDER = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t electron-builder)
-PLATFORM_PKG = $(shell ./scripts/build/platform-convert.sh -r $(PLATFORM) -t pkg)
-ENTRY_POINT_CLI = lib/cli/etcher.js
-ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE)
-ifeq ($(PLATFORM),win32)
-ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE).exe
-endif
-
-PRODUCT_NAME = etcher
-APPLICATION_NAME_LOWERCASE = $(shell echo $(APPLICATION_NAME) | tr A-Z a-z)
-APPLICATION_VERSION_DEBIAN = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
-APPLICATION_VERSION_REDHAT = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 
 # ---------------------------------------------------------------------
 # Rules
@@ -420,14 +425,14 @@ ifeq ($(RELEASE_TYPE),production)
 		-f $(publishable) \
 		-b $(S3_BUCKET) \
 		-v $(APPLICATION_VERSION) \
-		-p $(PRODUCT_NAME)))
+		-p $(APPLICATION_NAME_LOWERCASE)))
 endif
 ifeq ($(RELEASE_TYPE),snapshot)
 	$(foreach publishable,$^,$(call execute-command,./scripts/publish/aws-s3.sh \
 		-f $(publishable) \
 		-b $(S3_BUCKET) \
 		-v $(APPLICATION_VERSION) \
-		-p $(PRODUCT_NAME) \
+		-p $(APPLICATION_NAME_LOWERCASE) \
 		-k $(shell date +"%Y-%m-%d")))
 endif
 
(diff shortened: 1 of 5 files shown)


jviotti:
It works! Check https://bintray.com/resin-io/debian/etcher-devel/1.0.0%2Bf6fcdc7

jviotti:
It should be easy to do the same thing for RPM packages after this.

jviotti:
I updated the PR to do the same thing for RPMs

Makefile
```diff
@@ -36,11 +40,13 @@ PACKAGE_JSON_VERSION = $(shell jq -r '.version' package.json)
 ifeq ($(RELEASE_TYPE),production)
 APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)
 S3_BUCKET = resin-production-downloads
+BINTRAY_COMPONENT = etcher
```
lurch: Should this be `BINTRAY_COMPONENT = $(APPLICATION_NAME_LOWERCASE)` ?

Makefile
```diff
@@ -36,11 +40,13 @@ PACKAGE_JSON_VERSION = $(shell jq -r '.version' package.json)
...
 endif
 ifeq ($(RELEASE_TYPE),snapshot)
 CURRENT_COMMIT_HASH = $(shell git log -1 --format="%h")
 APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)+$(CURRENT_COMMIT_HASH)
 S3_BUCKET = resin-nightly-downloads
+BINTRAY_COMPONENT = etcher-devel
```
lurch: ...with this then being `BINTRAY_COMPONENT = $(APPLICATION_NAME_LOWERCASE)-devel` ?

scripts/ci/deploy.sh
```diff
@@ -48,8 +48,8 @@ if [ "$ARGV_OPERATING_SYSTEM" == "linux" ]; then
   ./scripts/build/docker/run-command.sh \
     -r "$ARGV_ARCHITECTURE" \
     -s "$(pwd)" \
-    -c "make publish-aws-s3"
+    -c "make publish-aws-s3 publish-bintray-debian publish-bintray-redhat"
 else
   ./scripts/build/check-dependency.sh make
-  make publish-aws-s3
+  make publish-aws-s3 publish-bintray-debian publish-bintray-redhat
```
lurch: I'm not sure about this change - it means that OSX will be trying to make `publish-bintray-debian` ?
What if we instead created a new `deploy` phoney target in the Makefile, which the Makefile could then conditionally add any of the `publish-aws-s3`, `publish-bintray-debian`, `publish-bintray-redhat` targets to as necessary, and then this script would just do `make deploy`?

scripts/publish/bintray-debian.sh
```diff
@@ -28,35 +28,43 @@ function usage() {
   echo "Options"
```
lurch: Should this script have been deleted?

scripts/publish/bintray.sh
```diff
@@ -0,0 +1,104 @@
...
+
+PACKAGE_FILE_NAME=$(basename $ARGV_FILE)
+PACKAGE_NAME=${PACKAGE_FILE_NAME%.*}
+PACKAGE_ARCHITECTURE=$(./scripts/build/architecture-convert.sh -r "$ARGV_ARCHITECTURE" -t debian)
+
+curl --upload-file $ARGV_FILE \
+  --user $BINTRAY_USER:$BINTRAY_API_KEY \
+  --header "X-Bintray-Debian-Distribution: $PACKAGE_DISTRIBUTION" \
+  --header "X-Bintray-Debian-Component: $ARGV_COMPONENT" \
+  --header "X-Bintray-Debian-Architecture: $PACKAGE_ARCHITECTURE" \
```
lurch: Are there not similar meta-tags for uploading the RPM package?

scripts/publish/bintray.sh
```diff
@@ -0,0 +1,104 @@
...
+elif [ "$ARGV_RELEASE_TYPE" == "snapshot" ]; then
+  PACKAGE_DISTRIBUTION=devel
+else
+  echo "Invalid release type: $ARGV_RELEASE_TYPE" 1>&2
+  exit 1
+fi
+
+PACKAGE_FILE_NAME=$(basename $ARGV_FILE)
+PACKAGE_NAME=${PACKAGE_FILE_NAME%.*}
+PACKAGE_ARCHITECTURE=$(./scripts/build/architecture-convert.sh -r "$ARGV_ARCHITECTURE" -t debian)
```
lurch: I guess the `debian` here should actually be `$ARGV_REPOSITORY` ?

Makefile
```diff
@@ -26,6 +26,10 @@ APPLICATION_NAME = $(shell jq -r '.displayName' package.json)
 APPLICATION_DESCRIPTION = $(shell jq -r '.description' package.json)
 APPLICATION_COPYRIGHT = $(shell cat electron-builder.yml | shyaml get-value copyright)
 
+BINTRAY_ORGANIZATION = resin-io
+BINTRAY_REPOSITORY_DEBIAN = debian
+BINTRAY_REPOSITORY_REDHAT = rpm
```
lurch: These two lines seem a bit inconsistent. They should either be:
```
BINTRAY_REPOSITORY_DEBIAN = debian
BINTRAY_REPOSITORY_REDHAT = redhat
```
or
```
BINTRAY_REPOSITORY_DEBIAN = deb
BINTRAY_REPOSITORY_REDHAT = rpm
```

scripts/publish/bintray.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+
```
lurch: Only need one blank line here, not two? ;)

lurch:
https://bintray.com/resin-io/debian/etcher-devel is displaying an html-escaped string
`Flash OS images to SD cards &amp; USB drives, safe &amp; easy (development version)`

I guess after this PR is merged, users will be able to switch between the stable and snapshot versions by just changing their `/etc/apt/sources.list.d/resinio-etcher.list`?

It would be nice to see this PR also updating https://github.com/resin-io/etcher/blob/master/README.md#installers with RPM-from-Bintray instructions :)

Makefile
lurch: Hmm, I see that on https://bintray.com/resin-io you've got `debian` and `rpm`. If this is just because "that's the way that bintray works" (which again seems inconsistent, but ho-hum) then I guess we could either refer to `redhat` in the `Makefile` and then convert it to `rpm` inside `bintray.sh` ; or another approach would be to rename `BINTRAY_REPOSITORY_REDHAT` to `BINTRAY_REPOSITORY_RPM`.

scripts/ci/deploy.sh
lurch: Hmmm, looks like this is already causing Appveyor to fail https://ci.appveyor.com/project/resin-io/etcher/build/1.0.5317/job/injturoyu7u2fkw7

But why is Appveyor trying to publish a not-yet-merged PR-build anyway??! :confused:

Makefile
jviotti: I think we can change the name of the Bintray repository. Let me see...

jviotti: Done! \o/

Makefile

Makefile

scripts/ci/deploy.sh
jviotti: Ah, haha, good catch. I'll fix it!

scripts/publish/bintray-debian.sh
jviotti: Done!

scripts/publish/bintray.sh
jviotti: Done!

scripts/publish/bintray.sh
jviotti: They are equal, but they are not necessarily the same thing. `debian` is the Bintray repository name we picked, while this is one of the types the architecture convert script takes as an argument.

scripts/publish/bintray.sh
jviotti: Turns out not. The only package type that takes extra headers is debian.

jviotti:
I fixed the HTML encoding issue. I'll push an RPM package and figure out the instruction set :)

scripts/ci/deploy.sh
jviotti: I also fixed the Appveyor issue.

Makefile
```diff
@@ -26,6 +26,30 @@ APPLICATION_NAME = $(shell jq -r '.displayName' package.json)
...
+TARGET_ARCH_APPIMAGE = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t appimage)
+TARGET_ARCH_ELECTRON_BUILDER = $(shell ./scripts/build/architecture-convert.sh -r $(TARGET_ARCH) -t electron-builder)
+TARGET_PLATFORM_PKG = $(shell ./scripts/build/target-convert.sh -r $(TARGET_PLATFORM) -t pkg)
+ENTRY_POINT_CLI = lib/cli/etcher.js
+ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE)
+ifeq ($(TARGET_PLATFORM),win32)
+ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE).exe
+endif
+
+PRODUCT_NAME = etcher
```
lurch: Any reason for a separate `PRODUCT_NAME` variable, when (AFAICT) it's the same as `APPLICATION_NAME_LOWERCASE` ?

scripts/publish/bintray.sh
lurch: Well I guess Redhat (or some other distro) might have a different package architecture then, so perhaps the `architecture-convert.sh` script should be moved to the Makefile, and then pass the `PACKAGE_ARCHITECTURE` as another parameter to `bintray.sh` ?

Makefile
jviotti: Yeah, lets get rid of it.

scripts/publish/bintray.sh
jviotti: Not sure I understand the comment. All our scripts, including `bintray.sh` take the generic architecture names we chose, and each of these scripts to the needed transformations inside.

We could pass the already transformed architecture to `bintray.sh`, but then that defeats the purpose of a change we did at some point in the past, where all scripts would take instead the generic arch name to avoid confusion.

lurch: I guess I'm just trying to get rid of as many hard-coded strings as possible...

> They are equal, but they are not necessarily the same thing. debian is the Bintray repository name we picked, while this is one of the types the architecture convert script takes as an argument.

In which case, could we have an extra argument to `bintray.sh` which is precisely "one of the types the architecture convert script takes as an argument" ?

scripts/ci/deploy.sh
```diff
@@ -48,7 +48,7 @@ if [ "$ARGV_OPERATING_SYSTEM" == "linux" ]; then
   ./scripts/build/docker/run-command.sh \
     -r "$ARGV_ARCHITECTURE" \
     -s "$(pwd)" \
-    -c "make publish-aws-s3"
+    -c "make publish-aws-s3 publish-bintray-debian publish-bintray-redhat"
```
lurch: As I mentioned in an earlier comment, IMHO both the `if` and `else` branches of this script should simply be doing `make publish-all`, and then the `Makefile` should be modified to build up that target dynamically (I guess via e.g. a `PUBLISH_TARGETS` variable).

jviotti: Looks like our `installers-all` target already did exactly this. I'm going to update the deploy script to use that instead.

jviotti:
I'm assuming there are no further comments here, but let me know if that's not the case. RPM instructions have been sent by a user: https://github.com/resin-io/etcher/pull/1652

scripts/ci/deploy.sh
lurch: Sorry for missing this, but I see you've now fixed it ;-)

lurch:
Travis Linux snapshot deployments are failing with:
```
./scripts/publish/bintray.sh -f dist/etcher-electron_1.1.1+4827267_amd64.deb -v 1.1.1+4827267 -r x64 -t snapshot -o resin-io -p debian -c etcher-devel
Please define the following environment variables:
  BINTRAY_USER
  BINTRAY_API_KEY
make: *** [publish-bintray-debian] Error 1
Script failed with status 2
```
I guess you need to add those encrypted variables to Travis?

scripts/publish/bintray.sh
lurch: Bumping this comment - I still don't like the face that there's a hardcoded `debian` string at the end of this `architecture-convert.sh` line.

(and Ideally, I think it'd be nice if the `X-Debian-` curl headers were only set on Debian uploads, but I guess this doesn't really matter as long as it all works)

jviotti: Ah, I think I finally understood this comment. I was assuming this line was on `Makefile`, which was the source of the confusion. See https://github.com/resin-io/etcher/pull/1670

jviotti:
Oops, I totally forgot to do so.

PR #1616, opened 2017-07-20
fix(windows): include local native modules in final asar

jviotti:
Etcher has local native code that gets built to `build/`, but that's not
being included inside the final asar.

The fix is to manually add the `*.node` files inside `build/Release` by
adding the appropriate wilcard to the `files` section of
`electron-builder.yml`.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -6,6 +6,7 @@ nodeGypRebuild: true
 publish: null
 files:
   - lib
+  - build/**/*.node
   - assets/icon.png
   - node_modules/**/*
 mac:


lurch:
Uh?!? How come this wasn't picked up before? :-/

jviotti:
@lurch I think it was introduced as part of some of the Windows installer refactorings/tweaks I made at some point, and was not caught until I run through a complete flash process. Hopefully integration tests will help us detect these kinds of things in the future.

lurch:
Could this be the cause of #1617 ?

jviotti:
I don't think so. I built v1 on top of these changes, and this only affects Windows. Can you try to reproduce https://github.com/resin-io/etcher/issues/1617 with the deb/rpm or AppImage (I don't have a real Linux laptop over here)? Maybe I got something wrong during the Linux packaging?

PR #1620, opened 2017-07-22
fix(GUI): throw a user error if the user is not in the sudoers file

jviotti:
Change-Type: patch
Changelog-Entry: Display a user-friendly error message if the user is not in the sudoers file.

### DIFF
--- a/lib/shared/permissions.js
+++ b/lib/shared/permissions.js
@@ -198,6 +198,13 @@ exports.elevateCommand = (command, options) => {
   // for now, we should make sure we double check if the error messages
   // have changed every time we upgrade `sudo-prompt`.
 
+  }).catch((error) => {
+    return _.includes(error.message, 'is not in the sudoers file');
+  }, () => {
+    throw errors.createUserError({
+      title: 'Your user doesn\'t have enough privileges to proceed',
+      description: 'This application requires sudo privileges to be able to write to drives'
+    });
   }).catch({
     message: 'User did not grant permission.'
   }, () => {


lib/shared/permissions.js
```diff
@@ -198,6 +198,13 @@ exports.elevateCommand = (command, options) => {
   // for now, we should make sure we double check if the error messages
   // have changed every time we upgrade `sudo-prompt`.
 
+  }).catch((error) => {
+    return _.includes(_.trim(error.message), 'is not in the sudoers file');
```
lurch: Given that we're using `_.includes`, is the `_.trim` necessary?

lib/shared/permissions.js
```diff
@@ -198,6 +198,13 @@ exports.elevateCommand = (command, options) => {
...
+  }, () => {
+    throw errors.createUserError({
+      title: 'Your user doesn\'t have enough permissions to proceed',
+      description: 'This application requires sudo permissions to interact with removable drives'
```
lurch: The `removable drives` bit is only true if the user hasn't enabled unsafe mode ;-)
I wonder if `This application requires sudo permissions to be able to write to drives` would be more generic / appropriate?

lib/shared/permissions.js
jviotti: Looks like the error message contains some ugly new lines I wanted to protect ourselves from.

lib/shared/permissions.js
jviotti: Fair enough, I'll change it

lurch:
minor nit: There's a typo in the commit-message title - should be `sudoers`  :wink:

jviotti:
Oops, fixed!

lurch:
Out of curiosity, in what scenario did you run into this error? In my testing on both Ubuntu 14.04 and Debian 9 this evening, the elevation-prompt always asked me for the password of a user who *was* in the sudoers file, when I was running Etcher as a user not in the sudoers file.

jviotti:
I saw various occurrences on Sentry, but have no idea what edge case can lead to something like this.

lib/shared/permissions.js
lurch: But `_.trim` only removes whitespace from the start and end of a string, and I'd expect `_.includes` to work the same regardless of leading or trailing whitespace?

jviotti: Fair enough, I'll remove the `_.trim` call.

lib/shared/permissions.js
```diff
@@ -198,6 +198,13 @@ exports.elevateCommand = (command, options) => {
   // for now, we should make sure we double check if the error messages
   // have changed every time we upgrade `sudo-prompt`.
 
+  }).catch((error) => {
+    return _.includes(error.message, 'is not in the sudoers file');
+  }, () => {
+    throw errors.createUserError({
+      title: 'Your user doesn\'t have enough permissions to proceed',
+      description: 'This application requires sudo permissions to be able to write to drives'
```
lurch: Nit: I think these lines should be referring to 'privileges' rather than 'permissions'
(there was something about the phrasing that was niggling me, and I just worked out what it was ;)  )

jviotti: Fixed!

lib/shared/permissions.js
```diff
@@ -198,6 +198,13 @@ exports.elevateCommand = (command, options) => {
   // for now, we should make sure we double check if the error messages
   // have changed every time we upgrade `sudo-prompt`.
 
+  }).catch((error) => {
+    return _.includes(error.message, 'is not in the sudoers file');
+  }, () => {
+    throw errors.createUserError({
+      title: 'Your user doesn\'t have enough permissions to proceed',
+      description: 'This application requires sudo privileges to be able to write to drives'
```
jhermsmeier: As we'll not only be writing to drives in the near future, should we change `[...] to drives` to `devices`, or `storage devices`?

lurch: The language everywhere else in Etcher uses "drives" so I think we should also use "drives" here for consistency. If we decide to change terminology in future, we'll need to make that change in many places in Etcher.

lib/shared/permissions.js
```diff
@@ -198,6 +198,13 @@ exports.elevateCommand = (command, options) => {
   // for now, we should make sure we double check if the error messages
   // have changed every time we upgrade `sudo-prompt`.
 
+  }).catch((error) => {
+    return _.includes(error.message, 'is not in the sudoers file');
+  }, () => {
+    throw errors.createUserError({
+      title: 'Your user doesn\'t have enough permissions to proceed',
```
lurch: This line should also be using `privileges` - sorry if that wasn't clear enough in my last comment.

jviotti: Oh, I missed that one. Fixed!

PR #1624, opened 2017-07-24
chore: `make distclean` now deletes `build` directory too

lurch:


Makefile
```diff
@@ -529,5 +529,6 @@ clean:
 
 distclean: clean
 	rm -rf node_modules
+	rm -rf build
```
jviotti: Should we use `$(BUILD_DIRECTORY)` here?

lurch: Nope, because `$(BUILD_DIRECTORY)` is `dist` (which already gets deleted by the `clean` rule).

jviotti: Ahh, I got confused between those directories :)

PR #1625, opened 2017-07-24
chore: use old custom build system to create AppImages

jviotti:
electron-builder seems to ship with an older AppImages version that
doesn't play very well with the custom AppImages elevation system we
created.

More particularly, we can't execute custom binaries inside the mounted
AppImage given that the mount point seems to lose permissions, owner,
and group file information.

This commit goes back to our old custom build system just for AppImages,
until we properly solve the problem, which will likely involve updating
the AppImages version in electron-builder.


lurch:
I've tested this `linux-appimage-old-packaging` branch on both Ubuntu 14.04 x64 (on real hardware, writing to a uSD card in a USB reader) and Debian 9 x86 (in a Virtual Machine, writing to a (blank) virtual SATA disk using Etcher's "unsafe mode") and in both setups the AppImage it builds seems to work totally fine, whereas neither setup works properly with the AppImage built from the current `master`.

Makefile
```diff
@@ -307,12 +307,39 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)_$(APPLICATION_VERSION_DEBIAN)_$(
...
+ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY = linux-$(TARGET_ARCH_ELECTRON_BUILDER)-unpacked
+endif
+
+$(BUILD_DIRECTORY)/$(ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY):
+	$(NPX) build --dir --linux $(ELECTRON_BUILDER_OPTIONS) \
+		--extraMetadata.name=$(APPLICATION_NAME_ELECTRON) \
 		--extraMetadata.version=$(APPLICATION_VERSION) \
 		--extraMetadata.packageType=AppImage
 
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(PLATFORM).AppDir: \
```
lurch: To keep things consistent, should this line (and line 335) be using `$(APPLICATION_NAME_LOWERCASE)` instead of `$(APPLICATION_NAME)` ?

Makefile
```diff
@@ -307,12 +307,39 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)_$(APPLICATION_VERSION_DEBIAN)_$(
...
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_ARCH_APPIMAGE).AppImage: \
-	| $(BUILD_DIRECTORY)
-	$(NPX) build --linux AppImage $(ELECTRON_BUILDER_OPTIONS) \
+ifeq ($(TARGET_ARCH),x64)
+ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY = linux-unpacked
+else
+ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY = linux-$(TARGET_ARCH_ELECTRON_BUILDER)-unpacked
+endif
+
+$(BUILD_DIRECTORY)/$(ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY):
```
lurch: This ought to have an order-only dependency on `$(BUILD_DIRECTORY)` (i.e. add `	| $(BUILD_DIRECTORY)` to the end of the line)

Makefile
```diff
@@ -307,12 +307,39 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)_$(APPLICATION_VERSION_DEBIAN)_$(
 		--extraMetadata.packageType=deb \
 		$(DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS)
 
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_ARCH_APPIMAGE).AppImage: \
-	| $(BUILD_DIRECTORY)
-	$(NPX) build --linux AppImage $(ELECTRON_BUILDER_OPTIONS) \
+ifeq ($(TARGET_ARCH),x64)
+ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY = linux-unpacked
```
lurch: AFAICT these names conflict with the directory created by the current `package-electron` rule. Is that deliberate, or is it something it'd be better to try and avoid?

Makefile
```diff
@@ -307,12 +307,39 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)_$(APPLICATION_VERSION_DEBIAN)_$(
...
+		-i assets/icon.png \
+		-o $@
+
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(TARGET_ARCH_APPIMAGE).AppImage: \
+	 $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(PLATFORM).AppDir \
+	 | $(BUILD_DIRECTORY) $(BUILD_TEMPORARY_DIRECTORY)
+	./scripts/build/electron-create-appimage-linux.sh \
+		-d $< \
+		-r $(TARGET_ARCH) \
+		-w $(BUILD_TEMPORARY_DIRECTORY) \
```
lurch: Haha, good thing that somebody reminded you not to delete the definition of this variable :stuck_out_tongue_winking_eye:

Makefile
lurch: As I've (probably) mentioned before, it's always better to try and avoid having directories as Makefile-targets, because they can easily mess things up.
Would an alternative be to have `$(BUILD_DIRECTORY)/$(ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY)/$(APPLICATION_NAME_ELECTRON)` as the target of this rule? (The last rule for this target would then need to be `touch $@` so that the datestamp gets updated when the target gets built)

jviotti:
Awesome, I'll merge so we can release, and will address the comments later.

Makefile
jviotti: In `electron-builder`, the packaging directory is a temporary stage before creating the installers, and thus its deliberate.

Makefile

Makefile
jviotti: Sounds reasonable!

Makefile
jviotti: Sounds good, done!

jviotti:
See https://github.com/resin-io/etcher/issues/1647

PR #1630, opened 2017-07-25
fix(CLI): pass required arguments to flashComplete message

jviotti:
The `flashComplete` message takes the drive object and the image
basename as arguments. This was updated on the GUI, but causes the CLI
to throw an error upon completion.

Change-Type: patch
Changelog-Entry: Fix "imageBasename is not defined" error on the CLI.

### DIFF
--- a/lib/gui/modules/drive-scanner.js
+++ b/lib/gui/modules/drive-scanner.js
@@ -17,7 +17,6 @@
 'use strict';
 
 const Rx = require('rx');
-const os = require('os');
 const _ = require('lodash');
 const EventEmitter = require('events').EventEmitter;
 const drivelist = require('drivelist');
@@ -38,19 +37,6 @@ const availableDrives = Rx.Observable.timer(
 
   .flatMap(() => {
     return Rx.Observable.fromNodeCallback(drivelist.list)();
-  })
-
-  // Build human friendly "description"
-  .map((drives) => {
-    return _.map(drives, (drive) => {
-      drive.name = drive.device;
-
-      if (os.platform() === 'win32' && !_.isEmpty(drive.mountpoints)) {
-        drive.name = _.join(_.map(drive.mountpoints, 'path'), ', ');
-      }
-
-      return drive;
-    });
   })
 
   .map((drives) => {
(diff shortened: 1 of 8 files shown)


lib/shared/messages.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
 
     flashComplete: _.template([
       '<%= imageBasename %> was successfully written to',
-      '<%= drive.description %> (<%= drive.name %>)'
+      '<%= drive.description %> (<%= drive.device %>)'
```
lurch: Won't this change also affect the GUI too? I assume `drive.name` was previously used here for good reasons?
(I can never remember what the `.name` and `.device` fields map to for different OSes)

jviotti: We use to manually add the `.name` property in the drive scanner, which contained the device path in Linux and macOS, and the list of mounted volumes joined by commas on Windows. In order to not duplicate that logic on the CLI, I moved the addition of the name property to `drivelist`, and upgraded it here.

lib/gui/modules/drive-scanner.js
```diff
@@ -40,19 +40,6 @@ const availableDrives = Rx.Observable.timer(
     return Rx.Observable.fromNodeCallback(drivelist.list)();
   })
 
-  // Build human friendly "description"
-  .map((drives) => {
-    return _.map(drives, (drive) => {
-      drive.name = drive.device;
-
-      if (os.platform() === 'win32' && !_.isEmpty(drive.mountpoints)) {
```
jhermsmeier: Looks like the `require('os')` needs removing now:
```
etcher\lib\gui\modules\drive-scanner.js
  20:7  error  'os' is assigned a value but never used  no-unused-vars
```

jviotti: Oops, fixed!

jviotti:
@jhermsmeier Is this good to merge?

lurch:
Great! It's a shame that the commit-message didn't mention the bump of `drivelist` though ;)

lib/gui/modules/drive-scanner.js
jhermsmeier: If I see this correctly, this block can now also be removed from https://github.com/resin-io/etcher/pull/1637, right @jviotti?

jviotti: Yep, we can get rid of it :)

PR #1635, opened 2017-07-26
[WIP] feat: introduce and use electron-redux when possible

Shou:
We make use of the `electron-redux` library when possible, and otherwise
fall back to a plain instance of redux.

Changelog-Entry: Introduce and use electron-redux when possible.

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -177,7 +177,7 @@ app.run(() => {
       `(eta ${currentFlashState.eta}s)`
     ].join(' '));
 
-    windowProgress.set(currentFlashState.percentage);
+    windowProgress.set(currentFlashState);
   });
 });
 
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -177,7 +177,7 @@ app.run(() => {
       `(eta ${currentFlashState.eta}s)`
     ].join(' '));
 
-    windowProgress.set(currentFlashState);
+    windowProgress.set(currentFlashState.percentage);
   });
 });
 
--- a/lib/shared/store.js
+++ b/lib/shared/store.js
@@ -487,13 +487,15 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
   }
 };
 
+const storeActionsAndDefaults = {
+  Actions: ACTIONS,
+  Defaults: DEFAULT_STATE
+};
+
 // We're not in Electron
 if (typeof process === 'undefined') {
   console.log('PLAIN REDUX');
-  module.exports = _.merge(redux.createStore(storeReducer, DEFAULT_STATE), {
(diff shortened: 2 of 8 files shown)


Shou:
All the redux-related tests are failing here, hence the WIP tag. Looks like either there is no main process in the tests, or asynchrony is being introduced and mucking things up. It works outside the tests.

lib/gui/app.js
```diff
@@ -177,7 +177,7 @@ app.run(() => {
       `(eta ${currentFlashState.eta}s)`
     ].join(' '));
 
-    windowProgress.set(currentFlashState.percentage);
+    windowProgress.set(currentFlashState);
```
lurch: Hmm? Did this change accidentally slip through from #1438 ?

lib/shared/store.js
```diff
@@ -486,7 +487,46 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
+// We are in the main Electron process
+} else if (process) {
+  console.log('MAIN REDUX');
+  const store = _.merge(redux.createStore(storeReducer, DEFAULT_STATE, redux.applyMiddleware(
+    electronRedux.triggerAlias,
+    electronRedux.forwardToRenderer
+  )), {
+    Actions: ACTIONS,
+    Defaults: DEFAULT_STATE
+  });
```
lurch: Looks like this `{ Actions: ACTIONS, Defaults: DEFAULT_STATE }` is repeated 3 separate times. I dunno what this code is doing, but would it be worth making a new constant for that?

lurch:
Looks like the majority of this diff is just renaming a `data` attribute to `payload`. Apologies if this is a dumb question, but is that a necessary change?

Shou:
`electron-redux` explains the change to `payload` as [FSA compliance](https://github.com/hardchor/electron-redux#actions), and indeed it doesn't work without the change.

lib/gui/app.js
Shou: Well spotted

Shou:
Shelving this for now.

lib/shared/store.js
```diff
@@ -486,7 +487,42 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
+};
+
+// We're not in Electron
+if (typeof process === 'undefined') {
+  console.log('PLAIN REDUX');
+  module.exports = _.merge(redux.createStore(storeReducer, DEFAULT_STATE), storeActionsAndDefaults);
+
+// We are in the renderer process of Electron
+} else if (_.get(process, [ 'type' ]) === 'renderer') {
+  console.log('RENDERER REDUX');
```
lurch: If this ever gets re-opened again, you'll need to remember to remove these `console.log`s ;-)

PR #1637, opened 2017-07-26
fix(gui): Avoid running drivelist scans in parallel

jhermsmeier:
This prevents `drivelist` scans from being run in parallel, in turn preventing "Couldn't initialize the scanner" errors.

Of course this only partially fixes the symptoms created by https://github.com/resin-io-modules/drivelist/issues/198 – but prevents causing it by itself.

Change-Type: patch
Connects To: https://github.com/resin-io/etcher/issues/1639, https://github.com/resin-io-modules/drivelist/issues/198


lib/gui/modules/drive-scanner.js
```diff
@@ -16,100 +16,144 @@
...
+ * @private
+ * @param {Array<Object>} drives - list of discovered drives
+ * @returns {Array<Object>}
  * @example
- * driveScanner.start();
+ * drives = DriveScanner.processDrives(drives);
  */
-emitter.start = () => {
-  availableDrives.resume();
+DriveScanner.processDrives = (drives) => {
```
Shou: Couldn't this be inside the class definition?

jhermsmeier: Yeah, I also had it in there first, but the linter complained about it being a class method without using `this`, so here it is now.

Shou: Oh yeah, if I recall correctly, when you use `static` it stops complaining.

jhermsmeier: Right, totally forgot about that keyword existing in JS now :) Mended it!

lurch:
A possibly-naive question, but would changing things so that `drivelist` got run e.g. every four seconds instead of every two seconds, help to potentially alleviate this problem?

jhermsmeier:
> EDIT: And does the removal of rx make Etcher less "cool"? ;)

😆  Even cooler. Sub-zero cool.

> A possibly-naive question, but would changing things so that drivelist got run e.g. every four seconds instead of every two seconds, help to potentially alleviate this problem?

Possibly, yes – but with Rx that'd heavily depend on how quick the thing runs & how powerful the machine is. This solution only kicks off another scan once the previous finished – so regardless of the delay, it'll never be run in parallel again (unless you've got two or more Etchers open, that is).

jhermsmeier:
I've now added scan & error counts to the DriveScanner, and made it only display an error once, in light of the discussed upcoming behavior around drivelist. By also counting the total scans, we could also get a success ratio from it, should we want to.

lib/gui/app.js
```diff
@@ -199,9 +199,21 @@ app.run(($timeout) => {
...
-    return exceptionReporter.report(error);
+    // Clear the drive list, to avoid displaying stale data
+    $timeout(() => {
+      availableDrives.setDrives([]);
+    });
+
+    // TODO: Once drivelist can recover from transient failures
+    // (i.e. when a device is unplugged while it's scanning),
+    // only send `analytics.logException(error);` and continue scanning
+    if (!driveScanner.erroCount) {
```
jviotti: Should this be `errorCount`?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,100 +16,150 @@
...
+/**
+ * @summary DriveScanner class
+ * @class
+ * @extends {EventEmitter}
+ * @example
+ * var scanner = new DriveScanner()
+ *   .on('error', (error) => { ... })
+ *   .on('drives', (drives) => { ... })
+ */
+class DriveScanner extends EventEmitter {
```
jviotti: Do we really need a whole class to perform this task? Just worrying that we're adding a lot of complexity (e.g. the number of code lines exploded compared to Rx) to perform a relatively simple task.

Something like this (with a couple extra variables to track things) should implement the same thing:

```js
const scan = (emitter) => {
  return drivelist.scanAsync().then((drives) => {
    emitter.emit('drives', drives);

    if (shouldContinue) {
      return Bluebird.delay(2000).then(scan);
    }
  });
};
```

jhermsmeier: > Do we really need a whole class to perform this task? Just worrying that we're adding a lot of complexity (e.g. the number of code lines exploded compared to Rx) to perform a relatively simple task.

I don't see this being overly complex? Sure, that little that is going on here is visible now, but more than 50% of the lines are comments and whitespace, and the actual LOC inflation is relatively minimal, considering this uses no libs/frameworky things. Wrapping things in Promises will sure hide some of that little complexity and shave off a line or two, but I don't see that being worth it, or necessarily easier to reason about.

Plus, I think you'll end up with roughly the same LOC if doing it without a class, as you'll still have set up the state keeping vars, start(), stop() and the event emitter. Best to inherit from the EventEmitter, and keep all the context in one place, imo.

Does that make sense?

lib/gui/app.js
jhermsmeier: Yes, good catch – fixed!

lib/gui/modules/drive-scanner.js
```diff
@@ -16,100 +16,150 @@
...
 const EventEmitter = require('events').EventEmitter;
 const drivelist = require('drivelist');
+const debug = require('debug')('etcher:drive-scanner');
 const settings = require('../models/settings');
 
-const DRIVE_SCANNER_INTERVAL_MS = 2000;
-const DRIVE_SCANNER_FIRST_SCAN_DELAY_MS = 0;
-const emitter = new EventEmitter();
+const DRIVE_SCANNER_DELAY_MS = 2000;
+const DRIVE_SCANNER_FIRST_SCAN_DELAY_MS = 125;
```
lurch: Why do we want / need a first-scan-delay? What determines this 125ms value?
(Just curious...)

lib/gui/modules/drive-scanner.js
```diff
@@ -16,100 +16,150 @@
...
 const _ = require('lodash');
 const EventEmitter = require('events').EventEmitter;
 const drivelist = require('drivelist');
+const debug = require('debug')('etcher:drive-scanner');
 const settings = require('../models/settings');
 
-const DRIVE_SCANNER_INTERVAL_MS = 2000;
-const DRIVE_SCANNER_FIRST_SCAN_DELAY_MS = 0;
-const emitter = new EventEmitter();
+const DRIVE_SCANNER_DELAY_MS = 2000;
```
lurch: For consistency with `DRIVE_SCANNER_FIRST_SCAN_DELAY_MS`, I wonder if it might make sense to rename this to `DRIVE_SCANNER_REPEAT_SCAN_DELAY_MS` ?  \*shrug\*

lib/gui/modules/drive-scanner.js
jhermsmeier: Ping @jviotti ^

lib/gui/modules/drive-scanner.js
lurch: Would it be 'better' to not start the drive-scanner at all, until Etcher has finished loading / initialising itself?

jhermsmeier: Added comments, detailing the use & purpose of the constants now, @lurch

lib/gui/modules/drive-scanner.js
jviotti: Fair enough, sounds good to me :+1:

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
 
-const availableDrives = Rx.Observable.timer(
-  DRIVE_SCANNER_FIRST_SCAN_DELAY_MS,
-  DRIVE_SCANNER_INTERVAL_MS
-)
+  /**
+   * @summary DriveScanner constructor
+   * @param {Object} [options] - EventEmitter options
+   * @example
+   * var scanner = new DriveScanner()
```
jviotti: Should we use `const` here?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
+   *   .on('error', (error) => { ... })
+   *   .on('drives', (drives) => { ... })
+   */
+  constructor(options) {
+    super(options);
+    this.isRunning = false;
+    this.timer = null;
+    this.scanCount = 0;
```
jviotti: Should we only add this when there is a use case for it?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
+    this.errorCount = 0;
+    this.run = this.run.bind(this);
```
jviotti: Is this really needed? AFAIK fat arrow functions and ES6 classes should make this unnecessary.

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
+  }
 
-/* eslint-enable lodash/prefer-lodash-method */
+  /**
+   * @summary Start the scanner
+   * @public
+   * @example
+   * scanner.start()
```
jviotti: Semicolons?

lib/gui/app.js
```diff
@@ -200,9 +200,21 @@ app.run(($timeout) => {
     // otherwise we risk presenting the same error over
     // and over again to the user, while also heavily
     // spamming our error reporting service.
+    // TODO: Continue scanning once drivelist is fixed (see below)
     driveScanner.stop();
 
-    return exceptionReporter.report(error);
+    // Clear the drive list, to avoid displaying stale data
```
jviotti: So basically, if there is an error, we clean the list of available drives?

lib/gui/app.js
```diff
@@ -200,9 +200,21 @@ app.run(($timeout) => {
...
+    $timeout(() => {
+      availableDrives.setDrives([]);
+    });
+
+    // TODO: Once drivelist can recover from transient failures
```
jviotti: Is this something we should fix? Can you fill a bug report?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
+/**
+ * @summary Time to wait in between scan calls
+ * @type {Number}
+ * @constant
+ */
+const DRIVE_SCANNER_DELAY_MS = 2000;
+
+/**
+ * @summary Time to delay the first scan (in milliseconds)
+ * @description This prevents the drive scans from kicking
```
jviotti: Interesting. I wonder if instead of hard-coding a magic number, we can listen to any Electron browser window event to trigger the loop once the window is shown, for example.

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
+  /**
+   * @summary DriveScanner constructor
+   * @param {Object} [options] - EventEmitter options
+   * @example
+   * var scanner = new DriveScanner()
+   *   .on('error', (error) => { ... })
+   *   .on('drives', (drives) => { ... })
+   */
+  constructor(options) {
+    super(options);
```
jviotti: Do we really want to call the `EventEmitter` constructor with these options?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
+    this.scanCount += 1;
+    DriveScanner.scan((error, drives) => {
+      debug('scan', error || drives);
+      if (error) {
+        this.errorCount += 1;
+        this.emit('error', error);
+      } else {
+        this.emit('drives', drives);
+      }
+      if (this.isRunning) {
```
jviotti: Is it necessary to check this here? If `isRunning` is false, then the next `this.run` call will already `return`, right?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
- * driveScanner.stop();
- */
-emitter.stop = () => {
-  availableDrives.pause();
-};
+  /**
+   * @summary Scan for drives, automatically filtering out system drives
+   * @private
+   * @static
+   * @param  {Function} callback - callback(error, drives)
```
jviotti: There are two spaces between `@param` and `{Function}`

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,140 @@
...
+   * @example
+   * DriveScanner.scan((error, drives) => {
+   *   ...
+   * });
+   */
+  static scan(callback) {
+    drivelist.list((error, drives) => {
```
jviotti: Should we promisify `drivelist` and use promises here, for consistency with the rest of the Etcher codebase?

lib/gui/modules/drive-scanner.js
jhermsmeier: Ooops, sneaky space bar - fixed!

lib/gui/modules/drive-scanner.js
jhermsmeier: Yeah, but we don't want to schedule a timer for a noop, I reckon

lib/gui/modules/drive-scanner.js
jhermsmeier: The only options that would be passed are EventEmitter options, but as it turns out those were removed from the Node docs now, and there's `.setMaxListeners(n)`, which does the same as passing that as option. Removed now.

lib/gui/app.js
jhermsmeier: This should be fixed by https://github.com/resin-io-modules/drivelist/pull/202 and can be removed once that's in & updated in Etcher

lib/gui/app.js
jhermsmeier: Just a measure against accidental flashing of the wrong device until https://github.com/resin-io-modules/drivelist/pull/202 is merged; because if devices are unplugged or new ones plugged in the device numbers could mismatch what's displayed in case of an error. And to avoid Etcher nuking devices it shouldn't, the list is cleared.

lib/gui/modules/drive-scanner.js
jhermsmeier: Yes, sir! :)

lib/gui/modules/drive-scanner.js
jhermsmeier: Yeah, but `setTimeout(this.run)` doesn't keep context, that's why it's being bound here.

lib/gui/modules/drive-scanner.js
jhermsmeier: Hmm, yeah - removed!

lib/gui/modules/drive-scanner.js
jhermsmeier: Done!

jviotti:
The drivelist fixes have been merged and I'm upgrading the version in https://github.com/resin-io/etcher/pull/1653. Let's rebase and make the appropriate changes once that PR is merged.

lib/gui/app.js
lurch: Good call, always better to err on the side of caution

lib/gui/modules/drive-scanner.js
lurch: Yeah, that's exactly what I was asking with my comment the other day ;-)

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,137 @@
...
+class DriveScanner extends EventEmitter {
 
-const availableDrives = Rx.Observable.timer(
-  DRIVE_SCANNER_FIRST_SCAN_DELAY_MS,
-  DRIVE_SCANNER_INTERVAL_MS
-)
+  /**
+   * @summary DriveScanner constructor
+   * @example
+   * const scanner = new DriveScanner()
```
lurch: Huh? Identical examples in this comment-block and the previous one. Is one of them redundant?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,137 @@
...
-/* eslint-enable lodash/prefer-lodash-method */
+  /**
+   * @summary Start the scanner
+   * @public
+   * @example
+   * scanner.start();
+   */
+  start() {
+    debug('start');
+    this.isRunning = true;
```
lurch: I wonder if `start` should check that `this.isRunning == false`, to guard against `start` being called twice?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,137 @@
...
- * @public
- *
- * @example
- * driveScanner.stop();
- */
-emitter.stop = () => {
-  availableDrives.pause();
-};
+  /**
+   * @summary Scan for drives, automatically filtering out system drives
```
lurch: nit: I think `automatically` here should be changed to `optionally`

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,137 @@
...
+   * @private
+   * @static
+   * @param {Function} callback - callback(error, drives)
+   * @example
+   * DriveScanner.scan((error, drives) => {
+   *   ...
+   * });
+   */
+  static scan(callback) {
```
lurch: Is it worth `scan` checking that `this.isRunning == true` ?

jhermsmeier: `scan()` is static (`DriveScanner.scan()`), has no access to `this` and can be called at any time, regardless of a scanner instance

lib/gui/modules/drive-scanner.js
jhermsmeier: That's because it's the same thing, and the linter complains otherwise :/

lib/gui/modules/drive-scanner.js
jhermsmeier: Yeah, I'll tend to this when rebasing after #1653

lib/gui/modules/drive-scanner.js
lurch: @jviotti IMHO we shouldn't be duplicating identical examples - what should Jonas do here to keep the linter happy?

lib/gui/modules/drive-scanner.js
lurch: > can be called at any time

Doesn't that kinda blow a hole in the efforts to prevent multiple scans being run in parallel? ;-)

lib/gui/modules/drive-scanner.js
jviotti: I think you can remove the `class` JSDoc and just document the constructor.

lib/gui/modules/drive-scanner.js
jhermsmeier: Not really, only theoretically – it's not used that way. Had to make it static, because of the linter yelling at me (rightfully).

lib/gui/modules/drive-scanner.js
jhermsmeier: Since it's inside some Angular run loop thing, I'm not sure how fiddely this'd be. Maybe it makes sense to do this with the React transition?

lib/gui/app.js
jhermsmeier: Removed this now, in light of the recent drivelist fixes having been applied.

lib/gui/modules/drive-scanner.js
jhermsmeier: Done (removed most of it).

lib/gui/modules/drive-scanner.js
jhermsmeier: Yup, done!

lib/gui/modules/drive-scanner.js
jhermsmeier: Mended!

jhermsmeier:
@jviotti @lurch all comments addressed – is this good to go now?

lib/gui/modules/drive-scanner.js
```diff
@@ -16,86 +16,130 @@
...
+   * @summary Scan for drives, optionally filtering out system drives
+   * @private
+   * @static
+   * @param {Function} callback - callback(error, drives)
+   * @example
+   * DriveScanner.scan((error, drives) => {
+   *   ...
+   * })
+   */
+  static scan (callback) {
```
jviotti: Can we make this a promise?

jhermsmeier: That'd just make things more complex, and duplicate timer & state handling in then() and catch() blocks – is that really necessary?

lib/gui/modules/drive-scanner.js
lurch: :laughing:

lib/gui/app.js
```diff
@@ -210,9 +210,16 @@ app.run(($timeout) => {
...
-    return exceptionReporter.report(error)
+    // Clear the drive list, to avoid displaying stale data
+    $timeout(() => {
+      availableDrives.setDrives([])
+    })
+
+    // Only send the first error, and continue scanning
+    if (!driveScanner.errorCount) {
+      analytics.logException(error)
+    }
```
lurch: I wonder if it'd be worth doing something a bit cleverer and only avoid sending _duplicate_ errors, rather than avoiding sending every error after the first?

lib/gui/modules/drive-scanner.js
jviotti: What about something like this? It gets rid of the timer, and the extra `.bind()`, plus reduces the code size a bit (disclaimer: I didn't actually run it :P)

```diff
diff --git a/lib/gui/modules/drive-scanner.js b/lib/gui/modules/drive-scanner.js
index 2193fe8e..9a72e84d 100644
--- a/lib/gui/modules/drive-scanner.js
+++ b/lib/gui/modules/drive-scanner.js
@@ -18,7 +18,8 @@
 
 const _ = require('lodash')
 const EventEmitter = require('events').EventEmitter
-const drivelist = require('drivelist')
+const Bluebird = require('bluebird');
+const drivelist = Bluebird.promisifyAll(require('drivelist'))
 const settings = require('../models/settings')
 const debug = require('debug')('etcher:drive-scanner')
 
@@ -71,8 +72,9 @@ class DriveScanner extends EventEmitter {
     if (this.isRunning) {
       return
     }
+
     this.isRunning = true
-    this.timer = setTimeout(this.run, DRIVE_SCANNER_FIRST_SCAN_DELAY_MS)
+    Bluebird.delay(DRIVE_SCANNER_FIRST_SCAN_DELAY_MS).then(this.run);
   }
 
   /**
@@ -86,18 +88,16 @@ class DriveScanner extends EventEmitter {
       return
     }
 
-    DriveScanner.scan((error, drives) => {
-      debug('scan', error || drives)
-      if (error) {
-        this.errorCount += 1
-        this.emit('error', error)
-      } else {
-        this.emit('drives', drives)
-      }
+    DriveScanner.scan().then((drives) => {
+      this.emit('drives', drives)
+
       if (this.isRunning) {
-        this.timer = setTimeout(this.run, DRIVE_SCANNER_DELAY_MS)
+        return Bluebird.delay(DRIVE_SCANNER_DELAY_MS).then(this.run);
       }
-    })
+    }).catch((error) => {
+      this.errorCount += 1
+      this.emit('error', error)
+    });
   }
 
   /**
@@ -109,7 +109,6 @@ class DriveScanner extends EventEmitter {
   stop () {
     debug('stop')
     this.isRunning = false
-    clearTimeout(this.timer)
   }
 
   /**
@@ -122,22 +121,15 @@ class DriveScanner extends EventEmitter {
    *   ...
    * })
    */
-  static scan (callback) {
-    drivelist.list((error, drives) => {
-      if (error) {
-        callback(error)
-        return
-      }
-
-      let list = drives
-
+  static scan () {
+    return drivelist.listAsync().then((drives) => {
       if (!settings.get('unsafeMode')) {
-        list = _.reject(drives, {
+        return _.reject(drives, {
           system: true
         })
       }
 
-      callback(null, list)
+      return drives
     })
   }
 }
```

lurch: I was initially a bit confused about there being no equivalent of the `clearTimeout` in the code above, but then I had a more careful look at the code, and realised that even Jonas's current solution would still work if the `clearTimeout` was removed ;-)

lurch: > It gets rid of ... the extra `.bind()` ...

Does it? ;)

jviotti: > Does it? ;)

Oh, I forgot to delete it :P It should be necessary since its far arrow functions now

jhermsmeier:
@jviotti is this PR obsolete now that this is tied into the SDK?

jviotti:
Yeah, I think it was sorted by the commit that moved the scanning function to the SDK.

Issue #1636, opened 2017-07-26
GUI: Image SVG icons fail to render


jhermsmeier:
- **Etcher version:** 1.1.1
- **Operating system and architecture:** All
- **Image flashed:** https://github.com/ev3dev/ev3dev/releases
- **Do you see any meaningful error information in the DevTools?**

**Error:**

```
ENOENT, lib/gui/components/<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http:/www.inkscape.org/) -->
<svg id="svg4040" xmlns:rdf="http:/www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns="http:/www.w3.org/2000/svg" viewBox="0 0 48 48" version="1.1" xmlns:cc="http:/creativecommons.org/ns#" xmlns:dc="http:/purl.org/dc/elements/1.1/">
 ...
</svg>
 not found in /Applications/Etcher.app/Contents/Resources/app.asar
```

**Stack:**

```
Error: ENOENT, lib/gui/components/<?xml version="1.0" encoding="UTF-8" standalone="no"?>
 not found in /Applications/Etcher.app/Contents/Resources/app.asar
    at notFoundError (ELECTRON_ASAR.js:115:19)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:514:9)
    at SVGIcon.render (/Applications/Etcher.app/Contents/Resources/app.asar/lib/gui/components/svg-icon.js:56:25)
    at /Applications/Etcher.app/Contents/Resources/app.asar/node_modules/react-dom/lib/ReactCompositeComponent.js:796:21
    at measureLifeCyclePerf (/Applications/Etcher.app/Contents/Resources/app.asar/node_modules/react-dom/lib/ReactCompositeComponent.js:75:12)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (/Applications/Etcher.app/Contents/Resources/app.asar/node_modules/react-dom/lib/ReactCompositeComponent.js:795:25)
    at ReactCompositeComponentWrapper._renderValidatedComponent (/Applications/Etcher.app/Contents/Resources/app.asar/node_modules/react-dom/lib/ReactCompositeComponent.js:822:32)
    at ReactCompositeComponentWrapper._updateRenderedComponent (/Applications/Etcher.app/Contents/Resources/app.asar/node_modules/react-dom/lib/ReactCompositeComponent.js:746:36)
[...]

PR #1642, opened 2017-07-28
fix(GUI): make archive-embedded svgs work again

Shou:
We make the svg-icon component accept XML in its path argument to handle
archive-embedded SVG icons.

Changelog-Entry: Make archive-embedded SVG icons work again.
Fixes: https://github.com/resin-io/etcher/issues/1636

### DIFF
--- a/lib/gui/components/svg-icon.js
+++ b/lib/gui/components/svg-icon.js
@@ -54,18 +54,10 @@ class SVGIcon extends react.Component {
     // relatively to the `index.html`.
     const imagePath = path.join(__dirname, this.props.path);
 
-    let contents;
+    let contents = '';
 
     if (_.startsWith(this.props.path, '<')) {
-      const parser = new DOMParser();
-      const doc = parser.parseFromString(this.props.path, 'image/svg+xml');
-      const svg = doc.querySelector('svg');
-      const foreignObjects = svg.querySelectorAll('foreignobject');
-      _.map(foreignObjects, (foreignObject) => {
-        foreignObject.parentNode.removeChild(foreignObject);
-      });
-
-      contents = svg.outerHTML;
+      contents = this.props.path;
 
     } else {
       contents = fs.readFileSync(imagePath, {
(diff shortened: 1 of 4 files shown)


lurch:
Haha, I was just about to leave a comment asking for a unit-test, when you added the second commit to this PR :laughing:

jhermsmeier:
Can confirm it mitigates the little bugger now:
screen shot 2017-07-28 at 16 51 16

lib/gui/components/svg-icon.js
```diff
@@ -53,9 +54,24 @@ class SVGIcon extends react.Component {
     // relatively to the `index.html`.
     const imagePath = path.join(__dirname, this.props.path);
 
-    const contents = fs.readFileSync(imagePath, {
-      encoding: 'utf8'
-    });
+    let contents;
```
jhermsmeier: The linter's yelling about this not being initialized to a value; maybe just set it to an empty string; `let contents = '';`

Shou:
Passes tests and linter locally now, so this is good for reviewing now! Should also repel rickrolls.

lurch:
Just out of curiosity, what happens if it's given a malformed SVG that the DOMParser can't parse correctly?

jhermsmeier:
Then the DOMParser will return an error document: https://developer.mozilla.org/en-US/docs/Web/API/DOMParser#Error_handling – you can also see that if you load up a malformed SVG in your browser, for example.

Shou:
Looks like that causes it to use the previous icon, which in some cases isn't the default, if you've selected another image with an embedded icon, so I need to fix that.

jhermsmeier:
@Shou I just started working on handling parser errors, so don't bother, I'm on it :)

lurch:
Presumably you'll also be checking how it copes with malformed PNGs too? ;-)

jhermsmeier:
Chromium handles those just fine on its own. Also, we don't support custom PNGs – it has to be an SVG, see [`lib/image-stream/archive.js#L112`](https://github.com/resin-io/etcher/blob/master/lib/image-stream/archive.js#L112):

```js
logo: extractEntryByPath(archive, `${basePath}/logo.svg`, options),
```

lurch:
Mea culpa - didn't realise that.

PR #1647, opened 2017-07-30
refactor: address review comments from #1625

jviotti:
That PR was merged in a rush, so this is the follow-up commit addressing
the review comments made there.

See: https://github.com/resin-io/etcher/pull/1625


Makefile
```diff
@@ -313,26 +313,27 @@ else
...
-$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-$(APPLICATION_VERSION)-$(PLATFORM).AppDir: \
-	$(BUILD_DIRECTORY)/$(ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY) \
+$(BUILD_DIRECTORY)/$(APPLICATION_NAME_LOWERCASE)-$(APPLICATION_VERSION)-$(PLATFORM).AppDir: \
+	$(BUILD_DIRECTORY)/$(ELECTRON_BUILDER_LINUX_UNPACKED_DIRECTORY)/$(APPLICATION_NAME_ELECTRON) \
 	| $(BUILD_DIRECTORY)
 	./scripts/build/electron-create-appdir.sh \
 		-n $(APPLICATION_NAME) \
 		-d "$(APPLICATION_DESCRIPTION)" \
-		-p $< \
+		-p $(dirname $<) \
```
lurch: I guess this should have been `dir` rather than `dirname` ;-)  https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html

PR #1648, opened 2017-07-31
feat: support rpi-sdcard image file type

moragues:
Support the rpi-sdcard image file type output by Yocto for
the Raspberry Pi device.

Changelog-Entry: Add support for .rpi-sdcard images.

### DIFF
--- a/lib/image-stream/supported.js
+++ b/lib/image-stream/supported.js
@@ -72,5 +72,9 @@ module.exports = [
   {
     extension: 'sdcard',
     type: 'image'
+  },
+  {
+    extension: 'rpi-sdimg',
+    type: 'image'
   }
 ];
--- a/tests/shared/supported-formats.spec.js
+++ b/tests/shared/supported-formats.spec.js
@@ -35,7 +35,7 @@ describe('Shared: SupportedFormats', function() {
 
     it('should return the supported non compressed extensions', function() {
       const extensions = supportedFormats.getNonCompressedExtensions();
-      m.chai.expect(extensions).to.deep.equal([ 'img', 'iso', 'dsk', 'hddimg', 'raw', 'dmg', 'sdcard' ]);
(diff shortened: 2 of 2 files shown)


moragues:
Tested both .rpi-sdcard and .rpi-sdcard.gz.
UI allowed selection of image and flashing was as expected.

Question:  Should I be including npm-shrinkwrap.json and package.json as modified by the build on my machine?

jviotti:
> Question: Should I be including npm-shrinkwrap.json and package.json as modified by the build on my machine?

That shouldn't be needed. They are probably caused by a different node version on your system. 

I'll restart the CI services since it seems GitHub had some issues that affected this build.

moragues:
I submitted this right as github appeared to be having difficulties.
This is my first submission to the project.  If I did something incorrectly, please let me know.

Thanks.
-Bob

jviotti:
No worries, you didn't do anything wrong. npm is a bit painful to work with sometimes, so the whole Etcher team and the CI services use the exact same Node.js version to avoid subtle differences.

jhermsmeier:
Restarted one build job again - Electron failed to download in one of the containers.

jviotti:
BTW, thanks for the awesome commit @moragues. You are one of the few that perfectly adhered to the commit guidelines :tada:

moragues:
You are welcome @jviotti.  It was actually my first project submission on github.  I generally work in a gerrit environment where we don't push back to the open source repos except for publishing the sources as we release devices.  This process was pretty easy compared to some that I have seen.

Have a great day.

lurch:
My only nitpick would be that the commit-message refers to `rpi-sdcard` but the actual PR adds `rpi-sdimg` ??

@jviotti I thought we'd decided against adding extra extensions, and were going to 'fix' this by adding a wildcard file-filter? See #1360

jviotti:
Yeah, but given no-one is working on it, I guess we can white-list extensions users are particularly interested in, and then remove some from the list once we add the wildcard file-filter.

PR #1650, opened 2017-07-31
chore: replace release types with a STABLE_RELEASE flag 

jviotti:
This is one of the many changes needed to properly migrate deployments
to GitHub Releases.

We remove the `RELEASE_TYPE` variable and add a `STABLE_RELEASE` flag
instead, which when set, will cause the builds to be uploaded to GitHub
without using the "prerelease" tag.

Change-Type: patch


lurch:
To avoid confusion, I'd rather rename "snapshot" as "prerelease".
Also, from what I understood, isn't the only differences between `regular`, `update` and `major` the fact that there'll be different settings used inside the auto-update file (which is itself stored on GH releases)? So it probably doesn't make sense to expose these as different settings within the Makefile?
Also, I thought that 'snapshot' releases were now going to be 'regular' releases where only the last digit of the semver got bumped, with a future eye towards moving towards VersionBot, and thus the commit-hash part of the filename would be disappearing? (which AIUI means the only difference between a 'snapshot' release and a 'regular' release, is the fact that the former will be added to GH releases as a prerelease.)

jviotti:
Fair enough. I think we can simplify a lot by just adding a boolean flag determining if the release should be tagged as pre-release or not.

> Also, I thought that 'snapshot' releases were now going to be 'regular' releases where only the last digit of the semver got bumped, with a future eye towards moving towards VersionBot, and thus the commit-hash part of the filename would be disappearing?

Right, but we have to keep the commit hash until VB is up and running in this repo.

jviotti:
I re-purposed this PR to simply remove `RELEASE_TYPE` and add a prerelease boolean flag, which simplifies a lot of things. Once we implement auto updates, we can add a new variable to control how many people should update to a certain build, and the combination of those variables will provide us all the release types we need.

lurch:
> I think we can simplify a lot by just adding a boolean flag determining if the release should be tagged as pre-release or not.

Exactly. AFAiCT the only differences between a regular, update or major release are all things that happen "after" the Makefile step, so IMHO the Makefile should only know about `prelease` and `regular` builds.

> Right, but we have to keep the commit hash until VB is up and running in this repo.

Oh? We can still use `versionist` to manually version-bump, before we enable VB can't we? Given the large changes this new system introduces, I think it might be better to do it all at once, rather than trying to introduce it in small steps, and end up with 'weird' setups during the in-between steps.
I think the only thing we'll need to do differently is not push every 'regular' release to S3, to prevent the current versions of Etcher from displaying the update prompt?

Makefile
```diff
@@ -53,22 +53,18 @@ APPLICATION_VERSION_REDHAT = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 # Release type
 # ---------------------------------------------------------------------
 
-# Add the current commit to the version if release type is "snapshot"
-RELEASE_TYPE ?= snapshot
 PACKAGE_JSON_VERSION = $(shell jq -r '.version' package.json)
-ifeq ($(RELEASE_TYPE),production)
+
+ifdef PRERELEASE
```
lurch: I think it'd be better to invert this, so that `prerelease` is the default-mode (just like `snapshot` used to be the default-mode), and you have to do "something special" (i.e. setting a make-variable) in order to get `regular` builds?

jviotti:
> AFAiCT the only differences between a regular, update or major release are all things that happen "after" the Makefile step, so IMHO the Makefile should only know about prelease and regular builds.

The issue is that `electron-builder` is the one that generates the file that determines how many people will auto update, so we need a way to pass that information to it from the Makefile, but I guess that's something to discuss once we have auto updates up and running.

> Oh? We can still use versionist to manually version-bump, before we enable VB can't we? Given the large changes this new system introduces, I think it might be better to do it all at once, rather than trying to introduce it in small steps, and end up with 'weird' setups during the in-between steps.
I think the only thing we'll need to do differently is not push every 'regular' release to S3, to prevent the current versions of Etcher from displaying the update prompt?

We can, but we'd have to rebase every PR every single time something is merged, and ensure we only run versionist when appropriate (on the PR), or add some code to the CI servers to run versionist once the PR is merged, and create a commit, etc which is basically re-inventing VersionBot :P 

We had a discussion at some point about ways to improve VersionBot to handle this particular case for Etcher, so I think we should wait for that to happen.

Makefile
jviotti: Sounds good. I'll think of a variable name :P

jviotti: I guess just `STABLE` should do it.

jviotti:
Hm, Appveyor is failing with weird errors:

> Build version 1.0.5503 already existed and a random string was appended to its version to avoid collision. Update next build number or change version format.

jviotti:
I restarted all the jobs and it seems to be OK now.

lurch:
> The issue is that electron-builder is the one that generates the file that determines how many people will auto update, so we need a way to pass that information to it from the Makefile, but I guess that's something to discuss once we have auto updates up and running.

We could add a separate `UPDATE_STRATEGY` Makefile variable for that? ;-)

> We had a discussion at some point about ways to improve VersionBot to handle this particular case for Etcher, so I think we should wait for that to happen.

Maybe this PR should also wait for that new version of VB too? ;)

jviotti:
> We could add a separate UPDATE_STRATEGY Makefile variable for that? ;-)

Yep, exactly.

> Maybe this PR should also wait for that new version of VB too? ;)

The GitHub Releases migration is blocking on a big feature, which is auto-updates. I rather get this up and running with pre-release builds having a commit hash so we can ship auto-updates (and then we can aggressively release other changes).

lurch:
But then don't we come back to the problem (if only using commit-hashes to differentiate prerelease builds) of not being able to tell which is the "newest" prerelease build? (or indeed whether the latest prerelease build on GH is newer or older than the prelease build we're running locally?)

Makefile
```diff
@@ -53,22 +53,18 @@ APPLICATION_VERSION_REDHAT = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 # Release type
 # ---------------------------------------------------------------------
 
-# Add the current commit to the version if release type is "snapshot"
-RELEASE_TYPE ?= snapshot
 PACKAGE_JSON_VERSION = $(shell jq -r '.version' package.json)
-ifeq ($(RELEASE_TYPE),production)
+
+ifdef STABLE
```
lurch: Just a minor niggle, but `STABLE` somehow feels like too short of a word for this (IYSWIM). What about something like `STABLE_RELEASE` ?

jviotti:
> But then don't we come back to the problem (if only using commit-hashes to differentiate prerelease builds) of not being able to tell which is the "newest" prerelease build? (or indeed whether the latest prerelease build on GH is newer or older than the prelease build we're running locally?)

GitHub Releases entries have their own timestamps, which we can use to determine the latest available version, independently on the actual version string.

lurch:
> GitHub Releases entries have their own timestamps, which we can use to determine the latest available version, independently on the actual version string.

Does that mean that we also ought to store the build timestamp in the `package.json`, so we can work out if the latest prerelease on GH is newer than the version we're running locally?

docs/PUBLISHING.md
```diff
@@ -37,14 +23,9 @@ Update Channels
 ---------------
 
 Etcher has a setting to include the unstable update channel. If this option is
```
lurch: Presumably in a later PR, this will change to `Etcher has a setting to include the prerelease update channel.` ? (with the name of the actual setting being changed too)

docs/PUBLISHING.md
```diff
@@ -37,14 +23,9 @@ Update Channels
...
-set, Etcher will consider both stable and unstable versions when showing the
-update notifier dialog. Unstable versions are the ones that contain a `beta`
-pre-release tag. For example:
-
-- Production unstable version: `1.4.0-beta.1`
-- Snapshot unstable version: `1.4.0-beta.1+7fde24a`
-- Production stable version: `1.4.0`
-- Snapshot stable version: `1.4.0+7fde24a`
```
lurch: If we're still including the commit-hash on prerelease builds, perhaps part of this wording should be kept?

Makefile
```diff
@@ -53,22 +53,18 @@ APPLICATION_VERSION_REDHAT = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
 # Release type
 # ---------------------------------------------------------------------
 
-# Add the current commit to the version if release type is "snapshot"
```

docs/PUBLISHING.md
jviotti: Sounds good, I'll send another PR soon.

docs/PUBLISHING.md
jviotti: Done!

jviotti:
> Does that mean that we also ought to store the build timestamp in the package.json, so we can work out if the latest prerelease on GH is newer than the version we're running locally?

We can find the latest version available on GH releases, and them to semver comparison with the current version, to avoid having to embed the timestamp.

> nit: I think the boolean should be removed from the commit-title, so that it's just chore: replace release types with a STABLE_RELEASE flag - otherwise people might think that make <target> STABLE_RELEASE=0 builds a prerelease version, and that's not the case.

Done!

lurch:
> We can find the latest version available on GH releases, and them to semver comparison with the current version, to avoid having to embed the timestamp.

But we might have the situation where the latest pre-release version on GH-releases is `1.1.2-abdcef` and the pre-release version we're running locally is `1.1.2-fedb23` and we have no way of knowing if the version on GH is newer than the version we're already running??

lurch:
Commit-message still refers to `boolean variable` - could we change that to just `flag`?

Makefile
```diff
@@ -53,22 +53,19 @@ APPLICATION_VERSION_REDHAT = $(shell echo $(APPLICATION_VERSION) | tr "-" "~")
...
 APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)
 S3_BUCKET = resin-production-downloads
 BINTRAY_COMPONENT = $(APPLICATION_NAME_LOWERCASE)
-endif
-ifeq ($(RELEASE_TYPE),snapshot)
-CURRENT_COMMIT_HASH = $(shell git log -1 --format="%h")
-APPLICATION_VERSION = $(PACKAGE_JSON_VERSION)+$(CURRENT_COMMIT_HASH)
+BINTRAY_PACKAGE_DISTRIBUTION = stable
+else
+# Add the current commit to the version if release type is "snapshot"
```
lurch: `current commit` -> `current commit hash` ?
`"snapshot"` -> `"prerelease"` ?

Makefile
```diff
@@ -473,10 +468,10 @@ publish-bintray-debian: $(PUBLISH_BINTRAY_DEBIAN)
 		-f $(publishable) \
 		-v $(APPLICATION_VERSION_DEBIAN) \
 		-r $(TARGET_ARCH) \
-		-t $(RELEASE_TYPE) \
 		-o $(BINTRAY_ORGANIZATION) \
 		-p $(BINTRAY_REPOSITORY_DEBIAN) \
-		-c $(BINTRAY_COMPONENT)))
+		-c $(BINTRAY_COMPONENT) \
+		-d $(BINTRAY_PACKAGE_DISTRIBUTION)))
```
lurch: I guess an alternative approach, which _might_ be more consistent with the way we handle `ARCH` and `PLATFORM`, would be to just pass the 'base' SNAPSHOT / PRERELEASE value into the S3-publish and Bintray-publish scripts, and then have them use a `releasetype-convert.sh` script, similar to the `architecture-convert.sh` and `platform-convert.sh` scripts? \*shrug\*
e.g. `scripts/build/relasetype-convert.sh -r stable -t S3`could return `resin-production-downloads` and `scripts/build/relasetype-convert.sh -r prerelease -t bintray` could return `devel` ?

Hmmm, although that's probably one level of abstraction too much, as there's actually **two** variables which you need to modify for Bintray when switching between prerelease and stable versions... :-/
...so I guess you could have:
```
scripts/build/relasetype-convert.sh -r stable -t S3_bucket -> resin-production-downloads
scripts/build/relasetype-convert.sh -r prerelease -t bintray_distribution -> devel
scripts/build/relasetype-convert.sh -r prerelease -t bintray_component -n etcher -> etcher-devel
```
Yeah, that's kinda ugly and not very extensible; ignore me and just leave it all in the `Makefile` ;-)

Makefile

jviotti:
This PR is stale. Closing...

PR #1651, opened 2017-07-31
feat(GUI): check GH Releases for versions when showing the update dialog

jviotti:
This commit makes Etcher look at GitHub Releases to determine whether it
should show the update notification dialog or not.

The build system currently works with S3 (there is an in progress PR to
switch to GitHub Releases), however we already mirror all the files in
GitHub Releases, and that's that the website already looks at.

Change-Type: minor
Changelog-Entry: Look at GitHub Releases to determine if the update notification dialog should be displayed.


jviotti:
This PR is stale. Closing...

Issue #1525, opened 2017-06-16
Error: unable to get local issuer certificate


kodweis:
- **Etcher version:** 1.0.0
- **Operating system and architecture:** Linux ghost 4.8.0-54-generic #57-Ubuntu SMP Wed May 24 10:21:44 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
- **Image flashed:** n/a
- **Do you see any meaningful error information in the DevTools?** : [screenshots](http://imgur.com/a/t0Co7)

It still writes images ok.. just this message pops up every time I open the app. Probably has to do with corporate proxy I'm in

jviotti:
Hi @kodweis ,

Are you behind a proxy or something like that? As the error message states, you should update your system root SSL certificates (`ca-certificates` package).

This can also happen because of an un-trusted certificate, but I'm ruling out that possibility since its unlikely Amazon S3 will have such a certificate.

jviotti:
> Probably has to do with corporate proxy I'm in

Oh, I missed that part. Yeah, looks like the proxy might not be well configured and might be messing up with the connection's SSL certificates.

jviotti:
There's not much we can do from our side here. I only see two options:

- Swallow the error (and maybe log it in DevTools just for debugging purposes)
- Present a friendly warning window describing what happened, every time Etcher starts

What do you think?

kodweis:
What functionality is lost due to this error? An update check?
Since it's not a critical error that prevents the app from performing it's main purpose, I would say just log it to DevTools. I don't think this error warrants a popup dialog every time someone starts the app.

jviotti:
Yeah, its just an update check. I agree with you, and will make the change soon.

jviotti:
Hi @kodweis , I agree. Thanks a lot for the feedback. I implemented the discussed changes, which should land in the next version.

kodweis:
Awesome! Thanks!

PR #1655, opened 2017-08-02
fix(GUI): log known user errors when querying S3 for updates

jviotti:
We send an HTTP request to S3 to determine the latest available version.
There are various error that can happen that we don't have control over
(like `ETIMEDOUT`).

The current approach is to whitelist certain errors and pretend there is
no update available, however this commit improves that whole situation.

Instead of swallowing these errors, we throw a user error from the
function that determines the latest available version. From the
application code, we check if that function throws a user error, and if
so, instead of showing it to the user, we log a mixpanel event and carry
on.

This change is motivated by the latest reporter error we can't do
anything about: `UNABLE_TO_GET_ISSUER_CERT_LOCALLY`.

Fixes: https://github.com/resin-io/etcher/issues/1525
Change-Type: patch
Changelog-Entry: Fix `UNABLE_TO_GET_ISSUER_CERT_LOCALLY` error at startup when behind certain proxies.

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -34,6 +34,7 @@ const messages = require('../shared/messages')
 const s3Packages = require('../shared/s3-packages')
 const release = require('../shared/release')
 const store = require('../shared/store')
+const errors = require('../shared/errors')
 const packageJSON = require('../../package.json')
 const flashState = require('../shared/models/flash-state')
 const settings = require('./models/settings')
@@ -152,6 +153,17 @@ app.run(() => {
 
       return updateNotifier.notify(latestVersion, {
         allowSleepUpdateCheck: currentReleaseType === release.RELEASE_TYPE.PRODUCTION
+      })
+
+    // If the error is an update user error, then we don't want
+    // to bother users each time they open the app.
+    // See: https://github.com/resin-io/etcher/issues/1525
+    }).catch((error) => {
+      return errors.isUserError(error) && error.code === 'UPDATE_USER_ERROR'
+    }, (error) => {
+      analytics.logEvent('Update check user error', {
+        title: errors.getTitle(error),
+        description: errors.getDescription(error)
       })
     })
   }).catch(exceptionReporter.report)
--- a/tests/shared/errors.spec.js
+++ b/tests/shared/errors.spec.js
@@ -445,6 +445,16 @@ describe('Shared: Errors', function () {
       m.chai.expect(errors.getDescription(error)).to.equal('Something happened')
     })
 
+    it('should correctly add a code', function () {
+      const error = errors.createError({
+        title: 'Foo',
+        description: 'Something happened',
+        code: 'HELLO'
+      })
+
+      m.chai.expect(error.code).to.equal('HELLO')
+    })
+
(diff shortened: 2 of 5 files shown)


lurch:
Huh, why go to the trouble of creating a user-error and then swallowing it again afterwards? Wouldn't it be easier to just log the mixpanel event directly?  (am I missing something obvious?)

jviotti:
@lurch The s3 packages module runs in the SDK, while the analytics are GUI specific, therefore we have to deal with it from the GUI code.

lib/gui/app.js
```diff
@@ -154,6 +155,16 @@ app.run(() => {
         allowSleepUpdateCheck: currentReleaseType === release.RELEASE_TYPE.PRODUCTION
       });
     });
+
+  // If the error is a user error, then we don't want
+  // to bother the user each time he opens the app.
+  // See: https://github.com/resin-io/etcher/issues/1525
+  }).catch(errors.isUserError, (error) => {
```
lurch: Should we also check the `title` of the error, so that we only swallow 'Unable to check for updates' errors, in case there are other errors which we _do_ want the user to see?

jviotti: Should good. I'm passing an error code, and checking that here instead.

lib/gui/app.js
```diff
@@ -154,6 +155,18 @@ app.run(() => {
         allowSleepUpdateCheck: currentReleaseType === release.RELEASE_TYPE.PRODUCTION
       });
     });
+
+  // If the error is an update user error, then we don't want
+  // to bother users each time they open the app.
+  // See: https://github.com/resin-io/etcher/issues/1525
+  }).catch({
```
lurch: Shouldn't we first check that we've got a userError, and _then_ check the error's `code` ?

jviotti: Sounds reasonable. Done!

lib/gui/app.js
```diff
@@ -154,6 +155,17 @@ app.run(() => {
...
       });
     });
+
+  // If the error is an update user error, then we don't want
+  // to bother users each time they open the app.
+  // See: https://github.com/resin-io/etcher/issues/1525
+  }).catch((error) => {
+    return errors.isUserError(error) && error.code === 'UPDATE_USER_ERROR';
+  }, (error) => {
+    analytics.logEvent('Update check user error', {
```
lurch: I guess in the majority of cases, if we get an error while trying to download the updates list from S3 (and it fails due to network issues), then chances are that the error we're trying to log here won't be able to make it back to Mixpanel either ;-)

jviotti: True, but at least it will be nicely logged in DevTools

tests/shared/errors.spec.js
```diff
@@ -445,6 +445,16 @@ describe('Shared: Errors', function () {
       m.chai.expect(errors.getDescription(error)).to.equal('Something happened')
     })
 
+    it('should correctly add a code', function () {
+      const error = errors.createError({
+        title: 'Foo',
+        description: 'Something happened',
```
lurch: LOL, I love the idea of `Something happened` being an error :rofl:

PR #1657, opened 2017-08-02
refactor: adhere mostly to StandardJS guidelines

jviotti:
This commit changes the whole codebase to adhere to all StandardJS
guidelines rules except semicolons, since the removal of semicolons
affect pretty much all lines, and the final diff is very hard to follow
(and to assess other more involved changes).

In a nutshell:

- When using `function`, we now require a space before the opening
  parenthesis
- If a line with operators is broken into multiple lines, the operator
  should now go after the line break
- Unnecessary padding lines are now forbidden

There were also some minor things that the `standard` CLI caught that I
updated here.

See: https://standardjs.com

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -184,7 +184,6 @@ app.run(() => {
 
 app.run(($timeout) => {
   driveScanner.on('drives', (drives) => {
-
     // Safely trigger a digest cycle.
     // In some cases, AngularJS doesn't acknowledge that the
     // available drives list has changed, and incorrectly
@@ -195,7 +194,6 @@ app.run(($timeout) => {
   });
 
   driveScanner.on('error', (error) => {
-
     // Stop the drive scanning loop in case of errors,
     // otherwise we risk presenting the same error over
     // and over again to the user, while also heavily
@@ -241,7 +239,6 @@ app.run(($window) => {
         // This circumvents the 'beforeunload' event unlike
         // electron.remote.app.quit() which does not.
         electron.remote.process.exit(EXIT_CODES.SUCCESS);
-
       }
 
       analytics.logEvent('Close rejected while flashing');
@@ -252,7 +249,6 @@ app.run(($window) => {
 
 app.run(($rootScope) => {
   $rootScope.$on('$stateChangeSuccess', (event, toState, toParams, fromState) => {
-
     // Ignore first navigation
     if (!fromState.name) {
       return;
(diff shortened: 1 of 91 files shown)


jhermsmeier:
What do you think about using [standard/eslint-config-standard](https://github.com/standard/eslint-config-standard) and extend that, to shrink down our custom defined rule set?

jviotti:
Yep, we definitely should. I'll handle it it another PR.

lurch:
Out of curiosity, have any / many other resin.io repos switched to this new style?

jviotti:
I think we're the pioneers :P

lib/gui/app.js
```diff
@@ -297,10 +292,9 @@ app.controller('HeaderController', function(OSOpenExternalService) {
     const supportUrl = selectionState.getImageSupportUrl() || DEFAULT_SUPPORT_URL;
     OSOpenExternalService.open(supportUrl);
   };
-
 });
 
-app.controller('StateController', function($rootScope, $scope) {
+app.controller('StateController', function ($rootScope, $scope) {
```
lurch: Out of curiosity (this may well be another one of my stupid questions) how come this is still using the `function` keyword rather than fat-arrow notation?

lib/gui/components/safe-webview.js
```diff
@@ -78,18 +78,17 @@ const API_VERSION = 1;
...
-  constructor(props) {
+  constructor (props) {
     super(props);
 
     this.state = {
       shouldShow: true
     };
 
-    const url = new URL(props.src);
+    const url = new window.URL(props.src);
```
lurch: @Shou is this correct?

lib/gui/app.js
jviotti: Angular relies on being able to set custom contexts with `.call()` and `.apply()`, which fat arrow functions don't support. Basically, Angular.js 2 is not ES6 ready.

jhermsmeier:
> Out of curiosity, have any / many other resin.io repos switched to this new style?

@lurch yup https://github.com/resin-io/resin-preload

lib/gui/components/safe-webview.js
Shou: Yup! URL lives on `window`, and this makes it explicit, so a good change.

lurch:
@jhermsmeier Yup, saw that ;-)

PR #1663, opened 2017-08-03
fix(svg-icon): Handle DOMParser parsererrors

jhermsmeier:
This checks for `parsererror`s in the DOMParser's returned document,
to prevent re-use of the previous image and avoid displaying an error
document in the SVG icon slot.

screen shot 2017-08-03 at 13 19 46

Change-Type: patch
Connects To: #1642

### DIFF
--- a/lib/gui/components/svg-icon.js
+++ b/lib/gui/components/svg-icon.js
@@ -35,6 +35,8 @@ const angularSVGIcon = angular.module(MODULE_NAME, [])
 
 const DEFAULT_SIZE = '40px'
 
+const domParser = new window.DOMParser()
+
 /**
  * @summary SVG element that takes both filepaths and file contents
  * @type {Object}
@@ -65,11 +67,11 @@ class SVGIcon extends react.Component {
     const width = this.props.width || DEFAULT_SIZE
     const height = this.props.height || DEFAULT_SIZE
 
-    const parser = new window.DOMParser()
-    const doc = parser.parseFromString(contents, 'image/svg+xml')
+    const doc = domParser.parseFromString(contents, 'image/svg+xml')
+    const parserError = doc.querySelector('parsererror')
     const svg = doc.querySelector('svg')
-
-    const svgData = `data:image/svg+xml,${encodeURIComponent(svg.outerHTML)}`
+    const svgXml = svg && _.isNil(parserError) ? svg.outerHTML : ''
+    const svgData = `data:image/svg+xml,${encodeURIComponent(svgXml)}`
 
     return react.createElement('img', {
       className: 'svg-icon',
--- a/tests/gui/components/svg-icon.spec.js
+++ b/tests/gui/components/svg-icon.spec.js
@@ -72,6 +72,17 @@ describe('Browser: SVGIcon', function () {
       m.chai.expect(element.children().attr('src')).to.equal(imgData)
     })
 
+    it('should use an empty src if there is a parsererror', function () {
+      // The following is invalid, because there's no closing tag for `foreignObject`
+      const iconContents = '<svg><foreignObject></svg>'
+      const imgData = `data:image/svg+xml,`
+      $rootScope.iconContents = iconContents
+
+      const element = $compile('<svg-icon path="iconContents">Resin.io</svg-icon>')($rootScope)
+      $rootScope.$digest()
+      m.chai.expect(element.children().attr('src')).to.equal(imgData)
+    })
+
     it('should default the size to 40x40 pixels', function () {
       const icon = '../../../lib/gui/assets/etcher.svg'
       const element = $compile(`<svg-icon path="'${icon}'">Resin.io</svg-icon>`)($rootScope)


lib/gui/components/svg-icon.js
```diff
@@ -35,6 +35,8 @@ const angularSVGIcon = angular.module(MODULE_NAME, []);
 
 const DEFAULT_SIZE = '40px';
 
+const domParser = new window.DOMParser();
```
jhermsmeier: We can re-use one instance for all the parsing

lib/gui/components/svg-icon.js
```diff
@@ -65,12 +67,13 @@ class SVGIcon extends react.Component {
     const width = this.props.width || DEFAULT_SIZE;
     const height = this.props.height || DEFAULT_SIZE;
 
-    const parser = new window.DOMParser();
-    const doc = parser.parseFromString(contents, 'image/svg+xml');
+    const img = document.createElement('img');
+    const doc = domParser.parseFromString(contents, 'image/svg+xml');
+    const parserError = doc.querySelector('parsererror');
     const svg = doc.querySelector('svg');
+    const svgXml = svg && _.isNil(parserError) ? svg.outerHTML : '';
```
jhermsmeier: Not sure if we might want to display a custom error-icon of ours instead?

lurch: I think we should just default to our regular icon, but maybe we could add a warning message or something to DevTools?

lurch:
Is there any way to add unit-tests for this?

lib/gui/components/svg-icon.js
jhermsmeier: @lurch this component is generic to all SVG icons used in Etcher, not just the external ones – defaulting to the regular drive icon is not what we'd want to do here.

jviotti: Maybe we can allow this directive to be passed as fallback? If the main one fails, we try the fallback, otherwise we do nothing (e.g. use an empty string)

lurch: Rather than passing a fallback SVG into this directive, it might be better for the code using the directive to check to check if the loading failed, and if so try to load an alternative SVG? \*shrug\*  That way it's not limited to only a single fallback... ;-)

Shou: Since we're going to introduce a new attribute for embedded contents, maybe we can prioritize the embedded SVG attribute and use the path as a fallback?

jviotti: The thing is that its hard for the client code to "catch" these kinds of problems (there is no good messaging protocol apart from events). Can we allow this directive to take an array of embedded SVG strings, or an array of paths? We will try the first one, and if that fails, the next one, etc and if nothing works we throw an error.

jhermsmeier:
Rebased, and added a test now @lurch

lurch:
As mentioned in flowdock (afraid I've not logged in for a couple of days) would this PR also be a good place to add separate methods for "load the SVG data from this file-path" and "here's the SVG data (as a string)" ?

tests/gui/components/svg-icon.spec.js
```diff
@@ -72,6 +72,17 @@ describe('Browser: SVGIcon', function () {
       m.chai.expect(element.children().attr('src')).to.equal(imgData)
     })
 
+    it('should use an empty src if there is a parsererror', function () {
+      // The following is invalid, because there's no closing tag for `foreignObject`
+      const iconContents = '<svg><foreignObject></svg>'
+      const imgData = `data:image/svg+xml,`
+      $rootScope.iconContents = iconContents
+
+      const element = $compile('<svg-icon path="iconContents">Resin.io</svg-icon>')($rootScope)
```
lurch: Out of curiosity, how come `$compile` here is using a single-quoted string, but on e.g. line 88 `$compile` is using a backtick-quoted string?

jviotti: The back-ticks are used when you need to interpolate variables in the string

tests/gui/components/svg-icon.spec.js
```diff
@@ -72,6 +72,17 @@ describe('Browser: SVGIcon', function () {
       m.chai.expect(element.children().attr('src')).to.equal(imgData)
     })
 
+    it('should use an empty src if there is a parsererror', function () {
+      // The following is invalid, because there's no closing tag for `foreignObject`
+      const iconContents = '<svg><foreignObject></svg>'
+      const imgData = `data:image/svg+xml,`
```
Shou: Shouldn't this use single-quotes because it doesn't need to process any variables?

Shou:
@lurch I think we should just get this merged, and we'll address the other issue in a separate PR, keep them simple so we can get it through fast and hopefully in time for 1.1.2.

tests/gui/components/svg-icon.spec.js
lurch: Yeah, that's what I thought. So why is variable-interpolation not needed here, when it is needed on line 88?
(i.e. to clarify: I'm confused why this line _isn't_ doing variable interpolation - sorry for not being clearer in my earlier comment.)

Shou: Here we take advantage of Angular's string variable processing, notice how we put `iconContents` on the `$rootScope`. We don't have to do that here, but it's generally safer to do so because it means we don't have to escape any characters.

lurch: Apologies for being super-dense (probably due to me having zero knowledge of Angular!), but I'm still struggling to understand why this line doesn't need string-interpolation, but lines 88, 96, etc. do :confused:

jviotti:
I'll merge so I can start the release process, but lets fix any comments in another PR :+1:

tests/gui/components/svg-icon.spec.js
Shou: We only need to do it the Angular way on L70, every other use isn't necessary and we can use backtick string-interpolation, either is fine outside of L70.

lurch: Aha! It's the "either is fine" that I wasn't understanding. Thanks @Shou :)

lurch:
@jhermsmeier Does the screenshot in your first message show the behaviour before or after this PR? If it's the behaviour from before, could you also show an equivalent screenshot from after this PR, please? :)

tests/gui/components/svg-icon.spec.js
lurch: (out of curiosity, should this sort of thing be picked up by the linter?)

Shou: I remember it used to be, maybe the change to StandardJS removed it.

jviotti: @Shou @lurch Can you help me find out what's the rule that's causing this? We should definitely enable it.

jviotti: See https://github.com/resin-io/etcher/pull/1679

PR #1664, opened 2017-08-03
chore: follow standardjs guidelines

jviotti:
- Extend the `standard` ESLint configuration
- Remove ESLint rules that are defined in the `standard` configuration
- Get rid of semi-colons

See: https://github.com/resin-io/etcher/pull/1657


lurch:
Hmm? How come all the semi-colons have been removed? I guess this means that pretty much every open PR will need to be rebased?

jviotti:
Yeah, that's the disadvantage, but we can get over that. Semi colons are disallowed in standard JS.

PR #1666, opened 2017-08-03
chore: add a publish-all make target

jviotti:
The CI servers currently call `installers-all` to deploy snapshot
builds, however that target builds the installers, but doesn't call the
phony rules that actually publishes them.


jviotti:
I'll merge this so we can start getting snapshot builds for other PRs, but happy to address any comment in another PR.

PR #1669, opened 2017-08-04
feat(GUI): use relative units in flash speed

Shou:
We utilize the relative unit functionality in the flashing speed status.

Changelog-Entry: Use relative units in the flashing speed status.

### DIFF
--- a/lib/shared/models/flash-state.js
+++ b/lib/shared/models/flash-state.js
@@ -114,25 +114,25 @@ exports.unsetFlashingFlag = (results) => {
  * });
  */
 exports.setProgressState = (state) => {
+  const pickedState = _.pick(state, [
+    'type',
+    'percentage',
+    'eta',
+    'speed'
+  ])
+
+  const humanSpeed = units.bytesToClosestUnit(state.speed, '/s')
+
+  const data = _.assign({}, pickedState, {
+    humanSpeed,
+    percentage: _.isFinite(state.percentage)
+      ? Math.floor(state.percentage)
+      : state.percentage
+  })
+
   store.dispatch({
     type: store.Actions.SET_FLASH_STATE,
-    data: {
-      type: state.type,
-      percentage: _.isNumber(state.percentage) && !_.isNaN(state.percentage)
-        ? Math.floor(state.percentage)
-        : state.percentage,
-      eta: state.eta,
-
-      speed: _.attempt(() => {
-        if (_.isNumber(state.speed) && !_.isNaN(state.speed)) {
-          // Preserve only two decimal places
-          const PRECISION = 2
-          return _.round(units.bytesToMegabytes(state.speed), PRECISION)
-        }
-
-        return null
-      })
-    }
+    data
   })
 }
 
--- a/lib/shared/store.js
+++ b/lib/shared/store.js
@@ -60,7 +60,8 @@ const flashStateNoNilFields = [
   'type',
   'percentage',
   'eta',
-  'speed'
+  'speed',
+  'humanSpeed'
 ]
 
 /**
@@ -87,7 +88,8 @@ const DEFAULT_STATE = Immutable.fromJS({
   flashResults: {},
   flashState: {
     percentage: 0,
-    speed: 0
+    speed: 0,
+    humanSpeed: ''
   },
   settings: {
     unsafeMode: false,
--- a/lib/shared/units.js
+++ b/lib/shared/units.js
@@ -21,17 +21,6 @@ const _ = require('lodash')
 const prettyBytes = require('pretty-bytes')
 
 /**
- * @summary Megabyte to byte ratio
- * @constant
- * @private
- * @type {Number}
- *
- * @description
- * 1 MB = 1e+6 B
- */
-const MEGABYTE_TO_BYTE_RATIO = 1e+6
-
-/**
  * @summary Milliseconds in a day
  * @constant
  * @private
@@ -43,35 +32,21 @@ const MEGABYTE_TO_BYTE_RATIO = 1e+6
 const MILLISECONDS_IN_A_DAY = 86400000
 
 /**
- * @summary Convert bytes to megabytes
- * @function
- * @public
- *
- * @param {Number} bytes - bytes
- * @returns {Number} megabytes
- *
- * @example
- * const result = units.bytesToMegabytes(7801405440);
- */
-exports.bytesToMegabytes = (bytes) => {
-  return bytes / MEGABYTE_TO_BYTE_RATIO
-}
-
-/**
  * @summary Convert bytes to most appropriate unit string
  * @function
  * @public
  *
  * @param {Number} bytes - bytes
+ * @param {String} [suffix=''] - suffix string e.g. '/s'
  * @returns {String} size and unit string
  *
  * @example
- * const humanReadable = units.bytesToClosestUnit(7801405440);
- * > '7.8 GB'
+ * const humanReadable = units.bytesToClosestUnit(7801405440, '/s');
+ * > '7.8 GB/s'
  */
-exports.bytesToClosestUnit = (bytes) => {
-  if (_.isNumber(bytes)) {
-    return prettyBytes(bytes)
+exports.bytesToClosestUnit = (bytes, suffix = '') => {
+  if (_.isFinite(bytes)) {
+    return prettyBytes(bytes) + suffix
   }
 
   return null
--- a/tests/shared/models/flash-state.spec.js
+++ b/tests/shared/models/flash-state.spec.js
@@ -26,7 +26,7 @@ describe('Model: flashState', function () {
 
   describe('flashState', function () {
     describe('.resetState()', function () {
-      it('should be able to reset the progress state', function () {
+      it('should be able to reset the progress state while flashing', function () {
         flashState.setFlashingFlag()
         flashState.setProgressState({
           type: 'write',
@@ -39,11 +39,12 @@ describe('Model: flashState', function () {
 
         m.chai.expect(flashState.getFlashState()).to.deep.equal({
           percentage: 0,
-          speed: 0
+          speed: 0,
+          humanSpeed: ''
         })
       })
 
-      it('should be able to reset the progress state', function () {
+      it('should be able to reset the progress state after flashing', function () {
         flashState.unsetFlashingFlag({
           cancelled: false,
           sourceChecksum: '1234'
@@ -234,6 +235,23 @@ describe('Model: flashState', function () {
         }).to.not.throw('Missing flash fields: speed')
       })
 
+      it('should set the .humanSpeed field', function () {
+        const state = {
+          type: 'write',
+          percentage: 50,
+          eta: 15,
+          speed: 10000000
+        }
+
+        flashState.setFlashingFlag()
+        flashState.setProgressState(state)
+
+        // eslint-disable-next-line lodash/prefer-lodash-method
+        m.chai.expect(flashState.getFlashState()).to.deep.equal(Object.assign({}, state, {
+          humanSpeed: '10 MB/s'
+        }))
+      })
+
       it('should floor the percentage number', function () {
         flashState.setFlashingFlag()
         flashState.setProgressState({
@@ -244,6 +262,16 @@ describe('Model: flashState', function () {
         })
 
         m.chai.expect(flashState.getFlashState().percentage).to.equal(50)
+      })
+
+      it('should check for path existence and whether undefined', function () {
+        flashState.setFlashingFlag()
+        m.chai.expect(function () {
+          flashState.setProgressState({
+            type: undefined,
+            percentage: undefined
+          })
+        }).to.throw('Missing flash fields: type, percentage, eta, speed, humanSpeed')
       })
     })
 
@@ -268,7 +296,8 @@ describe('Model: flashState', function () {
         const currentFlashState = flashState.getFlashState()
         m.chai.expect(currentFlashState).to.deep.equal({
           percentage: 0,
-          speed: 0
+          speed: 0,
+          humanSpeed: ''
         })
       })
 
@@ -283,7 +312,11 @@ describe('Model: flashState', function () {
         flashState.setFlashingFlag()
         flashState.setProgressState(state)
         const currentFlashState = flashState.getFlashState()
-        m.chai.expect(currentFlashState).to.deep.equal(state)
+
+        // eslint-disable-next-line lodash/prefer-lodash-method
+        m.chai.expect(currentFlashState).to.deep.equal(Object.assign({
+          humanSpeed: '0 B/s'
+        }, state))
       })
     })
 
@@ -378,7 +411,8 @@ describe('Model: flashState', function () {
 
         m.chai.expect(flashState.getFlashState()).to.not.deep.equal({
           percentage: 0,
-          speed: 0
+          speed: 0,
+          humanSpeed: '0 B/s'
         })
 
         flashState.unsetFlashingFlag({
@@ -388,7 +422,8 @@ describe('Model: flashState', function () {
 
         m.chai.expect(flashState.getFlashState()).to.deep.equal({
           percentage: 0,
-          speed: 0
+          speed: 0,
+          humanSpeed: ''
         })
       })
 
--- a/lib/gui/app/app.js
+++ b/lib/gui/app/app.js
@@ -189,7 +189,7 @@ app.run(() => {
     // `undefined` values.
     analytics.logDebug(
       `Progress (${currentFlashState.type}): ` +
-      `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s ` +
+      `${currentFlashState.percentage}% at ${currentFlashState.humanSpeed} ` +
       `(eta ${currentFlashState.eta}s)`
     )
 
--- a/lib/gui/app/pages/main/templates/main.tpl.html
+++ b/lib/gui/app/pages/main/templates/main.tpl.html
@@ -103,7 +103,7 @@
         </progress-button>
 
         <p class="step-footer step-footer-split" ng-if="main.state.getFlashState().speed && main.state.getFlashState().percentage != 100">
-          <span ng-bind="main.state.getFlashState().speed.toFixed(2) + ' MB/s'"></span>
+          <span ng-bind="main.state.getFlashState().humanSpeed"></span>
           <span>ETA: {{ main.state.getFlashState().eta | secondsToDate | amDateFormat:'m[m]ss[s]' }}</span>
         </p>
       </div>
--- a/tests/shared/units.spec.js
+++ b/tests/shared/units.spec.js
@@ -51,11 +51,4 @@ describe('Shared: Units', function () {
       m.chai.expect(units.bytesToClosestUnit(999)).to.equal('999 B')
     })
   })
-
-  describe('.bytesToMegabytes()', function () {
-    it('should convert bytes to megabytes', function () {
-      m.chai.expect(units.bytesToMegabytes(1.2e+7)).to.equal(12)
-      m.chai.expect(units.bytesToMegabytes(332000)).to.equal(0.332)
-    })
-  })
 })


lib/shared/models/flash-state.js
```diff
@@ -124,8 +124,7 @@ exports.setProgressState = (state) => {
       speed: _.attempt(() => {
         if (_.isNumber(state.speed) && !_.isNaN(state.speed)) {
           // Preserve only two decimal places
```
jviotti: I guess this comment is not relevant anymore, right?

lib/shared/units.js
```diff
@@ -43,21 +32,6 @@ const MEGABYTE_TO_BYTE_RATIO = 1e+6
...
- * @function
- * @public
- *
- * @param {Number} bytes - bytes
- * @returns {Number} megabytes
- *
- * @example
- * const result = units.bytesToMegabytes(7801405440);
- */
-exports.bytesToMegabytes = (bytes) => {
```
jviotti: Does this function has corresponding tests? Should they be removed as well?

jviotti:
Looks like there are some failing tests that are not really related to this change. Any clues?

Shou:
This passes tests locally now, so should be good to go once they're green here.

lib/shared/models/flash-state.js
```diff
@@ -123,9 +123,7 @@ exports.setProgressState = (state) => {
 
       speed: _.attempt(() => {
         if (_.isNumber(state.speed) && !_.isNaN(state.speed)) {
-          // Preserve only two decimal places
-          const PRECISION = 2
-          return _.round(units.bytesToMegabytes(state.speed), PRECISION)
+          return units.bytesToClosestUnit(state.speed)
```
lurch: Rather than appending "/s" elsewhere, perhaps it'd make more sense to do that here?
I believe this would also mean that the store would be initialised to `speed: '0 B/s'` which is probably a more meaningful 'speed' value than `0 B`? ;-)

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -95,7 +95,7 @@
         </progress-button>
 
         <p class="step-footer step-footer-split" ng-show="main.state.getFlashState().speed && main.state.getFlashState().percentage != 100">
```
lurch: Does the `main.state.getFlashState().speed` here also need changing to `main.state.getFlashState().speed != '0 B'`, similar to the change in `flash.js`?

tests/shared/models/flash-state.spec.js
```diff
@@ -271,7 +272,9 @@ describe('Model: flashState', function () {
         flashState.setFlashingFlag()
         flashState.setProgressState(state)
         const currentFlashState = flashState.getFlashState()
-        m.chai.expect(currentFlashState).to.deep.equal(state)
+        m.chai.expect(currentFlashState).to.deep.equal(_.assign(state, {
+          speed: `${state.speed} B`
+        }))
```
lurch: Could we also add some tests where we check other speed values get the correct suffix appended? (e.g. KB/s, MB/s)

jhermsmeier: I don't think we need to be testing `pretty-bytes` in Etcher... ;)

lurch: The suggestion here wasn't to test `pretty-bytes` itself, but to test that our code is correctly _using_ `pretty-bytes` the way we expect it to :stuck_out_tongue_winking_eye:

lib/gui/pages/main/templates/main.tpl.html
Shou: Yup! Done.

lib/shared/models/flash-state.js
Shou: Done this now.

lib/gui/app.js
```diff
@@ -189,9 +189,10 @@ app.run(() => {
     // property being set, and the flashing actually starting, which
     // might cause some non-sense flashing state logs including
     // `undefined` values.
+
     analytics.logDebug(
-      `Progress (${currentFlashState.type}): ` +
-      `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s ` +
+      `Progress (${currentFlashState.type}):` +
```
lurch: Was the space after the colon deliberately removed?

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -99,8 +99,8 @@
             <span ng-bind="flash.getProgressButtonLabel()"></span>
         </progress-button>
 
-        <p class="step-footer step-footer-split" ng-show="main.state.getFlashState().speed && main.state.getFlashState().percentage != 100">
-          <span ng-bind="main.state.getFlashState().speed.toFixed(2) + ' MB/s'"></span>
+        <p class="step-footer step-footer-split" ng-show="main.state.getFlashState().speed != '0 B/s' && main.state.getFlashState().percentage != 100">
```
jhermsmeier: I think we should still display the speed, even if it's zero, and even if it's at 100%. With compressed images, both cases can happen; i.e. speed will be zero if a lot of blocks are skipped due to blockmapping, or there will still be a non-zero write speed at 100% if a highly compressed image (like bz or gz) is being flashed due to the expansion.

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -99,8 +99,8 @@
...
+          <span ng-bind="main.state.getFlashState().speed"></span>
```
lurch: This looks so much neater now :+1:

lib/shared/models/flash-state.js
```diff
@@ -123,9 +123,7 @@ exports.setProgressState = (state) => {
 
       speed: _.attempt(() => {
         if (_.isNumber(state.speed) && !_.isNaN(state.speed)) {
-          // Preserve only two decimal places
-          const PRECISION = 2
-          return _.round(units.bytesToMegabytes(state.speed), PRECISION)
+          return units.bytesToClosestUnit(state.speed) + '/s'
```
jhermsmeier: No need to `_.attempt()` here anymore, since it's being checked whether it's a number. Also might want to use `.isFinite()` here instead of `!_.isNaN()`

tests/shared/models/flash-state.spec.js
```diff
@@ -39,7 +40,7 @@ describe('Model: flashState', function () {
 
         m.chai.expect(flashState.getFlashState()).to.deep.equal({
           percentage: 0,
-          speed: 0
+          speed: '0 B'
```
lurch: Shouldn't these `0 B` now be `0 B/s` as in `progress-status.spec.js` ?

lib/gui/pages/main/templates/main.tpl.html
Shou: We could set speed to an empty string or `'None'` initially and check for that instead

lib/shared/models/flash-state.js
Shou: Good call

lib/gui/modules/progress-status.js
```diff
@@ -41,7 +41,7 @@ exports.fromFlashState = (state) => {
   const shouldValidate = settings.get('validateWriteOnSuccess')
   const shouldUnmount = settings.get('unmountOnSuccess')
 
-  if (state.percentage === utils.PERCENTAGE_MINIMUM && !state.speed) {
+  if (state.percentage === utils.PERCENTAGE_MINIMUM && state.speed === '0 B/s') {
```
jhermsmeier: I think we should leave `state.speed` be, and keep it a number, but add `state.humanSpeed` or something that contains the speed as a human readable string, because we might want to use the `state.speed` in other places as a number.

tests/shared/models/flash-state.spec.js
lurch: I still think that it'd be nice to have at least one test that checks a value other than '0 B/s' (i.e. testing setProgressState with a state.speed other than 0).

lib/shared/store.js
```diff
@@ -212,9 +213,12 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         })
       }
 
-      if (_.isNil(action.data.speed)) {
+      if (_.some([
```
lurch: I think this would be more readable without using `_.some`, especially as you're already using other logical-operators below.

lib/gui/modules/progress-status.js
```diff
@@ -41,7 +41,7 @@ exports.fromFlashState = (state) => {
   const shouldValidate = settings.get('validateWriteOnSuccess')
   const shouldUnmount = settings.get('unmountOnSuccess')
 
-  if (state.percentage === utils.PERCENTAGE_MINIMUM && state.speed == null) {
+  if (state.percentage === utils.PERCENTAGE_MINIMUM && state.speed === null) {
```
jhermsmeier: Is `state.speed` going to be `null` , or could it also be undefined?

lib/shared/store.js
```diff
@@ -215,7 +215,7 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
 
       if (_.some([
         (!_.isNumber(action.data.speed) || !_.isFinite(action.data.speed)) && action.data.speed !== null,
-        action.data.speed === undefined
+        _.isUndefined(action.data.speed)
```
jhermsmeier: You should be able to just do `_.isNil(action.data.speed)` instead of checking against null & undefined here

lib/shared/store.js
```diff
@@ -212,9 +213,12 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         })
       }
 
-      if (_.isNil(action.data.speed)) {
+      if (_.some([
+        (!_.isNumber(action.data.speed) || !_.isFinite(action.data.speed)) && action.data.speed !== null,
```
lurch: And IMHO this should be rearranged to:
```
action.data.speed !== null && (!_.isNumber(action.data.speed) || !_.isFinite(action.data.speed))
```
because if `action.data.speed` _is_ null, then there's no point checking `_.isNumber` or `_.isFinite` ;-)

lurch:
Sorry to sound like a broken record, but I still think more unit-tests would be great here e.g. checking that `setProgressState` adds the `.humanSpeed` attribute.

lib/shared/store.js
Shou: Fixed these two now

Shou:
Will add some tests now

lib/shared/store.js
```diff
@@ -212,12 +213,18 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
       }
 
-      if (_.isNil(action.data.speed)) {
+      if (_.isUndefined(action.data.speed)) {
         throw errors.createError({
           title: 'Missing state speed'
         })
       }
 
+      if (action.data.speed !== null && (!_.isNumber(action.data.speed) || !_.isFinite(action.data.speed))) {
```
jhermsmeier: You can just do `!_.isNil(action.data.speed) || !_.isFinite(action.data.speed)` here, values that are not a number aren't finite either

lib/shared/units.js
```diff
@@ -70,7 +44,7 @@ exports.bytesToMegabytes = (bytes) => {
  * > '7.8 GB'
  */
 exports.bytesToClosestUnit = (bytes) => {
-  if (_.isNumber(bytes)) {
+  if (_.isNumber(bytes) && _.isFinite(bytes)) {
```
jhermsmeier: Same here, no need to check whether it's a number, as isFinite() checks that

lib/gui/modules/progress-status.js
Shou: It should never be `undefined`, but its default state is `null` as `0` doesn't necessarily indicate that progress isn't happening as you highlighted

lib/shared/units.js
Shou: Weirdly enough I thought to check if `_.isFinite` could replace the null-check but not this obvious case...

lib/shared/store.js
jhermsmeier: Oh, actually only isFinite() will suffice I realized

Shou:
Here we go, added a test. The old tests have already been retrofitted to check for `.humanSpeed` e.g. when resetting.

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -99,8 +99,8 @@
             <span ng-bind="flash.getProgressButtonLabel()"></span>
         </progress-button>
 
-        <p class="step-footer step-footer-split" ng-show="main.state.getFlashState().speed && main.state.getFlashState().percentage != 100">
-          <span ng-bind="main.state.getFlashState().speed.toFixed(2) + ' MB/s'"></span>
+        <p class="step-footer step-footer-split" ng-show="main.state.getFlashState().speed != undefined && main.state.getFlashState().percentage != 100">
```
jhermsmeier: The speed should still display at 100% (because of compressed images etc)

Shou: Done

jhermsmeier:
I gave this a spin, and it stops updating the speed etc. once it's gotten to the "Finishing..." state

jviotti:
@jhermsmeier @Shou I might be wrong, but it seems that this PR is getting complex (took a week of Benedict's work). I wonder if we should be using @Shou 's time in much more important tasks (like polishing multi-writes and integrating resin-components). This is a nicety we can pause for now, unless its very close to be merged.

Shou:
@jhermsmeier isn't that intentional? At "Finishing..." it's at 100% progress so as far as I know nothing should be updating at that step.

@jviotti I was on support so it wasn't fully a week but I'll put this to rest if it isn't done now.

jhermsmeier:
@Shou as mentioned, there's the state you end up in if you have a – let's say – 300MB compressed imaged that expands to ~2GB or so. And with that, there's the issue that for bz2 and gzip for example, we can't determine the uncompressed size, and hence we're measuring the progress by the source. This means that the source will yield 100% progress, while the last reads from the source actually expand to a couple hundred MB that need to be written – meaning that 100% progress != done.

My thinking is that we should still keep updating the speed etc. while displaying "Finishing...", because that is what it's doing (finishing the writing), and because otherwise there'll be no activity in the UI indicating that there's still something going on – possibly making people think it's stuck when it isn't.

lurch:
I wonder if the complexities / subtleties of this PR mean that it should be split into two? One that _just_ adds the `humanSpeed` property and then updates the UI to use that (with no other logic changes), and a separate one that updates the logic of what gets displayed with a 0 speed and/or 100% progress as @jhermsmeier is asking about?

lib/shared/models/flash-state.js
```diff
@@ -114,23 +114,23 @@ exports.unsetFlashingFlag = (results) => {
...
 exports.setProgressState = (state) => {
+  const pickedState = _.pick(state, [
+    'type',
+    'percentage',
+    'eta',
+    'speed'
+  ])
+
+  const humanSpeedUnits = units.bytesToClosestUnit(state.speed)
+  const humanSpeed = humanSpeedUnits ? `${humanSpeedUnits}/s` : null
```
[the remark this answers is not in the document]
Shou: That's a good idea, I've done this and will push it shortly

lib/shared/store.js
```diff
@@ -41,7 +41,8 @@ const DEFAULT_STATE = Immutable.fromJS({
   flashResults: {},
   flashState: {
     percentage: 0,
-    speed: 0
+    speed: null,
+    humanSpeed: ''
```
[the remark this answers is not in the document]
Shou: I'm using empty string to simplify the logic here, and since empty string is the `null` of its respective domain I it makes sense. The only reason we need `null` and can't only use `0` in `speed` is because zero might actually be hit with certain devices during flashing so it can't represent a value responsible for removing the flashing status subtitle with the speed and ETA... maybe we can use the flashing flag instead, however.

@jhermsmeier when we hit `0` speed I assume we should replace the status with something more helpful than simply "0 B/s" – e.g. "Unknown speed and ETA" or "Speed and ETA unavailable", whatever is more succinct to the user.

tests/gui/modules/progress-status.spec.js
```diff
@@ -18,29 +18,30 @@ describe('Browser: progressStatus', function () {
       settings.set('validateWriteOnSuccess', true)
     })
 
-    it('should report 0% if percentage == 0 but speed != 0', function () {
+    it('should report 0% if percentage == 0 but speed != null', function () {
+      this.state.speed = 0
```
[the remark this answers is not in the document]
Shou: Fixed that now

lib/shared/store.js
```diff
@@ -212,12 +213,18 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
       }
 
-      if (_.isNil(action.data.speed)) {
+      if (_.isUndefined(action.data.speed)) {
         throw errors.createError({
           title: 'Missing state speed'
         })
       }
 
+      if (action.data.speed !== null && !_.isFinite(action.data.speed)) {
```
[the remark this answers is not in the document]
Shou: Actually, we could make it an else to the `null` check then it'll just be the `!_.isFinite` check

lib/shared/store.js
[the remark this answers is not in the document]
lurch: I guess I should clarify that I was only asking about the `humanSpeed` property - in `lib/shared/store.js` it defaults to `''` whereas in `lib/shared/models/flash-state.js` it defaults to `null`, and I was asking if they should be using the same default value for consistency purposes.

Shou:
Let me know what you think of this now. I'm gonna make another PR for the logic behind `speed` and displaying its metadata, since it's a bit more complex than this PR is, plus potentially another PR.

(accidentally closed this with a commit)

jhermsmeier:
Looks like this will need yet some more updating @Shou

lib/shared/units.js
```diff
@@ -42,36 +31,22 @@ const MEGABYTE_TO_BYTE_RATIO = 1e+6
...
  */
-exports.bytesToClosestUnit = (bytes) => {
-  if (_.isNumber(bytes)) {
-    return prettyBytes(bytes)
+exports.bytesToClosestUnit = (bytes, suffix = '') => {
+  if (_.isFinite(bytes)) {
+    return prettyBytes(bytes) + suffix
   }
 
   return null
```
jhermsmeier: This method should be consistent in the data type returned;
Let's make this return an empty string in this case – that's what the flash state tests are apparently expecting anyways. Which bears the question: Why aren't they failing?

lib/shared/models/flash-state.js
```diff
@@ -114,25 +114,25 @@ exports.unsetFlashingFlag = (results) => {
...
+  ])
+
+  const humanSpeed = units.bytesToClosestUnit(state.speed, '/s')
+
+  const data = _.assign({}, pickedState, {
+    humanSpeed,
+    percentage: _.isFinite(state.percentage)
+      ? Math.floor(state.percentage)
+      : state.percentage
+  })
```
jhermsmeier: I think this is overcomplicating things heavily for no apparent reason. Why not just do the following?

```js
state.humanSpeed = units.bytesToClosestUnit(state.speed, '/s')

store.dispatch({
  type: store.Actions.SET_FLASH_STATE,
  data: state
})
```

tests/shared/models/flash-state.spec.js
```diff
@@ -378,7 +411,8 @@ describe('Model: flashState', function () {
 
         m.chai.expect(flashState.getFlashState()).to.not.deep.equal({
           percentage: 0,
-          speed: 0
+          speed: 0,
+          humanSpeed: '0 B/s'
```
lurch: Apologies if this is a stupid question, but why do some unit-tests have:
```
          speed: 0,
          humanSpeed: ''
```
and other unit-tests have:
```
          speed: 0,
          humanSpeed: '0 B/s'
```
?

lib/shared/models/flash-state.js
lurch: I'd assume that @Shou is trying to preserve the "const-ness" of `state` by creating a new object rather than modifying `state` in place? \*shrug\*

BTW do any of the unit-tests check what happens when `_.isFinite(state.percentage)` returns false?

And just to be pedantic I guess there's no need for `percentage` in `pickedState` because `data.percentage` is being set directly from `state.percentage` ;)

lib/shared/units.js
lurch: > Which bears the question: Why aren't they failing?

Perhaps that implies that the `bytesToClosestUnit` function needs more comprehensive unit-tests?

Issue #1532, opened 2017-06-20
Integrate pipage in the writer


jviotti:
See https://github.com/resin-io-modules/pipage

lurch:
Should there be a 'linked' issue at https://github.com/resin-io-modules/etcher-image-write/issues ?

jviotti:
Up to @jhermsmeier. IMHO, is nicer to unify tickets in the main Etcher project, since its easier to search/tag everything on a single place, but I have no strong opinions neither.

jhermsmeier:
@lurch the pipage stuff is happening in Etcher, and `etcher-image-write` won't be used anymore at that point, hence it'd make sense to only track it here.

lurch:
Ah, so the title should be changed from `Integrate pipage in etcher-image-write` to `Integrate pipage in Etcher` ? ;-) Or even `Replace etcher-image-write with pipage` ?

jviotti:
I rephrased etcher-image-write to "the writer."

jhermsmeier:
Done in #1671

Issue #1571, opened 2017-07-05
Etcher.exe creates/runs batch script every 2 seconds indefinitely, at idle


Sawaba:
- Version 1.0.0
- Windows 10 64-bit
- Burning images not necessary to recreate error
- Nothing in DevTools that helps

My hard drive was filling up, so I did some analysis and found an obscene number of temp files in %UserProfile%\AppData\Local\Temp. They're all the same exact file, but with a different random name, according to the convention:
`tmp-<pid><randomchar x12>.bat`

Every 2 seconds, I see this Etcher process create a cmd process to run this batch script. It is the second of three subthreads of Etcher.exe, run with --renderer.

![image](https://user-images.githubusercontent.com/1382808/27844912-7e98a7e2-60f6-11e7-9fb4-0a8e6349130e.png)

I haven't looked too much into what the script does, but I found a copy of it uploaded to both [Pastebin ](https://pastebin.com/U8NuJRY1) and [VirusTotal](https://virustotal.com/en/file/762faf35eaefadee795758666e968f187c706951e9c4eb83ba4f028756eb2701/analysis/).

Recreating the issue is as simple as just executing Etcher.exe and letting it run.

It has created over 1000 files in the time it took me to write this.
![image](https://user-images.githubusercontent.com/1382808/27844971-3f4cf380-60f7-11e7-9565-aa563158dcfc.png)

![image](https://user-images.githubusercontent.com/1382808/27844982-63767a92-60f7-11e7-87ba-38829fd159fe.png)

It should be noted that this doesn't noticeably impact the performance of Etcher - I've created a few images in the last day or two. The only reason this came to my attention is because it was filling up my disk. At this rate, it consumes 354MB of disk daily. That's less of a concern than the 43,200 files created daily. At about 100,000 files, Windows Explorer starts to choke when you try to browse a directory.

Again, this is a great piece of software - imaging stuff is so much easier. Just a bit alarmed to see this churning out more than just images ;)

jhermsmeier:
Hey @Sawaba, thank you for bringing this up! This is a beautiful bug report :)

The culprit here is [resin-io-modules/drivelist](https://github.com/resin-io-modules/drivelist), which Etcher uses to scan for storage devices. It does look a lot like there's something fishy going on due to the location and name of the files; but the only thing that happens is that drivelist creates a script to get the drives, as that was the easiest way to get a list of disks with sizes etc. at the time, without using native APIs.

[...]

Issue #673, opened 2016-09-01
Add 0's at the beginning of Beagle images when using BMAPs


jadonk:
- **Etcher version: master**
- **Operating system and architecture: all**
- **Do you see any meaningful error information on DevTools? N/A**

There are sometimes issues with the ROM bootloader on Beagle boards when you don't 0 out the first 1MB of the SD cards before writing the images using BMAPs. This is because the bootloader expects this and the BMAP tool sees this as sparse space. If you just write 0s to the first 1MB before writing the image, it'll boot fine every time.

alexandrosm:
@jviotti how about adding an offset config parameter?

jviotti:
@jadonk What is your current workaround? Do users flashing the image with `bmaptool` frequently encounter this issue? If the zeroes are explicitly being written in the first megabyte then `bmaptool` shouldn't interpret this as sparse space. 

This sounds like the something that should be fixed from the image generation side, since it'd be a special code path in Etcher, but I'm open to further discussion.

@alexandrosm You mean an offset option to determine how much to zero out?

alexandrosm:
@jviotti exactly!

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

jadonk:
It could be fixed as a bug in bmaptool, good point. The writing of zeros before writing with bmaptool is the work-around. I'll raise a bug there.

RobertCNelson:
@jviotti the current workaround was to stop publishing "*.bmap" files and just recommended users use etcher.io. ;)  I'll need to retest, as it was a few quarters ago, but the gist was:

On boards like the omap3-xm (dm3730), which don't officially support "raw boot", the bootloader was found in the first fat partition.  If i had a microSD card, that was flashed for a bbb (am335x which support raw boot, (the bootrom was 2 generations later then the dm3730's..)).  So if i then used bmaptool to then flash that same card for the omap3-xm, it would fail to boot.  However if i dd'ed the image directly to the microSD, it would boot fine.  At the time since Etcher.io just came out, i then recommended users use that instead.

My theory at the time, the dm3730 was tripping over something in the 'raw boot' sectors on the microSD.  Thus clearing out the first 1MB should fix that.

Regards,

alexandrosm:
I think the solution is actually to insert the zeroes into your image and
mark them in the bmap file as an area that has to be written. This can
probably be done with manual editing of the bmap file. However the easiest
[...]

PR #1671, opened 2017-08-04
feat(writer): Implement streaming pipelines

jhermsmeier:
This implements a new way of image write streaming under use of [pipage] and [blockmap], which paves the way for a few things like using network locations as sources, and imaging of storage devices (aka backups). As it allows for mutation of the streaming pipeline while it's writing, it also facilitates the development of dynamic block-mapping.

[pipage]: https://github.com/resin-io-modules/pipage
[blockmap]: https://github.com/resin-io-modules/blockmap

Change-Type: minor
Connects To: https://github.com/resin-io/etcher/issues/1532, https://github.com/resin-io/etcher/issues/1806


lib/writer/checksum-stream.js
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * @summary Create & pipe to the Hash streams
+   * @private
+   * @param {String[]} algorithm - hash algorithm
+   * @returns {Stream}
+   * @example
+   * const hash = this._createHash(algorithm)
+   */
+  _createHash (algorithm) {
+    let hash = null
```
jviotti: We can avoid the `let` by using `_.attempt`:

```js
const hash = _.attempt(getHashStream, algorithm);
if (_.isError(hash)) {
    ...
}
```

lib/writer/checksum-stream.js
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * @summary Check for all checksums to have been calculated
+     * @private
+     * @example
+     * hash.once('end', check)
+     */
+    const check = () => {
+      if (_.keys(this.results).length === this.algorithms.length) {
+        this.emit('checksum', _.assign({}, this.results))
```
jviotti: Should we replace `_.assign({}, this.results)` with `_.clone(this.results)`?

lib/writer/checksum-stream.js
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+
+    hash.once('error', (error) => {
+      return this.emit('error', error)
+    })
+
+    if (algorithm === 'crc32') {
+      hash.once('end', function () {
+        self.results[algorithm] = this.digest('hex')
```
jviotti: Could we do `hash.digest()`, and `hash.read()` below, and avoid the `self` and `function` altogether?

jhermsmeier: Yeah, not sure anymore why I went down that road here.

lib/writer/checksum-stream.js
jhermsmeier: Ah yes, good catch!

lib/writer/checksum-stream.js
jhermsmeier: Done!

lib/writer/index.js
```diff
@@ -0,0 +1,306 @@
...
+      bytesWritten: this.bytesWritten,
+      checksum: this.checksum
+    })
+  }
+
+  /**
+   * @summary Creates a write pipeline from given options
+   * @private
+   * @param {Object} options - options
+   * @param {Object} options.image - source image
```
Shou: Maybe we can generalize this to `options.source` and `options.target` in the `_.create*Pipeline`s?

lib/writer/block-read-stream.js
```diff
@@ -0,0 +1,218 @@
...
+ * limitations under the License.
+ */
+
+'use strict'
+
+const stream = require('readable-stream')
+const fs = require('fs')
+const debug = require('debug')('block-read-stream')
+
+/* eslint-disable lodash/prefer-lodash-method */
```
jviotti: Can we have a nested ESLint configuration (in `lib/writer/.eslintrc.yml`) that disables certain rules in the name of supporting low level code?

lib/writer/block-write-stream.js
```diff
@@ -0,0 +1,266 @@
...
+   * @private
+   * @param {Error} [error] - error
+   * @param {Function} done - callback
+   * @example
+   * stream.destroy()
+   */
+  _destroy (error, done) {
+    debug('_destroy', error)
+
+    // TODO: Figure out how the error handling works here
```
jviotti: Does this need to be addressed?

jhermsmeier: Yup, mended!

lib/writer/index.js
jhermsmeier: Yeah, I was planning to do this in the course of using network locations & devices as sources as that'll require a generalization of these anyways. I just kept it as is for now to stay compatible with the wide rest of `image-stream` etc.

lib/cli/diskpart.js
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @param {String} extension - temporary file extension
+ * @returns {String} filename
+ *
+ * @example
+ * const filename = tmpFilename('.sh');
+ */
+const tmpFilename = (extension) => {
+  const random = crypto.randomBytes(TMP_RANDOM_BYTES).toString('hex')
+  const filename = `etcher-diskpart-${random}${extension}`
```
jviotti: I wonder if we should use [`node-tmp`](https://github.com/raszi/node-tmp) instead of rolling our own unique file name generation code, given that I believe we have it on the dependency tree already.

lib/writer/block-write-stream.js
```diff
@@ -0,0 +1,267 @@
...
+ * limitations under the License.
+ */
+
+'use strict'
+
+const stream = require('readable-stream')
+const fs = require('fs')
+const debug = require('debug')('block-write-stream')
+
+/* eslint-disable lodash/prefer-lodash-method */
```
jviotti: Lets create the nested ESLint configuration file, so we don't have to repeat this on every file.

lib/cli/diskpart.js
jhermsmeier: Remember that issue with temporary file spam and running out of file handles on drivelist (https://github.com/resin-io/etcher/issues/1571, https://github.com/resin-io-modules/drivelist/issues/190, https://github.com/resin-io-modules/drivelist/pull/191)? These couple of lines are more resilient than that module, and I'd rather not we use node-tmp anywhere :)

jviotti: Ah, good point. I'll approve then, but it'd be nice to get rid of the custom ESLint comments in various files first (by using a nested ESLint configuration).

jhermsmeier: Yup, doing that in a minute or two :)

lib/writer/block-write-stream.js
jhermsmeier: Added!

lib/cli/diskpart.js
```diff
@@ -0,0 +1,124 @@
...
+  return path.join(os.tmpdir(), filename)
```
lurch: Should this be using the new `CONSTANTS.TMP_DIRECTORY` ?

jhermsmeier: There is no `CONSTANTS.TMP_DIRECTORY` in the cli or anywhere else yet; that might be something to do & consolidate later.

lurch: Ah, I was getting confused with the other PR that I was reviewing at the same time https://github.com/resin-io/etcher/pull/1843/files#diff-0472e5ca31ec513d2fe254fad48d1965R34

lib/writer/block-stream.js
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * @summary BlockStream class
+ * @class
+ */
+class BlockStream extends stream.Transform {
+  /**
+   * @summary BlockStream constructor
+   * @param {Object} [options] - options
+   * @param {Number} [options.blockSize] - block size in bytes
+   * @param {Number} [options.chunkSize] - chunk size in bytes
```
lurch: Perhaps you could clarify the difference between a 'block' and a 'chunk'?

lib/cli/writer.js
```diff
@@ -82,24 +85,20 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
...
-        device: drive.raw,
-        size: drive.size
-      }, {
-        stream: image.stream,
-        size: image.size.original
-      }, {
-        check: options.validateWriteOnSuccess,
-        transform: image.transform,
-        bmap: image.bmap,
-        bytesToZeroOutFromTheBeginning: image.bytesToZeroOutFromTheBeginning
```
lurch: Where's the equivalent handling of this now?

lib/writer/block-write-stream.js
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  /**
+   * @summary Write buffered data before a stream ends
+   * @private
+   * @param {Function} done - callback
+   * @example
+   * // Called by stream internals
+   */
+  _final (done) {
```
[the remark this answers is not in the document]
lurch: Ahh, so perhaps this is what BlockStream is for?

lib/writer/block-write-stream.js
```diff
@@ -0,0 +1,261 @@
...
+    this._firstBlocks = []
+
+    this.open()
+  }
+
+  /**
+   * @summary Internal write handler
+   * @private
+   * @param {Buffer} chunk - chunk buffer
+   * @param {String} encoding - chunk encoding
```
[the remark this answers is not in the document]
jhermsmeier: Yes, buffer encoding refers to character set encodings; we're not writing text here ;)

lib/writer/block-write-stream.js
```diff
@@ -0,0 +1,261 @@
...
+
+    this.bytesRead += chunk.length
+    this.blocksRead += 1
+
+    if (chunk.position == null) {
+      chunk.position = this.position
+    }
+
+    if (!this._flushing && (chunk.position < CHUNK_SIZE)) {
+      this._firstBlocks.push(chunk)
```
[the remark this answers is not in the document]
jhermsmeier: It does only get used once – `_firstBlocks` is what contains the first blocks (which we need to write *last*, because Windows) of the stream, which are then flushed before end; this also obsoletes `bytesToZeroOutFromTheBeginning`

lib/writer/block-write-stream.js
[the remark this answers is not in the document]
jhermsmeier: This is being handled by the block-stream; the block-write-stream will write what it's given (flushed in min. 64K chunks)

lib/writer/index.js
```diff
@@ -0,0 +1,311 @@
...
+
+    // If the image transform is a pass-through,
+    // ignore it to save on the overhead
+    if (image.transform && !isPassThrough) {
+      pipeline.append(image.transform)
+    }
+
+    // If the final size is known precisely and we're not
+    // using block maps, then use the final size for progress
+    if (!image.size.final.estimation && !image.bmap) {
```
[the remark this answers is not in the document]
jhermsmeier: > (Or maybe that is what you're doing by adding the progressStream either before or after the image.transform?)

Yup, that's how it works – depending on which size is used, the relevant stream is observed by the progress-stream.

lib/writer/index.js
```diff
@@ -0,0 +1,311 @@
...
+      debug('write:bmap', blockMap)
+      progressStream = new ProgressStream(progressOptions)
+      pipeline.append(progressStream)
+      pipeline.append(new BlockMap.FilterStream(blockMap))
+    } else {
+      debug('write:blockstream')
+      const checksumStream = new ChecksumStream({
+        algorithms: options.checksumAlgorithms
+      })
+      pipeline.append(checksumStream)
```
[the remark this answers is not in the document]
jhermsmeier: The checksum-stream's transform is a noop if no checksums are specified. Due to discoveries in perf opt recently, the checksum-stream will need a brutal overhaul soon anyways.

lib/writer/index.js
```diff
@@ -0,0 +1,311 @@
...
+      pipeline.bind(checksumStream, 'checksum')
+      pipeline.append(new BlockStream())
```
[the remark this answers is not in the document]
jhermsmeier: It ensures that buffers down the line are aligned to a min block size of 512 bytes

lib/writer/index.js
```diff
@@ -0,0 +1,311 @@
...
+      length: size,
+      time: 500
+    })
+
+    pipeline.append(progressStream)
+
+    if (options.image.bmap) {
+      debug('verify:bmap')
+      const blockMap = BlockMap.parse(options.image.bmap)
+      pipeline.append(new BlockMap.FilterStream(blockMap))
```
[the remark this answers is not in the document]
jhermsmeier: Yes, because bmaps contain their the range checksums; reading it back through that verifies it

lib/writer/progress-stream.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * @class ProgressStream
+ * @public
+ */
+class ProgressStream extends Stream.Transform {
+  /**
+   * @summary ProgressStream constructor
+   * @param {Object} options - options
+   * @param {Number} options.length - expected total
+   * @param {Number} [options.time] - time interval to report progress
```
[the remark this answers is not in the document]
jhermsmeier: This has been kept this way to stay compatible with the `progress-stream` module's API

lib/writer/progress-stream.js
```diff
@@ -0,0 +1,113 @@
...
+   *     console.log( state.percentage.toFixed(0) + '%' )
+   *   })
+   */
+  constructor (options) {
+    super(options)
+
+    this.start = 0
+    this.interval = options.time || DEFAULT_TIME_MS
+    this.timer = null
+    this.meter = speedometer()
```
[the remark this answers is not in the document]
jhermsmeier: See [speedometer](https://github.com/mafintosh/speedometer)'s docs; the time-counting doesn't start until the first delta update.

Regarding read/write speed for compressed sources; that's already been the case so far, so this doesn't change; Sure we should have proper write speed display; probably even both, read & write speeds

package.json
```diff
@@ -47,14 +47,16 @@
...
+    "bluebird-retry": "0.11.0",
     "bootstrap-sass": "3.3.6",
     "chalk": "1.1.3",
     "command-join": "2.0.0",
-    "debug": "2.6.0",
+    "crc32-stream": "2.0.0",
+    "debug": "2.6.8",
     "drivelist": "5.2.4",
     "electron-is-running-in-asar": "1.0.0",
-    "etcher-image-write": "9.1.3",
```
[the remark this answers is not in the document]
jhermsmeier: Indeed!

lib/writer/block-read-stream.js
```diff
@@ -0,0 +1,212 @@
...
+ * @class
+ */
+class BlockReadStream extends stream.Readable {
+  /**
+   * @summary BlockReadStream constructor
+   * @param {Object} [options] - options
+   * @param {Number} [options.fd] - file descriptor
+   * @param {String} [options.path] - file path
+   * @param {String} [options.flags] - file open flags
+   * @param {Number} [options.mode] - file mode
```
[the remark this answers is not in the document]
jhermsmeier: Yeah, most of the time they'll actually be `rs` or `rs+`, unless a `fd` is given. Omitting this capability wouldn't make sense. Also, as usual we're staying close to core the `fs` streams API.

lib/writer/block-read-stream.js
```diff
@@ -0,0 +1,212 @@
...
+   * @param {Number} [options.start] - start offset in bytes
+   * @param {Number} [options.end] - end offset in bytes
+   * @param {Boolean} [options.autoClose] - automatically close the stream on end
+   * @example
+   * new BlockReadStream()
```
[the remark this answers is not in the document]
jhermsmeier: Sure, it'll error, but it is a thing you can do :D

lib/writer/block-read-stream.js
```diff
@@ -0,0 +1,212 @@
...
+
+    const toRead = this.end - this.position
+
+    if (toRead <= 0) {
+      this.push(null)
+      return
+    }
+
+    const length = Math.min(CHUNK_SIZE, Math.max(MIN_CHUNK_SIZE, toRead))
+    const buffer = Buffer.alloc(length)
```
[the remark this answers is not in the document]
jhermsmeier: If you'd use a static buffer, you'd continuously overwrite it with the following reads, resulting in timing-based garbage – whatever would use the buffer would have to do so before the next read kicks off, which is not how streams work

lib/writer/block-read-stream.js
```diff
@@ -0,0 +1,212 @@
...
+   * @summary Open a handle to the file
+   * @private
+   * @example
+   * this.open()
+   */
+  open () {
+    debug('open')
+
+    if (this.fd != null) {
+      this.emit('open', this.fd)
```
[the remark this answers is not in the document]
jhermsmeier: > Should there be code to prevent .open being called twice? 

No, it's made to be called even with an open fd, so that the "open" event will be emitted correctly etc.

> ...and is/should it be permissible to call .open on a stream after you've .closed it? (and if not should that be checked for?)

This is handled by Node core streams (which we're inheriting from) – anything after close/end/destroy will cause "read/write after close" errors

lib/writer/index.js
```diff
@@ -0,0 +1,311 @@
...
+    const progressOptions = {
+      length: image.size.original,
+      time: 500
+    }
+
+    let progressStream = null
+
+    // If the final size is an estimation,
+    // use the original source size for progress metering
+    if (image.size.final.estimation) {
```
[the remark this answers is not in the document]
jhermsmeier: See comment below

lib/writer/block-read-stream.js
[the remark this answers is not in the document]
lurch: Ah right, like I said I wasn't sure if the buffer got copied elsewhere.

lib/writer/block-write-stream.js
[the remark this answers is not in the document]
lurch: Ah, I misunderstood what the code was doing then :-/  So is this a slight change in behaviour then, where previously we'd only write the first 512 bytes after everything else, and now we write the first 64KB after everything else?

I couldn't remember what `bytesToZeroOutFromTheBeginning` was about, so I did a bit of searching and it appears to be a bmap-thing rather than a Windows-thing #673

lib/writer/block-write-stream.js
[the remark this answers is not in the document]
lurch: Ahh, so if we flush chunks bigger than 64K it doesn't matter, and the streams stuff will just handle all that for us?

lib/writer/progress-stream.js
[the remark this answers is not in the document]
lurch: I _did_ look at the docs, but the docs don't make it clear that "the time-counting doesn't start until the first delta update"

PR #1675, opened 2017-08-07
upgrade: drivelist to v5.1.4

jviotti:
Change-Type: patch
Changelog-Entry: Improve Windows drive detection error codes.
See: https://github.com/resin-io-modules/drivelist/pull/204


jviotti:
I'll just merge this so I can start releasing.

PR #1677, opened 2017-08-07
feat(GUI): separate svg path and content attributes

Shou:
We separate the SVG component path and content into attributes
`paths` and `contents` which take lists of strings that are
tried until one succeeds. `contents` takes precedence over `paths`,
i.e. it is tried first.

Change-Type: patch
Changelog-Entry: Separate SVG component's path and content attributes.

### DIFF
--- a/lib/gui/app/components/svg-icon.js
+++ b/lib/gui/app/components/svg-icon.js
@@ -29,6 +29,7 @@ const propTypes = require('prop-types')
 const react2angular = require('react2angular').react2angular
 const path = require('path')
 const fs = require('fs')
+const analytics = require('../modules/analytics')
 
 const MODULE_NAME = 'Etcher.Components.SVGIcon'
 const angularSVGIcon = angular.module(MODULE_NAME, [])
@@ -36,6 +37,29 @@ const angularSVGIcon = angular.module(MODULE_NAME, [])
 const DEFAULT_SIZE = '40px'
 
 const domParser = new window.DOMParser()
+
+/**
+ * @summary Try to parse SVG contents and return it data encoded
+ *
+ * @param {String} contents - SVG XML contents
+ * @returns {String|null}
+ *
+ * @example
+ * const encodedSVG = tryParseSVGContents('<svg><path></path></svg>')
+ *
+ * img.src = encodedSVG
+ */
+const tryParseSVGContents = (contents) => {
+  const doc = domParser.parseFromString(contents, 'image/svg+xml')
+  const parserError = doc.querySelector('parsererror')
+  const svg = doc.querySelector('svg')
+
+  if (!parserError && svg) {
+    return `data:image/svg+xml,${encodeURIComponent(svg.outerHTML)}`
+  }
+
+  return null
+}
 
 /**
  * @summary SVG element that takes both filepaths and file contents
@@ -59,30 +83,51 @@ class SVGIcon extends react.Component {
       // eslint-disable-next-line no-underscore-dangle
       : global.__dirname
 
-    // This means the path to the icon should be
-    // relative to *this directory*.
-    // TODO: There might be a way to compute the path
-    // relatively to the `index.html`.
-    const imagePath = path.join(baseDirectory, 'assets', this.props.path)
+    let svgData = ''
 
-    let contents = ''
+    _.find(this.props.contents, (content) => {
+      const attempt = tryParseSVGContents(content)
 
-    if (_.startsWith(this.props.path, '<')) {
-      contents = this.props.path
-    } else {
-      contents = fs.readFileSync(imagePath, {
-        encoding: 'utf8'
+      if (attempt) {
+        svgData = attempt
+        return true
+      }
+
+      return false
+    })
+
+    if (!svgData) {
+      _.find(this.props.paths, (relativePath) => {
+        // This means the path to the icon should be
+        // relative to *this directory*.
+        // TODO: There might be a way to compute the path
+        // relatively to the `index.html`.
+        const imagePath = path.join(baseDirectory, 'assets', relativePath)
+
+        const contents = _.attempt(() => {
+          return fs.readFileSync(imagePath, {
+            encoding: 'utf8'
+          })
+        })
+
+        if (_.isError(contents)) {
+          analytics.logException(contents)
+          return false
+        }
+
+        const parsed = _.attempt(tryParseSVGContents, contents)
+
+        if (parsed) {
+          svgData = parsed
+          return true
+        }
+
+        return false
       })
     }
 
     const width = this.props.width || DEFAULT_SIZE
     const height = this.props.height || DEFAULT_SIZE
-
-    const doc = domParser.parseFromString(contents, 'image/svg+xml')
-    const parserError = doc.querySelector('parsererror')
-    const svg = doc.querySelector('svg')
-    const svgXml = svg && _.isNil(parserError) ? svg.outerHTML : ''
-    const svgData = `data:image/svg+xml,${encodeURIComponent(svgXml)}`
 
     return react.createElement('img', {
       className: 'svg-icon',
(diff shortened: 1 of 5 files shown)


tests/gui/components/svg-icon.spec.js
```diff
@@ -48,7 +48,7 @@ describe('Browser: SVGIcon', function () {
       iconContents[0] = `<!--${iconContents[0].slice(1, iconContents[0].length - 1)}-->`
       iconContents = iconContents.join('\n')
 
-      const element = $compile(`<svg-icon path="'${icon}'">Resin.io</svg-icon>`)($rootScope)
+      const element = $compile(`<svg-icon paths="['${icon}']">Resin.io</svg-icon>`)($rootScope)
```
lurch: Hmmm, so it looks like `paths` and `contents` both now actually accept arrays of values, rather than just a single value? And it then searches through each of these arrays, and displays the first value in the array that's valid? Sounds like lots more unit tests would be useful to validate this new behaviour... ;-)

(and this behaviour should also be mentioned in the jsdoc, especially the fact that 'contents' (if present) is checked before `paths` is checked)

lib/gui/pages/main/templates/main.tpl.html
```diff
@@ -3,7 +3,7 @@
     <div class="box text-center relative" os-dropzone="image.selectImageByPath($file)">
 
       <div class="center-block">
-        <svg-icon path="main.selection.getImageLogo() || '../assets/image.svg'"></svg-icon>
+        <svg-icon contents="main.selection.getImageLogo()" path="'../assets/image.svg'"></svg-icon>
```
lurch: I'm guessing that `path=` here should be `paths=` ? And should these also be using array-notation like the tests are using?

lurch: Hmmm, as well as `paths` and `contents` I wonder if it'd be worth also allowing `content` and `path` for the (common) situation where you only want to pass in a single path, not multiple paths? \*shrug\*
(obviously you'd need to explicitly document which order each of these properties are tried in)

jhermsmeier:
@Shou what's the state of this?

tests/gui/components/svg-icon.spec.js
Shou: I've added some tests for this now

lib/gui/pages/main/templates/main.tpl.html
Shou: That's true, I've fixed the `path` into `paths` now

Shou:
I've just rebased and fixed this, [here are the relevant changes][1]. Let me know what you think.

[1]: https://github.com/resin-io/etcher/compare/6b0ed64819cd1a178d18e9fd0605e98f69eb8b8a...svg-contents-attribute

jhermsmeier:
@Shou looks like the CIs aren't happy

jhermsmeier:
Ping @Shou

Shou:
Rebased on master and fixed linter errors

lib/gui/app/components/svg-icon.js
```diff
@@ -59,31 +83,52 @@ class SVGIcon extends react.Component {
...
-    } else {
-      contents = fs.readFileSync(imagePath, {
-        encoding: 'utf8'
+    if (!svgData) {
+      _.find(this.props.paths, (relativePath) => {
+        // This means the path to the icon should be
+        // relative to *this directory*.
+        // TODO: There might be a way to compute the path
+        // relatively to the `index.html`.
+        const imagePath = path.join(baseDirectory, 'assets', relativePath)
```
jhermsmeier: Regarding the TODO here, I think using `path.relative(from,to)` could work here

Shou: Let's fix that in a separate PR, as this is just a moved block, the TODO isn't new to this PR.

jhermsmeier: Oh, right – yeah, let's do that then

Shou: Sweet, is this good to merge then?

jhermsmeier:
What about @lurch's comment?

> Looks like this PR also adds the "allow multiple paths to be specified so that we can give fallback values" - IMHO that should be mentioned in the commit message too.

Shou:
@jhermsmeier updated the commit message and the OP as well

lib/gui/app/components/svg-icon.js
```diff
@@ -59,31 +83,52 @@ class SVGIcon extends react.Component {
...
+            encoding: 'utf8'
+          })
+        })
+
+        if (_.isError(contents)) {
+          analytics.logException(contents)
+          return false
+        }
+
+        const parsed = _.attempt(tryParseSVGContents, contents)
```
lurch: Does this _need_ to use `_.attempt`, given that the call on line 89 doesn't?

lib/gui/app/components/svg-icon.js
```diff
@@ -59,31 +83,52 @@ class SVGIcon extends react.Component {
...
     const width = this.props.width || DEFAULT_SIZE
     const height = this.props.height || DEFAULT_SIZE
 
-    const doc = domParser.parseFromString(contents, 'image/svg+xml')
-    const parserError = doc.querySelector('parsererror')
-    const svg = doc.querySelector('svg')
-    const svgXml = svg && _.isNil(parserError) ? svg.outerHTML : ''
-    const svgData = `data:image/svg+xml,${encodeURIComponent(svgXml)}`
-
     return react.createElement('img', {
```
lurch: What's the behaviour if neither `contents` or `paths` resolves to an actual SVG, i.e. `svgData` is still set to `''`?

lib/gui/app/components/svg-icon.js
```diff
@@ -108,9 +153,14 @@ class SVGIcon extends react.Component {
...
-   * @summary SVG contents or path to the resource
+   * @summary Paths to SVG files to be tried in succession if any fails
+   */
+  paths: propTypes.array,
+
+  /**
+   * @summary List of embedded SVG contents to be tried in succession if any fails
    */
-  path: propTypes.string.isRequired,
+  contents: propTypes.array,
```
lurch: Minor nit, but given that the code attempts to parse `contents` before parsing `paths`, maybe `contents` should be declared above `paths`?

lib/gui/app/pages/main/templates/main.tpl.html
```diff
@@ -3,7 +3,7 @@
     <div class="box text-center relative" os-dropzone="image.selectImageByPath($file)">
 
       <div class="center-block">
-        <svg-icon path="main.selection.getImageLogo() || '../../assets/image.svg'"></svg-icon>
+        <svg-icon contents="main.selection.getImageLogo()" paths="[ '../../assets/image.svg' ]"></svg-icon>
```
lurch: Isn't `contents` supposed to be an array?! See e.g. https://github.com/resin-io/etcher/pull/1677/files#diff-0bee66f458b50bfd24c37224e5891e78R105

lib/gui/app/components/svg-icon.js
```diff
@@ -37,6 +38,29 @@ const DEFAULT_SIZE = '40px'
...
+ *
+ * @example
+ * const encodedSVG = tryParseSVGContents('<svg><path></path></svg>')
+ *
+ * img.src = encodedSVG
+ */
+const tryParseSVGContents = (contents) => {
+  const doc = domParser.parseFromString(contents, 'image/svg+xml')
+  const parserError = doc.querySelector('parsererror')
+  const svg = doc.querySelector('svg')
```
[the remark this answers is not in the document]
jhermsmeier: @lurch it's counter-intuitive, but a parser error does not mean that no document is constructed; in the case of SVG, there will at least be an "error document", which is an SVG containing the error message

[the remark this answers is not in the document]
lurch: Okay. Perhaps I should have left this comment on this line instead ;)  https://github.com/resin-io/etcher/pull/1677/files#diff-9d4c83308fdfbb2b3567f9a9ecff3136R57
(but as I said, only a minor nit)

PR #1680, opened 2017-08-08
feat(GUI): remove angular from image-writer

Shou:
We remove Angular from the `image-writer` module by using Redux store
updates, subscribing to them while flashing.

Changelog-Entry: Remove Angular dependency from image-writer.

### DIFF
--- a/lib/gui/modules/image-writer.js
+++ b/lib/gui/modules/image-writer.js
@@ -16,11 +16,7 @@
 
 'use strict'
 
-/**
- * @module Etcher.Modules.ImageWriter
- */
-
-const angular = require('angular')
+const Bluebird = require('bluebird')
 const _ = require('lodash')
 const childWriter = require('../../child-writer')
 const settings = require('../models/settings')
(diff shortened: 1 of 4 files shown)


lib/gui/modules/image-writer.js
```diff
@@ -29,118 +25,105 @@ const errors = require('../../shared/errors')
...
-      unmountOnSuccess: settings.get('unmountOnSuccess'),
-      validateWriteOnSuccess: settings.get('validateWriteOnSuccess')
-    }
+/**
+ * @summary Flash an image to a drive
+ * @function
+ * @public
+ *
+ * @description
+ * This function will update `ImageWriterService.state` with the current writing state.
```
jviotti: I guess `ImageWriterService` is not accurate anymore.

lib/gui/modules/image-writer.js
```diff
@@ -29,118 +25,105 @@ const errors = require('../../shared/errors')
...
+ *
+ * @param {String} image - image path
+ * @param {Object} drive - drive
+ * @returns {Promise}
+ *
+ * @example
+ * ImageWriterService.flash('foo.img', {
```
jviotti: We should also update the example.

tests/gui/modules/image-writer.spec.js
```diff
@@ -1,132 +1,99 @@
...
-          ImageWriterService.flash('foo.img', '/dev/disk2').catch(function (error) {
-            rejectError = error
-          })
-
-          $rootScope.$apply()
+      it('should reject the second flash attempt', function () {
+        imageWriter.flash('foo.img', '/dev/disk2')
 
+        let rejectError = null
+        imageWriter.flash('foo.img', '/dev/disk2').catch(function (error) {
```
jviotti: I think we can now use fat arrows in most of these tests. Angular stuff heavily rely on `function`, and now that we're using `Bluebird` instead of `$q`, I think we can now use fat arrows.

Shou:
Done!

lurch:
Does any of this need to be updated now that Pipage has been integrated?

Shou:
The macOS test is stuck here and on many PRs it seems

jhermsmeier:
You'll need to rebase this for the tests to work again, as @jviotti mentioned a few days ago

jhermsmeier:
Oh, no – the Mac OS worker crashed again, and is offline from what it looks like.
Feel free to merge, as long as required statuses are green.

Issue #1541, opened 2017-06-23
Integrate usbboot into Etcher


jviotti:
See https://github.com/raspberrypi/usbboot
See https://www.raspberrypi.org/documentation/hardware/computemodule/cm-emmc-flashing.md

`usbboot` is a small C program that is able to mount the eMMC of a Raspberry Pi as a mass storage device, to which Etcher can flash to, effectively flashing directly to the internal storage.

Let's see if we can build a Node.js add-on around it (so we don't have to spawn anything), and then integrate with Etcher as follows:

- Try to find Pi devices connected to the computer during the drive listing loop
- If the user selects such device, `usbboot` will run and mount the device when the user clicks "Flash"

jviotti:
We need to investigate how the Windows version works. Apparently, the same C program is compiled under Cygwin, but we have to double check.

alexandrosm:
we should also make sure to build such a module in a way that is compatible
with the scan()/update() interface that the micros implementation will use.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

On Fri, Jun 23, 2017 at 12:04 PM, Juan Cruz Viotti <notifications@github.com

lurch:
> Try to find Pi devices connected to the computer during the drive listing loop

It's been a while since I've used `usbboot`, but IIRC that wouldn't be possible. I think the usage-pattern is to set the ComputeModule into Slave-boot mode by setting a jumper, then start `usbboot`, and then finally power up the CM. There's then some kind of USB-handshaking done between the CM and usbboot (also called rpiboot), and rpiboot sends the CM its boot image (firmware) which makes it appear as a USB MassStorageDevice. Although I guess it _might_ be possible to split up the 'all-in-one' action of rpiboot into separate stages... I guess it depends if there's any timing-requirements.

> Let's see if we can build a Node.js add-on around it

A quick search finds https://github.com/schakko/node-usb and https://github.com/tessel/node-usb so that should be possible.

alexandrosm:
I guess if Etcher ran usbboot, users would have to connect their slave-set
device after running etcher, and would have the right result, right?

lurch:
I think so, yeah. But I don't think we should run always run usbboot automatically, but instead maybe from a button-click.
I think I remember you mentioning elsewhere that node.js / Electron apps are single-threaded, which might mean that the Etcher interface can't do anything else while it's waiting for usbboot to discover a newly-connected CM?

[...]

PR #1686, opened 2017-08-10
feat: implement usbboot adapter 

jviotti:
This commit installs `node-usb` v1.3.0 from GitHub, since that version
was never published to NPM, and is the only one that works with Visual
Studio 2015 (see https://github.com/tessel/node-usb/issues/109).

The usbboot communicates with a Raspberry Pi through USB and eventually
mounts it as a block device we can write to.

This feature bundles bootcode.bin and start.elf from the original
usbboot implementation.

The flow is the following:

- On each scan, the usbboot scanner will try to get a usbboot compatible
  USB device to the next "phase", until they are all transformed to
  block devices the user can flash to as usual

Change-Type: minor
Fixes: https://github.com/resin-io/etcher/issues/1541
See: https://github.com/raspberrypi/usbboot


jviotti:
I've been making a lot of changes and now the test suite is broken. I'll be fixing it very soon.

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,148 @@
+/*
+ * Copyright 2016 Raspberry Pi Foundation
```
lurch: I wonder if this Copyright line should be moved under the https://github.com/raspberrypi/usbboot link?
\*shrug\* I'm not a Lawyer ;-)

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,148 @@
...
+ * })
+ */
+const getDeviceStringFromIndex = (device, index) => {
+  return Bluebird.fromCallback((callback) => {
+    device.getStringDescriptor(index, callback)
+  })
+}
+
+/**
+ * @summary Check of a node-usb device object is usbboot-capable
```
lurch: of -> if? ;)

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,232 @@
...
+      })
+
+      it('should resolve an empty array', function () {
+        return usbboot.scan().then((devices) => {
+          m.chai.expect(devices).to.deep.equal([])
+        })
+      })
+    })
+
+    describe('given a single BCM2763 device', function () {
```
lurch: There's no such thing as a BCM2763 ;-)

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,232 @@
...
+                idVendor: 0x0a5c,
+                idProduct: 0x2763
+              }
+            }
+          ])
+        })
+      })
+    })
+
+    describe('given a single BCM2764 device', function () {
```
lurch: Nor is there a `BCM2764`

npm-shrinkwrap.json
```diff
@@ -4900,6 +4884,23 @@
       "from": "node-ipc@8.9.2",
       "resolved": "https://registry.npmjs.org/node-ipc/-/node-ipc-8.9.2.tgz"
     },
+    "node-pre-gyp": {
```
lurch: Hopefully we won't get more gyp-version-conflicts! ;)

lib/shared/sdk/usbboot.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Oops, fixed!

jviotti:
@lurch Take a look again, I made various changes to both the scanning code and the test suite

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,160 @@
...
+    },
+    open: _.noop,
+    close: _.noop,
+    getStringDescriptor: (index, callback) => {
+      callback(null, options.strings[index])
+    }
+  }
+}
+
+const RASPBERRY_PI_ZERO_STUB = createUSBDeviceStub({
```
jviotti: We can collect real libusb details from various models to test the scanning routines.

lurch: Could we rename this to `BCM2835_STUB` ? As my logfiles show, all Pi1 devices behave the same.

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,160 @@
...
+  vendorID: 2652,
+  productID: 10083,
```
lurch: IMHO these two values should be kept in hexadecimal

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,155 @@
...
+const USBBOOT_CAPABLE_USB_DEVICES = [
+
+  // Raspberry Pi 1
+
+  {
+    vendorID: USB_VENDOR_ID_BROADCOM_CORPORATION,
+    productID: 0x2763
+  },
+
+  // Raspberry Pi 2 & 3
```
lurch: I dunno if it's worth taking out the mention of "Raspberry Pi 2" here, since there are no RPi2 devices that actually work with usbboot.

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // We need to open the device in order to access _configDescriptor
+    device.open()
+
+    // Handle 2837 where it can start with two interfaces, the first
+    // is mass storage the second is the vendor interface for programming
+
+    /* eslint-disable no-underscore-dangle */
+    if (device._configDescriptor.bNumInterfaces === INTERFACES_SOC_BCM2837) {
```
lurch: Interestingly, if you look at the CM3 and CM3L logfiles that I sent you, it seems that they always have only 1 `bNumInterfaces`! But we should obviously preserve this code, just in case ;-)

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,155 @@
...
+const usb = require('usb')
+
+/**
+ * @summary The number of USB interfaces in devices with a BCM2837 SoC
+ * @description
+ * Raspberry Pi 3, and CM3 use this SoC.
+ * @type {Number}
+ * @constant
+ */
+const INTERFACES_SOC_BCM2837 = 1
```
lurch: Actually, this is `INTERFACES_SOC_BCM2835` - the [code comment](https://github.com/raspberrypi/usbboot/blob/master/main.c#L108) says "Handle 2837 where it can start with two interfaces", i.e. the 2837-handling is actually in the 'else' part of that code ;-)

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,155 @@
...
+const USB_VENDOR_ID_BROADCOM_CORPORATION = 0x0a5c
+
+/**
+ * @summary List of usbboot capable devices
+ * @type {Object[]}
+ * @constant
+ */
+const USBBOOT_CAPABLE_USB_DEVICES = [
+
+  // Raspberry Pi 1
```
lurch: Might be nicer to refer to just the SoC (i.e. `BCM2835`) rather than `Raspberry Pi 1`, since the latter implies a particular board? Whereas the same SoC is shared across multiple boards. (and as far as usb-booting is concerned, it's only the SoC that matters)

tests/shared/sdk/usbboot.spec.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Oh, great catch. Fixed :)

tests/shared/sdk/usbboot.spec.js
jviotti: Right, let me fix that

lurch:
Note to self: never comment on a commit directly in future, because @jviotti rebases and your comments got lost. Remember to only comment on the PR! ;-)

lurch:
> This commit installs node-usb v1.3.0 from GitHub .... (see tessel/node-usb#109).

...and also https://github.com/tessel/node-usb/issues/162 and https://github.com/tessel/node-usb/issues/193 it seems ;-)

jviotti:
lol, yeah, its usually not a good idea to comment on commits, since the author doesn't even get the notification (unless you watch everything)

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+'use strict'
+
+const _ = require('lodash')
+const Bluebird = require('bluebird')
+const usb = require('usb')
+
+/**
+ * @summary The number of USB interfaces in devices with a BCM2835 SoC
```
lurch: I wonder if this could be clarified somehow to indicate that this is the number of interfaces in the USB endpoint, and not the physical number of USB ports that the SoC actually has? Doesn't matter if not, only a very minor nit.

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param {Object} device - device
+ * @returns {Boolean} whether the device is usbboot-capable
+ *
+ * @example
+ * if (isUsbBootCapableUSBDevice({ ... })) {
+ *   console.log('We can use usbboot on this device')
+ * }
+ */
+const isUsbBootCapableUSBDevice = (device) => {
+  return !_.isUndefined(_.find(USBBOOT_CAPABLE_USB_DEVICES, {
```
lurch: Lodash-newbie-alert, but could this use `_.some`?

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,160 @@
...
+      })
+    })
+
+    describe('given a supported and a non supported device', function () {
+      beforeEach(function () {
+        this.usbGetDeviceListStub = m.sinon.stub(usb, 'getDeviceList')
+        this.usbGetDeviceListStub.returns([
+          BCM2835_STUB,
+          createUSBDeviceStub({
+            vendorID: 9999,
```
lurch: For consistency, these should probably be `0x9999` ;-)

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,160 @@
...
+      })
+
+      it('should resolve a friendly device object', function () {
+        return usbboot.scan().then((devices) => {
+          m.chai.expect(devices.length).to.equal(1)
+          const device = _.first(devices)
+          m.chai.expect(device.name).to.equal('Broadcom BCM2708 Boot')
+          m.chai.expect(device.interface).to.equal(0)
+          m.chai.expect(device.outEndpoint).to.equal(1)
+          m.chai.expect(device.device).to.deep.equal(BCM2835_STUB)
```
lurch: `device.device` looks a tiny bit strange here - might be worth giving the outer object a different name, just to clarify things?

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,160 @@
...
+          m.chai.expect(devices).to.deep.equal([])
+        })
+      })
+    })
+
+    describe('given a non supported device', function () {
+      beforeEach(function () {
+        this.usbGetDeviceListStub = m.sinon.stub(usb, 'getDeviceList')
+        this.usbGetDeviceListStub.returns([
+          createUSBDeviceStub({
```
lurch: Is it worth creating an `UNKNOWN_STUB` / `UNSUPPORTED_STUB` const-Object?

jviotti: Sounds good, I'll do it.

tests/shared/sdk/usbboot.spec.js
jviotti: Fixed!

lib/shared/sdk/usbboot.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Indeed, you're a Lodash expert by now :P

tests/shared/sdk/usbboot.spec.js
jviotti: My other WIP PR renames it to `instance`. I'll make those changes here as well.

jviotti:
All comments addressed!

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,153 @@
...
+const _ = require('lodash')
+const Bluebird = require('bluebird')
+const usb = require('usb')
+
+/**
+ * @summary The number of USB endpoint interfaces in devices with a BCM2835 SoC
+ * @type {Number}
+ * @constant
+ */
+const USB_ENDPOINT_INTERFACES_SOC_BCM2835 = 1
```
lurch: Not that it really matters, but perhaps this should be listed after the other two `const`s?

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,149 @@
...
+    },
+    open: _.noop,
+    close: _.noop,
+    getStringDescriptor: (index, callback) => {
+      callback(null, options.strings[index])
+    }
+  }
+}
+
+const BCM2835_STUB = createUSBDeviceStub({
```
lurch: Just for the sake of completeness, probably wouldn't hurt to also add a `BCM2837_STUB` ?

tests/shared/sdk/usbboot.spec.js
lurch: Which "other WIP PR" is that?

lib/shared/sdk/usbboot.js
jviotti: Done!

tests/shared/sdk/usbboot.spec.js
jviotti: What Pi model is that one?

jviotti: Nerver mind, found it!

lurch:
I hope you don't mind, but I'd like to see the other PR (that presumably builds on top of the commit in this one), before approving this PR.

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,153 @@
...
+ * })
+ */
+const getDeviceStringFromIndex = (device, index) => {
+  return Bluebird.fromCallback((callback) => {
+    device.getStringDescriptor(index, callback)
+  })
+}
+
+/**
+ * @summary Scan for usbboot USB capable devices
```
lurch: Minor nit, but I'd change this to `Scan for usbboot capable USB devices`

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,153 @@
...
+ * @type {Object[]}
+ * @constant
+ */
+const USBBOOT_CAPABLE_USB_DEVICES = [
+
+  // BCM2835
+
+  {
+    vendorID: USB_VENDOR_ID_BROADCOM_CORPORATION,
+    productID: 0x2763
```
lurch: I wonder if it'd be worth creating constants for `USB_PRODUCT_ID_BCM2708_BOOT` and `USB_PRODUCT_ID_BCM2710_BOOT` ?

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * @summary The number of USB endpoint interfaces in devices with a BCM2835 SoC
+ * @type {Number}
+ * @constant
+ */
+const USB_ENDPOINT_INTERFACES_SOC_BCM2835 = 1
+
+/**
+ * @summary Check if a node-usb device object is usbboot-capable
```
lurch: All the other doc-strings just talk about 'USB devices', so perhaps `node-usb device` here should just be changed to `USB device` ?  (the mention of `node-usb` looks a bit out of place otherwise, IMHO)

lib/shared/sdk/usbboot.js
```diff
@@ -0,0 +1,153 @@
...
+  return usb.listDevices().filter(isUsbBootCapableUSBDevice).map((device) => {
+  /* eslint-enable lodash/prefer-lodash-method */
+
+    // We must keep the original device object from `node-usb`,
+    // since it has functions in its prototype we must use later on
+    const deviceWrapper = {
+      instance: device
+    }
+
+    // We need to open the device in order to access _configDescriptor
```
lurch: Does it also need to be open to access the deviceDescriptor ?

lib/shared/sdk/utils/usb.js
```diff
@@ -0,0 +1,52 @@
...
+'use strict'
+
+const _ = require('lodash')
+const Bluebird = require('bluebird')
+
+// The USB module calls `libusb_init`, which will fail
+// if the device we're running in has no USB controller
+// plugged in (e.g. in certain CI services).
+// In order to workaround that, we need to return a
+// stub if such error occurs.
```
lurch: :+1:

tests/shared/sdk/usbboot.spec.js
```diff
@@ -0,0 +1,191 @@
...
+  productIndex: 2,
+  interfaces: 1,
+  strings: [
+    null,
+    'Broadcom',
+    'BCM2708 Boot'
+  ]
+})
+
+const BCM2837_STUB = createUSBDeviceStub({
```
lurch: According to the logfiles I gathered, the CM3 (BCM2837) always starts up with `bNumInterfaces: 1`, but according to a [code comment](https://github.com/raspberrypi/usbboot/blob/master/main.c#L108) in the upstream `rpiboot` there _might_ also be some circumstances where the BCM2837 starts up with two interfaces. Perhaps it'd be worth also creating and testing a `BCM2837_TWO_INTERFACES_STUB` ?

lib/shared/sdk/usbboot.js
lurch: \*bump\*

lib/shared/sdk/usbboot.js
lurch: \*bump\*

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+    return device
+  })
+}
+
+const closeDevice = (device) => {
+  debug(`Closing device: ${device.name}`)
+  device.close()
+}
+
+const writeBootCode = (device, endpoint, bootCodeBuffer) => {
```
lurch: Missing jsdoc? ;)

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+}
+
+exports.flash = (device, options = {}) => {
+  // We need to open the device in order to access _configDescriptor
+  debug(`Opening device: ${device.name}`)
+  device.open()
+
+  // Handle 2837 where it can start with two interfaces, the first
+  // is mass storage the second is the vendor interface for programming
+  const addresses = {}
```
lurch: Small nit, but perhaps give this variable a different name? Normally with a variable named e.g. `addresses` I'd expect it to be an array?

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+  const deviceInterface = device.interface(addresses.interface)
+  debug(`Claiming interface: ${addresses.interface}`)
+  deviceInterface.claim()
+
+  const endpoint = deviceInterface.endpoint(addresses.endpoint)
+
+  const serialNumberIndex = device.deviceDescriptor.iSerialNumber
+  debug(`Serial number index: ${serialNumberIndex}`)
+
+  if (serialNumberIndex === USB_DESCRIPTOR_NULL_INDEX) {
```
lurch: Hmmm, I'm not sure `NULL_INDEX` is the appropriate term here. Maybe something like `USB_DESCRIPTOR_FIRST_SERIALINDEX`? And then the `iSerialNumber: 1` below could be `iSerialNumber: USB_DESCRIPTOR_SECOND_SERIALINDEX`?

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+    return writeBootCode(device, endpoint, _.get(options.files, [ 'bootcode.bin' ]))
+      .finally(_.partial(closeDevice, device))
+      .delay(USB_OPERATION_DELAY_MS).then(() => {
+        debug('Waiting for device to come back')
+        return waitForDevice({
+          deviceDescriptor: {
+            idVendor: USB_VENDOR_ID_BROADCOM_CORPORATION,
+            idProduct: 0x2764,
```
lurch: See my earlier comment about giving this value a named constant.

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+            iSerialNumber: 1
+          }
+        }, {
+          retries: 5,
+          delay: 1000
```
lurch: Should this be using `USB_OPERATION_DELAY_MS` ?

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+        })
+      }).then((newDevice) => {
+        if (_.isNil(newDevice)) {
+          throw new Error(`Device ${device.name} never came back`)
+        }
+
+        return exports.flash(newDevice, options)
```
lurch: Hmmm, so (unless I'm misunderstanding) this is a recursive function? Is that right?

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+
+        return Bluebird.reject(new Error(`Unrecognized command: ${fileMessage.command}`))
+      }).then(() => {
+        debug('Starting again')
+        return startFileServer(device, endpoint, files)
+      })
+    })
+}
+
+exports.flash = (device, options = {}) => {
```
lurch: JSDoc would be useful... ;)

lib/shared/sdk/usbboot/protocol.js
```diff
@@ -0,0 +1,314 @@
...
+ */
+
+'use strict'
+
+const _ = require('lodash')
+const Bluebird = require('bluebird')
+const usb = require('./usb')
+
+// The equivalent of a NULL buffer, given that node-usb complains
+// if the daata argument is not an instance of Buffer
```
lurch: daata -> data

lib/shared/sdk/usbboot/protocol.js
```diff
@@ -0,0 +1,314 @@
...
+ * @constant
+ */
+const FILE_MESSAGE_COMMAND_OFFSET = 0
+
+/**
+ * @summary The size of the file message command section
+ * @type {Number}
+ * @constant
+ */
+const FILE_MESSAGE_COMMAND_LENGTH = 4
```
lurch: For consistency, this should be `FILE_MESSAGE_COMMAND_SIZE`

lib/shared/sdk/usbboot/protocol.js
```diff
@@ -0,0 +1,314 @@
...
+ * @constant
+ */
+const FILE_MESSAGE_FILE_NAME_OFFSET = FILE_MESSAGE_COMMAND_LENGTH
+
+/**
+ * @summary The size of the file message file name section
+ * @type {Number}
+ * @constant
+ */
+const FILE_MESSAGE_FILE_NAME_LENGTH = 256
```
lurch: Again, LENGTH -> SIZE ?

lib/shared/sdk/usbboot/protocol.js
```diff
@@ -0,0 +1,314 @@
...
+ * @constant
+ */
+exports.FILE_MESSAGE_COMMAND_READ_FILE = 'ReadFile'
+
+/**
+ * @summary The "done" file message command name
+ * @type {String}
+ * @constant
+ */
+exports.FILE_MESSAGE_COMMAND_DONE = 'Done'
```
lurch: Instead of 3 separate constants, I guess you could do:
```javascript
const FILE_MESSAGE_COMMANDS = {
    GET_FILE_SIZE: 'GetFileSize',
    READ_FILE: 'ReadFile',
    DONE: 'Done'
}
```
?

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * @summary The delay to wait between each USB read/write operation
+ * @type {Number}
+ * @constant
+ * @description
+ * The USB bus seems to hang if we execute many operations at
+ * the same time.
+ */
+const USB_OPERATION_DELAY_MS = 1000
```
lurch: Should this be using the constant from `protocol` instead?

lib/shared/sdk/usbboot/protocol.js
```diff
@@ -0,0 +1,314 @@
...
+        endpoint.timeout = USB_BULK_TRANSFER_TIMEOUT_MS
+        endpoint.transfer(buffer, callback)
+      })
+    })
+}
+
+exports.sendErrorSignal = (device) => {
+  // The original implementation sends a control transfer
+  // with the size of a null buffer, and then the null
+  // buffer itself, however sending the actual null buffer
```
lurch: > The original implementation sends ... and then the null buffer itself

Are you sure? Doesn't https://github.com/raspberrypi/usbboot/blob/master/main.c#L149 mean that the NULL buffer _doesn't_ get written?

lib/shared/sdk/usbboot/protocol.js
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  const fileName = _.chain(fileMessageBuffer.toString('ascii', FILE_MESSAGE_FILE_NAME_OFFSET))
+
+    // The parsed string will likely contain tons of trailing
+    // null bytes that we should get rid of for convenience
+    // See https://github.com/nodejs/node/issues/4775
+    .takeWhile((character) => {
+      return character !== '\0'
+    })
```
lurch: Does this strip out _every_ null byte? Would it be "more correct" to _only_ strip out the trailing null bytes?
Alternatively, to more closely mimic the upstream `C` implementation, I guess we could assume that the _first_ null byte is the end-of-string marker and simply discard everything after that?

lib/shared/sdk/usbboot/usb.js
```diff
@@ -0,0 +1,170 @@
...
+], (constant) => {
+  exports[constant] = usb[constant]
+})
+
+/**
+ * @summary The timeout for USB control transfers, in milliseconds
+ * @type {Number}
+ * @constant
+ */
+const USB_CONTROL_TRANSFER_TIMEOUT = 1000
```
lurch: Yet another 'duplicated' timeout constant?

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+
+const startFileServer = (device, endpoint, files) => {
+  debug('Listening for file messages')
+  return protocol
+    .read(device, protocol.FILE_MESSAGE_LENGTH)
+    .then(protocol.parseFileMessageBuffer)
+
+    // We get this error message when reading a command
+    // from the device when the communication has ended
+    .catch({
```
lurch: I'm afraid Promise-based code isn't my strong point, but could this `.catch` go between the previous `.read` and `.then`?

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,318 @@
...
+  debug(`Claiming interface: ${addresses.interface}`)
+  deviceInterface.claim()
+
+  const endpoint = deviceInterface.endpoint(addresses.endpoint)
+
+  const serialNumberIndex = device.deviceDescriptor.iSerialNumber
+  debug(`Serial number index: ${serialNumberIndex}`)
+
+  if (serialNumberIndex === USB_DESCRIPTOR_NULL_INDEX) {
+    return writeBootCode(device, endpoint, _.get(options.files, [ 'bootcode.bin' ]))
```
lurch: Perhaps there should be a named constant for this `bootcode.bin` here?

test.js
```diff
@@ -0,0 +1,25 @@
+const _ = require('lodash')
+const usbboot = require('./lib/shared/sdk/usbboot')
+const usb = require('usb')
+const fs = require('fs')
+
+usbboot.scan().then((devices) => {
+  console.log(devices[0].instance)
+  return usbboot.flash(_.first(devices), {
+    files: {
+      'bootcode.bin': fs.readFileSync('../usbboot/msd/bootcode.bin'),
```
lurch: Maybe we need to add a script that gets run during Etcher's build process, to download the latest version of these files from https://github.com/raspberrypi/usbboot ? (or maybe we should version-lock to a specific commit hash?)
If we include these files within Etcher itself, I wonder if we need to update our LICENSE file? \*shrug\*

lurch:
Seems like there's still a few places that use `.instance`, despite the `Get rid of device.instance` commit.

lib/shared/sdk/usbboot.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Done!

lib/shared/sdk/usbboot.js
jviotti: Nope, that is always available

lib/shared/sdk/usbboot/index.js
jviotti: Does `address` make it better? I honestly can't think of a better name, but I'm open to suggestions :)

lib/shared/sdk/usbboot/index.js
jviotti: As far as I can see (I ran a couple of tests), an index of 0 means there is no data to fetch. Maybe `USB_DESCRIPTOR_NO_DATA_INDEX` makes it better?

lib/shared/sdk/usbboot/index.js
jviotti: Yep, that's right. The code has been architecture a lot, so now only the function that implements the second phase is recursive

lib/shared/sdk/usbboot/protocol.js
jviotti: Fixed!

lib/shared/sdk/usbboot/protocol.js
jviotti: Done!

lib/shared/sdk/usbboot/protocol.js
jviotti: Done!

lib/shared/sdk/usbboot/protocol.js
jviotti: Aaah, great find. I guess that clarifies the weird behavior I was getting by trying to write the null buffer :P I'll revise the comment

lib/shared/sdk/usbboot/protocol.js
jviotti: > Alternatively, to more closely mimic the upstream C implementation, I guess we could assume that the first null byte is the end-of-string marker and simply discard everything after that?

That's exactly what its doing. The `takeWhile` function will consume all the characters it can before hitting the first null byte, and discard everything else.

lib/shared/sdk/usbboot/usb.js
jviotti: Unified!

lib/shared/sdk/usbboot/index.js
jviotti: It could, however we'd need to return a buffer from the `.catch` block that `parseFileMessageBuffer` can parse, which is a bit inconvenient. By putting it afterwards, we can send the already parsed object.

lib/shared/sdk/usbboot/index.js
jviotti: Done!

jviotti:
@lurch The code was heavily re-architected based on recent discussions, so a lot of the review comments don't make sense anymore. I went through the old messages and addressed the ones that were still relevant. Let me know what you think now!

lib/gui/modules/drive-scanner.js
```diff
@@ -83,6 +108,12 @@ const scan = () => {
   return sdk.scan({
     standard: {
       includeSystemDrives: settings.get('unsafeMode')
+    },
+    usbboot: {
+      files: {
+        'bootcode.bin': USBBOOT_BOOTCODE_BIN_BUFFER,
+        'start.elf': USBBOOT_START_ELF_BUFFER
```
Shou: I don't know how likely they are to change, but maybe we could use constants here in case these names change?

    files: {
      [BOOTCODE_BIN]: ...,
      [START_ELF]: ...
    }

Comes with the added bonus of making it seem more like there are bins and elves involved.

lib/shared/sdk/usbboot/protocol.js
```diff
@@ -0,0 +1,367 @@
...
+ * } boot_message_t;
+ *
+ * This needs to be sent to the out endpoint of the USB device
+ * as a 24 bytes big-endian buffer where:
+ *
+ * - The first 4 bytes contain the size of the bootcode.bin buffer
+ * - The remaining 20 bytes contain the boot signature, which
+ *   we don't make use of in this implementation
+ *
+ * @param {Buffer} bootCodeBufferLength - bootcode.bin buffer length
```
Shou: Isn't this a `Number`?

lib/shared/store.js
```diff
@@ -120,6 +120,9 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         })
       }
 
+      // Convert object instances to plain objects
+      action.data = JSON.parse(JSON.stringify(action.data))
```
Shou: Wouldn't something like `_.assign({}, action.data)` or `{ ...action.data }` work here?

lib/shared/sdk/usbboot/protocol.js
jviotti: Fixed!

lib/shared/store.js
jviotti: Good catch! `_.assign` seems to do the trick.

lib/gui/modules/drive-scanner.js
jviotti: Haha, the thing is that the files you pass here depends on what you want to accomplish. We pass these for our use case, but we could be passing different ones depending on our use case.

lib/shared/sdk/usbboot/protocol.js
jviotti: Done!

lib/blobs/usbboot/README.md
```diff
@@ -0,0 +1,4 @@
+usbboot
+=======
+
+The files in this directory were taken https://github.com/raspberrypi/usbboot.
```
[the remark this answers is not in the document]
jviotti: Oops, fixed!

lib/shared/sdk/usbboot/index.js
```diff
@@ -0,0 +1,375 @@
...
+ * @param {Number} id - USB id
+ * @returns {String} string id
+ *
+ * @example
+ * console.log(usbIdToString(2652))
+ * > '0x0a5c'
+ */
+const usbIdToString = (id) => {
+  const USB_ID_RADIX = 16
+  const USB_ID_LENGTH = 4
```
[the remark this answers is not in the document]
jviotti: Done!

lib/blobs/usbboot/README.md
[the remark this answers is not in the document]
jhermsmeier: ping @jviotti ^

lib/shared/store.js
Shou: `_.assign` didn't work as expected here?

PR #1694, opened 2017-08-15
chore: cache Travis CI docker builds

jviotti:
Let's see if we can reduce the time it gets to build and test Etcher on
GNU/Linux.

See: https://giorgos.sealabs.net/docker-cache-on-travis-and-docker-112.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ cache:
     - $HOME/Library/Caches/electron
     - $HOME/Library/Caches/electron-builder
     - $HOME/.pkg-cache
+    - $HOME/docker
     - node_modules
 
 services:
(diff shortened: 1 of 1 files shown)


jviotti:
I'm merging so we can see if the usbboot branch finally builds correctly.

.travis.yml
```diff
@@ -63,6 +68,10 @@ install:
 script:
   - ./scripts/ci/test.sh -o $HOST_OS -r $TARGET_ARCH
   - ./scripts/ci/build-installers.sh -o $HOST_OS -r $TARGET_ARCH
+  - if [[ ${TRAVIS_BRANCH} == "master" ]] && [[ ${TRAVIS_PULL_REQUEST} == "false" ]]; then
```
lurch: Hmmm, so we only update the cache on merge-commits to `master`? Should this also check that `[[ "$TRAVIS_OS_NAME" == "linux" ]]` ?

jviotti: Yeah, otherwise any user could put malicious Docker commands in the cache. See https://github.com/resin-io/etcher/pull/1695 for the Linux check fix, plus other nasty things I found while debugging this.

PR #1695, opened 2017-08-15
chore: fix build system

jviotti:
- Bintray deployments are broken because of some bash nested quoting
  issue
- Travis CI will attempt to cache Docker layers on macOS
- Docker caches from different architectures will override each other

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,8 @@ os:
   - osx
 
 before_install:
-  - if [ -f ${DOCKER_CACHE_FILE} ] && [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-      gunzip -c ${DOCKER_CACHE_FILE} | docker load;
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -52,8 +52,8 @@ os:
   - osx
 
 before_install:
-  - if [ -f ${DOCKER_CACHE_FILE} ] && [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-      gunzip -c ${DOCKER_CACHE_FILE} | docker load;
+  - if [ -f ${DOCKER_CACHE_FILE}.${TARGET_ARCH} ] && [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
```
lurch: This means we end up with filenames like `$HOME/docker/cache.tar.gz.x86` which isn't very pretty. IMHO a filename like `$HOME/docker/cache-x86.tar.gz` would be 'nicer'.
(but obviously if this works, then it doesn't really matter)

PR #1702, opened 2017-08-21
feat: bring usbboot into drive scanner

Shou:
We leverage the Etcher usbboot SDK and bring it into drive scanner.

Changelog-Entry: Bring usbboot into drive scanner.
Depends: https://github.com/resin-io/etcher/pull/1686

### DIFF
--- a/lib/gui/modules/drive-scanner.js
+++ b/lib/gui/modules/drive-scanner.js
@@ -21,12 +21,59 @@ const _ = require('lodash')
 const EventEmitter = require('events').EventEmitter
 const drivelist = require('drivelist')
 const settings = require('../models/settings')
+const usbboot = require('../../shared/sdk/usbboot')
 
 const DRIVE_SCANNER_INTERVAL_MS = 2000
 const DRIVE_SCANNER_FIRST_SCAN_DELAY_MS = 0
 const emitter = new EventEmitter()
 
 /* eslint-disable lodash/prefer-lodash-method */
+
+/**
+ * @summary
+ *
+ * @param {Function} callback - two argument function callback of (err, result)
+ *
+ * @example
+ */
+const lister = (callback) => {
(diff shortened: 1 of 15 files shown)


Shou:
Is this going the right way in terms of design, or do we want to split the two (drives, USB devices) up more?

jviotti:
@Shou Can you show us a screenshot? There are various on-going discussions about running usbboot on the scan phase and only present block devices, so lets see what we decide on before continuing work on this.

lib/gui/modules/drive-scanner.js
```diff
@@ -21,13 +21,60 @@ const _ = require('lodash')
...
+
+    const typeDrives = _.map(drives, (drive) => {
+      return _.assign({}, drive, {
+        type: 'BLOCK_DEVICE'
+      })
+    })
+
+    return usbboot.scan().then((usbDevices) => {
+      const typeUsbDevices = _.map(usbDevices, (usbDevice) => {
+        const description = `Bus ${usbDevice.busNumber} / Address ${usbDevice.deviceAddress}`
```
lurch: If such a field is needed, it'd probably be better if it was added into the `usbDevice` object in @jviotti 's usbboot PR? Much like the `usbDevice.name` field you're also making use of below.

Shou:
@jviotti sure, here's the drive selector.

image

lib/gui/modules/drive-scanner.js
Shou: Sounds good with me, so it's up to @jviotti and his PR.

lurch:
Is this PR still needed, or has it been superseded by something else?

jviotti:
Yeah, I think it was superseded. I'll close it!

Shou:
Can confirm it was superseded.

PR #1707, opened 2017-08-24
refactor: use an SDK orchestrator to implement drive scanning

jviotti:
This is a major first step towards adopting an SDK architecture. This
commit creates an SDK adaptor with a `.scan()` function that uses
`drivelist` under the hood. Then, an SDK orchestrator is used to provide
drive scanning capabilities to the GUI.

Here's a list of some particularly interesting changes:

- The drives returned by the SDK adaptor now have a "pending" and an
  "adaptor" property. The "pending" property is a boolean flag that
  determines if the drive is ready to be used (this will come handy for
  usbboot), while the "adaptor" property simply contains the name of the
  adaptor that drive came from

- The GUI drive scanner Rx implementation was replaces with a "promise
  loop." Before, the drive scanning routine would be called every 2
  seconds (without waiting for the previous scan to complete), while
  now, the next scan happens *after* the previous scan completes. For
  this reason, I reduced the drive scanning interval timeout to match
  the timing we had before

Change-Type: patch
See: https://github.com/resin-io/etcher/pull/1686

### DIFF
--- a/lib/shared/sdk/standard/index.js
+++ b/lib/shared/sdk/standard/index.js
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2017 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


lib/shared/sdk/standard/index.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @example
+ * standard.scan({
+ *   includeSystemDrives: true
+ * }).each((device) => {
+ *   console.log(device)
+ * })
+ */
+exports.scan = (options = {}) => {
+  /* eslint-disable lodash/prefer-lodash-method */
```
Shou: Just a tiny thing, but we could use `// eslint-disable-next-line lodash/prefer-lodash-method` instead here, then you don't need to close it later.

jviotti: Oh, that's perfect, thanks a lot :)

PR #1708, opened 2017-08-24
feat: add sudo-prompt to start on linux

Shou:
We prompt the user with a sudo-prompt upon launch on Linux-based systems
to ensure the program has enough permissions for features needed
throughout the program's runtime.

Changelog-Entry: Add a sudo-prompt upon launch on Linux-based systems.

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -83,10 +83,12 @@ deb:
     - libxss1
     - libxtst6
     - polkit-1-auth-agent | policykit-1-gnome | polkit-kde-1
+    - gksu | kdesudo
 rpm:
   icon: assets/icon.png
   depends:
     - lsb
     - libXScrnSaver
+    - beesu
 appImage:
   icon: assets/icon.png
--- a/lib/child-writer/renderer-utils.js
+++ b/lib/child-writer/renderer-utils.js
@@ -43,8 +43,5 @@ exports.getApplicationEntryPoint = () => {
   const ENTRY_POINT_ARGV_INDEX = 1
   const relativeEntryPoint = electron.remote.process.argv[ENTRY_POINT_ARGV_INDEX]
 
-  // On GNU/Linux, `pkexec` resolves relative paths
-  // from `/root`, therefore we pass an absolute path,
-  // in order to be on the safe side.
   return path.join(CONSTANTS.PROJECT_ROOT, relativeEntryPoint)
 }
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -65,7 +65,7 @@ const OPTIONS_INDEX_START = 2
 const etcherArguments = process.argv.slice(OPTIONS_INDEX_START)
 
 permissions.isElevated().then((elevated) => {
-  if (!elevated) {
+  if (!elevated || !process.env.ETCHER_RUNNING) {
     console.log('Attempting to elevate')
 
     const commandArguments = _.attempt(() => {
@@ -92,6 +92,11 @@ permissions.isElevated().then((elevated) => {
         ELECTRON_RUN_AS_NODE: 1,
         IPC_SERVER_ID: process.env.IPC_SERVER_ID,
         IPC_CLIENT_ID: process.env.IPC_CLIENT_ID,
+
+        // TODO: This flag shouldn't be necessary
+        // Signal that Etcher is already running, so the writer shouldn't
+        // try to spawn any more instances, and also knows it's a writer.
+        ETCHER_RUNNING: 1,
 
         // This environment variable prevents the AppImages
         // desktop integration script from presenting the
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -65,7 +65,7 @@ const OPTIONS_INDEX_START = 2
 const etcherArguments = process.argv.slice(OPTIONS_INDEX_START)
 
 permissions.isElevated().then((elevated) => {
-  if (!elevated || !process.env.ETCHER_RUNNING) {
+  if (!elevated) {
     console.log('Attempting to elevate')
 
     const commandArguments = _.attempt(() => {
@@ -92,11 +92,6 @@ permissions.isElevated().then((elevated) => {
         ELECTRON_RUN_AS_NODE: 1,
         IPC_SERVER_ID: process.env.IPC_SERVER_ID,
         IPC_CLIENT_ID: process.env.IPC_CLIENT_ID,
-
-        // TODO: This flag shouldn't be necessary
-        // Signal that Etcher is already running, so the writer shouldn't
-        // try to spawn any more instances, and also knows it's a writer.
-        ETCHER_RUNNING: 1,
 
         // This environment variable prevents the AppImages
         // desktop integration script from presenting the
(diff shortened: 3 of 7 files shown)


Shou:
Here's the polkit policy file I'm using. `/usr/share/polkit-1/actions/etcher.policy`

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE policyconfig PUBLIC
 "-//freedesktop//DTD PolicyKit Policy Configuration 1.0//EN"
 "http://www.freedesktop.org/standards/PolicyKit/1/policyconfig.dtd">
<policyconfig>

  <vendor>Examples for the PolicyKit Project</vendor>
  <vendor_url>http://hal.freedesktop.org/docs/PolicyKit/</vendor_url>

  <action id="org.freedesktop.policykit.pkexec.etcher">
    <description>Run Etcher</description>
    <message>Authentication is required to run Etcher</message>
    <icon_name>audio-x-generic</icon_name> 
    <defaults>
      <allow_any>auth_admin</allow_any>
      <allow_inactive>auth_admin</allow_inactive>
      <allow_active>auth_admin</allow_active>
    </defaults>
    <annotate key="org.freedesktop.policykit.exec.path">/home/shou/Prog/Javascript/etcher-alt/node_modules/electron/dist/electron</annotate>
    <annotate key="org.freedesktop.policykit.exec.allow_gui">true</annotate>
  </action>

</policyconfig>
```

Shou:
Here we go, AppImages should work – thoughts on these changes?

lib/start.js
```diff
@@ -29,5 +29,69 @@
...
+        name: 'gksudo',
+        options: [
+          '--preserve-env',
+          '--message',
+          'Please provide your password. Etcher needs special permissions to interact with raw devices.'
+        ]
+      },
+      {
+        name: 'kdesudo',
+        options: []
```
jviotti: Can we customize `kdesudo`'s window message like we do with `gksu`?

lib/start.js
```diff
@@ -29,5 +29,69 @@
...
+          throw new Error(`Command does not exist: ${command.name}`)
+        }
+
+        return command
+      })
+    })).then((command) => {
+      return new Bluebird((resolve, reject) => {
+        const args = process.env.APPIMAGE ? [ process.env.APPIMAGE ] : process.argv
+        const appinstallSkip = !process.env.APPIMAGE ? {} : {
+          SKIP: 1
```
jviotti: We could always pass this variable (since its harmless) and reduce a bit the complexity

lib/child-writer/writer-proxy.js
```diff
@@ -92,6 +92,7 @@ permissions.isElevated().then((elevated) => {
         ELECTRON_RUN_AS_NODE: 1,
         IPC_SERVER_ID: process.env.IPC_SERVER_ID,
         IPC_CLIENT_ID: process.env.IPC_CLIENT_ID,
+        ETCHER_RUNNING: 1,
```
jviotti: What is this variable for? What does it mean?

electron-builder.yml
```diff
@@ -83,6 +83,7 @@ deb:
     - libxss1
     - libxtst6
     - polkit-1-auth-agent | policykit-1-gnome | polkit-kde-1
+    - gksu | kdesudo
```
jviotti: Should we add the rpm equivalents below?

lib/child-writer/writer-proxy.js
Shou: Basically, we need some way to communicate to the writer-proxy that it was called. Previously it was just a check of whether it was running with privileges or not, but now it always does. Knowing this, it creates the IPC channel. We could use `SKIP` instead but I thought it'd be easier to understand with a new variable. I'll add a comment explaining this.

jviotti: So basically if you elevate the app with `gksudo`, then `is-elevated` still returns `false`?

Shou: Oh no, the opposite, it always returns true which causes it to skip the step and no IPC channel is established which means no progress.

Shou:
Testing `beesu` on Fedora, the `--preserve-environment` flag doesn't work, nor the shorthand `-m`, which presents the AppImage installation dialogue a second time, otherwise seems to work.

Shou:
This is done and ready for reviews, the CI tests have been stuck for a while.

jhermsmeier:
@Shou can you commit --amend and push to this again to re-trigger the CI builds? AppVeyor should be unstuck now.

lib/child-writer/writer-proxy.js
jhermsmeier: @Shou have you added the comment explaining this yet? I'm not seeing it, or I'm missing something.

lib/start.js
```diff
@@ -29,5 +29,110 @@
...
+  const os = require('os')
+  const _ = require('lodash')
+  const Bluebird = require('bluebird')
+  const commandExists = Bluebird.promisify(require('command-exists'))
+  const childProcess = require('child_process')
+  const permissions = require('./shared/permissions')
+  const errors = require('./shared/errors')
+  const EXIT_CODES = require('./shared/exit-codes')
+
+  const MESSAGE = 'Etcher needs special permissions to interact with raw devices. Please provide your password.'
```
jhermsmeier: Should this live in `shared/messages` instead?

electron-builder.yml
```diff
@@ -83,10 +83,12 @@ deb:
...
 rpm:
   icon: assets/icon.png
   depends:
     - lsb
     - libXScrnSaver
+    - beesu
```
jviotti: So `kdesudo` is not available on Fedora?

lib/child-writer/renderer-utils.js
```diff
@@ -46,5 +46,5 @@ exports.getApplicationEntryPoint = () => {
   // On GNU/Linux, `pkexec` resolves relative paths
   // from `/root`, therefore we pass an absolute path,
   // in order to be on the safe side.
-  return path.join(CONSTANTS.PROJECT_ROOT, relativeEntryPoint)
+  return path.resolve(CONSTANTS.PROJECT_ROOT, relativeEntryPoint)
```
jviotti: Can you explain why this change is needed? Will it break other platforms? If not, should we send it as a different PR?

jviotti:
I'll take over this PR so we can release today.

lib/child-writer/renderer-utils.js
```diff
@@ -43,8 +43,5 @@ exports.getApplicationEntryPoint = () => {
   const ENTRY_POINT_ARGV_INDEX = 1
   const relativeEntryPoint = electron.remote.process.argv[ENTRY_POINT_ARGV_INDEX]
 
-  // On GNU/Linux, `pkexec` resolves relative paths
-  // from `/root`, therefore we pass an absolute path,
-  // in order to be on the safe side.
```
lurch: I've not been following the progress of this PR, but why has this comment been removed?

jviotti:
I tested this on Ubuntu (gtksu and kdesudo in AppImage and deb) and Fedora (beesu on rpm) and works fine. 

@Shou Can you add a bit TODO comment about the `ETCHER_RUNNING` environment variable, and potentially create an issue for it once this gets merged? I'd love to take a closer look later on.

lib/child-writer/renderer-utils.js
Shou: We don't use `pkexec` anymore so this explanation isn't necessary

jhermsmeier:
@Shou how're we looking on this? There are still some requested changes & the branch also needs some rebasing / conflict resolution.

jviotti:
@Shou @jhermsmeier @lurch I can confirm that removing the `ETCHER_RUNNING` has no effect. The flash process works fine without it on my Ubuntu 16.04 system, so I think we can safely get rid of it.

PR #1709, opened 2017-08-24
feat(GUI): support new "pending" drive flag

jviotti:
We recently added a "pending" flag to all drives that represents whether
the drive is ready for selection or not. This flag will be used by the
"usbboot" flashing adaptor, which will emit various "pending" USB
devices while it converts them to block devices that can actually be
flashed.

In terms of the GUI, the following visible changes were made:

- Drives with a `pending: true` property will be disabled in the drive
  selector window
- Drives with a `pending: true` property have a "PENDING" red badge

See: https://github.com/resin-io/etcher/pull/1707
See: https://github.com/resin-io/etcher/pull/1686

### DIFF
--- a/tests/shared/drive-constraints.spec.js
+++ b/tests/shared/drive-constraints.spec.js
@@ -527,6 +527,43 @@ describe('Shared: DriveConstraints', function () {
     })
   })
 
+  describe('.isDrivePending()', function () {
+    it('should return true if the drive is pending', function () {
+      const result = constraints.isDrivePending({
+        device: '/dev/disk1',
+        name: 'USB Drive',
+        size: 1000000000,
+        protected: false,
+        pending: true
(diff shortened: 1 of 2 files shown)


jviotti:
I'll merge this so we can aim for a fully functional Etcher + usbboot build on Friday, but happy to address any comments in a future PR.

tests/shared/drive-constraints.spec.js
```diff
@@ -809,6 +975,19 @@ describe('Shared: DriveConstraints', function () {
       })
     })
 
+    describe('given the drive is pending', () => {
+      it('should return an empty list', function () {
```
lurch: I think this `should return an empty list` should actually be `should return the pending error`, which would better match what the next unit-test does?

PR #1712, opened 2017-08-28
upgrade: mountutils to v1.2.2

jviotti:
See: https://github.com/resin-io-modules/mountutils/pull/45
Change-Type: patch
Changelog-Entry: Retry ejection various times before giving up on Windows.


lurch:
Actually, I don't think this does fix #1689 .
It looks like **gregcotech** 'resolved his problem' before he even tried your custom build, and **CACTANAKA** is still getting the same error, even with the 5 retries.

jviotti:
Good catch. I removed the `Fixes` tag. Lets get this merged anyway, since I think it will fix *some* cases.

lurch:
Haha, fair enough :)

Issue #1706, opened 2017-08-23
"Could not locate the bindings" on Windows 10


atn832:
- **Etcher version: \Etcher-cli-1.1.2-win32-x64**
- **Operating system and architecture: Windows 10**
- **Image flashed: iot_rpi3.img **

I downloaded the build for Windows x64 from the etcher/cli website (https://etcher.io/cli/, https://github.com/resin-io/etcher/releases/download/v1.1.2/Etcher-cli-1.1.2-win32-x64.zip). I extracted the zip to C:\etcher-cli like in the instructions, and also added that path to my PATH environment variable. Etcher-cli crashes before even listing the available drives to flash.

```
C:\etcher-cli>etcher
Usage: etcher.exe C:\snapshot\etcher\dist\Etcher-cli-1.1.2-win32-x64-app\lib\cli\etcher.js [options] <image>

Options:
  --help, -h     show help  [boolean]
  --version, -v  show version number  [boolean]
  --drive, -d    drive  [string]
  --check, -c    validate write  [boolean] [default: true]
  --yes, -y      confirm non-interactively  [boolean]
  --unmount, -u  unmount on success  [boolean] [default: true]

Examples:
  etcher.exe C:\snapshot\etcher\dist\Etcher-cli-1.1.2-win32-x64-app\lib\cli\etcher.js raspberry-pi.img
  etcher.exe C:\snapshot\etcher\dist\Etcher-cli-1.1.2-win32-x64-app\lib\cli\etcher.js --no-check raspberry-pi.img
  etcher.exe C:\snapshot\etcher\dist\Etcher-cli-1.1.2-win32-x64-app\lib\cli\etcher.js -d /dev/disk2 ubuntu.iso
  etcher.exe C:\snapshot\etcher\dist\Etcher-cli-1.1.2-win32-x64-app\lib\cli\etcher.js -d /dev/disk2 -y rpi.img

Exit codes:
  0 - Success
  1 - General Error
  2 - Validation Error
  3 - Cancelled

If you need help, don't hesitate in contacting us at:

  GitHub: https://github.com/resin-io/etcher/issues/new
  Gitter: https://gitter.im/resin-io/etcher

Missing image

C:\etcher-cli>etcher C:\snapshot\etcher\dist\Etcher-cli-1.1.2-win32-x64-app\lib\cli\etcher.js iot_rpi3.img
Could not locate the bindings file. Tried:
[...]

PR #1713, opened 2017-08-28
chore: fix CLI packaging snapshot relative directory

jviotti:
The current CLI releases are broken. Seems that `pkg` creates the
application snapshot based on the current working directory, so at the
moment, the snapshot gets created based on the root of the project,
rather than based on the dist/Etcher-cli-* directories, causing the
native add-ons to not be resolved correctly.

Fixes: https://github.com/resin-io/etcher/issues/1706
Change-Type: patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ TARGET_ARCH_ELECTRON_BUILDER = $(shell ./scripts/build/architecture-convert.sh -
 PLATFORM_PKG = $(shell ./scripts/build/platform-convert.sh -r $(PLATFORM) -t pkg)
 ENTRY_POINT_CLI = lib/cli/etcher.js
 ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE)
-ifeq ($(TARGET_PLATFORM),win32)
+ifeq ($(PLATFORM),win32)
 ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE).exe
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -107,7 +107,7 @@ TARGET_ARCH_ELECTRON_BUILDER = $(shell ./scripts/build/architecture-convert.sh -
 PLATFORM_PKG = $(shell ./scripts/build/platform-convert.sh -r $(PLATFORM) -t pkg)
 ENTRY_POINT_CLI = lib/cli/etcher.js
 ETCHER_CLI_BINARY = $(APPLICATION_NAME_LOWERCASE)
-ifeq ($(TARGET_PLATFORM),win32)
+ifeq ($(PLATFORM),win32)
```
lurch: Haha, whoops! Good catch.

Makefile
```diff
@@ -228,7 +228,7 @@ $(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(PLATFORM)-$(
 	$(BUILD_DIRECTORY)/$(APPLICATION_NAME)-cli-$(APPLICATION_VERSION)-$(PLATFORM)-$(TARGET_ARCH)-app \
 	| $(BUILD_DIRECTORY)
 	mkdir $@
-	$(NPX) pkg --output $@/$(ETCHER_CLI_BINARY) -t node6-$(PLATFORM_PKG)-$(TARGET_ARCH) $</$(ENTRY_POINT_CLI)
+	cd $< && ../../$(NPX) pkg --output ../../$@/$(ETCHER_CLI_BINARY) -t node6-$(PLATFORM_PKG)-$(TARGET_ARCH) $(ENTRY_POINT_CLI)
```
lurch: This looks a little bit ugly - I wonder if it'd be less ugly to have a pkg-wrapper script, that could 'hide' this ugliness from the `Makefile`?
Would it be also worth adding an issue to the `pkg` repo, suggesting they add an additional command-line parameter, that would then make pkg 'natively' do what you want (without this wrapper script)?

jviotti: See https://github.com/zeit/pkg/issues/231. I'll merge this so we can release (we're planning a release very soon), but hopefully `pkg` fixes the issue soon, and we can update this.

PR #1717, opened 2017-08-29
chore: pass a dictionary to codespell.py

jviotti:
The `-` option loads the default dictionary.


jhermsmeier:
Little typo in the commit message: dictionaty -> dictionary

Makefile
```diff
@@ -530,7 +530,7 @@ lint-html:
 	node scripts/html-lint.js
 
 lint-spell:
-	codespell.py \
+	codespell.py --dictionary - --dictionary dictionary \
```
jhermsmeier: I'm not sure I'm following what `--dictionary -` is for here?
Unrelated nit: What do you think about renaming the dictionary file to `dictionary.txt` to avoid confusion about what it is?

jviotti: Its used to load the default dictionary (see https://github.com/lucasdemarchi/codespell/blob/master/codespell_lib/_codespell.py#L547).

> Unrelated nit: What do you think about renaming the dictionary file to dictionary.txt to avoid confusion about what it is?

Sounds good, I'll handle it

jviotti:
Fixed!

Makefile
jhermsmeier: Haha, mildly cryptic and undocumented

lurch:
Please don't merge this yet, I'll have to double-check things...
ISTR we _did_ once use this option in the past, but then for some other reason we had to version-lock to an older version of `codespell` (I think it was something to do with either the `.py` extension being added or removed), and in that older version it didn't support multiple dictionaries.
Or perhaps I'm just getting muddled up with something else? Which is why I need to double-check ;-)

jhermsmeier:
Looks like @lurch remembers correctly – v1.9.2 (which is what we're using because of the `.py` issue) doesn't appear to have support for multiple dictionaries: https://github.com/lucasdemarchi/codespell/compare/v1.9.2...master#diff-c4984ff60bb6d58a3a2f9940d1a396f9L510

jviotti:
See https://github.com/lucasdemarchi/codespell/issues/82

lurch:
https://github.com/lucasdemarchi/codespell/issues/82#issuecomment-344946735 says the new version of codespell will be pushed to pypi next week, which will finally allow us to update and then merge this PR :)

jviotti:
PR updated with new version.

lurch:
I fear I may have broken the `Signed-off-by` thing by squashing my commits into the original?

lurch:
Still waiting for the OSX builds, but all the other CI builds are now green :-)

lurch:
Apologies for hitting the "Merge" button on this guys - I didn't realise how much it'd add to the CI builds backlog (I should have waited until after v1.2.1 was published) :disappointed:

Issue #1723, opened 2017-09-07
Guix SD .xz file not detected


kwill:
- **Etcher version:** 1.1.2
- **Operating system and architecture:** Ubuntu 17.04 x64
- **Image flashed:** guixsd-usb-install-0.13.0.x86_64-linux.xz
- **Do you see any meaningful error information in the DevTools?** Yes, see below.

**Steps to reproduce:**

* Download Guix SD image: https://www.gnu.org/software/guix/
* Open Etcher
* Click "Select image"
* Select Guix SD file

**What should happen?**

Image is loaded. Flash process continues.

**What happens instead?**

After a significant pause the following error message appears:

Error opening image
Something went wrong while opening guixsd-usb-install-0.13.0.x86_64-linux.xz
Error: File format not recognized

And the following error is reported in devtools:

<code>
[...]

PR #1724, opened 2017-09-08
fix: Support raw images without secondary file extension

jhermsmeier:
This allows selection of images without a secondary file extension
(i.e. `example.gz`, compared to `example.img.gz`) by defaulting to `img`
in the image-stream handlers, should no secondary extension be found.

Further this adjusts `.getPenultimateFileExtension()` to return `null`
if the detected penultimate extension is not a known file extension.

Change-Type: patch
Connects To: #1723

### DIFF
--- a/lib/image-stream/handlers.js
+++ b/lib/image-stream/handlers.js
@@ -34,6 +34,13 @@ const path = require('path')
 const errors = require('../shared/errors')
 
 /**
+ * @summary Default image extension to be assumed
+ * @type {String}
+ * @constant
+ */
+const DEFAULT_EXT = 'img'
+
+/**
  * @summary Image handlers
  * @namespace handlers
  * @public
@@ -57,7 +64,7 @@ module.exports = {
     return Bluebird.props({
       path: imagePath,
       archiveExtension: fileExtensions.getLastFileExtension(imagePath),
-      extension: fileExtensions.getPenultimateFileExtension(imagePath),
+      extension: fileExtensions.getPenultimateFileExtension(imagePath) || DEFAULT_EXT,
       stream: fs.createReadStream(imagePath),
       size: {
         original: options.size,
@@ -97,7 +104,7 @@ module.exports = {
       return Bluebird.props({
         path: imagePath,
         archiveExtension: fileExtensions.getLastFileExtension(imagePath),
-        extension: fileExtensions.getPenultimateFileExtension(imagePath),
+        extension: fileExtensions.getPenultimateFileExtension(imagePath) || DEFAULT_EXT,
         stream: fs.createReadStream(imagePath),
         size: {
           original: options.size,
@@ -138,7 +145,7 @@ module.exports = {
       return {
         path: imagePath,
         archiveExtension: fileExtensions.getLastFileExtension(imagePath),
-        extension: fileExtensions.getPenultimateFileExtension(imagePath),
+        extension: fileExtensions.getPenultimateFileExtension(imagePath) || DEFAULT_EXT,
         stream: fs.createReadStream(imagePath),
         size: {
           original: options.size,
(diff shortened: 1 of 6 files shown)


lib/image-stream/handlers.js
```diff
@@ -97,7 +97,7 @@ module.exports = {
       return Bluebird.props({
         path: imagePath,
         archiveExtension: fileExtensions.getLastFileExtension(imagePath),
-        extension: fileExtensions.getPenultimateFileExtension(imagePath),
+        extension: fileExtensions.getPenultimateFileExtension(imagePath) || 'img',
```
jviotti: Could we change `getPenultimateFileExtension` itself to fallback to `img`?

jhermsmeier: I think that'd cause the archive handling to not work as expected, but would need to double-check. This whole file extension handling needs a rework imo, while this is just a quick fix for some distros that omit the secondary extension for some reason.

Shou: Also maybe we could make `'img'` into a constant `DEFAULT_EXT` or something along those lines?

jhermsmeier: @Shou yup, definitely – updated!

tests/shared/models/selection-state.spec.js
```diff
@@ -370,6 +370,42 @@ describe('Model: selectionState', function () {
         m.chai.expect(imagePath).to.equal('foo.zip')
       })
 
+      it('should infer a compressed raw image if the penultimate extension is missing', function () {
```
lurch: I'm confused - what part of this unit-test is checking that it's inferring a raw image type?

lib/image-stream/handlers.js
lurch: > some distros that omit the secondary extension

Just out of curiosity, could you link to some examples? (just as a comment here, not as a code-comment in the PR)

lib/image-stream/handlers.js
```diff
@@ -138,7 +138,7 @@ module.exports = {
       return {
```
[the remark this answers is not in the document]
jhermsmeier: Remnants of a bygone era, see https://github.com/resin-io/etcher/pull/1726

tests/shared/models/selection-state.spec.js
jhermsmeier: It's in `selectionState.setImage()` – there are a bunch of checks in `lib/shared/store.js` (just below the TODO comment I added there) that'd previously go off with "invalid archive extension", which makes the tests very opaque here.

lib/shared/file-extensions.js
```diff
@@ -62,13 +63,14 @@ exports.getLastFileExtension = (filePath) => {
...
  * @example
  * const extension = fileExtensions.getPenultimateFileExtension('path/to/foo.img.gz');
  * console.log(extension);
- * > [ 'img' ]
+ * > 'img'
  */
 exports.getPenultimateFileExtension = (filePath) => {
-  return _.last(_.initial(exports.getFileExtensions(filePath)))
+  const ext = _.last(_.initial(exports.getFileExtensions(filePath)))
+  return !_.isNil(ext) && mime.lookup(ext) ? ext : null
```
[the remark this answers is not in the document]
jhermsmeier: Yeah, this is basically a guard against our extension checks going off on things like `imagename.arch-x86.gz`, where `arch-x86` would be determined as the penultimate extension, but isn't actually descriptive to the file format.

jviotti:
Looks good. I guess the that a way to solve this whole thing would be to make image objects report the MIME of the archive, and the file, and avoid extensions altogether.

jhermsmeier:
Yeah, was thinking something similar – shifting the format / type handling into the handlers seems like the sensible thing to do.

tests/shared/models/selection-state.spec.js
lurch: > which makes the tests very opaque here.

Ahh, fair enough :)

lurch:
I guess this also touches on #1360

PR #1736, opened 2017-09-20
feat: add drive multi-selection in store

Shou:
We lay the foundation for multi-selecting drives by implementing it into
the `store` and relevant modules interacting with the `store`.

Changelog-Entry: Add drive multi-selection to the store.

### DIFF
--- a/lib/gui/app/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/app/components/drive-selector/controllers/drive-selector.js
@@ -118,6 +118,7 @@ module.exports = function (
           previouslySelected: selectionState.isCurrentDrive(drive.device)
         })
 
+        selectionState.deselectOtherDrives(drive.device)
         selectionState.toggleDrive(drive.device)
       }
     })
@@ -132,7 +133,7 @@ module.exports = function (
    * DriveSelectorController.closeModal();
    */
   this.closeModal = () => {
-    const selectedDrive = selectionState.getDrive()
+    const selectedDrive = selectionState.getCurrentDrive()
 
     // Sanity check to cover the case where a drive is selected,
     // the drive is then unplugged from the computer and the modal
--- a/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
+++ b/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
@@ -38,7 +38,7 @@
         </div>
         <span class="list-group-item-section tick tick--success"
           ng-show="modal.constraints.isDriveValid(drive, modal.state.getImage())"
-          ng-disabled="!modal.state.isCurrentDrive(drive.device)"></span>
+          ng-disabled="!modal.state.isDriveSelected(drive.device)"></span>
     </li>
     <li class="list-group-item"
       ng-show="!modal.drives.hasAvailableDrives()">
--- a/lib/gui/app/pages/main/templates/main.tpl.html
+++ b/lib/gui/app/pages/main/templates/main.tpl.html
@@ -66,11 +66,11 @@
               'text-disabled': main.shouldDriveStepBeDisabled()
             }">
             <span class="drive-step step-name"
-              uib-tooltip="{{ main.selection.getDrive().description }} ({{ main.selection.getDrive().displayName }})">
+              uib-tooltip="{{ main.selection.getCurrentDrive().description }} ({{ main.selection.getCurrentDrive().displayName }})">
               <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
-              {{ (main.selection.getDrive().description || "") | middleEllipses:11 }}
+              {{ (main.selection.getCurrentDrive().description || "") | middleEllipses:11 }}
             </span>
-            <span class="step-drive step-size">{{ main.selection.getDrive().size | closestUnit }}</span>
+            <span class="step-drive step-size">{{ main.selection.getCurrentDrive().size | closestUnit }}</span>
             <span class="step-drive step-warning glyphicon glyphicon-exclamation-sign"
               uib-tooltip="{{ main.constraints.getDriveImageCompatibilityStatuses(main.selection.getDrive(), main.selection.getImage())[0].message }}"
               ng-show="main.constraints.hasDriveImageCompatibilityStatus(main.selection.getDrive(), main.selection.getImage())"></span>
@@ -97,7 +97,7 @@
           percentage="main.state.getFlashState().percentage"
           striped="{{ main.state.getFlashState().type == 'check' }}"
           ng-attr-active="{{ main.state.isFlashing() }}"
-          ng-click="flash.flashImageToDrive(main.selection.getImage(), main.selection.getDrive())"
+          ng-click="flash.flashImageToDrive(main.selection.getImage(), main.selection.getCurrentDrive())"
           ng-disabled="main.shouldFlashStepBeDisabled() || main.state.getLastFlashErrorCode()">
             <span ng-bind="flash.getProgressButtonLabel()"></span>
         </progress-button>
--- a/lib/shared/models/selection-state.js
+++ b/lib/shared/models/selection-state.js
@@ -48,10 +48,38 @@ exports.selectDrive = (driveDevice) => {
  * selectionState.toggleDrive('/dev/disk2');
  */
 exports.toggleDrive = (driveDevice) => {
-  if (exports.isCurrentDrive(driveDevice)) {
-    exports.deselectDrive()
+  if (exports.isDriveSelected(driveDevice)) {
+    exports.deselectDrive(driveDevice)
   } else {
     exports.selectDrive(driveDevice)
+  }
+}
+
+/**
+ * @summary Deselect all other drives and keep the current drive's status
+ * @function
+ * @public
+ * @deprecated
+ *
+ * @description
+ * This is a temporary function during the transition to multi-writes,
+ * remove this and its uses when multi-selection should become user-facing.
+ *
+ * @param {String} driveDevice - drive device identifier
+ *
+ * @example
+ * console.log(selectionState.getSelectedDevices())
+ * > [ '/dev/disk1', '/dev/disk2', '/dev/disk3' ]
+ * selectionState.deselectOtherDrives('/dev/disk2')
+ * console.log(selectionState.getSelectedDevices())
+ * > [ '/dev/disk2' ]
+ */
+exports.deselectOtherDrives = (driveDevice) => {
+  if (exports.isDriveSelected(driveDevice)) {
+    const otherDevices = _.reject(exports.getSelectedDevices(), _.partial(_.isEqual, driveDevice))
+    _.each(otherDevices, exports.deselectDrive)
+  } else {
+    exports.deselectAllDrives()
   }
 }
 
@@ -82,19 +110,38 @@ exports.selectImage = (image) => {
 }
 
 /**
- * @summary Get drive
+ * @summary Get all selected drives' devices
+ * @function
+ * @public
+ *
+ * @returns {String[]} selected drives' devices
+ *
+ * @example
+ * for (driveDevice of selectionState.getSelectedDevices()) {
+ *   console.log(driveDevice)
+ * }
+ * > '/dev/disk1'
+ * > '/dev/disk2'
+ */
+exports.getSelectedDevices = () => {
+  return store.getState().getIn([ 'selection', 'devices' ]).toJS()
+}
+
+/**
+ * @summary Get the head of the list of selected drives
  * @function
  * @public
  *
  * @returns {Object} drive
  *
  * @example
- * const drive = selectionState.getDrive();
+ * const drive = selectionState.getCurrentDrive();
+ * console.log(drive)
+ * > { device: '/dev/disk1', name: 'Flash drive', ... }
  */
-exports.getDrive = () => {
-  return _.find(availableDrives.getDrives(), {
-    device: store.getState().getIn([ 'selection', 'drive' ])
-  })
+exports.getCurrentDrive = () => {
+  const device = _.head(exports.getSelectedDevices())
+  return _.find(availableDrives.getDrives(), { device })
 }
 
 /**
@@ -252,7 +299,7 @@ exports.getImageRecommendedDriveSize = () => {
  * }
  */
 exports.hasDrive = () => {
-  return Boolean(exports.getDrive())
+  return Boolean(exports.getSelectedDevices().length)
 }
 
 /**
@@ -272,16 +319,22 @@ exports.hasImage = () => {
 }
 
 /**
- * @summary Remove drive
+ * @summary Remove drive from selection
  * @function
  * @public
  *
+ * @param {String} driveDevice - drive device identifier
+ *
  * @example
- * selectionState.deselectDrive();
+ * selectionState.deselectDrive('/dev/sdc');
+ *
+ * @example
+ * selectionState.deselectDrive('\\\\.\\PHYSICALDRIVE3');
  */
-exports.deselectDrive = () => {
+exports.deselectDrive = (driveDevice) => {
   store.dispatch({
-    type: store.Actions.DESELECT_DRIVE
+    type: store.Actions.DESELECT_DRIVE,
+    data: driveDevice
   })
 }
 
@@ -300,19 +353,28 @@ exports.deselectImage = () => {
 }
 
 /**
+ * @summary Unselect all drives
+ * @function
+ * @public
+ *
+ * @example
+ * selectionState.deselectAllDrives()
+ */
+exports.deselectAllDrives = () => {
+  _.each(exports.getSelectedDevices(), exports.deselectDrive)
+}
+
+/**
  * @summary Clear selections
  * @function
  * @public
  *
  * @example
  * selectionState.clear();
- *
- * @example
- * selectionState.clear({ preserveImage: true });
  */
 exports.clear = () => {
   exports.deselectImage()
-  exports.deselectDrive()
+  exports.deselectAllDrives()
 }
 
 /**
@@ -333,5 +395,29 @@ exports.isCurrentDrive = (driveDevice) => {
     return false
   }
 
-  return driveDevice === _.get(exports.getDrive(), [ 'device' ])
+  return driveDevice === _.get(exports.getCurrentDrive(), [ 'device' ])
+}
+
+/**
+ * @summary Check whether a given device is selected.
+ * @function
+ * @public
+ *
+ * @param {String} driveDevice - drive device identifier
+ * @returns {Boolean}
+ *
+ * @example
+ * const isSelected = selectionState.isDriveSelected('/dev/sdb')
+ *
+ * if (isSelected) {
+ *   selectionState.deselectDrive(driveDevice)
+ * }
+ */
+exports.isDriveSelected = (driveDevice) => {
+  if (!driveDevice) {
+    return false
+  }
+
+  const selectedDriveDevices = exports.getSelectedDevices()
+  return _.includes(selectedDriveDevices, driveDevice)
 }
--- a/lib/shared/store.js
+++ b/lib/shared/store.js
@@ -82,7 +82,9 @@ const selectImageNoNilFields = [
  */
 const DEFAULT_STATE = Immutable.fromJS({
   availableDrives: [],
-  selection: {},
+  selection: {
+    devices: new Immutable.OrderedSet()
+  },
   isFlashing: false,
   flashResults: {},
   flashState: {
@@ -122,25 +124,20 @@ const ACTIONS = _.fromPairs(_.map([
 }))
 
 /**
- * @summary Find a drive from the list of available drives
+ * @summary Get available drives from the state
  * @function
- * @private
+ * @public
  *
- * @param {Object} state - application state
- * @param {String} device - drive device
- * @returns {(Object|Undefined)} drive
+ * @param {Object} state - state object
+ * @returns {Object} new state
  *
  * @example
- * const drive = findDrive(state, '/dev/disk2');
+ * const drives = getAvailableDrives(state)
+ * _.find(drives, { device: '/dev/sda' })
  */
-const findDrive = (state, device) => {
-  /* eslint-disable lodash/prefer-lodash-method */
-
-  return state.get('availableDrives').find((drive) => {
-    return drive.get('device') === device
-  })
-
-  /* eslint-enable lodash/prefer-lodash-method */
+const getAvailableDrives = (state) => {
+  // eslint-disable-next-line lodash/prefer-lodash-method
+  return state.get('availableDrives').toJS()
 }
 
 /**
@@ -160,6 +157,8 @@ const findDrive = (state, device) => {
 const storeReducer = (state = DEFAULT_STATE, action) => {
   switch (action.type) {
     case ACTIONS.SET_AVAILABLE_DRIVES: {
+      // Type: action.data : Array<DriveObject>
+
       if (!action.data) {
         throw errors.createError({
           title: 'Missing drives'
@@ -167,20 +166,20 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
       }
 
       // Convert object instances to plain objects
-      action.data = JSON.parse(JSON.stringify(action.data))
+      const drives = JSON.parse(JSON.stringify(action.data))
 
-      if (!_.isArray(action.data) || !_.every(action.data, _.isPlainObject)) {
+      if (!_.isArray(drives) || !_.every(drives, _.isPlainObject)) {
         throw errors.createError({
-          title: `Invalid drives: ${action.data}`
+          title: `Invalid drives: ${drives}`
         })
       }
 
-      const newState = state.set('availableDrives', Immutable.fromJS(action.data))
+      const newState = state.set('availableDrives', Immutable.fromJS(drives))
 
       const AUTOSELECT_DRIVE_COUNT = 1
-      const numberOfDrives = action.data.length
+      const numberOfDrives = drives.length
       if (numberOfDrives === AUTOSELECT_DRIVE_COUNT) {
-        const drive = _.first(action.data)
+        const [ drive ] = drives
 
         // Even if there's no image selected, we need to call several
         // drive/image related checks, and `{}` works fine with them
@@ -198,27 +197,42 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
           !constraints.isSystemDrive(drive)
 
         ])) {
+          // Auto-select this drive
           return storeReducer(newState, {
             type: ACTIONS.SELECT_DRIVE,
             data: drive.device
           })
         }
-      }
 
-      const selectedDevice = newState.getIn([ 'selection', 'drive' ])
-
-      if (selectedDevice && !_.find(action.data, {
-        device: selectedDevice
-      })) {
+        // Deselect this drive in case it still is selected
         return storeReducer(newState, {
-          type: ACTIONS.DESELECT_DRIVE
+          type: ACTIONS.DESELECT_DRIVE,
+          data: drive.device
         })
       }
 
-      return newState
+      const selectedDevices = newState.getIn([ 'selection', 'devices' ]).toJS()
+
+      // Remove selected drives that are stale, i.e. missing from availableDrives
+      return _.reduce(selectedDevices, (accState, device) => {
+        // Check whether the drive still exists in availableDrives
+        if (device && !_.find(drives, {
+          device
+        })) {
+          // Deselect this drive gone from availableDrives
+          return storeReducer(accState, {
+            type: ACTIONS.DESELECT_DRIVE,
+            data: device
+          })
+        }
+
+        return accState
+      }, newState)
     }
 
     case ACTIONS.SET_FLASH_STATE: {
+      // Type: action.data : FlashStateObject
+
       if (!state.get('isFlashing')) {
         throw errors.createError({
           title: 'Can\'t set the flashing state when not flashing'
@@ -264,6 +278,8 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
     }
 
     case ACTIONS.UNSET_FLASHING_FLAG: {
+      // Type: action.data : FlashResultsObject
+
       if (!action.data) {
         throw errors.createError({
           title: 'Missing results'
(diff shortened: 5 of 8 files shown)


Shou:
I've rebased this with some fixes to pass the tests.

tests/gui/pages/main.spec.js
```diff
@@ -154,6 +154,9 @@ describe('Browser: MainPage', function () {
           }
         })
 
+        console.log(`hasDrive: ${selectionState.hasDrive()}`)
+        console.log(`hasImage: ${selectionState.hasImage()}`)
```
lurch: Is this leftover debugging code?

tests/shared/models/selection-state.spec.js
```diff
@@ -129,16 +131,101 @@ describe('Model: selectionState', function () {
...
+    })
+
+    it('should be able to remove a drive', function () {
+      m.chai.expect(selectionState.getSelectedDevices().length).to.equal(2)
+      selectionState.toggleDrive(this.drives[0].device)
+      m.chai.expect(selectionState.hasDrive()).to.be.true
+      m.chai.expect(selectionState.getSelectedDevices().length).to.equal(1)
+      m.chai.expect(selectionState.getSelectedDevices()).to.deep.equal([ this.drives[1].device ])
+    })
+  })
```
lurch: Some of these unit-tests feel like a bit of a mish-mash? Could you come up with a neat & logical set of more thorough unit-tests?

tests/shared/models/selection-state.spec.js
```diff
@@ -743,7 +830,7 @@ describe('Model: selectionState', function () {
         selectionState.removeImage()
       })
 
-      it('should de-select a previously selected not-recommended drive', function () {
+      it('should not de-select a previously selected not-recommended drive', function () {
```
lurch: Is this a behaviour change, or was the test-description previously incorrect?

lib/shared/models/selection-state.js
```diff
@@ -75,21 +75,54 @@ exports.setImage = (image) => {
...
+exports.getDevices = () => {
+  return store.getState().getIn([ 'selection', 'drives' ]).toJS()
+}
+
+/**
+ * @summary Get all selected drive objects
+ * @function
+ * @public
+ *
+ * @returns {String[]} selected devices
```
lurch: Hmm, the `@summary` says `selected drive objects`, but the `@returns` says a String array of `selected devices`??

I see that the old `.getDrive()` function returned a drive Object, which seems a bit inconsistent naming-wise, given that all the other Drive-related functions just accept a String parameter. So perhaps for clarity, it'd be worth having a `.getSelectedDrives()` function which returns an array of Strings, and a `.getSelectedDriveObjects()` function (if it's indeed needed) which returns an array of drive Objects?

lib/shared/models/selection-state.js
```diff
@@ -308,33 +344,54 @@ exports.removeImage = () => {
...
-    store.dispatch({
-      type: store.Actions.REMOVE_IMAGE
-    })
+    exports.removeImage()
   }
 
-  store.dispatch({
-    type: store.Actions.REMOVE_DRIVE
-  })
+  _.map(exports.getSelectedDevices(), exports.removeDrive)
```
lurch: Would it be worth having a `.removeAllDrives()` function?

lib/shared/models/selection-state.js
```diff
@@ -21,37 +21,37 @@ const store = require('../store')
...
+ * @param {String} driveDevice - drive device identifier
  *
  * @example
- * selectionState.setDrive('/dev/disk2');
+ * selectionState.addDrive('/dev/disk2');
  */
-exports.setDrive = (drive) => {
+exports.addDrive = (driveDevice) => {
   store.dispatch({
     type: store.Actions.SELECT_DRIVE,
```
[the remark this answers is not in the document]
lurch: Alternatively, we could rename `store.Actions.REMOVE_DRIVE` to `store.Actions.UNSELECT_DRIVE` ;-)

lib/shared/store.js
```diff
@@ -151,19 +153,27 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
-      }
-
-      const selectedDevice = newState.getIn([ 'selection', 'drive' ])
 
-      if (selectedDevice && !_.find(action.data, {
-        device: selectedDevice
-      })) {
         return storeReducer(newState, {
-          type: ACTIONS.REMOVE_DRIVE
+          type: ACTIONS.REMOVE_DRIVE,
```
lurch: I still can't figure out why you've got REMOVE_DRIVE both here and on line 170 :-/

lib/shared/store.js
```diff
@@ -414,24 +426,37 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         })
       }
 
-      const selectedDrive = findDrive(state, state.getIn([ 'selection', 'drive' ]))
+      const selectedDevices = state.getIn([ 'selection', 'drives' ])
```
lurch: NIT: should this variable be named `selectedDrives` for consistency?

lib/shared/store.js
```diff
@@ -414,24 +426,37 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
+          return storeReducer(accState, {
+            type: ACTIONS.REMOVE_DRIVE,
+            data: driveDevice
           })
         }
 
-        return state
-      }).setIn([ 'selection', 'image' ], Immutable.fromJS(action.data))
+        return accState
+      }, state).setIn([ 'selection', 'image' ], Immutable.fromJS(action.data))
```
lurch: So it looks like this e.g. removes every selected drive that is now smaller than the newly-selected image? There should definitely be unit-tests for this ;-)

lib/shared/store.js
```diff
@@ -308,7 +318,9 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         })
       }
 
-      return state.setIn([ 'selection', 'drive' ], Immutable.fromJS(action.data))
+      const stateDrives = state.getIn([ 'selection', 'drives' ])
```
lurch: Perhaps this variable should be renamed to `selectionDrives` for consistency with line 457?

lib/shared/models/selection-state.js
```diff
@@ -75,21 +75,54 @@ exports.setImage = (image) => {
...
+ * @summary Get the head of the list of selected drives
+ * @function
+ * @public
+ *
+ * @returns {Object} drive
+ *
+ * @example
+ * const drive = selectionState.getCurrentDrive();
+ */
+exports.getCurrentDrive = () => {
```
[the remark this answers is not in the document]
Shou: Basically it's at the head of the list. We need this – for now – to keep behaviour the same, applying a single drive to be flashed and not a list of drives. It's just a convenience function until we get multi-writing for real.

lib/shared/models/selection-state.js
```diff
@@ -308,33 +344,54 @@ exports.removeImage = () => {
...
+ * }
+ */
+exports.isSelectedDrive = (driveDevice) => {
+  if (!driveDevice) {
     return false
   }
 
-  return drive === _.get(exports.getDrive(), [ 'device' ])
+  const selectedDriveDevices = exports.getSelectedDevices()
+  return Boolean(_.find(selectedDriveDevices, _.partial(_.isEqual, driveDevice)))
```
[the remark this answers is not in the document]
Shou: I copied what worked elsewhere and just wrapped it in Boolean but yours is a better, more concise way, thanks!

lib/shared/store.js
```diff
@@ -36,7 +36,9 @@ const packageJSON = require('../../package.json')
  */
 const DEFAULT_STATE = Immutable.fromJS({
   availableDrives: [],
-  selection: {},
+  selection: {
+    drives: new Immutable.OrderedSet()
```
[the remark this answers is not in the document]
Shou: For now it needs to be ordered, and maybe we'd prefer/require ordered sets later. This is due to it pretending to be single-selection while we're getting multi-writes in fully.

tests/gui/pages/main.spec.js
Shou: Yup!

lib/shared/store.js
Shou: Since it returns a `String[]` containing the `device` fields instead of drive objects, I think it makes more sense with `selectedDevices`.

lurch: For clarity, perhaps we need an Etcher-wide definition of what's a `device` or a `drive` or a `drive object` or a `driveDevice` etc., and then (try to) use those terms consistently throughout the codebase? Ping @jhermsmeier

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
lurch: Hmmm, so does that effectively make this a "WIP PR" until we actually have multi-writing for real? (if so, could you add the "WIP" prefix to the PR title?)

IMHO it makes little sense to merge functionality allowing a user to select multiple drives, while still only allowing the **first** drive they selected to be written to.

lib/shared/store.js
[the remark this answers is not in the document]
lurch: > maybe we'd prefer/require ordered sets later

The reason I'd be :-1: on an OrderedSet is because the user has no way (without unselecting everything again) of changing that ordering.

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
Shou: It's not a WIP PR, more like a transitionary PR towards multi-writes. We can't allow users to select multiple drives when we don't actually allow flashing those multiple drives, if that makes sense.

lib/shared/store.js
Shou: With this PR I've tried to keep `device : String` and `drive : Object`, as a starting point

lib/shared/models/selection-state.js
Shou: I recall using `selectionState.clear({ preserveImage: true })` as a `.removeAllDrives()` so good question, I think we should add the function and replace those `.clear()`s for clarity.

lib/shared/store.js
lurch: But it'd still be good to nail down exactly what `device : String` and `drive : Object` actually mean, with examples of what they might contain in different scenarios, on different OSes, etc :)

lib/shared/store.js
Shou: The first is ensuring the drive should be auto-selected, the other is making sure the selected drives exist since this is an update to the available drives

lib/shared/models/available-drives.js
```diff
@@ -68,3 +68,20 @@ exports.setDrives = (drives) => {
...
+/**
+ * @summary Find a drive by its device field
+ * @function
+ * @public
+ *
+ * @param {String} driveDevice - drive device identifier
+ * @returns {Object|undefined} drive object, or undefined
+ *
+ * @example
+ * const drive = availableDrives.findnDriveByDevice('/dev/sdb')
```
jhermsmeier: Typo: findnDriveByDevice -> findDriveByDevice

jhermsmeier:
Given the number of helper functions that are concerned with finding a drive by its "device" property; maybe we should consolidate them, and put them somewhere accessible to all consuming code?

lib/shared/models/selection-state.js
```diff
@@ -272,76 +312,105 @@ exports.hasImage = () => {
...
  *
- * @param {Object} options - options
- * @param {Boolean} [options.preserveImage] - preserve image
- *
  * @example
- * selectionState.clear();
+ * selectionState.unselectAllDrives()
+ */
+exports.unselectAllDrives = () => {
+  _.map(exports.getSelectedDevices(), exports.unselectDrive)
```
jviotti: Should this be `_.each`?

Shou:
@jhermsmeier I looked into replacing all the finding with one generic helper function, but it would essentially have to be as generic as `_.find` itself, so instead I opted to just use that – the complexity is really in one state getter pattern used in the store so instead I made a helper function for that.

Once https://github.com/resin-io/etcher/pull/2043 is in, this PR is good for reviews.

lib/shared/store.js
```diff
@@ -167,20 +163,20 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
       }
 
       // Convert object instances to plain objects
-      action.data = JSON.parse(JSON.stringify(action.data))
+      const drives = JSON.parse(JSON.stringify(action.data))
```
lurch: This might be another stupid question, but is there a more efficient way of doing this than transitioning to and from JSON?

lib/shared/store.js
```diff
@@ -198,24 +194,35 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
-          type: ACTIONS.DESELECT_DRIVE
+          type: ACTIONS.DESELECT_DRIVE,
+          data: drive.device
         })
       }
 
-      return newState
+      const selectedDevices = newState.getIn([ 'selection', 'devices' ]).toJS()
+
+      return _.reduce(selectedDevices, (accState, device) => {
```
lurch: Could you add a code-comment explaining what this is doing?
I _think_ it's deselecting all drives? Would it be easier to just have a `ACTIONS.DESELECT_ALL_DRIVES` ? (which you could do just once _before_ testing the number of drives, so that you don't have two separate code-paths (as you have now) each doing `ACTIONS.DESELECT_DRIVE`)

lib/shared/store.js
```diff
@@ -305,40 +312,44 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
     }
 
     case ACTIONS.SELECT_DRIVE: {
-      if (!action.data) {
+      const device = action.data
```
lurch: Sorry to keep banging on about consistency (but hey at least I'm consistent :wink: ), but on line 166 you have `const drives` but here you have `const device` :confused:

lib/shared/store.js
```diff
@@ -432,24 +443,39 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         })
       }
 
-      const selectedDrive = findDrive(state, state.getIn([ 'selection', 'drive' ]))
+      const selectedDevices = state.getIn([ 'selection', 'devices' ])
```
lurch: Line 350 (and line 475) has `const selectionDevices` but here you have `const selectedDevices`
(IMHO the latter sounds better)

lib/shared/store.js
```diff
@@ -432,24 +443,39 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
...
 
-        return state
-      }).setIn([ 'selection', 'image' ], Immutable.fromJS(action.data))
+        return accState
+      }, state).setIn([ 'selection', 'image' ], Immutable.fromJS(action.data))
     }
 
     case ACTIONS.DESELECT_DRIVE: {
-      return state.deleteIn([ 'selection', 'drive' ])
+      if (!action.data) {
```
lurch: For consistency with the other actions, you could also assign this `action.data` to a const variable?

jhermsmeier: Let's not assign stuff to vars just for the sake of it; object property access is just fine, and should be preferred, unless too unweildy

lib/shared/store.js
jhermsmeier: @Shou why is this object being copied again? I remember there was some reason for it, but since it's not in the comment...

lib/shared/store.js
lurch: I dunno if the same is true in JavaScript-land, but I remember reading that in Python when you have performance-critical code, doing
```
foo = bar.thing
do_something(foo)
do_something_else(foo)
foo.wibble()
```
can be slightly faster than
```
do_something(bar.thing)
do_something_else(bar.thing)
bar.thing.wibble()
```
because it then does the property-lookup on `bar` just once rather than multiple times. \*shrug\*

> imo the `const something = action.data` should be removed in favor of accessing the props directly

Even if it makes the code less readable? \*shrug\* again ;)

lib/gui/app/components/drive-selector/controllers/drive-selector.js
```diff
@@ -118,6 +118,7 @@ module.exports = function (
           previouslySelected: selectionState.isCurrentDrive(drive.device)
         })
 
+        selectionState.deselectOtherDrives(drive.device)
```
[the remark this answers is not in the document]
Shou: Given that this function isn't meant to be kept when we fully enable multi-selection for multi-writes it isn't something we need to worry about for now

lib/shared/models/selection-state.js
```diff
@@ -82,19 +106,34 @@ exports.selectImage = (image) => {
...
- * @summary Get drive
+ * @summary Get all selected drives' devices
+ * @function
+ * @public
+ *
+ * @returns {String[]} selected drives' devices
+ *
+ * @example
+ * for (driveDevice of selectionState.getSelectedDevices()) {
+ *   console.log(driveDevice)
```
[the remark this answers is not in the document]
Shou: One specifies what it returns (drive device field) and the other specifies the state it affects (drives selected), but I guess we could clarify by renaming it `deselectOtherDrivesByDevice` although that is a bit verbose and I think the `@param` field documents what it selects a drive by quite well already.

lib/shared/models/selection-state.js
```diff
@@ -333,5 +387,29 @@ exports.isCurrentDrive = (driveDevice) => {
...
+ * @returns {Boolean}
+ *
+ * @example
+ * const isSelected = selectionState.isSelectedDrive('/dev/sdb')
+ *
+ * if (isSelected) {
+ *   selectionState.deselectDrive(driveDevice)
+ * }
+ */
+exports.isSelectedDrive = (driveDevice) => {
```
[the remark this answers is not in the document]
Shou: I'm curious, what makes one preferred over the other?

lib/shared/store.js
Shou: We tried using `Object.assign` once but it didn't fulfil the job completely – @jviotti knows the specifics. Maybe we can combine `Object.assign` with a function that takes care of the edge-case you found @jviotti? I imagine it'll be a lot faster than JSON conversion. We can do this in another PR though!

lib/shared/store.js
Shou: `action.data` is used in a polymorphic manner and here it's a device field `String`, whereas on L166 it's a list of drives: `Array<Object>`. We could potentially change the field used in `action` and is definitely worth looking into in another issue/PR.

lib/shared/store.js
Shou: The reason for the `const`s above is because it makes code in the same block easier to write (object shorthand syntax), whereas here it wouldn't help as much, but the discussion of `const` performance is worth continuing in an issue.

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
lurch: Maybe it's just me, but IMHO "isSelectedDrive" sounds like "is this the (one and only) selected drive", whereas "isDriveSelected" sounds like "is this drive selected (while other drives might be selected at the same time too)". IYSWIM :wink:

lib/shared/store.js
lurch: Right. Can you see why I'm getting confused between 'drives' and 'devices' though? :slightly_smiling_face: 
(perhaps more verbose comments is all that's needed to clear things up?)

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
lurch: Maybe I'm just being a bit dim in grokking this code as I've not worked on it myself, but perhaps verbosity is a good thing? \*shrug\*

lib/shared/store.js
Shou: I've renamed them `selectedDevices` now

lib/shared/store.js
Shou: It's deselecting drives that don't exist in `availableDrives` anymore – I've added a few comments explaining that block now.

lib/shared/store.js
```diff
@@ -122,25 +124,19 @@ const ACTIONS = _.fromPairs(_.map([
...
  *
- * @param {Object} state - application state
- * @param {String} device - drive device
- * @returns {(Object|Undefined)} drive
+ * @param {Object} state - state object
+ * @returns {Object} new state
  *
  * @example
- * const drive = findDrive(state, '/dev/disk2');
+ * utils.getDriveByDevice(getAvailableDrives(state), '/dev/sda')
```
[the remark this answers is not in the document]
Shou: Done

lib/shared/models/selection-state.js
```diff
@@ -82,19 +106,34 @@ exports.selectImage = (image) => {
...
+/**
+ * @summary Get the head of the list of selected drives
  * @function
  * @public
  *
  * @returns {Object} drive
  *
  * @example
- * const drive = selectionState.getDrive();
+ * const drive = selectionState.getCurrentDrive();
```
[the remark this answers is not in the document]
Shou: Done

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
Shou: I've improved the docs here now

lib/shared/models/selection-state.js
```diff
@@ -48,13 +48,37 @@ exports.selectDrive = (driveDevice) => {
...
+ * @description
+ * This is a temporary function during the transition to multi-writes,
+ * remove this and its uses when multi-selection should become user-facing.
+ *
+ * @param {String} driveDevice - drive device identifier
+ *
+ * @example
+ * selectionState.deselectOtherDrives('/dev/disk2')
+ */
+exports.deselectOtherDrives = (driveDevice) => {
```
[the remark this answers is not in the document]
Shou: Done! I've also added tests just in case, because I don't trust this function anymore

lib/shared/store.js
Shou: I wish we had TS here and that'd clarify it in the type :P

lib/shared/models/selection-state.js
```diff
@@ -65,15 +65,19 @@ exports.toggleDrive = (driveDevice) => {
...
+ * > [ '/dev/disk1', '/dev/disk2', '/dev/disk3' ]
  * selectionState.deselectOtherDrives('/dev/disk2')
+ * console.log(selectionState.getSelectedDevices())
+ * > [ '/dev/disk2' ]
  */
-exports.deselectOtherDrives = (driveDevice) => {
-  if (exports.isSelectedDrive(driveDevice)) {
-    exports.deselectAllDrives()
-    exports.selectDrive(driveDevice)
+exports.deselectOtherDrives = (device) => {
```
lurch: Umm, perhaps I'm missing something, but why does the `@example` for `deselectOtherDrives` show examples of `getSelectedDevices` ?

Shou: It's to clarify what exactly it does. Do you think it's too verbose?

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
Shou: That makes sense, I've changed it to `isDriveSelected` now

tests/shared/models/selection-state.spec.js
```diff
@@ -225,6 +225,18 @@ describe('Model: selectionState', function () {
...
+
+    describe('.deselectOtherDrives()', function () {
+      it('should deselect other drives', function () {
+        selectionState.deselectOtherDrives('/dev/disk1')
+        m.chai.expect(selectionState.getSelectedDevices()).to.not.include.members([ '/dev/disk2' ])
+      })
+
+      it('should not remove the specified drive', function () {
+        selectionState.deselectOtherDrives('/dev/sdb')
+        m.chai.expect(selectionState.getSelectedDevices()).to.deep.equal([ this.drives[0].device ])
```
lurch: You've already got `/dev/sdb` on the line above, so it might be more logical for this line to say:
```
m.chai.expect(selectionState.getSelectedDevices()).to.deep.equal([ '/dev/sdb' ])
```
?

lib/shared/models/selection-state.js
lurch: Sorry, my fault. I was misreading the diff and hadn't spotted that you'd still got `deselectOtherDrives` in the middle of the example :laughing: 
image

tests/shared/models/selection-state.spec.js
Shou: I've opted for the other route as it looks more consistent with the tests above – use `this.drives` more.

lib/shared/store.js
```diff
@@ -229,6 +231,8 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
     }
 
     case ACTIONS.SET_FLASH_STATE: {
+      // action.data : Object
```
lurch: Nice, but given there's now so many lines that say `// action.data : Object`, I wonder if these too should be a bit more verbose e.g. `// action.data : Object (flash state)` ? (or `// action.data : flash state Object`)
Or perhaps that's unnecessary as that info is implicit from the `case` line (e.g. `ACTIONS.SET_FLASH_STATE`) ?

Shou: Done that now (and also the linter requires sentence-case in comments, so that's fixed too)

lib/shared/store.js
```diff
@@ -157,7 +157,7 @@ const getAvailableDrives = (state) => {
 const storeReducer = (state = DEFAULT_STATE, action) => {
   switch (action.type) {
     case ACTIONS.SET_AVAILABLE_DRIVES: {
-      // action.data : Array<DriveObject>
+      // Type: action.data : Array<DriveObject>
```
lurch: Is this some 'convention' I'm not aware of? If not, I think having two colons on the one line looks a bit strange?

lib/shared/store.js
jviotti: I don't remember the details anymore, sorry :(

jhermsmeier: Yes, good point, let's figure that out in a different PR

tests/shared/models/selection-state.spec.js
lurch: Sweet. And I've just realised that `this.drives` has both `/dev/sdb` (i.e. Linux-style device strings) and `/dev/disk2` (i.e. OSX-style device strings) which is slightly confusing! :laughing: (but that's obviously outside the scope of this PR)

lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -38,7 +38,7 @@ <h4 class="list-group-item-heading">{{ drive.description }}
         </div>
         <span class="list-group-item-section tick tick--success"
           ng-show="modal.constraints.isDriveValid(drive, modal.state.getImage())"
-          ng-disabled="!modal.state.isSelectedDrive(drive.device)"></span>
+          ng-disabled="!modal.state.isDriveSelected(drive.device)"></span>
```
lurch: :+1:  `isDriveSelected` also looks nicer next to `isDriveValid` :)

lib/shared/store.js
Shou: `Type:` isn't really a convention AFAIK, it's just a nod to the user reading it, but I can see why it would confuse and I'm not sure how to resolve this one now. `Type(action.data : Array<DriveObject>)`? `Type - action.data : Array<DriveObject>`?

lib/shared/models/selection-state.js
```diff
@@ -48,13 +48,41 @@ exports.selectDrive = (driveDevice) => {
...
+ * @summary Deselect all other drives and keep the current drive's status
+ * @function
+ * @public
+ * @deprecated
+ *
+ * @description
+ * This is a temporary function during the transition to multi-writes,
+ * remove this and its uses when multi-selection should become user-facing.
+ *
+ * @param {String} device - drive device identifier
```
lurch: Sorry to be a pain, but here you've got `@param {String} device - drive device identifier` and on e.g. line 406 you've got `@param {String} driveDevice - drive device identifier` - perhaps it makes sense to choose one or the other and be consistent?

Shou: No worries, you're right, I made that change to leverage object shorthand syntax but now that I've ended up not using it, it can be `driveDevice` with no drawbacks. It's done now.

lib/shared/store.js
lurch: Anything wrong with just `action.data : Array<DriveObject>` - should be fairly self-evident what it means?

lib/shared/store.js
```diff
@@ -122,25 +124,20 @@ const ACTIONS = _.fromPairs(_.map([
...
  */
-const findDrive = (state, device) => {
-  /* eslint-disable lodash/prefer-lodash-method */
-
-  return state.get('availableDrives').find((drive) => {
-    return drive.get('device') === device
-  })
-
-  /* eslint-enable lodash/prefer-lodash-method */
+const getAvailableDrives = (state) => {
```
lurch: I dunno how all the store-related classes are supposed to interact, but is this a "duplicate" of https://github.com/resin-io/etcher/blob/master/lib/shared/models/available-drives.js#L68 ?

Shou: Yup, but `availableDrives` relies on `store` so we can't expose that, and I don't think we want to expose anything else from the `store` either

tests/shared/models/selection-state.spec.js
lurch: Or maybe it _is_ within the scope if this PR? I was only looking at a single commit before and didn't realise that this PR was adding the entirety of the `this.drives` object.

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
lurch: I _think_ I'm starting to finally understand it all now - and the extra comments help, thank you :)

So we have a `drive` which is an Object, and a `driveDevice` which is a String, and each `drive` Object has a `device` String property (amongst others) which corresponds to the `driveDevice` strings. And "availableDrives" (in the `store`) is a list of these `drive` Objects, and "selectedDevices" (also in the `store`) is a list of `driveDevice` strings. And there's various methods to manipulate/translate these, e.g. getting the `drive` Object(s) that correspond to a particular set of `driveDevice` strings, and vice-versa.

Why is the "availableDrives" stored as an array and not as an object mapping `driveDevice` String -> `drive` Object? (seems like that'd make a lot of the manipulations / translations a bit easier?)

And listing the drive/device-related functions in `selection-state.js` (using a modified notation) we have:
```
selectDrive(String driveDevice)
toggleDrive(String driveDevice)
deselectOtherDrives(String driveDevice)
String[] driveDevices getSelectedDevices()
Object drive getCurrentDrive()
Boolean hasDrive()
deselectDrive(String driveDevice)
deselectAllDrives()
Boolean isCurrentDrive(String driveDevice)
Boolean isDriveSelected(String driveDevice)
```
...so I think the reason I was getting so confused is that we have lots of functions named `xxxDrive` which actually take a `driveDevice` as a parameter, another function named `xxxDrive()` (`getCurrentDrive`) that returns a `drive` Object, and only one function-name named `xxxDevice` (`getSelectedDevices`) that returns a list of `driveDevice`s. :upside_down_face:  (and there's also the asymmetry of `getCurrentDrive` and `isCurrentDrive` returning / taking different types, even though they're named so similarly!)


And again it's something well outside the scope of this particular PR, but I wonder if (in future) it might make sense to split `selection-state.js` up into `drive-selection-state.js` and `image-selection-state.js` ?

lib/shared/store.js
```diff
@@ -122,25 +124,20 @@ const ACTIONS = _.fromPairs(_.map([
...
+ * @summary Get available drives from the state
  * @function
- * @private
+ * @public
  *
- * @param {Object} state - application state
- * @param {String} device - drive device
- * @returns {(Object|Undefined)} drive
+ * @param {Object} state - state object
+ * @returns {Object} new state
```
lurch: The `@returns` here is still wrong, isn't it?

lib/gui/app/pages/main/templates/main.tpl.html
```diff
@@ -66,11 +66,11 @@
               'text-disabled': main.shouldDriveStepBeDisabled()
             }">
             <span class="drive-step step-name"
-              uib-tooltip="{{ main.selection.getDrive().description }} ({{ main.selection.getDrive().displayName }})">
+              uib-tooltip="{{ main.selection.getCurrentDrive().description }} ({{ main.selection.getCurrentDrive().displayName }})">
               <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
-              {{ (main.selection.getDrive().description || "") | middleEllipses:11 }}
+              {{ (main.selection.getCurrentDrive().description || "") | middleEllipses:11 }}
```
lurch: In `selection-state.js` there's a bunch of "helper methods" like `getImageName` and `getImageSize`. Would it also make sense for there to be "helper methods" like `getCurrentDriveDescription()` and `getCurrentDriveSize()` ?

lib/shared/models/selection-state.js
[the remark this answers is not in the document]
Shou: Using a dictionary/map/object instead of a list is a good idea as you should never have two identical mount points – I assume it's faster too. @jhermsmeier what do you think?

As for splitting it, that's definitely something I think we'll go towards once we get drives and HTTP as sources.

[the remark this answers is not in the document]
Shou: And yeah, you've got it right, here :)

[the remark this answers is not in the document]
jhermsmeier: @Shou yeah, I think it's a good idea – that will very much simplify things I reckon; performance wise the impact is negligible compared to everything else going on, as this code isn't called very often (i.e. not 1000+ op/s)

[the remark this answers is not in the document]
jhermsmeier: Do you wanna do that in a follow up PR? This one has become quite the one to navigate :D

lurch:
Sorry all my comments made this PR drag on for so long, but hopefully the code is much better for it :slightly_smiling_face:

PR #1744, opened 2017-09-25
[WIP] feat(GUI): drive multi-selection and multi-writing

Shou:
We add multi-selection and multi-writing, including to the GUI.

Changelog-Entry: Add multi-selection and multi-writing.
Depends: #1736

### DIFF
--- a/lib/child-writer/cli.js
+++ b/lib/child-writer/cli.js
@@ -81,11 +81,14 @@ exports.getBooleanArgumentForm = (argumentName, value) => {
  * });
  */
 exports.getArguments = (options) => {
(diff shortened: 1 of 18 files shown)


Shou:
I'm not requesting any code reviews on this yet, it's up if you want to have a look at how it runs!

jhermsmeier:
@Shou Given the staleness of this, I think we should close this and re-do it, while only adding this functionality to the UI, and not altering the writer (just pick the first drive for the time being), as the writer will undergo some significant changes shortly.

Shou:
@jhermsmeier sounds good, and we'll keep the branch alive as a reference for now. Closed.

Issue #1734, opened 2017-09-19
Accessibility: Improve keyboard access


jhermsmeier:
Currently accessing UI elements via the keyboard is certainly possible; but there's no visual indication of which element is active & selected.

We should add visual cues to allow for easy keyboard-only access & use.

jhermsmeier:
Duplicate of #1745

lurch:
I guess this can be closed?

lurch:
Re-opening in light of #1854

lurch:
Related to resin-io/hq#1024

PR #1745, opened 2017-09-26
feat(GUI): use tabindex and focus to navigate

Shou:
We make navigating with the tab key easier by highlighting focused
elements more visibly, adding `tabindex` attributes to elements, and
making `open-external` links respond to keyboard events.

Change-Type: minor
Changelog-Entry: Improve tab-key navigation through tabindex and visual
improvements.
Connects-To: #1734

### DIFF
--- a/lib/gui/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/components/drive-selector/controllers/drive-selector.js
@@ -251,4 +251,29 @@ module.exports = function (
   this.getDriveStatuses = this.memoizeImmutableListReference((drive) => {
     return this.constraints.getDriveImageCompatibilityStatuses(drive, this.state.getImage())
   })
+
+  /**
+   * @summary Keyboard event drive toggling
+   * @function
+   * @public
+   *
+   * @description
+   * Keyboard-event specific entry to the toggleDrive function.
+   *
+   * @param {Object} drive - drive
+   * @param {Object} $event - event
+   *
+   * @example
+   * <div tabindex="1" ng-keypress="this.keywordToggleDrive(drive, $event)">
+   *   Tab-select me and press enter or space!
+   * </div>
+   */
+  this.keywordToggleDrive = (drive, $event) => {
+    console.log($event.keyCode)
+    const ENTER = 13
+    const SPACE = 32
+    if (_.includes([ ENTER, SPACE ], $event.keyCode)) {
+      this.toggleDrive(drive)
+    }
+  }
 }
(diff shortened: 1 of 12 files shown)


Shou:
This is part of an effort to make Etcher more accessible. @konmouz are there any "Resin-standard" ways of highlighting elements in an accessible manner? Right now I think only the footer elements are visible enough. Examples below.

Unselected button vs Selected button
image image
Selected SVG icon (same for text)
image
Selected help icon (unselected cog for comparison)
image

Shou:
Looks like the HTML linter doesn't like Angular templates. @jviotti do you know how to get around this, e.g. disable it for a single line or something like that?

konmouz:
@Shou, I can't find anything relevant in the style guide.

jhermsmeier:
Relevant issue: https://github.com/resin-io/etcher/issues/1734

jhermsmeier:
@Shou I fear that the difference in selection state of the button & icons is barely visible (I can barely make it out in the screenshots) – can we make the effect a bit more pronounced, or use an outline for them as well?

Shou:
@jhermsmeier I agree with you fully. I was hoping to spawn discussion about it here, although on second thought we should use your issue instead as issues are more apt for potentially lengthy discussion.

konmouz:
The orange frame seems to work fine. Can we have the same behaviour on all buttons?

jhermsmeier:
Ah well, it's happened now. +1 on the orange outline for everything – that's the most familiar for everyone from browserland I guess.
And can you add that issue to the PR description as "Connects To:" (Change-Type is also missing), so we don't forget to close it once this gets merged?

Shou:
How's this? Any tabbable element now uses the orange outlines, but they're also impermanent, with a 10 second timeout before the outlines disappear.

Also added `Change-Type` and `Connects-To`.

jhermsmeier:
I'm getting blue outlines on Mac OS here, as their defaults are OS dependent – we'll need to explicitly set the color on them

screen shot 2017-09-28 at 17 40 25
screen shot 2017-09-28 at 17 40 30

Shou:
Added an orange outline now.

jhermsmeier:
Can we also get the padding on the top right buttons down so that the outline is evenly spaced?

screen shot 2017-10-03 at 15 19 52

Shou:
This should be done now, but let me know if there's anything else

jhermsmeier:
Just checked it out, looks good now! :)

PR #1748, opened 2017-09-29
feat: add babel module for jsx support

Shou:
We add Babel to support JSX, and convert existing ReactDOM files to JSX.
    
Changelog-Entry: Add Babel for JSX support.
Change-Type: minor


Makefile
```diff
@@ -3,6 +3,7 @@
 # ---------------------------------------------------------------------
 
 NPX = ./node_modules/.bin/npx
+babel = ./node_modules/.bin/babel
```
jhermsmeier: This could be run through `npx` as well, i.e. `$(NPX) babel --copy-files ....`

package.json
```diff
@@ -102,6 +106,7 @@
     "eslint-plugin-lodash": "2.3.6",
     "eslint-plugin-node": "5.1.1",
     "eslint-plugin-promise": "3.5.0",
+    "eslint-plugin-react": "^7.4.0",
```
jhermsmeier: The version range selector here will need removing ;)

.babelrc
```diff
@@ -0,0 +1,15 @@
+{
+  "presets": [
+    [
+      "env", {
+        "targets": {
+          "electron": "1.6.6"
```
lurch: Does this need changing, as #1784 has been merged?

Makefile
```diff
@@ -519,9 +519,14 @@ electron-develop:
 		-t electron \
 		-s "$(PLATFORM)"
 
+babel:
+	$(NPX) babel --copy-files --out-dir lib lib && $(NPX) babel --copy-files --out-dir tests tests
+
 sass:
 	$(NPX) node-sass lib/gui/scss/main.scss > lib/gui/css/main.css
 
+transpile: babel sass
```
lurch: You should also add these `babel` and `transpile` targets to the `TARGETS` list.

scripts/ci/test.sh
```diff
@@ -44,6 +44,8 @@ if [ -z "$ARGV_OPERATING_SYSTEM" ] || [ -z "$ARGV_ARCHITECTURE" ]; then
   usage
 fi
 
+make babel
```
lurch: I'm not sure this is the right place for this, should it be added to the existing `make lint test ...` lines below?

Makefile
```diff
@@ -519,9 +519,14 @@ electron-develop:
 		-t electron \
 		-s "$(PLATFORM)"
 
+babel:
+	$(NPX) babel --copy-files --out-dir lib lib && $(NPX) babel --copy-files --out-dir tests tests
```
lurch: So, from what I can tell from a quick skim through this PR, this will convert e.g. `lib/gui/components/safe-webview.jsx` into `lib/gui/components/safe-webview.js` ? Should we also be checking these 'output' files into Git, just like we do for the 'output' sass files? (and if so, I guess we'll also want a `ensure-staged-babel.sh` CI script?) Or if these files should remain outside of git, I guess they should be added to `.gitignore`?

.babelrc
```diff
@@ -0,0 +1,15 @@
...
+        }
+      }
+    ],
+    "react"
+  ],
+  "only": [
+    "lib/**/*.jsx"
```
lurch: In the `babel` Makefile target it seems you also process the `tests` directory - does that need to be added here too?

lurch:
Given the seemingly-minor changes you needed to make to the two `.js` files to convert them to `.jsx`, what benefit does this PR give us? Or is it more of a "setup" phase to make later steps easier?

.babelrc
Shou: Yup! I noticed yesterday and changed it while working on the file selector (based on this PR). I'll push it.

.babelrc
Shou: Yeah, good catch.

Makefile
Shou: @jviotti what do you think about checking them into git vs not doing that?

scripts/ci/test.sh
Shou: It could possibly as long as those are done sequentially, but I do think it's more clear that it happens before those by having it independently above. It's important that the linter and tests run after this.

Shou:
@lurch that's right, it's more of a setup for future contributions. The file browser branch is based on this PR right now.

scripts/ci/test.sh
lurch: In that case this shouldn't be in the `test.sh` script, but elsewhere.

Makefile
LucianBuzzo: I think with this change you should really be compiling the etcher UI into a build directory. Keeping source and compiled files side-by-side is not good.
Using webpack to compile the code into a bundle would be my preferred approach.

jviotti: Yeah, having them together will make it difficult to only ship the built ones in the asar.

Shou: So bundle it is? Sounds good to me

lurch: > having them together will make it difficult to only ship the built ones in the asar.

I guess the same thing applies to the css and scss fles?

Shou:
We're instead working on using Webpack primarily now

Close #1748

PR #1759, opened 2017-10-10
fix(sdk): increase bulk transfer usb timeouts

jviotti:
We experienced timeouts when sending big files (ie ~14 MBs). Setting the
timeout to 0 makes the timeout infinite.

Change-Type: patch


lurch:
Does this mean that if someone unplugs the USB device while flashing (or maybe it accidentally disconnects due to a dodgy cable) then Etcher will just hang forever?

jviotti:
I'll merge this for the demo, but we should surely make sure that's not the case. Do you mind giving it a shot tomorrow?

Issue #1761, opened 2017-10-10
Integrate landr


hobochild:
- **Etcher version:**
- **Operating system and architecture:**
- **Image flashed:**
- **Do you see any meaningful error information in the DevTools?**

PR #1762, opened 2017-10-10
(WIP) Chore: Integrate landr (#1761)

hobochild:
Currently working for `npx` but a global install should work fine. 

open todos: 
     - [ ] crossbrowser testing
     - [ ] render the [banner used by etcher](https://github.com/resin-io/etcher-homepage/tree/master/pages/success-banner)
     - [ ] improve bundle size (see resin-components [discussion](https://github.com/resin-io-modules/resin-components))
     - [ ] auto detect `lead` & `motivation` currently pulled from `landr.conf.js`
     - [ ] clean up styles
     - [ ] add extra navigation (mailing list/gitter)
 
Change-Type: patch


landr.conf.js
```diff
@@ -0,0 +1,62 @@
...
+        'image': 'cross-platform.png',
+        'description': 'Works for everyone,<br/>no more complicated install instructions.'
+      },
+      {
+        'title': 'More on the way',
+        'image': 'coming-soon.png',
+        'description': '50% faster burns, simultaneous writing for multiple drives.<br/><a target="_blank" href="https://github.com/resin-io/etcher/milestones">View our roadmap</a>'
+      }
+    ],
+    motivation: [
```
jviotti: This should be automatically extracted from the repo. Can you create a new PR that adds this as a "Why" section in the README, and then we can pretend it comes from there?

landr.conf.js
```diff
@@ -0,0 +1,62 @@
+const prepAssets = require('./www/prepAssets')
+
+module.exports = {
+  theme: 'landr-theme-basic',
```
jviotti: Where is the source object that contains the extracted data from the repo that this theme is using?

landr.conf.js
```diff
@@ -0,0 +1,62 @@
...
+  middleware: (store, action, next) => {
```
jviotti: Can we replace this with a function that simply takes a filename, and returns an object? For example:

```js
parseReleaseFile: (filename) => {
    // Some magic here...

    return {
        os: 'darwin',
        type: 'cli',
        ...
    }
}
```

www/prepAssets.js
```diff
@@ -0,0 +1,113 @@
...
+        const addition = fn(str)
+        addition && acc.push(addition)
+        return acc
+      },
+      [start]
+    )
+    .join(' ')
+}
+
+const prepAssets = release => {
```
jviotti: I have the impression that we can drastically reduce the size of this file by using a set of regular expressions instead of code. Basically we can have a set of regular expressions to which the input is tested, one by one.

www/static/etcher.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
jviotti: Can we move all these files to `www/assets` instead?

jviotti:
How do I run `landr` here? I tried:

```js
$ ./node_modules/.bin/landr
{ [Error: {"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"}]
  message: '{"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"}',
  code: 401,
  status: 'Unauthorized',
  headers:
   { server: 'GitHub.com',
     date: 'Thu, 12 Oct 2017 10:10:46 GMT',
     'content-type': 'application/json; charset=utf-8',
     'content-length': '83',
     connection: 'close',
     status: '401 Unauthorized',
     'x-github-media-type': 'github.drax-preview; format=json',
     'x-ratelimit-limit': '60',
     'x-ratelimit-remaining': '55',
     'x-ratelimit-reset': '1507806454',
     'access-control-expose-headers': 'ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval',
     'access-control-allow-origin': '*',
     'content-security-policy': 'default-src \'none\'',
     'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
     'x-co% 
```

Should I set a token somewhere? If so, can we provide a better error message that explains what action should be taken?

***

Also, can we add a `make landr` target that runs this?

landr.conf.js
craig-mulligan: I've written it to www/locals.json for now.

jviotti:
Ping @craig-mulligan. What is the state of this PR? Is anything blocking progress?

craig-mulligan:
@jviotti, mainly just time for me to work on it, which has been scarce. There is some on going work on etcher.io which we will need to factor into landr's design for instance adding random pages (etcher pro), so we'll either need create custom theme for etcher or allow the theme user to create pages themselves.

jviotti:
@craig-mulligan In this particular case, the "random page" is the one we will remotely load from the Etcher application, and therefore its look 'n feel is completely independent from the theme we use to build the normal Etcher website. For that reason, I think we can allow users to put HTML files inside `www` (like `www/banner.html`), which will be copied as-is by Landr when generating the website. What do you think?

craig-mulligan:
@jviotti there is a another page being added for etcher-pro advertising/signup forms it is being directed to from the success-banner. The html file approach easy, but it looses the benefits of having access to interrogator data, styles and layout wrappers like the navbar/analytics scripts etc. I think a better approach is add pages as react components. And then allow users to add pages in their landr.conf.js like one does in the theme's landr.conf.js. eg `createRoutes (intergatorData) => { // return and array of routes that is merge with theme routes  }`

jviotti:
@craig-mulligan Fair enough. So basically users will need to provide a React component that gets access to all the props, right? Can we make the route registration process easier though? Maybe we can come up with a naming convention. If the component is declared at `www/banner.jsx`, then the route is `/banner`.

jhermsmeier:
@jviotti what has happened to landr in the meantime? Is this still planned, or shall we shelve this PR?

jviotti:
Lets close this PR, since Landr was put on hold for the time being.

lurch:
Connects to #1761

PR #1768, opened 2017-10-13
fix(GUI): make sure progress button is always rounded

shroudedcode:
At the moment the progress button which has slightly rounded corners allows the `__bar` to overflow. This causes the corners to become angular again which looks weird. I set the button's `overflow` to `hidden` to fix this issue.

*I wasn't able to test whether this fixes the issue for sure because I wasn't able to [make the development dependencies work](https://github.com/resin-io/etcher/blob/master/docs/CONTRIBUTING.md#developing) on my machine, but it should work.*

### DIFF
--- a/lib/gui/components/progress-button/styles/_progress-button.scss
+++ b/lib/gui/components/progress-button/styles/_progress-button.scss
@@ -47,6 +47,8 @@ $progress-button-stripes-animation-duration: 1s;
 .progress-button {
   @extend .button;
   @extend .button-primary;
+  
+  overflow: hidden;
 
   &[active="true"] {
     background-color: $palette-theme-warning-background;
--- a/lib/gui/css/main.css
+++ b/lib/gui/css/main.css
@@ -6315,16 +6315,16 @@ body {
  *   <span class="progress-button__bar" style="width: 50%;"></span>
  * </button>
  */
-.progress-button[active="true"] {
-  background-color: #ff912f; }
-
-.progress-button .progress-button__bar {
-  background-color: #ff9e49; }
-
-.progress-button.progress-button--striped {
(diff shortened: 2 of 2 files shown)


jhermsmeier:
> I wasn't able to test whether this fixes the issue for sure because I wasn't able to make the development dependencies work on my machine, but it should work.

Oh :/ Which one of them didn't work?

jhermsmeier:
Tests are failing because the compiled sass isn't committed; you should be able to just run `make sass` and commit the result as well and it should be fine :)

jhermsmeier:
Also, can you please add `Change-Type: patch` to the bottom of your commit message? We use that for versioning, changelog generation and such (details in the [commit guidelines](https://github.com/resin-io/etcher/blob/master/docs/COMMIT-GUIDELINES.md))

shroudedcode:
I've compiled the styles now but probably messed up a few things, so my commit is now duplicated ...

jhermsmeier:
Looks like you merged the branch into itself :)
No need to close this; you can just revert to your initial commit and then add the css again by doing the following:

```sh
git reset --hard 60bd840
make sass
git add --update
git commit --amend
```

That should reset you to your initial PR commit, and then add the compiled SASS to that commit (without creating a new one), then just `git push --force` to override this PR's commits.

shroudedcode:
I should have fixed everything now.
Thanks for the help @jhermsmeier, much to learn I still have about Git.

jhermsmeier:
Yep, looks good – gave it a test run, and the corners do stay round :)
Thanks for fixing this, good catch!

jhermsmeier:
Oh, and regarding whatever didn't work with the development dependencies – please feel free to open an issue about it; we'd like the dev setup to be as painless as possible, so if there's something that's confusing about the instructions, or something doesn't work straight up, it'd be good to know so we can fix it or at least update the docs.

shroudedcode:
@jhermsmeier Ok, I'll do that now.

PR #1775, opened 2017-10-16
feat(usbboot): add new files that provide better speed

jviotti:
We currently ship with `bootcode.bin` and `start.elf` from the Raspberry
Pi Foundation, which provide a writing speed of about 6 MB/s. This PR
includes new boot files by resin.io that boost the speed to ~20 MB/s.

Change-Type: minor
Changelog-Entry: Increase the flashing speed of usbboot discovered devices.


lib/blobs/usbboot/LICENSE
```diff
@@ -1,30 +0,0 @@
-Copyright (c) 2006, Broadcom Corporation.
```
lurch: This LICENSE shouldn't be removed, as it still applies to the `bootcode.bin` and `start.elf` files.

lib/gui/modules/drive-scanner.js
```diff
@@ -111,8 +111,21 @@ const scan = () => {
...
+        'cmdline.txt': readBlob('usbboot', 'cmdline.txt'),
+        'kernel7.img': readBlob('usbboot', 'kernel7.img'),
+        'bcm2708-rpi-0-w.dtb': readBlob('usbboot', 'bcm2708-rpi-0-w.dtb'),
+        'bcm2708-rpi-b.dtb': readBlob('usbboot', 'bcm2708-rpi-b.dtb'),
+        'bcm2708-rpi-b-plus.dtb': readBlob('usbboot', 'bcm2708-rpi-b-plus.dtb'),
+        'bcm2708-rpi-cm.dtb': readBlob('usbboot', 'bcm2708-rpi-cm.dtb'),
+        'bcm2709-rpi-2-b.dtb': readBlob('usbboot', 'bcm2709-rpi-2-b.dtb'),
+        'bcm2710-rpi-3-b.dtb': readBlob('usbboot', 'bcm2710-rpi-3-b.dtb'),
+        'bcm2710-rpi-cm3.dtb': readBlob('usbboot', 'bcm2710-rpi-cm3.dtb'),
+        'overlays/dwc2.dtbo': readBlob('usbboot', 'dwc2.dtbo')
```
lurch: Any reason not to put this `dwc2.dtbo` in `usbboot/overlays/`?
And then instead of always reading in every blob at start-up, you could only read the blobs that usb-boot actually requests on-demand, with only resorting to reading-from-disk for blobs that haven't already been internally cached.

lib/blobs/usbboot/config.txt
```diff
@@ -0,0 +1,6 @@
+kernel=kernel7.img
```
lurch: This line ought to be removed, so that we can use the same config.txt for both CM1 and CM3.

lib/blobs/usbboot/cmdline.txt
```diff
@@ -0,0 +1 @@
+console=serial0,115200
```
lurch: This is only useful for debugging, so should be removed before we actually make the next production version of Etcher.

lib/blobs/usbboot/LICENSE
lurch: Also, we might need to include a copy of the GPL2, as that'll apply to the kernel.img files we're including?

lib/blobs/usbboot/config.txt
jviotti: Will it still work if I remove this?

lib/blobs/usbboot/cmdline.txt
jviotti: Should I keep an empty `cmdline.txt`, or remove the file altogether?

lib/blobs/usbboot/LICENSE
jviotti: Done!

lib/blobs/usbboot/config.txt
lurch: On the CM3 it defaults to loading `kernel7.img`, and on the CM1 it defaults to loading `kernel.img`, which is exactly what we want. If we leave this line in place, it'll try to use `kernel7.img` for both CM1 and CM3, which will fail when we try to boot an ARMv7 `kernel7.img` on the ARMv6 of the CM1.

lib/blobs/usbboot/cmdline.txt
lurch: usb-boot seems to happily skip over any files that it can't find. So should be safe to remove it.

lib/blobs/usbboot/README.md
```diff
@@ -1,5 +0,0 @@
-usbboot
-=======
-
-The files in this directory were taken from
```
lurch: Might be worth bringing this file back too, to clarify which files are covered by which license?

lib/gui/modules/drive-scanner.js
jviotti: Done!

lib/blobs/usbboot/README.md
jviotti: I divided the files in subdirectories with their own licenses. Do you think that's enough?

lib/gui/modules/drive-scanner.js
```diff
@@ -110,9 +95,16 @@ const scan = () => {
       includeSystemDrives: settings.get('unsafeMode')
     },
     usbboot: {
-      files: {
-        'bootcode.bin': USBBOOT_BOOTCODE_BIN_BUFFER,
-        'start.elf': USBBOOT_START_ELF_BUFFER
+      readFile: (name) => {
+        const blobPath = _.attempt(() => {
+          if (name === 'bootcode.bin' || name === 'start_cd.elf') {
```
lurch: fixup_cd.dat ought to be in the same folder as start_cd.elf

lib/blobs/usbboot/config.txt
```diff
@@ -0,0 +1,6 @@
...
+gpu_mem=16
+enable_uart=1
```
lurch: ...and if we're getting rid of the `console=` in cmdline.txt, we can get rid of this enable_uart line too ;)

lib/blobs/usbboot/cmdline.txt
jviotti: Done!

lib/gui/modules/drive-scanner.js
jviotti: Done!

lib/blobs/usbboot/config.txt
jviotti: Awesome, done!

lib/blobs/usbboot/config.txt
jviotti: Done!

lib/gui/modules/drive-scanner.js
```diff
@@ -110,9 +95,16 @@ const scan = () => {
       includeSystemDrives: settings.get('unsafeMode')
     },
     usbboot: {
-      files: {
-        'bootcode.bin': USBBOOT_BOOTCODE_BIN_BUFFER,
-        'start.elf': USBBOOT_START_ELF_BUFFER
+      readFile: (name) => {
+        const blobPath = _.attempt(() => {
+          if (name === 'bootcode.bin' || name === 'start_cd.elf' || 'fixup_cd.dat') {
```
lurch: I guess this was supposed to be `|| name === 'fixup_cd.dat') {` ? ;-)

Just in case we decide to use the non-cutdown start.elf in future (CD = Cut Down) I guess it probably doesn't hurt to add ` || name === 'start.elf' || name === 'fixup.dat'` in here too? (or perhaps we should only do that if / when we add start.elf)

jviotti: lol, great catch :P Yeah, lets add the conditional when we introduce the file.

lib/shared/sdk/standard/index.js
```diff
@@ -46,19 +46,21 @@ exports.name = 'standard'
  */
 exports.scan = (options = {}) => {
   // eslint-disable-next-line lodash/prefer-lodash-method
-  return drivelist.listAsync().filter((drive) => {
-    return options.includeSystemDrives || !drive.system
-  }).map((drive) => {
+  return drivelist.listAsync().map((drive) => {
+    drive.pending = false
```
lurch: Does this conflict with #1777 ?

lib/gui/modules/drive-scanner.js
```diff
@@ -110,9 +95,16 @@ const scan = () => {
...
+            return path.join('raspberrypi', name)
+          }
+
+          return name
+        })
+
+        return fs.readFileAsync(path.join(BLOBS_DIRECTORY, 'usbboot', blobPath))
```
lurch: I can't quite tell, but does this cache the files, so that a second request for the same filename (e.g. config.txt) doesn't result in multiple fs.readFileAsync calls?

lib/shared/sdk/standard/index.js
jviotti: Accidental addition :P Fixed now.

lib/gui/modules/drive-scanner.js
jviotti: Good call. Done!

lib/gui/modules/drive-scanner.js
```diff
@@ -110,9 +97,22 @@ const scan = () => {
       includeSystemDrives: settings.get('unsafeMode')
     },
     usbboot: {
-      files: {
-        'bootcode.bin': USBBOOT_BOOTCODE_BIN_BUFFER,
-        'start.elf': USBBOOT_START_ELF_BUFFER
+      readFile: (name) => {
+        const blobPath = _.attempt(() => {
```
lurch: This blobPath stuff can be moved to after the `if (BLOBS_CACHE[name]) {` block :)

lib/gui/modules/drive-scanner.js
lurch: I guess we could even change this to:
```
if (_.includes(['bootcode.bin', 'start_cd.elf', 'fixup_cd.dat'], name)) {
```
:question:

lib/gui/modules/drive-scanner.js
jviotti: True. Fixed!

lib/gui/modules/drive-scanner.js
jviotti: Good call. Done!

jviotti:
@lurch What do you about merging this as-is, with just CM3 support, and then creating a new PR for CM1 support?

lib/gui/modules/drive-scanner.js
```diff
@@ -63,21 +64,7 @@ const emitter = new EventEmitter()
...
-  path.join(BLOBS_DIRECTORY, 'usbboot', 'bootcode.bin'))
-
-/**
- * @summary The usbboot "start.elf" buffer
- * @type {Buffer}
- * @constant
- */
-const USBBOOT_START_ELF_BUFFER = fs.readFileSync(
-  path.join(BLOBS_DIRECTORY, 'usbboot', 'start.elf'))
+const BLOBS_CACHE = {}
```
jhermsmeier: Not sure how I feel about the blobs being cached in a GUI module – any alternatives?

lib/gui/modules/drive-scanner.js
```diff
@@ -110,9 +97,20 @@ const scan = () => {
...
+
+        const blobPath = _.includes([
+          'bootcode.bin',
+          'start_cd.elf',
+          'fixup_cd.dat'
+        ], name) ? path.join('raspberrypi', name) : name
+
+        return fs.readFileAsync(path.join(BLOBS_DIRECTORY, 'usbboot', blobPath)).tap((buffer) => {
+          BLOBS_CACHE[name] = buffer
+        })
```
jhermsmeier: Same as above regarding the device specific config here

lib/gui/modules/drive-scanner.js
jviotti: Can you explain the concern? We could not cache them, which I think its fine anyways (the speed improvement of this particular caching case is probably minimal)

lib/gui/modules/drive-scanner.js
jhermsmeier: Also the config being a function feels a bit off; what do you think about moving this into the usbboot adapter?

jviotti: We could move the logic to the adapter, and make the user pass a path from where files will be read as needed, but then it means that we lose the potential flexibility of streaming the files from somewhere else (something we might want to do at some point), or any other way of reading the files.

lib/gui/modules/drive-scanner.js
jhermsmeier: Just as ref from the Flowdock thread; I'd move it into the usbboot adapter itself (as the UI shouldn't be concerned with config or internal state mgmt of something that's not the UI)

lib/gui/modules/drive-scanner.js
jhermsmeier: Ah, we'll figure something out for the streams. We could still pass them through the options, as they're one-time consumables anyway, for example.

jviotti: Done, take a look now!

lib/blobs/usbboot/README.md
lurch: Well having the files in separate directories makes it harder to go back to `rpiboot` for testing purposes, but I guess that's not really much of a problem.

jviotti:
Awesome, thanks a lot for the files!

lib/shared/sdk/usbboot/index.js
```diff
@@ -142,6 +142,41 @@ const USB_ID_RADIX = 16
...
+ */
+const queryBlobFromCache = (name, fallback) => {
+  if (BLOBS_CACHE[name]) {
+    return Bluebird.resolve(BLOBS_CACHE[name])
+  }
+
+  return fallback(name).tap((buffer) => {
+    BLOBS_CACHE[name] = buffer
+  })
+}
```
lurch: Doesn't really matter, but when I've written code like this before I tend to do something like the following pseudocode:
```
function readCachedFile(filename) {
    if (filename not in cache) {
        cache[filename] = readFile(filename);
    }
    return cache[filename];
}
```
but perhaps that wouldn't work with the Promise-based stuff you're doing here, which I still struggle with sometimes ;-)

lib/shared/sdk/usbboot/index.js
```diff
@@ -260,35 +296,38 @@ const startFileServer = (device, endpoint, files) => {
       return Bluebird.try(() => {
         if (fileMessage.command === protocol.FILE_MESSAGE_COMMANDS.GET_FILE_SIZE) {
           debug(`Getting the size of ${fileMessage.fileName}`)
-          const fileSize = _.get(files, [ fileMessage.fileName, 'length' ])
 
-          if (_.isNil(fileSize)) {
+          return queryBlobFromCache(fileMessage.fileName, readFile).then((fileBuffer) => {
```
lurch: Oh yeah, I'd forgotten that the usbboot protocol requests the filesize first - that's another very good reason to support file-caching :)

lib/shared/sdk/usbboot/index.js
jviotti: Yeah, promises make that a bit hard.

lurch:
The usbboot-server stuff itself seems a bit flaky (which is a separate discussion...), but with the latest commit I've just pushed, I've successfully tested these kernels as working in Etcher under both Ubuntu Linux 14.04 x64 and Windows 7 x64 :grinning:   (but the latter requires drivers to be installed first, which is also a separate conversation :wink: )

PR #1776, opened 2017-10-16
fix(usbboot): opening device debug message prints undefined

jviotti:
We have a debug message that prints `device.name`, which is not a valid
property, and therefore the debug logs show `undefined` instead of the
USB id pair.

Change-Type: patch

### DIFF
--- a/lib/shared/sdk/usbboot/index.js
+++ b/lib/shared/sdk/usbboot/index.js
@@ -337,7 +337,7 @@ exports.scan = (options) => {
     device.adaptor = exports.name
 
     // We need to open the device in order to access _configDescriptor
-    debug(`Opening device: ${device.name}`)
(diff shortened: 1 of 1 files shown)


lib/shared/sdk/usbboot/index.js
```diff
@@ -337,7 +337,7 @@ exports.scan = (options) => {
     device.adaptor = exports.name
 
     // We need to open the device in order to access _configDescriptor
-    debug(`Opening device: ${device.name}`)
+    debug(`Opening device: ${idPair}`)
```
lurch: Looking at the variables above, would the correct-fix here be to change `${device.name}` to `${device.displayName}` ?

jviotti: Yeah, I think that's correct. Fixed!

PR #1779, opened 2017-10-16
fix(GUI): don't show the "too small" badge if the size is null

jviotti:
Some devices don't have a size, like USB devices in the usbboot adaptor.
The `.isDriveLargeEnough()` correctly returns `false` in this case,
however we don't want to show the `TOO SMALL` badge for aesthetics
purposes.

So if a drive has a size that equals `null`, we don't allow such drive
to be selected, and we don't show a badge for it.

### DIFF
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -389,7 +389,7 @@ exports.getDriveImageCompatibilityStatuses = (drive, image) => {
       type: exports.COMPATIBILITY_STATUS_TYPES.ERROR,
       message: exports.COMPATIBILITY_STATUS_MESSAGES.LOCKED
     })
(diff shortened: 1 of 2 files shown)


lurch:
Would a "cleaner" approach be to make `.isDriveLargeEnough()` return `null` if the drive-size is `null`? That would prevent the drive-size being checked in two separate places... (but it might cause other problems if there's code elsewhere which relies on `.isDriveLargeEnough()` always returning either `true` or `false`?)

jviotti:
Yeah, I think that would be a bit confusing in the JavaScript world, where `null` is considered a falsy value, and thus it'd be a nightmare to deal with a function that returns `null` and `false` on different cases.

lurch:
Fair enough, I'll approve then :)

PR #1781, opened 2017-10-16
refactor(SDK): make adapter scan functions event based

jviotti:
This change will allow us to start emitting progress events out of the
adaptors scan functions.

### DIFF
--- a/lib/shared/sdk/index.js
+++ b/lib/shared/sdk/index.js
@@ -54,6 +54,10 @@ const ADAPTORS = [
  */
 exports.scan = (options) => {
   return Bluebird.all(_.map(ADAPTORS, (adaptor) => {
-    return adaptor.scan(_.get(options, [ adaptor.name ], {}))
+    return new Bluebird((resolve, reject) => {
+      adaptor.scan(_.get(options, [ adaptor.name ], {}))
+        .on('done', resolve)
+        .on('error', reject)
+    })
   })).then(_.flatten)
 }
--- a/lib/shared/sdk/index.js
+++ b/lib/shared/sdk/index.js
@@ -18,25 +18,36 @@
 
 const Bluebird = require('bluebird')
 const _ = require('lodash')
+const sdk = module.exports
 
 /**
- * @summary The list of loaded adaptors
+ * @summary The list of loaded adapters
  * @type {Object[]}
  * @constant
  */
-const ADAPTORS = [
+const ADAPTERS = [
   require('./standard'),
   require('./usbboot')
 ]
 
 /**
- * @summary Scan for drives using all registered adaptors
+ * @summary Initialised adapters
+ * @type {Object<String,Adapter>}
+ * @constant
+ */
+sdk.adapters = _.reduce(ADAPTERS, (adapters, Adapter) => {
+  adapters[Adapter.name] = new Adapter()
+  return adapters
+}, {})
+
+/**
+ * @summary Scan for drives using all registered adapters
  * @function
  * @public
  *
  * @description
  * The options object contains options for all the registered
- * adaptors. For the `standard` adaptor, for example, place
+ * adapters. For the `standard` adapter, for example, place
  * options in `options.standard`.
  *
  * @param {Object} options - options
@@ -52,12 +63,16 @@ const ADAPTORS = [
  *   console.log(drives)
  * })
  */
-exports.scan = (options) => {
-  return Bluebird.all(_.map(ADAPTORS, (adaptor) => {
+sdk.scan = (options) => {
+  return Bluebird.all(_.map(sdk.adapters, (adapter) => {
     return new Bluebird((resolve, reject) => {
-      adaptor.scan(_.get(options, [ adaptor.name ], {}))
-        .on('done', resolve)
-        .on('error', reject)
+      adapter.scan(_.get(options, [ adapter.id ], {}), (error, devices) => {
+        if (error) {
+          reject(error)
+        } else {
+          resolve(devices)
+        }
+      })
     })
   })).then(_.flatten)
 }
--- a/lib/shared/sdk/standard/index.js
+++ b/lib/shared/sdk/standard/index.js
@@ -18,63 +18,78 @@
 
 const _ = require('lodash')
 const Bluebird = require('bluebird')
-const EventEmitter = require('events').EventEmitter
+const EventEmitter = require('events')
 const drivelist = Bluebird.promisifyAll(require('drivelist'))
 
 /**
- * @summary The name of this adaptor
+ * @summary StandardAdapter
+ * @class
+ */
+class StandardAdapter extends EventEmitter {
+  /**
+   * @summary StandardAdapter constructor
+   * @class
+   * @example
+   * const adapter = new StandardAdapter()
+   */
+  constructor () {
+    super()
+
+    /** @type {String} Adapter name */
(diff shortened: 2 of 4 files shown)


lib/shared/sdk/standard/index.js
```diff
@@ -60,5 +66,13 @@ exports.scan = (options = {}) => {
     }
 
     return drive
+  }).filter((drive) => {
+    return options.includeSystemDrives || !drive.system
+  }).then((drives) => {
+    emitter.emit('done', drives)
+  }).catch((error) => {
+    emitter.emit('error', error)
```
jhermsmeier: Shouldn't the `.catch()` happen before the `.then()`?

lib/shared/sdk/usbboot/index.js
```diff
@@ -383,5 +391,11 @@ exports.scan = (options) => {
   // See http://bluebirdjs.com/docs/api/promise.map.html
   }, {
     concurrency: 5
+  }).then((devices) => {
+    emitter.emit('done', devices)
+  }).catch((error) => {
+    emitter.emit('error', error)
```
jhermsmeier: Same here (?)

jhermsmeier:
Generally I think we should have the adapters be instances of an EventEmitter, and have the adapters themselves emit events for things like device detections etc. – not for individual progress i.e. though.

jhermsmeier:
Can't request a review from you @jviotti because you're the author, so yeah – if you wanna have a look now :)

jhermsmeier:
Btw, I also changed the word adaptor -> adapter, as that's the main used variant, which is also in the dictionaries.

lib/shared/sdk/index.js
```diff
@@ -52,8 +63,16 @@ const ADAPTORS = [
  *   console.log(drives)
  * })
  */
-exports.scan = (options) => {
-  return Bluebird.all(_.map(ADAPTORS, (adaptor) => {
-    return adaptor.scan(_.get(options, [ adaptor.name ], {}))
+sdk.scan = (options) => {
+  return Bluebird.all(_.map(sdk.adapters, (adapter) => {
+    return new Bluebird((resolve, reject) => {
+      adapter.scan(_.get(options, [ adapter.id ], {}), (error, devices) => {
```
jhermsmeier: I'm thinking this (the options config) would eventually eventually move to something like `sdk.init(options)` or so, which would then initialise the adapters with the given options, like on line 38 here. That'll enable continuous scanning on the adapter's behalf and on its own schedule without having to call `.scan()` on each adapter every time, as well as simpler error handling in one place.

Not sure whether to do that in this PR, or do that in a follow up, as the general set up needs some more thinking and spec-ing wrt the Adapters Spec.

lib/shared/sdk/usbboot/index.js
```diff
@@ -158,230 +133,284 @@ const usbIdToString = (id) => {
...
-      message: 'LIBUSB_TRANSFER_STALL'
-    }, {
-      message: 'LIBUSB_TRANSFER_ERROR'
-    }, (error) => {
-      debug(`Got ${error.message} when reading a command, assuming everything is done`)
-      return {
-        command: protocol.FILE_MESSAGE_COMMANDS.DONE
+class USBBootAdapter extends EventEmitter {
+  /**
+   * @summary StandardAdapter constructor
```
lurch: Typo here? Should it be `USBBootAdapter constructor` ?

jhermsmeier: Whoops. Fixed!

lib/shared/sdk/index.js
jviotti: I think that we should go for the lowest hanging fruit now, in order to make a release, and then we can take the time to discuss more radical changes.

lib/shared/sdk/standard/index.js
```diff
@@ -17,48 +17,77 @@
...
+   * @summary StandardAdapter constructor
+   * @class
+   * @example
+   * const adapter = new StandardAdapter()
+   */
+  constructor () {
+    super()
+
+    /** @type {String} Adapter name */
+    this.id = this.constructor.id
```
jviotti: Interesting. I wonder if we should have a class called `BaseAdapter` that extends `EventEmitter` that does this out of the box, and the the adapters extend that one instead.

jviotti:
I can't approve because I'm the author, but looks good code-wise. I'll test it out in a minute.

jviotti:
Works great! I tested both the standard and the usbboot adapter.

lib/shared/sdk/standard/index.js
jhermsmeier: I was thinking the same for a while, but that'll make creating third-party adapters a little messy. The ultimate lazy would be to just use the constructor.name, but that might break (i.e. if people name their constructors / classes the same), so just using costructor.id (without the instance property) would be the most minimal, while still being explicit I guess.

jviotti:
@jhermsmeier You can proceed to rebase this one :)

jhermsmeier:
@jviotti rebased, but not 1000% sure I didn't screw up the rebase – so if you could double-check it would be ace

lib/shared/sdk/usbboot/index.js
```diff
@@ -193,251 +133,318 @@ const usbIdToString = (id) => {
...
+   * adapter.startFileServer(device, device.interfaces(0).endpoint(1), (name) => {
+   *   return fs.readFileAsync(name)
+   * }).then(() => {
+   *   console.log('Done!')
+   * })
+   */
+  startFileServer (device, endpoint, readFile) {
+    debug('Listening for file messages')
+    return protocol
+        .read(device, protocol.FILE_MESSAGE_SIZE)
```
jviotti: I think there is way too much indentation here.

lib/shared/sdk/usbboot/index.js
```diff
@@ -193,251 +133,318 @@ const usbIdToString = (id) => {
...
+        addresses.interface = 0
+        addresses.endpoint = 1
+      } else {
+        addresses.interface = 1
+        addresses.endpoint = 3
       }
 
+      const deviceInterface = device.interface(addresses.interface)
+      debug(`Claiming interface: ${addresses.interface}`)
+      deviceInterface.claim()
```
jviotti: Looks like we lost the `catch` block that is on `master`: https://github.com/resin-io/etcher/blob/master/lib/shared/sdk/usbboot/index.js#L404

jviotti:
Left a couple of comments. I'll send a new PR that splits the main code chunks in `usbboot/index.js` to another file, so then we can change to adapter specific one without having to deal with that code.

lib/shared/sdk/index.js
Shou: I think if you do this here it'll pave way for https://github.com/resin-io/etcher/pull/1797 and https://github.com/resin-io/etcher/pull/1800 since we can simply add the progress to the devices and `emit` them.

jviotti:
@jhermsmeier So we will delay this PR for the next release, right?

jhermsmeier:
@jviotti nah, this should be good to put in now – and I don't see why we should delay it, or why do you mean?

lib/shared/sdk/usbboot/index.js
jhermsmeier: Oh yes, fixed now!

PR #1783, opened 2017-10-16
upgrade(package): Update outdated dependencies

jhermsmeier:
Updated dependencies
--------------------

**Major:**

- chalk 1.1.3 -> 2.1.0
- debug 2.6.0 -> 3.1.0
- file-type 4.1.0 -> 7.1.0
- ~~lzma-native 1.5.2 -> 3.0.1~~ (subsequently reverted)
- node-ipc 8.9.2 -> 9.1.1
- sudo-prompt 6.1.0 -> 7.1.1
- yargs 4.7.1 -> 9.0.1

**Minor:**

- angular-moment 1.0.1 -> 1.1.0
- bindings 1.2.1 -> 1.3.0
- bluebird 3.4.1 -> 3.5.1
- lodash 4.13.1 -> 4.17.4
- mountutils 1.2.2 -> 1.3.2
- nan 2.3.5 -> 2.7.0
- prop-types 15.5.9 -> 15.6.0
- react 15.5.4 -> 15.6.2
- react-dom 15.5.4 -> 15.6.2
- redux 3.5.2 -> 3.7.2
- request 2.81.0 -> 2.83.0
- resin-cli-visuals 1.3.1 -> 1.4.0
- semver 5.1.1 -> 5.4.1
- trackjs 2.3.1 -> 2.9.0
- unbzip2-stream 1.0.11 -> 1.2.5
- uuid 3.0.1 -> 3.1.0
- yauzl 2.6.0 -> 2.8.0

**Patch:**

- angular 1.6.3 -> 1.6.6
- angular-if-state 1.0.0 -> 1.0.1
- angular-middle-ellipses 1.0.0 -> 1.0.1
- angular-seconds-to-date 1.0.0 -> 1.0.1
- angular-ui-bootstrap 2.5.0 -> 2.5.6
- angular-ui-router 0.4.2 -> 0.4.3
- bootstrap-sass 3.3.6 -> 3.3.7
- drivelist 5.2.4 -> 5.2.6
- flexboxgrid 6.3.0 -> 6.3.1
- immutable 3.8.1 -> 3.8.2
- mime-types 2.1.15 -> 2.1.17
- xml2js 0.4.17 -> 0.4.19

Dev-dependencies
----------------

**Major:**

- electron-mocha 3.3.0 -> 4.0.3
- file-exists 1.0.0 -> 5.0.0
- npx 5.2.0 -> 9.6.0

**Minor:**

- asar 0.10.0 -> 0.13.0
- ~~electron-builder 19.9.1 -> 19.37.2~~ (rolled back)
- eslint-plugin-lodash 2.3.6 -> 2.4.5
- eslint-plugin-node 5.1.1 -> 5.2.0
- eslint-plugin-promise 3.5.0 -> 3.6.0
- html-angular-validate 0.1.9 -> 0.2.1
- node-gyp 3.5.0 -> 3.6.2
- pkg 4.1.1 -> 4.2.5
- sass-lint 1.10.2 -> 1.12.0
- versionist 2.8.1 -> 2.11.2

**Patch:**

- angular-mocks 1.6.3 -> 1.6.6
- eslint 3.19.0 -> 4.9.0
- eslint-plugin-jsdoc 3.1.1 -> 3.1.3
- nock 9.0.9 -> 9.0.22
- tmp 0.0.31 -> 0.0.33

Dependencies NOT upgraded due to incompatibility
------------------------------------------------

- ~~electron 1.6.6 -> 1.7.9~~ (happened in #1784)
- electron-builder 19.9.1 -> 19.37.2
- mochainon 1.0.0 -> 2.0.0
- pretty-bytes 1.0.4 -> 4.0.2
- react 15.x.x -> 16.0.0
- react2angular 1.1.3 -> 3.0.1

Change-Type: patch


Shou:
What's holding electron back?

jhermsmeier:
Not entirely sure; haven't dug into it, but I think there are some API changes that cause this:

screen shot 2017-10-16 at 19 42 24

jhermsmeier:
I'll get to upgrading Electron itself in a separate PR

jhermsmeier:
@Shou turns out it was only due to the order in which I did the bumps (native deps...) – Electron update PR is #1784

lurch:
Appveyor issues:
```
There are unstaged sass changes. Please commit the result of:
    make sass
```

Travis issues:
```
node version: v6.10.3
npm version: 3.10.10
npm ERR! Darwin 15.6.0
npm ERR! argv "/Users/travis/.nvm/versions/node/v6.10.3/bin/node" "/Users/travis/.nvm/versions/node/v6.10.3/bin/npm" "shrinkwrap" "--dev"
npm ERR! node v6.10.3
npm ERR! npm  v3.10.10
npm ERR! Problems were encountered
npm ERR! Please correct and try again.
```
and
```
Using electron-builder.yml configuration file
Error: Configuration is invalid.
```

jhermsmeier:
The issues with sass on AppVeyor are due to it now using CRLF line endings on Windows, I think. Will see if there's an easy and quick fix for that.
Regarding electron-builder – I might just revert that to the previously known to work version.

lurch:
> The issues with sass on AppVeyor are due to it now using CRLF line endings on Windows, I think.

\*sigh\* I thought I'd already fixed that with the .gitattributes file? Can't remember the full details though. Perhaps we should modify the `make sass` target automatically run something like `dos2unix` over the css file it outputs?

lurch:
...and just out or curiosity, what went wrong with the lzma upgrade? I've not looked at it in a while, but I thought lzma and node-gyp had collaborated and fixed everything so it should all be working properly again? Ahhh, or was it because we need to patch lzma in order to be able to use it in the CLI or something?

lurch:
I guess `mountutils` should now be 1.3.3 https://github.com/resin-io-modules/mountutils/pull/56
and `drivelist` should now be 5.2.7 https://github.com/resin-io-modules/drivelist/pull/232

Ah, all the fun of trying to manage multiple moving targets all at once :laughing:

jhermsmeier:
There are several issue with the lzma@3.x.x; those you mentioned and that it's looking for the wrong binding (it's looking for the node runtime binding, whereas it should be electron) when loaded... that'll need some investigation and probably PRs to get that going.

scripts/ci/ensure-npm-shrinkwrap-versions.sh
```diff
@@ -26,15 +26,15 @@ HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+MODULE_NAMES=()
+MODULE_VERSIONS=()
 
 if [[ ${#MODULE_NAMES[@]} -ne ${#MODULE_VERSIONS[@]} ]]; then
     echo "Lengths of MODULE_NAMES and MODULE_VERSIONS arrays must match"
     exit 1
 fi
 
-for i in ${!MODULE_NAMES[@]}; do
+for i in "${!MODULE_NAMES[@]}"; do
```
lurch: Are you sure the change on this line is necessary? I've just been doing some local testing, and regardless of whether there's 0, 1 or 2 entries in the `MODULE_NAMES` array, this line works just fine without the quotes.

jhermsmeier: That's just a mend from a [shellcheck](https://github.com/koalaman/shellcheck) warning to "avoid array expansions and re-splitting elements [SC2068]".

jhermsmeier:
Turns out `electron-builder` was blowing up due to a breaking change in a minor version (sigh), where the `icon` key was moved to the top-level `linux` key in the config.

Now it's detonating because of

```
Unhandled rejection Error: linux.icon must be set to an icon set directory,
but /etcher/assets/icon.png is not a directory.
Please see https://electron.build/configuration/configuration#LinuxBuildOptions-icon
```

Good fun included in that *minor version bump* 😭

jhermsmeier:
And `npm@3.10.10` continues on it's way of making me lose my sanity with shrinkwrap diffs between Mac OS and Linux.

jhermsmeier:
Oh ffs, electron-builder...

```
{:timestamp=>"2017-10-19T10:01:08.763034+0000", :message=>"Process failed: tar failed (exit code 2). Full command was:[\"tar\", \"-C\", \"/tmp/package-dir-staging-593ea4f2e20f8c9e5eb59a46992fa22020946abb1e57aee242fd443c5050\", \"-I'/etcher/node_modules/7zip-bin/7x.sh'\", \"--numeric-owner\", \"--owner\", \"0\", \"--numeric-owner\", \"--group\", \"0\", \"-cf\", \"/tmp/package-deb-build-7f0090f84bfa442cd512510a3b5fb6ddd647362e681160d3b9d6720147d8/data.tar.xz\", \".\"]", :level=>:error}
```

Rolling that one back.

scripts/ci/ensure-npm-shrinkwrap-versions.sh
lurch: Side-note: Would it be worth adding shellcheck to our lint-step?

jhermsmeier: Generally I'd like to, but installing it on Windows might be a bit of a pain, plus sorting out the false positives via config – we're probably better off occasionally running it ourselves.

I think the general solution here is to just not have bash scripts in the first place – ideally we'd port them to JS, as we have a cross-platform runtime w/ Node anyways, and we wouldn't have to worry about diff bash versions, and could also eliminate it as a dep on Windows.

lurch: Ahh, I thought that the reason that these scripts were written in Bash was so that we had an "out of band" mechanism, so that the scripts would still run if something went wonky with the nodeJS setup? (But perhaps that was a false assumption on my part? I agree that these would be much easier to maintain if written in a "proper" programming language, rather than trying to squeeze them into Bash's limitations)

lurch:
Regarding the sass errors - as I (kinda) mentioned in #1192 , AFACIT `make sass` simply merges together a whole bunch of different CSS files, and unfortunately it seems like some of the input CSS files use LF line-endings, and others use CRLF line-endings, and node-sass just seems to mash all the input files together (doing no line-end conversions itself), meaning that the final output file is an unholy mess of both LF **and** CRLF :-/  That's why I suggested in an earlier comment that we might need to run `dos2unix` (or `unix2dos`?) over the final output file, as that'd then hopefully unify everything?
(or perhaps this behaviour is fixed in the newer node-sass you're using here?)

scripts/ci/ensure-npm-shrinkwrap-versions.sh
jhermsmeier: > so that the scripts would still run if something went wonky with the nodeJS setup?

If something with that was to go wrong, then running anything else wouldn't make much sense imo (or am I missing something?) – so we might as well dump bash.

jhermsmeier:
I've modified the `make sass` task to spew out a diff now; and lo and behold:
```diff
diff --git a/lib/gui/css/main.css b/lib/gui/css/main.css
index c26d212..38dcae0 100644
--- a/lib/gui/css/main.css
+++ b/lib/gui/css/main.css
@@ -1,3 +1,4 @@
+C:\projects\etcher\node_modules\node-sass\bin\node-sass
 @charset "UTF-8";
 /*
  * Copyright 2016 resin.io
```
Now the question is: What prepends the command to `main.css` on Windows? :D

scripts/ci/ensure-npm-shrinkwrap-versions.sh
lurch: Ping @jviotti - is this something worth attempting? (after we've got the next release out of the way, obviously)

lurch:
LOL, wtf?! ;-D

scripts/ci/ensure-npm-shrinkwrap-versions.sh
jhermsmeier: Doing it gradually would be a relaxed approach, doing it all at once might be a bit heavy and bug-prone

jhermsmeier:
That might actually be `npx` printing the command to stderr...

lurch:
I didn't realise this hadn't got merged while I was away - is it not going to be included in the next Etcher release?

jhermsmeier:
I'm not sure anymore if we wanted to do this with the coming release; I thought there was some argument against it? Also, as a side-note this kind of depends on https://github.com/resin-io/etcher/pull/1816

jhermsmeier:
Closing this for now; as some of these updates are already stale again, and it probably makes more sense to update in smaller batches anyways.

lurch:
I made a couple of edits to the first comment, linking to relevant PRs :slightly_smiling_face:

PR #1784, opened 2017-10-16
upgrade(package): Update Electron v1.6.6 -> v1.7.9

jhermsmeier:
This updates Electron from v1.6.6 to v1.7.9

Change-Type: patch
Changelog-Entry: Update Electron to v1.7.9

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -37,6 +37,12 @@
       "from": "@types/lodash.some@>=4.6.2 <5.0.0",
       "resolved": "https://registry.npmjs.org/@types/lodash.some/-/lodash.some-4.6.2.tgz"
     },
+    "@types/node": {
+      "version": "7.0.44",
+      "from": "@types/node@>=7.0.18 <8.0.0",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-7.0.44.tgz",
+      "dev": true
+    },
     "@types/react": {
       "version": "15.0.24",
       "from": "@types/react@>=15.0.23 <16.0.0",
@@ -1354,9 +1360,9 @@
       "optional": true
     },
     "electron": {
-      "version": "1.6.6",
-      "from": "electron@1.6.6",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-1.6.6.tgz",
+      "version": "1.7.9",
+      "from": "electron@1.7.9",
+      "resolved": "https://registry.npmjs.org/electron/-/electron-1.7.9.tgz",
       "dev": true,
       "dependencies": {
         "electron-download": {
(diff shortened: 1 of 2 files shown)


lurch:
Forgive me for sounding nervous, but how well have you tested this @jhermsmeier ? ;-)  Given that we're planning to release "very soon" would it be safer to stay with 1.6.x and update to electron v1.6.15 ?  Or are there specific features in the 1.7.x line that we'd find useful too?

npm-shrinkwrap.json
```diff
@@ -127,7 +133,7 @@
     },
     "angular-if-state": {
       "version": "1.0.0",
-      "from": "angular-if-state@>=1.0.0 <2.0.0",
+      "from": "angular-if-state@1.0.0",
```
lurch: Travis is getting "upset" with so many changes in `npm-shrinkwrap.json` :-/  Can you check you're using the 'right' version of NPM?

lurch: (and ping @jviotti too, in case he remembers how we resolved this problem the last time it cropped up)

jhermsmeier: I was using the "right" npm version (which is node 6 with npm 3.10.x), the issue with this that the from field depends on the previous recorded version of the package, and will inevitably change when updating dependencies.

jhermsmeier:
> Forgive me for sounding nervous, but how well have you tested this @jhermsmeier ?

I have, and everything appears to work just like it did before – but I'd encourage you to test it as well.

jhermsmeier:
> would it be safer to stay with 1.6.x and update to electron v1.6.15 ? Or are there specific features in the 1.7.x line that we'd find useful too?

As we need up upgrade Electron due to the RCE issue anyways, and it's not breaking anything from what the tests and I can tell, I don't see why we should stick to the older 1.6.x – unless there are strong objections?

npm-shrinkwrap.json
lurch: :cry: 
But even so, why is the shrinkwrap file generated on Travis different to the shrinkwrap file in this PR? Or is it a horrible order-of-installation problem?

jhermsmeier: So, I've just triple-checked – I'm on the same versions of node & npm as the travis box is, I shrinkwrapped multiple times, and no change – `make sanity-checks` comes back clean.

lurch: CI servers seem happy now (and the shrinkwrap diff is much smaller, which is nice)
:-)

lurch:
Tested it on Ubuntu 14.04 - tested the `npm start`, the electron-packaged version, the AppImage, and even (just for the hell of it) the CLI. All seem to work fine :-)

PR #1789, opened 2017-10-18
fix: set debug env variable on remote electron process

Shou:
We fix the DEBUG environment variable by setting it on the `electron.remote`
instead, and we also move the code to `lib/gui/app.js` and away from
`lib/gui/index.html`.

Changelog-Entry: Set the DEBUG environment variable on the remote
electron process.
Change-Type: patch

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -47,6 +47,14 @@ const driveScanner = require('./modules/drive-scanner')
 const osDialog = require('./os/dialog')
 const exceptionReporter = require('./modules/exception-reporter')
 
+// Enable debug information from all modules that use `debug`
+// See https://github.com/visionmedia/debug#browser-support
+//
+// Enable drivelist debugging information
+// See https://github.com/resin-io-modules/drivelist
+process.env.DRIVELIST_DEBUG = /drivelist|^\*$/i.test(electron.remote.process.env.DEBUG) ? '1' : ''
+window.localStorage.debug = electron.remote.process.env.DEBUG
+
 const app = angular.module('Etcher', [
   require('angular-ui-router'),
   require('angular-ui-bootstrap'),
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -52,7 +52,7 @@ const exceptionReporter = require('./modules/exception-reporter')
 //
 // Enable drivelist debugging information
 // See https://github.com/resin-io-modules/drivelist
-process.env.DRIVELIST_DEBUG = /drivelist|^\*$/i.test(electron.remote.process.env.DEBUG) ? '1' : ''
+electron.remote.process.env.DRIVELIST_DEBUG = /drivelist|^\*$/i.test(electron.remote.process.env.DEBUG) ? '1' : ''
 window.localStorage.debug = electron.remote.process.env.DEBUG
 
 const app = angular.module('Etcher', [
--- a/lib/start.js
+++ b/lib/start.js
@@ -16,7 +16,7 @@
 
 'use strict'
 
-process.env.DEBUG = `sdk:usbboot,${process.env.DEBUG}`
+process.env.DEBUG = `sdk:usbboot${process.env.DEBUG ? `,${process.env.DEBUG}` : ''}`
 
 // See http://electron.atom.io/docs/v0.37.7/api/environment-variables/#electronrunasnode
 //
--- a/lib/start.js
+++ b/lib/start.js
@@ -16,8 +16,6 @@
 
 'use strict'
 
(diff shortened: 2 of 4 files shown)


lib/start.js
```diff
@@ -16,7 +16,7 @@
 
 'use strict'
 
-process.env.DEBUG = `sdk:usbboot,${process.env.DEBUG}`
+process.env.DEBUG = `sdk:usbboot${process.env.DEBUG ? `,${process.env.DEBUG}` : ''}`
```
jviotti: Can you test if something like this works?

```js
process.env.DEBUG = `sdk:usbboot${process.env.DEBUG || ''}`
```

jviotti: Ah, but that leaves the trailing comma (which doesn't really matter, but its a bit ugly). Up to you, I'll approve it anyways :)

lurch: I wonder if this version is any more readable?
```js
process.env.DEBUG += `${process.env.DEBUG ? `,` : ''}sdk:usbboot`
```
(it feels more logical to me :wink: )

lurch:
I'm a bit confused - the PR title says "set debug env variable **on** remote electron process" but `window.localStorage.debug = electron.remote.process.env.DEBUG` makes it look like we're setting the env var **from** the remote process?

lib/start.js
lurch: Nope, ignore me, looks like `+=` stringifies `null` into `'null'` :(

Shou: The missing comma would complicate things wouldn't it? I was thinking

```js
[' sdk:usbboot', process.env.DEBUG ].filter(_ => _).join(',')
```
but the linter would complain about non-lodash usage.

Shou: This doesn't work if `process.env.DEBUG` is `undefined` unfortunately.

Shou:
@lurch good point! I was confusing `remote` with `renderer` process, and "from remote" is clearer so your suggestion is better, I'll go with that.

lib/start.js
lurch: Yeah, which is what I was showing with my code-snippet ;-)

lurch: I guess the 'proper' way to do it would be to split `process.env.DEBUG` by `,` into an array, append `sdk:usbboot` to that array, and then join the array with `,` back into a string. But that's obviously a lot of work when @Shou 's simple inline-conditional works fine :)

Shou: Yeah Github seems to like dynamically loading the first comment to trick the user (me) into a false sense of security but then subsequently load no more comments or edits

jhermsmeier: Can we please not muck with the `process.env.DEBUG` here, and enable certain scopes we definitely want to be on all the time in their respective modules? For example, to have `sdk:usbboot` permamently enabled we can do the following in `lib/sdk/usbboot/index.js`:

```diff
 const debug = require('debug')('sdk:usbboot')
 const usb = require('./usb')
 const protocol = require('./protocol')

+debug.enabled = true
+
 /**
  * @summary The name of this adaptor
  * @public
```

jhermsmeier: Can we please not muck with the `process.env.DEBUG` here, and enable certain scopes we definitely want to be on all the time in their respective modules? For example, to have `sdk:usbboot` permamently enabled we can do the following in `lib/sdk/usbboot/index.js`:

```diff
 const debug = require('debug')('sdk:usbboot')
 const usb = require('./usb')
 const protocol = require('./protocol')

+debug.enabled = true
+
 /**
  * @summary The name of this adaptor
  * @public
```

jhermsmeier: Ping @shou @jviotti ^

jviotti:
Ping @jhermsmeier. Is this good now?

lib/gui/app.js
```diff
@@ -47,6 +47,14 @@ const driveScanner = require('./modules/drive-scanner')
 const osDialog = require('./os/dialog')
 const exceptionReporter = require('./modules/exception-reporter')
 
+// Enable debug information from all modules that use `debug`
+// See https://github.com/visionmedia/debug#browser-support
+//
+// Enable drivelist debugging information
+// See https://github.com/resin-io-modules/drivelist
+electron.remote.process.env.DRIVELIST_DEBUG = /drivelist|^\*$/i.test(electron.remote.process.env.DEBUG) ? '1' : ''
+window.localStorage.debug = electron.remote.process.env.DEBUG
```
lurch: Just a minor nit-pick, but could we move the `Enable debug information from ...` comment to just above this line?
( @jhermsmeier assuming that's the correct way that the comments line up to the code? )

jhermsmeier: Yeah, this is fine though

PR #1787, opened 2017-10-17
doc(CONTRIBUTING): Add libudev requirement to docs

jhermsmeier:
This adds `libudev-dev` to the list of requirements for Linux in `CONTRIBUTING.md`

Change-Type: patch
Connects To: #1771

Issue #1771, opened 2017-10-14
Problems installing the development dependencies


shroudedcode:
As I already mentioned in my pull request #1768, I have problems getting the Etcher development dependencies to work on my machine although I closely followed the [contributing guide](https://github.com/resin-io/etcher/blob/master/docs/CONTRIBUTING.md).

### My system
- elementary OS 0.4.1 Loki (64-bit)
- Built on Ubuntu 16.04.3 LTS

### Common prerequisites
- `node --version` – `v8.7.0`
- `python --version` – `Python 2.7.12`
- `jq --version` – `jq-1.5-1-a5b5cbe`
- `curl --version` – `curl 7.47.0 [...]`

Installing the `requirements.txt` using `pip` also worked.

### Installing npm dependencies
This is where it started to get confusing. First of all the guide says that you should install the prerequisites before running the `install` script, a script which **isn't mentioned later on**. I just skipped that and ran `make electron-develop`. This threw a bunch of errors, so I [recorded it with asciinema](https://asciinema.org/a/sLIvvJV1B3OGNMA5bfjAt8yon).

After that, I tried running `npm start` anyways, but [it didn't work](https://asciinema.org/a/FTkLWQSa5zbj7bgMbbiAPM1SQ).

jhermsmeier:
Looks like libusb from our `usb` dependency blew up during installation;

<details>
<summary>
../libusb/libusb/os/linux_udev.c:40:21: fatal error: libudev.h: No such file or directory                              
compilation terminated.
</summary>

```
> usb@1.3.0 install node_modules/usb                                                                                   
> node-pre-gyp install --fallback-to-build                                                                             
                                                                                                                       
make[1]: Entering directory '/mnt/data/GitHub/etcher/node_modules/usb/build'                                           
  CC(target) Release/obj.target/libusb/libusb/libusb/core.o                                                            
  CC(target) Release/obj.target/libusb/libusb/libusb/descriptor.o                                                      
  CC(target) Release/obj.target/libusb/libusb/libusb/hotplug.o                                                         
  CC(target) Release/obj.target/libusb/libusb/libusb/io.o                                                              
  CC(target) Release/obj.target/libusb/libusb/libusb/strerror.o                                                        
  CC(target) Release/obj.target/libusb/libusb/libusb/sync.o                                                            
  CC(target) Release/obj.target/libusb/libusb/libusb/os/poll_posix.o                                                   
[...]

PR #1790, opened 2017-10-18
doc(CONTRIBUTING): Remove refence to "install script"

jhermsmeier:
This removes a confusing reference to an "install script"
in the dependency section. A change that was forgotten in #1787

Change-Type: patch
Connects To: #1771, #1787


lurch:
That change was originally made in https://github.com/resin-io/etcher/pull/746 and I'm afraid I think it might have been my suggestion which prompted it.

PR #1799, opened 2017-10-20
style(usbboot): wait before scanning drives after the file server phase

jviotti:
This is a workaround to prevent the USB device from disappearing after
the file server phase, until the resulting block device comes up.

By adding a delay after the file server phase, we prevent the USB
scanner from getting triggered again, therefore keeping the current USB
device visible in the drive selector modal.

Change-Type: patch

### DIFF
--- a/lib/shared/sdk/usbboot/index.js
+++ b/lib/shared/sdk/usbboot/index.js
@@ -273,6 +273,11 @@ class USBBootAdapter extends EventEmitter {
       }
 
       result.progress = this.progress[result.raw]
+
+      if (result.progress === utils.PERCENTAGE_MAXIMUM) {
+        return Bluebird.delay(DEVICE_REBOOT_DELAY).return(result)
+      }
+
       return result
 
     // See http://bluebirdjs.com/docs/api/promise.map.html


jviotti:
I decided to go for this hack, which should provide a great experience in 90% of the cases, since there were some issues with the “hypothetical” device approach:

- It tightly couples the usbboot adapter with the standard adapter, since suddenly the usbboot adapter will be creating devices what should have been detected by the standard adapter
- Even if we assume such coupling and create a new device, there is no way to easily know that the block device that appears afterwards is meant to replace the hypothetical device, since there is no information, apart from the name, that we could use to associate them
- Even if we rely on the name, it all gets more complex if we consider multiple usbboot devices running at the same time. What block device should replace which hypothetical device? How do we distinguish the hypothetical devices from each other, if we don’t know any unique information about them?

We can probably solve this problem way better if we first solve the association problem, which we should have a way forward now.

jviotti:
> By adding a delay after the file server phase, we prevent the USB
scanner from getting triggered again,

This works great as long as people are not plugging new non-usbboot drives in the 5s time range. If so, the detection will take longer, but then people are likely to just focus on one usbboot device at a time, and therefore this shouldn't be a problem in practice.

jviotti:
@jhermsmeier What do you think of this as a quick hack so we can release this week?

jhermsmeier:
Is this still of use after #1802? I'd avoid putting in timing-based hacks, generally, and also not for a release's sake – these things are fragile, and I'd rather we find a good way around this.

jviotti:
@jhermsmeier It remains relevant, even though I will need to re-think this after the https://github.com/resin-io/etcher/pull/1803 usbboot re-architect. The real solution to this problem is to be able to associate block devices with the USB device that originate it, which we can't do at the moment.

jviotti:
Ping @jhermsmeier

PR #1800, opened 2017-10-23
feat(GUI): add optional progress bars to drive-selector drives

Shou:
We show a progress bar for any drive objects with a `progress` field
that isn't falsy, e.g. `undefined` or `0`.

Changelog-Type: minor
Changelog-Entry: Add optional progress bars to drive-selector drives.

### DIFF
--- a/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
+++ b/lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
@@ -32,7 +32,7 @@
             </span>
 
           </footer>
-          <progress ng-if="drive.progress || true" value="{{ drive.progress || 50 }}" max="100"></progress>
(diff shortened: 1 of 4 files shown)


lib/gui/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -32,6 +32,7 @@ <h4 class="list-group-item-heading">
             </span>
 
           </footer>
+          <progress ng-if="drive.progress || true" value="{{ drive.progress || 50 }}" max="100"></progress>
```
jviotti: Would it make more sense to not display the progress bar at all if the drive has no progress?

Shou: Yup, my bad! The conditionals are only there for testing, I don't remember why I decided to keep them when I pushed...

Shou:
image

Issue #1691, opened 2017-08-14
Emit events if the flashing process gets stuck


jviotti:
The GUI application expects to be updated about the writer process state very often through IPC. Given how many people get stuck at "Starting..." or at the middle of the flash, lets trigger a timeout every time the flash process start or every time we get an update from the writer process, and emit an informational event to Mixpanel if the timeout expires and we didn't get another update.

lurch:
I guess this also links (indirectly) to @Shou 's task about adding an 'Elevating...' state?

And I dunno if it relates directly to this issue or not, but I noticed that sometimes (when writing a smaller image) in the DevTools console the flashing progress might start at e.g. 4% (presumably because it wrote the first 4% before the first flashing-status got emitted) and might end at e.g. 98% before the validation process then starts at e.g. 7%. It might be nice if explicit "0% Flashing", "100% Flashing", "0% Validating" and "100% Validating" statuses were written to the DevTools console?

lurch:
Another issue that could be merged into #1100 ?

PR #1801, opened 2017-10-24
fix: timeout the writer IPC channel

Shou:
We add a timeout to the writer IPC channel, if there is a space larger
than 5 seconds after a message we emit an error.

Changelog-Entry: Timeout the writer IPC channel
Change-Type: patch

### DIFF
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -26,6 +26,8 @@ const CONSTANTS = require('./constants')
 const EXIT_CODES = require('../shared/exit-codes')
 const robot = require('../shared/robot')
 
+const TIMEOUT_TIME = 5000
+
 /**
  * @summary Perform a write
  * @function
@@ -170,10 +172,32 @@ exports.write = (image, drive, options) => {
     }
   }
 
+  // Timeout function that only runs if we exceed the specified timeout,
+  // terminating the server.
+  let terminationTimeout = null
+
+  /**
+   * @summary Reset the timeout countdown
+   * @function
+   * @private
+   *
+   * @example
+   * resetTerminationTimeout()
+   */
+  const resetTerminationTimeout = () => {
+    clearTimeout(terminationTimeout)
+    const TIMEOUT_ERROR = new Error(`Writer timeout, ${TIMEOUT_TIME}ms exceeded.`)
+    terminationTimeout = setTimeout(_.partial(emitError, TIMEOUT_ERROR), TIMEOUT_TIME)
+  }
+
   ipc.server.on('error', emitError)
-  ipc.server.on('message', bridgeRobotMessage)
+  ipc.server.on('message', (message) => {
+    resetTerminationTimeout()
+    bridgeRobotMessage(message)
+  })
 
   ipc.server.on('start', () => {
+    resetTerminationTimeout()
     const child = childProcess.fork(CONSTANTS.WRITER_PROXY_SCRIPT, argv, {
       silent: true,
       env: process.env
@@ -194,6 +218,7 @@ exports.write = (image, drive, options) => {
 
     child.on('close', (code) => {
       terminateServer()
+      clearTimeout(terminationTimeout)
 
       if (code === EXIT_CODES.CANCELLED) {
         return emitter.emit('done', {
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -186,8 +186,11 @@ exports.write = (image, drive, options) => {
    */
   const resetTerminationTimeout = () => {
     clearTimeout(terminationTimeout)
-    const TIMEOUT_ERROR = new Error(`Writer timeout, ${TIMEOUT_TIME}ms exceeded.`)
-    terminationTimeout = setTimeout(_.partial(emitError, TIMEOUT_ERROR), TIMEOUT_TIME)
+    terminationTimeout = setTimeout(() => {
+      emitter.emit('timeout', {
+        TIMEOUT_TIME
+      })
+    }, TIMEOUT_TIME)
   }
 
   ipc.server.on('error', emitError)
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -184,10 +184,24 @@ exports.write = (image, drive, options) => {
 
   // Timeout function that only runs if we exceed the specified timeout,
   // terminating the server.
+  // @type {Timeout}
   let terminationTimeout = null
 
   // If the child process is created successfully it will be stored here.
+  // @type {ChildProcess}
   let maybeChild = null
+
+  // Whether the child process is running
+  // @type {Boolean}
+  let isChildRunning = null
+
+  // Server's socket connection to the child process
+  // @type {Socket}
+  let childSocket = null
+
+  // Child socket connection status
+  // @type {Boolean}
+  let isChildSocketConnected = null
 
   /**
    * @summary Finish the child writer
@@ -238,6 +252,12 @@ exports.write = (image, drive, options) => {
     terminationTimeout = setTimeout(() => {
       const error = new Error('Write process timed out')
       error.timeout = CHILD_WRITER_TIMEOUT_MS
+      error.isChildRunning = isChildRunning
+      error.childSocket = _.pick(childSocket, [
+        'connecting',
+        'destroyed'
+      ])
+      error.childSocket.isConnected = isChildSocketConnected
 
       emitter.emit('timeout', error)
 
@@ -253,12 +273,20 @@ exports.write = (image, drive, options) => {
     resetTerminationTimeout()
     bridgeRobotMessage(message)
   })
+  ipc.server.on('connect', (socket) => {
+    childSocket = socket
+    isChildSocketConnected = true
+  })
+  ipc.server.on('disconnect', (socket) => {
+    isChildSocketConnected = false
+  })
 
   ipc.server.on('start', () => {
     const child = childProcess.fork(CONSTANTS.WRITER_PROXY_SCRIPT, argv, {
       silent: true,
       env: process.env
     })
+    isChildRunning = true
     resetTerminationTimeout(child)
 
     child.stdout.on('data', (data) => {
@@ -275,6 +303,9 @@ exports.write = (image, drive, options) => {
     child.on('error', emitError)
 
     child.on('close', finish)
+    child.on('exit', () => {
+      isChildRunning = false
+    })
   })
 
   ipc.server.start()
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -26,6 +26,16 @@ const CONSTANTS = require('./constants')
 const EXIT_CODES = require('../shared/exit-codes')
 const robot = require('../shared/robot')
 
+/**
+ * @summary Writer timeout in milliseconds
+ * @constant
+ * @private
+ *
+ * @description
+ * If there are no incoming messages over the IPC channel after the time
+ * specified below has passed, then we assume the connection has been lost,
+ * as this happens occasionally.
+ */
 const TIMEOUT_TIME = 5000
 
 /**
@@ -176,20 +186,33 @@ exports.write = (image, drive, options) => {
   // terminating the server.
   let terminationTimeout = null
 
+  // If the child process is created successfully it will be stored here.
+  let maybeChild = null
+
   /**
    * @summary Reset the timeout countdown
    * @function
    * @private
    *
+   * @param {ChildProcess} child - child process
+   *
    * @example
    * resetTerminationTimeout()
    */
-  const resetTerminationTimeout = () => {
+  const resetTerminationTimeout = (child) => {
+    if (child) {
+      maybeChild = child
+    }
+
     clearTimeout(terminationTimeout)
     terminationTimeout = setTimeout(() => {
       emitter.emit('timeout', {
-        TIMEOUT_TIME
+        timeout: TIMEOUT_TIME
       })
+
+      // Remove listeners so no more error popups are spawned
+      _.invoke(maybeChild, [ 'removeAllListeners' ])
+      _.invoke(maybeChild, [ 'kill' ])
     }, TIMEOUT_TIME)
   }
 
@@ -200,11 +223,11 @@ exports.write = (image, drive, options) => {
   })
 
   ipc.server.on('start', () => {
-    resetTerminationTimeout()
     const child = childProcess.fork(CONSTANTS.WRITER_PROXY_SCRIPT, argv, {
       silent: true,
       env: process.env
     })
+    resetTerminationTimeout(child)
 
     child.stdout.on('data', (data) => {
       console.info(`WRITER: ${data.toString()}`)
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -236,9 +236,10 @@ exports.write = (image, drive, options) => {
 
     clearTimeout(terminationTimeout)
     terminationTimeout = setTimeout(() => {
-      emitter.emit('timeout', {
-        timeout: CHILD_WRITER_TIMEOUT_MS
-      })
+      const error = new Error('Write process timed out')
+      error.timeout = CHILD_WRITER_TIMEOUT_MS
+
+      emitter.emit('timeout', error)
 
       // Remove listeners so no more error popups are spawned
       _.invoke(maybeChild, [ 'removeAllListeners' ])
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -36,7 +36,7 @@ const robot = require('../shared/robot')
  * specified below has passed, then we assume the connection has been lost,
  * as this happens occasionally.
  */
-const TIMEOUT_TIME = 5000
+const CHILD_WRITER_TIMEOUT_MS = 5000
 
 /**
  * @summary Perform a write
@@ -190,6 +190,36 @@ exports.write = (image, drive, options) => {
   let maybeChild = null
 
   /**
+   * @summary Finish the child writer
+   * @function
+   * @private
+   *
+   * @param {Number} code - exit code
+   * @returns {Void}
+   *
+   * @example
+   * finish(EXIT_CODES.SUCCESS)
+   */
+  const finish = (code) => {
+    terminateServer()
+    clearTimeout(terminationTimeout)
+
+    if (code === EXIT_CODES.CANCELLED) {
+      return emitter.emit('done', {
+        cancelled: true
+      })
+    }
+
+    // We shouldn't emit the `done` event manually here
+    // since the writer process will take care of it.
+    if (code === EXIT_CODES.SUCCESS || code === EXIT_CODES.VALIDATION_ERROR) {
+      return null
+    }
+
+    return emitError(new Error(`Child process exited with error code: ${code}`))
+  }
+
+  /**
    * @summary Reset the timeout countdown
    * @function
    * @private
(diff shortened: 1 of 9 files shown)


jviotti:
This will make Etcher throw an error if no activity from the IPC client is received, but doesn't fix the underlying issue (we should avoid the IPC client from getting disconnected on the first place).

I believe this is happening because of https://github.com/resin-io/etcher/blob/master/lib/child-writer/writer-proxy.js#L121. If there a connection error, we don't try to reconnect. We should find a way to emit the `disconnect` event once the writing is done, while still allowing the IPC module to reconnect in case of errors.

Shou:
Should I retrofit it to resolve https://github.com/resin-io/etcher/issues/1691 instead then?

jviotti:
Yeah, let's go for it!

Shou:
Done that now!

lib/child-writer/index.js
```diff
@@ -170,10 +172,35 @@ exports.write = (image, drive, options) => {
...
+   *
+   * @example
+   * resetTerminationTimeout()
+   */
+  const resetTerminationTimeout = () => {
+    clearTimeout(terminationTimeout)
+    terminationTimeout = setTimeout(() => {
+      emitter.emit('timeout', {
+        TIMEOUT_TIME
+      })
```
jhermsmeier: Why is `TIMEOUT_TIME` wrapped in an object here? And why is the timeout value emitted through the event?

lib/child-writer/index.js
```diff
@@ -26,6 +26,8 @@ const CONSTANTS = require('./constants')
 const EXIT_CODES = require('../shared/exit-codes')
 const robot = require('../shared/robot')
 
+const TIMEOUT_TIME = 5000
```
jviotti: Can we document this constant and explain why its needed?

lib/child-writer/index.js
```diff
@@ -170,10 +172,35 @@ exports.write = (image, drive, options) => {
...
+   * @function
+   * @private
+   *
+   * @example
+   * resetTerminationTimeout()
+   */
+  const resetTerminationTimeout = () => {
+    clearTimeout(terminationTimeout)
+    terminationTimeout = setTimeout(() => {
+      emitter.emit('timeout', {
```
jviotti: I think we should also make sure we kill the child (if its still alive) if the timeout occurred.

lib/gui/modules/image-writer.js
```diff
@@ -63,6 +63,7 @@ imageWriter.service('ImageWriterService', function ($q, $rootScope) {
       })
       child.on('error', reject)
       child.on('done', resolve)
+      child.on('timeout', _.partial(analytics.logEvent, 'Writer timeout'))
```
jviotti: Can we also show a nice error modal to the user explaining what happened?

lib/child-writer/index.js
Shou: I assume it's good data to have should we change the value in the future, the object is because I was wondering if there's more we could possibly need, if not we can just send the value by itself.

jviotti: Yeah, I think its fine. We can use it to display a nice modal to the user saying something like "We stopped the flashing process because we didn't receive any updates in X seconds".

My only objection is that right now the property name is `TIMEOUT_TIME`, so maybe something like:

```js
{ timeout: TIMEOUT_TIME }
```

will read better?

Shou:
Thoughts on this now?

image

jviotti:
So nice! :tada:

lib/gui/modules/image-writer.js
```diff
@@ -63,6 +66,15 @@ imageWriter.service('ImageWriterService', function ($q, $rootScope) {
       })
       child.on('error', reject)
       child.on('done', resolve)
+      child.on('timeout', (timeoutData) => {
+        analytics.logEvent('Writer timeout', timeoutData)
+        // eslint-disable-next-line no-magic-numbers
+        const timeoutDataSeconds = _.update(timeoutData, [ 'timeout' ], _.partial(_.divide, _, 1000))
```
jviotti: Should we create a utility function (maybe in `lib/shared/utils`) to transform milliseconds to seconds?

lib/child-writer/index.js
```diff
@@ -27,6 +27,18 @@ const EXIT_CODES = require('../shared/exit-codes')
...
+ * @summary Writer timeout in milliseconds
+ * @constant
+ * @private
+ *
+ * @description
+ * If there are no incoming messages over the IPC channel after the time
+ * specified below has passed, then we assume the connection has been lost,
+ * as this happens occasionally.
+ */
+const TIMEOUT_TIME = 5000
```
jviotti: Maybe `CHILD_WRITER_TIMEOUT_MS` is a more descriptive name?

lib/child-writer/index.js
```diff
@@ -194,6 +244,7 @@ exports.write = (image, drive, options) => {
 
     child.on('close', (code) => {
```
jviotti: Just out of curiosity, is this event triggered when we kill the child?

lib/child-writer/index.js
```diff
@@ -170,14 +182,52 @@ exports.write = (image, drive, options) => {
...
+   * @private
+   *
+   * @param {ChildProcess} child - child process
+   *
+   * @example
+   * resetTerminationTimeout()
+   */
+  const resetTerminationTimeout = (child) => {
+    if (child) {
+      maybeChild = child
```
jviotti: I think you could avoid having `maybeChild` and simply invoke the functions below on `child`. If `child` was passed as an argument here, then it will remain accessible by the timeout handler by the time it gets called.

Shou: Unfortunately `child` won't be passed in later calls to `resetTerminationTimeout` so it'll only work for the first five seconds assuming it doesn't get reset, so we need to store `child` for the later calls.

lib/gui/modules/image-writer.js
Shou: Done!

lib/child-writer/index.js
Shou: It was before `removeAllListeners` gets called, but not anymore, good catch! It exits cleanly now.

lib/child-writer/index.js
```diff
@@ -170,14 +182,83 @@ exports.write = (image, drive, options) => {
...
+    }
+
+    clearTimeout(terminationTimeout)
+    terminationTimeout = setTimeout(() => {
+      emitter.emit('timeout', {
+        timeout: CHILD_WRITER_TIMEOUT_MS
```
jhermsmeier: What do you think about making this an `Error` object instead? For example:

```js
var error = new Error( 'Write process timed out' )
error.timeout = CHILD_WRITER_TIMEOUT_MS

emitter.emit('timeout', error)
```

Shou: Done!

lib/child-writer/index.js
```diff
@@ -170,14 +182,112 @@ exports.write = (image, drive, options) => {
...
+  // @type {Timeout}
+  let terminationTimeout = null
+
+  // If the child process is created successfully it will be stored here.
+  // @type {ChildProcess}
+  let maybeChild = null
+
+  // Whether the child process is running
+  // @type {Boolean}
+  let isChildRunning = null
```
jviotti: Is there a way we can reduce the amount of mutable variables tracking state that we have here? It obviously works, but I feel its a matter of time until we will get weird edge cases where these variables are not updated appropriately, and that will be really hard to debug.

Shou:
How's this? It doesn't remove the state but it does use less explicit variables to achieve the same result.

lib/child-writer/index.js
```diff
@@ -170,14 +182,102 @@ exports.write = (image, drive, options) => {
...
+  // Timeout function that only runs if we exceed the specified timeout,
+  // terminating the server.
+  // @type {Timeout}
+  let terminationTimeout = null
+
+  // @type {Object} state
+  // @type {ChildProcess | undefined} [state.child]
+  // @type {Socket | undefined} [state.childSocket]
+  // @type {Boolean | undefined} [state.isChildRunning]
+  // @type {Boolean | undefined} [state.isChildSocketConnected]
```
jhermsmeier: Shouldn't this and the above be in the fashion of

```js
/**
 * [state description]
 * @private
 * @type {Object} state
 * @type {ChildProcess | undefined} [state.child]
 * @type {Socket | undefined} [state.childSocket]
 * @type {Boolean | undefined} [state.isChildRunning]
 * @type {Boolean | undefined} [state.isChildSocketConnected]
 */
```

lib/child-writer/index.js
```diff
@@ -170,14 +182,102 @@ exports.write = (image, drive, options) => {
     }
   }
 
+  // Timeout function that only runs if we exceed the specified timeout,
```
jhermsmeier: The wording here is a bit confusing, as it's the timeout ID that's being stored in here, not a function in that sense – can we adjust this accordingly?

jhermsmeier: Also, is there a reason this isn't on the `state` object, or can we move that there as well?

lib/child-writer/index.js
```diff
@@ -170,14 +182,102 @@ exports.write = (image, drive, options) => {
...
+      error.isChildRunning = state.isChildRunning
+      error.childSocket = _.pick(state.childSocket, [
+        'connecting',
+        'destroyed',
+        'bytesRead',
+        'bytesWritten'
+      ])
+      error.childSocket.isConnected = state.isChildSocketConnected
+      error.child = _.pick(state.child, [ 'signalCode' ])
+      console.log(_.assign({}, error))
```
jhermsmeier: Why copy the error's enumerable props when logging it? This will not copy over `.message`, `.stack` etc. And why are we `console.log()`ing this at all?, it'll be emitted through the `timeout` event, right? Or is this just a debug leftover? So many questions! :D

lib/child-writer/index.js
```diff
@@ -170,14 +182,102 @@ exports.write = (image, drive, options) => {
...
+
+      emitter.emit('timeout', error)
+
+      // Remove listeners so no more error popups are spawned
+      _.invoke(state.child, [ 'removeAllListeners' ])
+      _.invoke(state.child, [ 'kill' ])
```
jhermsmeier: No need for `_.invoke()` here I believe

Shou: It removes the need for a conditional checking whether `state.child` exists, because it'll otherwise error with `Cannot read property 'x' of undefined`

lib/child-writer/index.js
Shou: Hah yeah, debugging leftovers – let me remove that.

lib/child-writer/index.js
Shou: Yup, I forgot this as I copy-pasted the lines.

lib/child-writer/index.js
Shou: True – I wanted to link what the timeout actually does with its id variable when I wrote that. I'll prefix it with "Id of" to clarify.

We could move it there as well, but since `state` is mostly used for debugging I was thinking to keep the two separate.

lib/child-writer/index.js
jhermsmeier: Oh I see – the documentation mentions none of that, only "Invokes the method at path of object.", which is like "you can also just call the thing instead". I'd prefer to have an

```js
if( state.child ) {
  state.child.removeAllListeners()
  state.child.kill()
}
```

to make it obvious and clear what's going on here, and not swallow any errors.

Shou: Sure. It's a sneaky property of Lodash, you can assume any method will just "fail" silently and return `undefined` if so.

jhermsmeier:
The timeout error happens when you take more than 5 seconds to enter your credentials – that'll need mending :D

screen shot 2017-11-03 at 16 25 44

Shou:
Let me know what you think of this now. I've added a heartbeat to the writer proxy starting from before elevation.

jhermsmeier:
@Shou same thing happens – it times out during password entry

Shou:
So it looks like `sudoPrompt` bites us here too – not knowing when the elevation ends means we can't accurately do this. I've had to sacrifice IPC timeouts unless a message is sent because of that, otherwise the auth prompt causes a timeout, but it should work now. Let me know.

If we make the child writer persistent it should greatly simplify this elevation stuff.

lib/shared/utils.js
```diff
@@ -78,3 +78,28 @@ exports.percentageToFloat = (percentage) => {
...
+ */
+exports.millisecondsToSeconds = (time) => {
+  if (!_.isNumber(time)) {
+    throw errors.createError({
+      title: `Not a number: ${time}`
+    })
+  }
+
+  // eslint-disable-next-line no-magic-numbers
+  return time / 1000
```
lurch: Probably no harm in adding a `MILLISECONDS_IN_A_SECOND` constant somewhere?

I just did a search, and found `MILLISECONDS_IN_A_DAY` (and `daysToMilliseconds`) in `shared/units.js` so this whole function should probably be moved there?

lurch:
Do we have any way (or do you know of a way) to 'force' an IPC disconnect or an IPC timeout, so that we can test this code works correctly when things go wrong?

jhermsmeier:
It does seem to work on Mac OS now, but Windows still ain't a happy customer:

windows-ipc-timeout

jhermsmeier:
I'm also thinking we might want to set the timeout to something way higher, to a minute, in case of slow machines, sluggish writes or things like that. But first we should get this fail on Windows sorted.

Shou:
I think the Windows behaviour might be related to the child timeout – could you try commenting line 307 and 311 in `lib/child-writer/index.js` and see if it works on Windows?

lib/child-writer/writer-proxy.js
```diff
@@ -27,14 +27,15 @@ const EXIT_CODES = require('../shared/exit-codes')
...
+const units = require('../shared/units')
 
 /**
  * @summary Frequency of process heartbeat signal in milliseconds
  * @constant
  * @private
  * @type {Number}
  */
-const HEARTBEAT_TIME = 1000
+const HEARTBEAT_TIME_MS = units.MILLISECONDS_IN_A_SECOND
```
lurch: I'm not sure about this change (and also the `60 * units.MILLISECONDS_IN_A_SECOND` one) - elsewhere in the codebase we've got values specified in ms, so perhaps this should be changed back to `const HEARTBEAT_TIME_MS = 1000` ?
Although @jhermsmeier and @jviotti will have a much better idea than me what the "standard nodeJS practice" is.

lib/shared/units.js
```diff
@@ -91,3 +100,27 @@ exports.bytesToClosestUnit = (bytes) => {
...
+ * }, DELAY)
+ */
+exports.millisecondsToSeconds = (time) => {
+  if (!Number.isFinite(time)) {
+    throw errors.createError({
+      title: `Not a finite number: ${time}`
+    })
+  }
+
+  return time / exports.MILLISECONDS_IN_A_SECOND
```
lurch: Purely out of curiosity (I'm not querying what you've done, I'm just trying to learn), why does this use `exports.MILLISECONDS_IN_A_SECOND` but inside `daysToMilliseconds` there's no `exports.` on the front of `MILLISECONDS_IN_A_DAY` ?

lib/shared/messages.js
```diff
@@ -123,7 +123,7 @@ module.exports = {
 
     writerTimeout: _.template([
       'The writer process timed out after <%= timeout %>s.',
-      'We can\'t guarantee the drive was completely written nor verify its data is correct, please try again.'
+      'There was a synchronization issue while flashing your drive, please try again.'
```
lurch: Just a minor nit, but when we get this timeout-error, can we tell (based on the messages we've received previously) whether we're still in the flashing-stage, or whether we're in the verifying stage? Would it be worth displaying a slightly different error message in each case?

lib/shared/units.js
Shou: We only export `MILLISECONDS_IN_A_SECOND` because we use it later in the `60 * MILLISECONDS_IN_A_SECOND` expression. If `MILLISECONDS_IN_A_DAY` becomes needed elsewhere we could also export that

lurch: Ahhh, thanks! I hadn't noticed that `MILLISECONDS_IN_A_DAY` was a 'private' constant and not exported :+1:

tests/shared/units.spec.js
```diff
@@ -58,4 +58,33 @@ describe('Shared: Units', function () {
       m.chai.expect(units.bytesToMegabytes(332000)).to.equal(0.332)
     })
   })
+
+  describe('.millisecondsToSeconds()', function () {
+    it('should convert milliseconds to seconds', function () {
+      const seconds = units.millisecondsToSeconds(10000)
+      m.chai.expect(seconds).to.equal(10)
+    })
```
lurch: Would it be worth also adding a test for a not-exact number of seconds, e.g. 5500 milliseconds?

lib/child-writer/writer-proxy.js
```diff
@@ -27,6 +27,15 @@ const EXIT_CODES = require('../shared/exit-codes')
...
 const packageJSON = require('../../package.json')
+const units = require('../shared/units')
+
+/**
+ * @summary Frequency of process heartbeat signal in milliseconds
+ * @constant
+ * @private
+ * @type {Number}
+ */
+const HEARTBEAT_TIME_MS = units.MILLISECONDS_IN_A_SECOND
```
lurch: I guess this'll need fixing @Shou

Shou: Thanks, and we might as well increase it considering the new timeout is 60s.

lib/child-writer/writer-proxy.js
jhermsmeier: I'd just do `60 * 1e3` and be done with it, but our linter wouldn't like that I think :)

tests/shared/units.spec.js
jhermsmeier: Nah, timers run in ms intervals anyways, 500 more or less shouldn't matter.

lurch: Nah, I meant specifically for this `millisecondsToSeconds` unit-test, i.e. add a check that `units.millisecondsToSeconds(5500) == 5.5`

lib/gui/pages/main/controllers/flash.js
```diff
@@ -100,8 +100,8 @@ module.exports = function (
       } else if (error.code === 'ENOSPC') {
         FlashErrorModalService.show(messages.error.notEnoughSpaceInDrive())
       } else {
-        FlashErrorModalService.show(messages.error.genericFlashError())
-        exceptionReporter.report(error)
+        FlashErrorModalService.show(`${messages.error.genericFlashError()}\n\n${error}`)
+        analytics.logException(error)
```
lurch: I don't fully understand this code, or why this change is necessary, but given that `exceptionReporter` is also used in multiple other places, would a more-comprehensive solution to be add some kind of 'wrapper' around the `osDialog.showError` call (if that's indeed where this is falling over), so that you can detect if a native dialog is being displayed, and not trigger the timeout-error if so?
(And again, apologies if I'm asking stupid questions)

Shou: No worries man, they're not stupid questions at all. I feel like since Windows is the only OS with a dialog that has this effect we'd be maintaining a non-trivial workaround just for Windows, when we instead have dialogs that don't block.

That is indeed how it used to work before

lurch: So are we sure that none of the other uses of [exceptionReporter](https://github.com/resin-io/etcher/search?q=exceptionReporter&type=Code&utf8=%E2%9C%93) need workarounds similar to what you're doing here?

Is this potentially a bug in Electron that the native-dialog-code behaves differently on Windows than it does on OSX or Linux?
I can't remember the details or the reason why, but IIRC being able to use native dialogs was a "big thing" when @jviotti was first adding it? (Or possibly I'm getting confused with some other part)... Ah, looking at [CHANGELOG.md](https://github.com/resin-io/etcher/blob/master/CHANGELOG.md) it looks like I was thinking of the Update dialog rather than the Error dialog ;-)

> > the user would have seen two dialogs?

> That is indeed how it used to work before

Weird, any idea why it was done that way?

jhermsmeier:
Closing this, as superseded by https://github.com/resin-io/etcher/pull/1801

PR #1802, opened 2017-10-24
refactor(sdk): add bus number, device address, and ids in usb devices

jviotti:
The combination of bus number and device address is the only way to
uniquely identify a USB device, so we'll use that for the `device` and
`raw` properties.

Also, we store the USB vendor and product IDs as properties of the
drives, since they will be handy when implementing the prepare function.

Change-Type: patch


jviotti:
I'll merge this so I can keep going. I'll address any comments in a separate PR.

PR #1803, opened 2017-10-24
feat(usbboot): add progress property to usbboot scanned drives 

jviotti:
This commit re-architects the usbboot adapter to prepare the drives in
the background, while emitting scan results every 2s, where each drive
has a `progress` percentage property.

Change-Type: minor

### DIFF
--- a/lib/shared/sdk/usbboot/index.js
+++ b/lib/shared/sdk/usbboot/index.js
@@ -27,6 +27,7 @@ const Bluebird = require('bluebird')
 const debug = require('debug')('sdk:usbboot')
 const usb = require('./usb')
 const protocol = require('./protocol')
+const utils = require('../../utils')
 
 debug.enabled = true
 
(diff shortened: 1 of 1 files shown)


jviotti:
@jhermsmeier I rebased this PR on top of your event related changes.

lib/shared/sdk/usbboot/index.js
```diff
@@ -228,23 +247,97 @@ class USBBootAdapter extends EventEmitter {
...
+     * @private
+     *
+     * @param {Number} percentage - percentage
+     *
+     * @example
+     * setProgress(90)
+     */
+    const setProgress = (percentage) => {
+      debug(`%c[${device.raw}] -> ${Math.floor(percentage)}%%`, 'color:red;')
+      this.progress[device.raw] = percentage
```
jhermsmeier: Should the progress also be a property on the device object as `device.progress`?

jviotti: This is the internal progress cache is used as the USB device changes. There is another progress property that belongs to the drive.

jhermsmeier: Oh, I didn't see the `result.progress = this.progress[result.raw]` at first, because it used to be `device`. Nevermind then.

Issue #1136, opened 2017-02-25
Display the determined uncompressed size on the "too small for image" warning


jviotti:
We should make sure the user understands why the drive is not large enough, which can be a bit confusing in the case of compressed images, where the compressed package is usually smaller than the drive.

Shou:
Something like this? Although maybe "image size too big" "actual image size too big" or some such?

![image](https://cloud.githubusercontent.com/assets/819413/23359519/7716ee38-fcde-11e6-8d6e-a8fa7cba9712.png)

lurch:
Might also be worth having an option to display the exact byte-sizes of both the uncompressed image and the drive in that scenario too, since with the units currently being always-rounded to GB, a "7.5GB" image may not actually fit on a "7.5GB" drive, if the image is even a few kilobytes larger than the drive.

lurch:
Hmm, looking at @Shou 's screenshot I guess we need to be careful to try and keep the "warning label" small - maybe @konmouz has some ideas?

Shou:
@lurch yeah, good call. Perhaps "Image Size Bigger By x units" and we just display in the appropriate human readable unit?

jviotti:
Let's bring @konmouz here.

konmouz:
2 options:
- we can have a 'insufficient space' label and on hover a tooltip with details (what @lurch suggested)
- avoid the label and have just an the alert/ danger icon with the explanation, all in red - no tooltip

BTW, @Shou why the above element has two labels? System and boot? It would be better to stick to 1 label per drive.

jviotti:
I think I like the second option more, given that a tooltip might not be very discoverable unless the user however on the badge. In order to continue the brainstorm session: what about displaying the size that doesn't fit instead? E.g: You need 300 more MBs for the image.

konmouz:
@jviotti this sounds even better. So using the label font size and the alert icon, we can display such a message.  - Or, 300 MBs more are required for the selected image. -

lurch:
> BTW, @Shou why the above element has two labels? System and boot?

It's actually System and Source ;-)  But yeah, it probably makes sense to only show one warning-label-per-drive, and I guess in this instance `Source` is 'more important' than `System` (a system drive *may* be writable, but a source drive (where the image is being read from) never will be).

With regards to the "too small" label message, what about "Drive is 300MB smaller than image" ?

konmouz:
[...]

PR #1807, opened 2017-10-24
feat(gui): Display image size when drive too small

jhermsmeier:
This adds a display of the determined image size to the
drive label when the drive has been determined to be too small.

screen shot 2018-03-21 at 20 40 59

Change-Type: patch
Changelog-Entry: Display image size in drive selector if drive is too small
Connects To: #1136, https://github.com/resin-io/etcher/issues/2004

### DIFF
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -18,6 +18,8 @@
 
 const _ = require('lodash')
 const pathIsInside = require('path-is-inside')
+const prettyBytes = require('pretty-bytes')
+const messages = require('./messages')
 
 /**
  * @summary The default unknown size for things such as images and drives
@@ -300,72 +302,6 @@ exports.isDriveSizeLarge = (drive) => {
 }
 
 /**
- * @summary Drive/image compatibility status messages.
- * @public
- * @type {Object}
- *
- * @description
- * Status messages intended to be shown to the user.
- */
-exports.COMPATIBILITY_STATUS_MESSAGES = {
-
-  /**
-   * @property {String} SIZE_NOT_RECOMMENDED
-   * @memberof COMPATIBILITY_STATUS_MESSAGES
-   *
-   * @description
-   * The image and drive compatibility is not recommended; happens when the
-   * actual drive size is smaller than the image's recommended drive size.
-   */
-  SIZE_NOT_RECOMMENDED: 'Not Recommended',
-
-  /**
-   * @property {String} TOO_SMALL
-   * @memberof COMPATIBILITY_STATUS_MESSAGES
-   *
-   * @description
-   * The drive is too small for the image.
-   */
-  TOO_SMALL: 'Too Small For Image',
-
-  /**
-   * @property {String} LOCKED
-   * @memberof COMPATIBILITY_STATUS_MESSAGES
-   *
-   * @description
-   * The drive is locked (e.g. the lock-tab on SD cards) and cannot be written to.
-   */
-  LOCKED: 'Locked',
-
-  /**
-   * @property {String} SYSTEM
-   * @memberof COMPATIBILITY_STATUS_MESSAGES
-   *
-   * @description
-   * The drive is a system drive and should not be written to.
-   */
-  SYSTEM: 'System Drive',
-
-  /**
-   * @property {String} CONTAINS_IMAGE
-   * @memberof COMPATIBILITY_STATUS_MESSAGES
-   *
-   * @description
-   * The drive contains the image and therefore cannot be written to.
-   */
-  CONTAINS_IMAGE: 'Drive Contains Image',
-
-  /**
-   * @property {String} LARGE_DRIVE
-   * @memberof COMPATIBILITY_STATUS_MESSAGES
-   *
-   * @description
-   * The drive is large and therefore likely not a medium you want to write to.
-   */
-  LARGE_DRIVE: 'Large Drive'
-}
-
-/**
  * @summary Drive/image compatibility status types.
  * @public
  * @type {Object}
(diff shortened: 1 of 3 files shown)


lib/shared/drive-constraints.js
```diff
@@ -376,9 +377,10 @@ exports.getDriveImageCompatibilityStatuses = (drive, image) => {
       message: exports.COMPATIBILITY_STATUS_MESSAGES.LOCKED
     })
   } else if (!_.isNil(drive) && !_.isNil(drive.size) && !exports.isDriveLargeEnough(drive, image)) {
+    const imageSize = _.get(image, [ 'size', 'final', 'estimation' ]) ? image.size.final.value : image.size.original
```
jviotti: This logic looks weird. Basically, if the image final size is an estimation, use that, otherwise use the original one, which means we will use the original size even for compressed sizes that we know their final size for sure (i.e. that are not estimations).

Maybe the intended logic is: If the final one is *not* an estimation, use that, otherwise use the original size.

jhermsmeier: Oh boy, yeah, I messed that up – fixed now :)

lurch:
I'm not sure if this is actually needed? In @jhermsmeier 's screenshot above you can clearly see that the (uncompressed as necessary) image size is already shown in the main Etcher GUI anyway...

jhermsmeier:
Yes, you're right – I checked the conversation again, and have now changed it to display the *relative bytes* missing for the image to fit:


screen shot 2017-10-25 at 13 23 22
screen shot 2017-10-25 at 13 27 16

jhermsmeier:
I don't think this is completely unnecessary though, as the issue discusses, it's aimed at helping people figure out why a 8GB image won't fit on a 8GB device, for example.

lurch:
Instead of "Too Small For Image - 46.15 MB", what about something like "Too Small For Image by 46.15 MB" - _might_ be slightly clearer?

Shou:
I was just about to say the same @lurch! Also maybe we could make the constant into a template that takes a `size` field?

jhermsmeier:
Yeah, added the "by" :)

> Also maybe we could make the constant into a template that takes a size field?

I thought about it when I started this, but the messages used aren't in `shared/messages`, but the `shared/drive-constraints` itself – and moving them all over for no gain seems excessive to me. In this case, string concat is more than enough imho, and not worth overcomplicating (also, I'd actually rather get rid of those lodash templates, but that's a different discussion)

screen shot 2017-10-25 at 13 42 59

Shou:
> (also, I'd actually rather get rid of those lodash templates, but that's a different discussion)

Made https://github.com/resin-io/etcher/issues/1810 to discuss this

jviotti:
I must confess I'm confused when looking at the screenshot. For example, we have a drive that is 130 MB and the label says "Too small for image by 1.72 GB." I can't tell for sure if it means I should have a drive that is at least 1.72 GB, or that my drive would need to be 1.72 GB larger (in which case I have to do some arithmetic to find that I need a 1.85 GB drive).

Can we rephrase the whole label to something like "Too small. Should be at least XXX MBs" or something like that?

lurch:
> Can we rephrase the whole label to something like "Too small. Should be at least XXX MBs" or something like that?

But as I mentioned earlier, that would simply duplicate the (uncompressed) image size, which is already visible in the main Etcher GUI anyway?

Perhaps @jhermsmeier 's didn't choose the best examples (who has a 130MB drive these days?! :wink: ), but imagine that you have a compressed image that's 8001MB when uncompressed, but your SD card actually has a capacity of 8000MB. With @jhermsmeier 's last proposal Etcher would display "Too Small For Image by 1 MB" whereas @jviotti 's proposal would display "Too small. Should be at least 8 GBs"
\*shrug\*

(P.S. Yes, I can never remember whether Etcher uses 1024 or 1000 when converting from bytes to kilobytes, megabytes, etc. so please ignore any mistake I've made in that regard!)

lib/shared/drive-constraints.js
lurch: > Maybe the intended logic is: If the final one is not an estimation, use that, otherwise use the original size.

Isn't that exactly what the 'estimated' size is anyway?

jhermsmeier:
> who has a 130MB drive these days?! 😉

I do :P It's an old marketing gift thing I picked up somewhere, came in handy for testing every now and then :)

> P.S. Yes, I can never remember whether Etcher uses 1024 or 1000 when converting from bytes to kilobytes, megabytes, etc. so please ignore any mistake I've made in that regard!

[Pretty-bytes](https://github.com/sindresorhus/pretty-bytes) (which we're using atm) uses base-10 (aka "hipster-units"), so generally 1KB = 1000B in Etcher (I think there might be an internal place or two where be use 1024B for reasons).

jhermsmeier:
I concur with @lurch – the "Should be at least XXX MB" would just dupe the image size; How about "Insufficient space, additional XX MB required" or something along those lines?

lurch:
This would also fix #2004

jhermsmeier:
Changed the message to `'Insufficient space, additional XX required'` & rebased against master

lib/shared/drive-constraints.js
```diff
@@ -404,9 +405,11 @@ exports.getDriveImageCompatibilityStatuses = (drive, image) => {
       message: exports.COMPATIBILITY_STATUS_MESSAGES.LOCKED
     })
   } else if (!_.isNil(drive) && !_.isNil(drive.size) && !exports.isDriveLargeEnough(drive, image)) {
+    const imageSize = _.get(image, [ 'size', 'final', 'estimation' ]) ? image.size.original : image.size.final.value
+    const relativeBytes = imageSize - drive.size
     statusList.push({
       type: exports.COMPATIBILITY_STATUS_TYPES.ERROR,
-      message: exports.COMPATIBILITY_STATUS_MESSAGES.TOO_SMALL
+      message: exports.COMPATIBILITY_STATUS_MESSAGES.TOO_SMALL.replace('XX', prettyBytes(relativeBytes))
```
Shou: This is failing with `Prefer '_.replace' over the native function  lodash/prefer-lodash-method`

lib/shared/drive-constraints.js
```diff
@@ -303,7 +304,7 @@ exports.COMPATIBILITY_STATUS_MESSAGES = {
    * @description
    * The drive is too small for the image.
    */
-  TOO_SMALL: 'Too Small For Image',
+  TOO_SMALL: 'Insufficient space, additional XX required',
```
Shou: Maybe we should make these into template arrow functions? We can keep it as-is now and open a new PR for that, though.

lib/shared/drive-constraints.js
jhermsmeier: Fixed

lib/shared/drive-constraints.js
jhermsmeier: I've now moved the compatibility messages from `drive-constraints` to `shared/messages` and made them functions

jhermsmeier:
@jviotti can you have another look at this?

PR #1814, opened 2017-10-27
feat(sdk): Implement continuous scanning

jhermsmeier:
This implements an SDK.Scanner which handles any given
adapters and manages the scans. This change enables continuous
scanning without the need to `.scan()` scheduling in other places.

Change-Type: minor


lib/shared/sdk/index.js
```diff
@@ -35,44 +38,223 @@ const ADAPTERS = [
...
+  /**
+   * @summary Initialize adapters
+   * @private
+   * @example
+   * // Only to be used internally
+   * this.init()
+   */
+  init () {
+    debug('scanner:init', this)
+    _.map(_.keys(this.options), (adapterId) => {
```
jviotti: Should this be `_.each`, given we're not returning from the callback, nor assigning the results?

jhermsmeier: Fixed!

lib/shared/sdk/usbboot/index.js
```diff
@@ -269,6 +274,8 @@ class USBBootAdapter extends EventEmitter {
         // TODO: Emit an error event if this fails
```
jviotti: Can we delete this now-obsolete comments?

lib/shared/sdk/usbboot/index.js
```diff
@@ -284,10 +291,11 @@ class USBBootAdapter extends EventEmitter {
     }, {
       concurrency: 5
     }).catch((error) => {
-      callback(error)
+      this.emit('error', error)
+      callback && callback(error)
```
jviotti: Should we document that using a callback means that you won't get notified of preparation errors? Maybe better to clear the confusion by just supporting events?

jviotti:
usbboot works fine!

lib/shared/sdk/usbboot/index.js
jhermsmeier: Yeah, good point – let me do that in a follow-up PR.

lurch:
@jhermsmeier There are a whole bunch of mentions in the code here of `subsribe` - are these all typos for `subscribe` ?

jhermsmeier:
@lurch oh my, indeed – fixed in https://github.com/resin-io/etcher/pull/1832

PR #1815, opened 2017-10-27
fix(usbboot): increase device reboot delay a little bit

jviotti:
Change-Type: patch

### DIFF
--- a/lib/shared/sdk/usbboot/index.js
+++ b/lib/shared/sdk/usbboot/index.js
@@ -122,7 +122,7 @@ const USBBOOT_CAPABLE_USB_DEVICES = [
  * @type {Number}
  * @constant
  */
-const DEVICE_REBOOT_DELAY = 5000
+const DEVICE_REBOOT_DELAY = 6000
 
 /**
  * @summary The initial step of the file server usbboot phase


jhermsmeier:
What's the reason behind this?

jviotti:
I did a lot more experimentation and in some very few cases the device took some milliseconds more than this delay.

jhermsmeier:
That highlights what I feared though; that this "timeout-guessing" is probably going to fail in a lot of situations / on other machines etc.

jviotti:
Definitely, but its the best we can do until we fix the device associativity issue.

jviotti:
This is exhibiting the pip issue you mentioned before. I'll merge anyways, since Windows & macOS are passing, and its a very trivial change.

lib/shared/sdk/usbboot/index.js
```diff
@@ -122,7 +122,7 @@ const USBBOOT_CAPABLE_USB_DEVICES = [
  * @type {Number}
  * @constant
  */
-const DEVICE_REBOOT_DELAY = 5000
+const DEVICE_REBOOT_DELAY = 6000
```
lurch: What units is this? ;-)

lurch:
It would be nice to clarify _which_ 'reboot' this is - is it from power-on to stage1 (where we send bootcode.bin), is it from stage1 to stage2 (where we send start.elf and our custom kernel), or is it from stage2 to "stage3" (where a MSD gets exposed back to the host, and we finally have a 'real' blockdevice) ?

And technically this isn't actually a reboot, as a reboot would force the Pi back into stage1.

PR #1818, opened 2017-10-27
fix(scripts): Fix pip install in docker

jhermsmeier:
This works around the "Cannot fetch index base URL http://pypi.python.org/simple/"
error by installing pip==9.0.1 directly from the pypi.python.org/packages/

Change-Type: patch

### DIFF
--- a/scripts/build/docker/Dockerfile-armv7hf
+++ b/scripts/build/docker/Dockerfile-armv7hf
@@ -20,8 +20,10 @@ RUN apt-get update \
     libasound2 \
     libgconf-2-4 \
     libgtk2.0-0 \
-    libx11-xcb1 \
+    libudev-dev \
+    libusb-1.0-0-dev \
     libnss3 \
+    libx11-xcb1 \
     libxss1 \
     libxtst6 \
     libyaml-dev \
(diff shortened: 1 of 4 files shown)


lurch:
Would it be "better" to use https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py ?

lurch:
I assume that's a "no" then :wink:

jhermsmeier:
Whoops, didn't see your comment when I hit the button – but that might not have done what we wanted; not keen to pull & run arbitrary blobs in our CIs, and also can't find a single mention of the pip version there.

lurch:
> can't find a single mention of the pip version there

Presumably it always installs the "latest" stable version of pip.

jhermsmeier:
Nope – the [get-pip.py](https://bootstrap.pypa.io/get-pip.py) file contains a .zip of pip in base85 – not trusting that without checking; and even then that could still change without notice, so don't fancy using it for CI at all.

lurch:
Fair enough - although https://pip.pypa.io/en/stable/development/#release-process implies that the `get-pip.py` script is autogenerated.
\*shrug\*

PR #1824, opened 2017-10-31
chore(scripts): Update Dockerfiles' base to Ubuntu 16.04

jhermsmeier:
This updates the docker base images from Ubuntu 12.04 to the current LTS, Ubuntu 16.04.

Change-Type: patch


lurch:
I'm not sure we want to do this? See e.g. #1364 and #1481 

And also https://electron.atom.io/docs/tutorial/supported-platforms/

jviotti:
Ubuntu 16.04 has a very modern glibc version that will make Etcher incompatible with a lot of other Linux distributions. We should stick with 12.04 (or any other old release) as most of as we can.

Issue #1819, opened 2017-10-27
Use PolKit for Wayland support


Shou:
With Ubuntu 17.10, Wayland is the default display server, and by default Etcher doesn't work.

The error  message (with gksudo):
![image](https://user-images.githubusercontent.com/819413/32118744-af65e14c-bb4a-11e7-9980-6db93b4563a2.png)

```sh
Running gksudo                                                                                                     
Error copying '/run/user/1000/gdm/Xauthority' to '/tmp/libgksu-ibBRqB': No such file or directory  
```

[Issue on Redhat bugzilla closed as WONTFIX][1]:
> OK, to avoid the potential for misunderstood expectations: there are currently no plans to support running graphical apps with sudo under Wayland, and it seems quite unlikely that this will change anytime soon, so I'm going to close this as WONTFIX. The admin:// protocol mentioned by Ray is the way forward for applications that use gvfs to edit root-owned files. **Other applications need to use polkit to move privileged operations out of process.** This should really have happened many years ago, as polkit is no longer some new or mysterious technology.

[Insecure] workaround:
```sh
xhost +si:localuser:root
sudo etcher-electron
```

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1274451

jviotti:
Uh, the WONTFIX part sucks. How are we supposed to support polkit without requiring users to copy a certain configuration file into a certain part of the system? That obviously works for RPMs and DEBs, since we can add it to the package, but not for AppImages. 

Given USB is the only thing that needs this, can we put more time into documenting clear steps to allow a normal user to open USB devices, and then catching access errors on `node-usb`'s `.open()` and showing a nice user error documenting what the user needs to do? And then we can get rid of the elevation at startup Linux hack.

Shou:
It's not super pretty but could we run a stat for the config on startup and copy if it doesn't exist before spawning a process for the graphical parts?

Shou:
So with PolKit we're going to need a service running as root, where I assume we'll just re-use the child writer. Instead we could get around the whole thing by just making the child writer persistent like a service, a change that is platform agnostic so we don't have to maintain separate mechanisms – PolKit over DBus on Linux, IPC on other systems. Either way this ties heavily into the IPC channel stability.

lurch:
@Shou Does this mean that the current builds of Etcher AppImages don't work on Ubuntu 17.10 ?
[...]

PR #1831, opened 2017-11-07
fix: rid linux startup elevation

Shou:
We remove the Linux elevation meant for usbboot device access.

Reverts: #1708
See: #1819
Change-Type: minor
Changelog-Entry: Remove Linux elevation meant for usbboot.


Shou:
image

lib/gui/app.js
```diff
@@ -214,6 +215,24 @@ app.run(($timeout) => {
   })
 
   driveScanner.on('error', (error) => {
+    if (error.message === 'LIBUSB_ERROR_ACCESS') {
+      return permissions.isElevated().then((elevated) => {
+        if (!elevated) {
+          // Stop all adapters but notably usbboot, remove usbboot, and start the remaining adapters
```
lurch: @jhermsmeier Does this usage-pattern suggest there should be a "better" way to dynamically add and remove adapters?

lib/shared/messages.js
```diff
@@ -71,7 +71,11 @@ module.exports = {
       'It looks like this is not a bootable image.\n\n',
       'The image does not appear to contain a partition table,',
       'and might not be recognized or bootable by your device.'
-    ].join(' '))
+    ].join(' ')),
+
+    runAsRootUsbboot: (vars) => {
```
lurch: Perhaps this should be renamed to `needRootForUsbboot` ?

lurch:
@Shou As I asked elsewhere, could you remind me on which distros, and in which situations, this error triggers?

Perhaps the commit-comment could mention that this reverts #1708 and is a fix/workaround for #1819 ?
In fact, looking at the changes in #1708 , should this PR also remove the `providePassword` message, and revert the changes to the dependencies in `electron-builder.yml` ?

Shou:
I haven't tried any distros where root isn't needed without modification for the USB access we do here. Essentially, when `.open()` from the `usb` Node package is called, the error is triggered if you don't have the right permissions, and of course root gets around that.

Also, will add those changes now.

lib/gui/app.js
jhermsmeier: Yeah, the current implementation isn't complete or anything, just what we needed to make it work, and get closer to what we want in terms of the adapters.

jhermsmeier: @Shou does this mean anyone starting Etcher on Linux will get this warning at startup?

Shou: Only if they have a usbboot device plugged in.

Shou:
Done those changes.

lib/gui/app.js
lurch: ...and even then, if they have 'enough' permissions to open USB devices without being root, AIUI they won't see this message either.

Shou:
Thanks! Don't worry, squash and merge seems to be the only merge we can do in this repo so that's automatic.

lib/gui/app.js
```diff
@@ -214,6 +215,24 @@ app.run(($timeout) => {
...
+          driveScanner.stop()
+          driveScanner.adapters.delete('usbboot')
```
jhermsmeier: You should be able to use `driveScanner.unsubscribe('usbboot')` instead of `driveScanner.adapters.delete('usbboot')` here

lib/gui/app.js
jhermsmeier: @lurch just realized that we do have a method for this: `scanner.unsubsribe(adapter)`

lurch: @jhermsmeier You meant `unsubscribe`, right? :rofl:

lib/gui/app.js
lurch: Does he also need to stop and start the driveScanner before unsubscribing, or is that all handled internally?

Shou: Done this now, and also fixed a small bug in `.unsubscribe()`.

lib/gui/app.js
jhermsmeier: Oh jesus... man, how do I keep mucking that up? 😆

lib/gui/app.js
jhermsmeier: @lurch yeah, at this time you still need to stop the scanner before unsubscribing – which doesn't make sense – I'll fix that in a new PR.

jhermsmeier:
This'll need a little rebase after #1832 (sorry 'bout that)

jhermsmeier:
@Shou can you rebase this?

PR #1832, opened 2017-11-08
 fix(sdk): Correct "subscribe" typos

jhermsmeier:
Change-Type: patch

### DIFF
--- a/dictionary
+++ b/dictionary
@@ -3,3 +3,5 @@ aknowledge->acknowledge
 seleted->selected
 reming->remind
 locl->local
+subsribe->subscribe
+unsubsribe->unsubscribe
--- a/lib/shared/sdk/index.js
+++ b/lib/shared/sdk/index.js
@@ -192,7 +192,7 @@ SDK.Scanner = class Scanner extends EventEmitter {
   }
 
   /**
-   * @summary subscribe to an adapter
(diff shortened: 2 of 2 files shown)


dictionary
```diff
@@ -3,3 +3,4 @@ aknowledge->acknowledge
 seleted->selected
 reming->remind
 locl->local
+subsribe->subscribe
```
lurch: Could you add `unsubsribe` too? ;)

jhermsmeier: Sure thing, added!

lib/shared/sdk/index.js
```diff
@@ -192,28 +192,28 @@ SDK.Scanner = class Scanner extends EventEmitter {
   }
 
   /**
-   * @summary Subsribe to an adapter
+   * @summary subscribe to an adapter
```
lurch: Minor nit: should there be a capital 'S' on 'subscribe' here?

jhermsmeier: Yup, fixed! The pleasures of global find & replace tricked me here.

lurch: Sometimes those find'n'replace dialogs have a handy "match exact case only" tickbox :stuck_out_tongue_winking_eye:

PR #1833, opened 2017-11-08
fix: Correct image.size usage in tests and code-comments

lurch:
image.size is always an object, never a plain number

Change-type: patch


jhermsmeier:
@lurch this needs a rebase to fix pip on Travis

lurch:
Hmmm, I thought I had rebased it, but I must have slipped up somewhere...

lurch:
Ahh, I'd updated my local master, but then actually forgotten to rebase the branch I was working on! ;-)  Should be mended now...

PR #1835, opened 2017-11-08
minifix: remove duplicate debug enable in usbboot

Shou:
Changelog-Entry: Remove duplicate debug enabling in usbboot module.
Change-Type: patch

### DIFF
--- a/lib/shared/sdk/usbboot/index.js
+++ b/lib/shared/sdk/usbboot/index.js
@@ -31,8 +31,6 @@ const utils = require('../../utils')
 
 debug.enabled = true
 
-debug.enabled = true
-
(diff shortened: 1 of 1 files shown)


lurch:
This is so bizarre - AFAICT Github went crazy when #1789 got merged?
(see 9815f9a5a1fa5fce6c6e2a62dc4d376a27aaa5ee and then compare that to the individual commits in #1789 )
:confused:

lurch:
Ahhh, looks like that change was also made in 02f7a5f55be3505776bc7f9c6a6dede2b5d7405c so presumably #1789 hadn't been rebased before it got merged?

PR #1836, opened 2017-11-09
upgrade(ipc): Update node-ipc 8.9.2 -> 9.1.1

jhermsmeier:
Todo:

- [ ] Check if `disconnect` events work with reconnects now

Change-Type: patch

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -824,11 +824,6 @@
       "resolved": "https://registry.npmjs.org/color-convert/-/color-convert-0.5.3.tgz",
       "dev": true
     },
-    "colors": {
-      "version": "1.1.2",
-      "from": "colors@>=1.1.2 <2.0.0",
-      "resolved": "https://registry.npmjs.org/colors/-/colors-1.1.2.tgz"
-    },
     "columnify": {
       "version": "1.5.4",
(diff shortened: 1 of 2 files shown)


npm-shrinkwrap.json
```diff
@@ -4682,7 +4682,7 @@
         "form-data": {
           "version": "0.2.0",
           "from": "form-data@>=0.2.0 <0.3.0",
-          "resolved": "http://registry.npmjs.org/form-data/-/form-data-0.2.0.tgz",
+          "resolved": "https://registry.npmjs.org/form-data/-/form-data-0.2.0.tgz",
```
lurch: Weird how a bunch of these have spontaneously switched to `https:` ?

lurch: ...and that change is now causing errors on Travis :sob:

jhermsmeier: Yeah, but they should've never been `http:` in the first place as far as I can see.

jviotti: Lets omit the `https` -> `http` changes from this PR to make Travis happy and get this merged ASAP

jviotti:
We should have probably included a changelog entry on this PR, like "Fix stuck in Starting... if there was an error when initializing the IPC server." I guess we can manually add it to the CHANGELOG.

PR #1841, opened 2017-11-13
fix(SDK): disable usbboot adapter on GNU/Linux

jviotti:
Various GNU/Linux distributions require root access to be able to open
USB devices. This means that Etcher would need to be ran as root, which
is now not possible on Wayland based systems.

We declare usbboot as unsupported on GNU/Linux from the time being, but
we are currently working on a solution.

Change-Type: patch

### DIFF
--- a/lib/shared/sdk/index.js
+++ b/lib/shared/sdk/index.js
@@ -20,6 +20,7 @@ const EventEmitter = require('events')
 const _ = require('lodash')
 const SDK = module.exports
 const debug = require('debug')('sdk')
+const os = require('os')
 
 debug.enabled = true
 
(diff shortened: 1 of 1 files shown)


lurch:
Looks like you need to modify the tests so that we don't try to test usbboot on Linux?

jhermsmeier:
Guess we have two options here now:

1. Adjust the options
2. Only warn on options for an unknown adapter

Both have their pros, and cons – mainly visibility to us in case of misconfiguration, but I think I'm leaning towards option 2

jviotti:
Oh, I see. I agree that option 2 sounds reasonable.

jviotti:
Updated!

PR #1843, opened 2017-11-14
fix(child-writer): Handle exits due to a signal

jhermsmeier:
This adds handling for cases where the writer child process
exits due to reception of a signal, while also adjusting some peripherals,
like the IPC socket directory and inherited process environment.

Another addition is that the child process is explicitly killed
should an error arise on the IPC.

Change-Type: patch
Connects To: #1801
Obsoletes: #1801

### DIFF
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -26,6 +26,16 @@ const CONSTANTS = require('./constants')
 const EXIT_CODES = require('../shared/exit-codes')
 const robot = require('../shared/robot')
 
+// There might be multiple Etcher instances running at
+// the same time, therefore we must ensure each IPC
+// server/client has a different name.
+process.env.IPC_SERVER_ID = `etcher-server-${process.pid}`
+process.env.IPC_CLIENT_ID = `etcher-client-${process.pid}`
+
+ipc.config.id = process.env.IPC_SERVER_ID
+ipc.config.socketRoot = CONSTANTS.TMP_DIRECTORY
+ipc.config.silent = true
+
 /**
  * @summary Perform a write
  * @function
@@ -67,14 +77,6 @@ exports.write = (image, drive, options) => {
     unmountOnSuccess: options.unmountOnSuccess
   })
 
-  // There might be multiple Etcher instances running at
-  // the same time, therefore we must ensure each IPC
-  // server/client has a different name.
-  process.env.IPC_SERVER_ID = `etcher-server-${process.pid}`
-  process.env.IPC_CLIENT_ID = `etcher-client-${process.pid}`
-
-  ipc.config.id = process.env.IPC_SERVER_ID
-  ipc.config.silent = true
   ipc.serve()
 
   /**
@@ -192,7 +194,7 @@ exports.write = (image, drive, options) => {
 
     child.on('error', emitError)
 
-    child.on('close', (code) => {
+    child.on('exit', (code, signal) => {
       terminateServer()
 
       if (code === EXIT_CODES.CANCELLED) {
@@ -207,7 +209,11 @@ exports.write = (image, drive, options) => {
         return null
       }
 
-      return emitError(new Error(`Child process exited with error code: ${code}`))
+      const error = new Error(`Child process exited with code ${code}, signal ${signal}`)
+      error.code = code
+      error.signal = signal
+
+      return emitError(error)
     })
   })
 
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -27,6 +27,9 @@ const EXIT_CODES = require('../shared/exit-codes')
 const robot = require('../shared/robot')
 const permissions = require('../shared/permissions')
 const packageJSON = require('../../package.json')
+const CONSTANTS = require('./constants')
+
+/* eslint-disable no-eq-null */
 
 // This script is in charge of spawning the writer process and
 // ensuring it has the necessary privileges. It might look a bit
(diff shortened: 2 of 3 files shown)


lib/child-writer/writer-proxy.js
```diff
@@ -109,62 +122,77 @@ permissions.isElevated().then((elevated) => {
...
+     *   foo: 'bar'
+     * })));
+     */
+    const emitMessage = (data) => {
+      // Output from stdout/stderr coming from the CLI might be buffered,
+      // causing several progress lines to come up at once as single message.
+      // Trying to parse multiple JSON objects separated by new lines will
+      // of course make the parser confused, causing errors later on.
+      _.each(utils.splitObjectLines(data.toString()), (object) => {
+        ipc.of[process.env.IPC_SERVER_ID].emit('message', object)
```
lurch: As `ipc.of[process.env.IPC_SERVER_ID].emit('message', object)` is also used in `onError` below, would it be worth abstracting that part out?

lib/child-writer/writer-proxy.js
```diff
@@ -109,62 +122,77 @@ permissions.isElevated().then((elevated) => {
...
+     * @param {Buffer} data - json message data
+     *
+     * @example
+     * emitMessage(Buffer.from(JSON.stringify({
+     *   foo: 'bar'
+     * })));
+     */
+    const emitMessage = (data) => {
+      // Output from stdout/stderr coming from the CLI might be buffered,
+      // causing several progress lines to come up at once as single message.
```
lurch: Could the buffering mentioned here mean that we might also get incomplete lines? e.g.
```
oo"}\n{"id":"bar"}\n{"id":"baz"}
```
or
```
{"id":"foo"}\n{"id":"bar"}\n{"id":"baz"
```
AFAICT `utils.splitObjectLines` has no handling for that.

Or are we sure we'll always get some number of "complete" lines?

lib/child-writer/index.js
```diff
@@ -192,7 +194,7 @@ exports.write = (image, drive, options) => {
 
     child.on('error', emitError)
 
-    child.on('close', (code) => {
+    child.on('exit', (code, signal) => {
```
jviotti: Is `signal` being used?

lib/child-writer/index.js
```diff
@@ -26,6 +26,16 @@ const CONSTANTS = require('./constants')
...
 const robot = require('../shared/robot')
 
+// There might be multiple Etcher instances running at
+// the same time, therefore we must ensure each IPC
+// server/client has a different name.
+process.env.IPC_SERVER_ID = `etcher-server-${process.pid}`
+process.env.IPC_CLIENT_ID = `etcher-client-${process.pid}`
+
+ipc.config.id = process.env.IPC_SERVER_ID
+ipc.config.socketRoot = CONSTANTS.TMP_DIRECTORY
```
jviotti: Just trying to understand: what is the reason for this change?

lib/child-writer/index.js
lurch: Does `signal` need to be handled here at all?

lib/child-writer/index.js
jhermsmeier: I've just pulled them out to lowest the scope they need to be in, as they only need to be set once, and logically it didn't make sense to me to have them buried that deep in the handlers.

lib/child-writer/index.js
jhermsmeier: Whoops. It is now :)

lib/child-writer/writer-proxy.js
jhermsmeier: We don't write partial JSON to stdio, so it'll always be multiple lines containing valid, parseable JSON; btw this is also not new code, just moved it a bit.

lib/child-writer/writer-proxy.js
jhermsmeier: I don't think 2 repetitions of a single (partial) line, which is just a method call warrant any abstraction

lib/child-writer/index.js
jhermsmeier: I'm not sure I follow? Why wouldn't it be? The writer-proxy is also prone to being annihilated by a SIGKILL or so, or what am I missing?

lib/child-writer/writer-proxy.js
lurch: > We don't write partial JSON to stdio

I wondered if maybe it was possible that whatever "buffering" is being used is getting full, and so flushing before we expect it to?

> btw this is also not new code, just moved it a bit.

Yer I know!  I was just wondering if this might be another possible corner-case that we're not dealing with ;)

lib/child-writer/index.js
lurch: Hah, @jviotti beat me to it ;)

lib/child-writer/writer-proxy.js
jhermsmeier: > I wondered if maybe it was possible that whatever "buffering" is being used is getting full, and so flushing before we expect it to?

The `console.log()`ing flushes it, so the buffering in terms of partials won't be an issue.

> Yer I know! I was just wondering if this might be another possible corner-case that we're not dealing with ;)

Yeah, was wondering the same while I was reading over it, but then realized that we're controlling that, so didn't worry about it anymore. Still not worried. Also think that we definitely would've seen it, should it have happened.

lib/child-writer/index.js
jhermsmeier: Oh, that's about that one! I only saw your comment after I already pushed it, hence the great confusion :)

PR #1847, opened 2017-11-15
fix(versionist): Ignore case of Change-Type value

jhermsmeier:
There's a commit that slipped in (sorry) with a capitalised
Change-Type value of "Patch". This avoids versionist erroring,
by always lowercasing that value.

Change-Type: PATCH

### DIFF
--- a/versionist.conf.js
+++ b/versionist.conf.js
@@ -38,7 +38,11 @@ module.exports = {
   },
 
   getIncrementLevelFromCommit: (commit) => {
-    return commit.footer['Change-Type']
+    if (/none/i.test(commit.footer['Change-Type'])) {
+      return null
+    }
+    return commit.footer['Change-Type'] &&
+      _.toLower(commit.footer['Change-Type'])
   },
 
   transformTemplateData: (data) => {


lurch:
I guess you need to add an eslint exclude-rule to make the CIs happy.

lurch:
Does this mean that we'll _always_ need this customised `versionist.conf.js`, given that this error slipped through into Etcher's git history? (and is that why this is in the v1.2.1 milestone?)

jhermsmeier:
Yeah, either that, or we remove it to use versionist's default config, which already had these issues fixed.

lurch:
IIRC we decided that versioning was going to be done 'differently' on Etcher (there was talk of using vanity semver), which is why we're not using versionist's default config? (I'm afraid I can't really remember now)

versionist.conf.js
```diff
@@ -38,7 +38,11 @@ module.exports = {
   },
 
   getIncrementLevelFromCommit: (commit) => {
-    return commit.footer['Change-Type']
+    if (/none/i.test(commit.footer['Change-Type'])) {
```
lurch: If Etcher now "officially supports" `Change-Type: none` (it didn't before), should we also update https://github.com/resin-io/etcher/blob/master/docs/COMMIT-GUIDELINES.md#change-type-type ?

jviotti: I merged this so we can try to release today, but I agree a follow-up PR with such information would be nice.

PR #1850, opened 2017-11-16
fix(store): Fix "Missing state eta" error on Infinity/NaN

jhermsmeier:
If the speed is zero, the eta becomes Infinity, which isn't transmitted
properly over IPC for at this time unknown reasons.
To prevent the "Missing state eta" error from popping up when the speed is
zero, we set the eta to zero as well in those cases.

Change-Type: patch
Changelog-Entry: Fix "Missing state eta" error when speed is zero

### DIFF
--- a/lib/writer/progress-stream.js
+++ b/lib/writer/progress-stream.js
@@ -70,7 +70,10 @@ class ProgressStream extends Stream.Transform {
       this.state.remaining = this.state.length - this.state.transferred
       this.state.runtime = Date.now() - this.start
       this.state.speed = this.meter(this.state.delta)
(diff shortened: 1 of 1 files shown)


jviotti:
@jhermsmeier The linter is complaining: https://ci.appveyor.com/project/resin-io/etcher/build/1.0.6440/job/d776ewmv0f73g4ki

jviotti:
I cancelled the remaining builds so we have macOS builds available for the other PR.

jhermsmeier:
Ah damnit. Thanks, updated now.

jviotti:
Lets add a changelog entry when merging.

PR #1851, opened 2017-11-16
fix(child-writer): Pass parent env during elevation

jhermsmeier:
Change-Type: patch
Connects To: #1843


lurch:
Doesn't matter now (given that it was a bit of an urgent hotfix), but I guess in hindsight it would have been nice if the commit message mentioned what problem this was fixing.

jhermsmeier:
Turns out this was a stupid idea, as we're not escaping env vars properly in `permissions.getEnvironmentCommandPrefix()` from what it looks like – and this causes it to fail (rather get stuck) on Windows.

PR #1852, opened 2017-11-17
fix(drive-scanner): Dynamically get unsafe mode setting

jhermsmeier:
Due to the SDK keeping it's options once initialized,
the unsafe mode setting change was not reflected on subsequent scans.

Change-Type: patch


lurch:
It wasn't _just_ "subsequent scans" that were affected, but also the first scan - even if you enabled unsafe mode and restarted Etcher, the drive-selector didn't show any system drives.
But it's all good now.

PR #1853, opened 2017-11-17
fix(gui): Re-enable printing progress in console

jhermsmeier:
Change-Type: patch

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -185,20 +185,15 @@ app.run(() => {
 
     const currentFlashState = flashState.getFlashState()
 
-    // There is usually a short time period between the `isFlashing()`
+    // NOTE: There is usually a short time period between the `isFlashing()`
     // property being set, and the flashing actually starting, which
     // might cause some non-sense flashing state logs including
     // `undefined` values.
-    //
-    // We use the presence of `.eta` to determine that the actual
-    // writing started.
-    if (!currentFlashState.eta) {
-      analytics.logDebug(
-        `Progress (${currentFlashState.type}): ` +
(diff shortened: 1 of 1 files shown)


lib/gui/app.js
```diff
@@ -185,20 +185,15 @@ app.run(() => {
...
 
-    // There is usually a short time period between the `isFlashing()`
+    // NOTE: There is usually a short time period between the `isFlashing()`
     // property being set, and the flashing actually starting, which
     // might cause some non-sense flashing state logs including
     // `undefined` values.
-    //
-    // We use the presence of `.eta` to determine that the actual
-    // writing started.
-    if (!currentFlashState.eta) {
```
lurch: As I mentioned in flowdock, I believe this was a simple logic error and the `if (!currentFlashState.eta)` here should have been `if (currentFlashState.eta)`. I've just tested that locally, and it brings back the progress-spam to DevTools :wink:

jhermsmeier: Yeah, but eta can be zero (if the speed is zero), and we'll want to have that context in case of a bug, hence that check is removed completely here.

lurch: Ah, okay.

lurch:
https://ci.appveyor.com/project/resin-io/etcher/history indeed shows that Appveyor has only built the 'branch' build for this, and never even attempted the 'pr' build :-/

lurch:
FYI @jhermsmeier I restarted the x86 Linux builds on both Travis `push` and Travis `pr` - one errored about `nvm` failing, and one errored about `apt-get` failing, so I guess they must have both been hit with transient network failures at the same time. Both seem to be progressing okay now...

Issue #860, opened 2016-11-10
Improve visual feedback on element :focus


Shou:
Navigation using the tab key is tough because visual feedback is either too subtle or nonexistent when an element gains `:focus`, e.g. the "Select image" button, or warning dialog button(s).

jviotti:
Pinging @taahirisaacs and @konmouz for input on visual feedback.

jhermsmeier:
This is happening in https://github.com/resin-io/etcher/pull/1745

lurch:
Closing as duplicate of #1734

PR #1854, opened 2017-11-17
fix(gui): Disable outlines

jhermsmeier:
For the time being, the focus outlines are removed, until
we can improve the UX on this.

Change-Type: patch
Connects To: #1745 #1734 #860


lurch:
This sorta temporarily reverts #1745

PR #1855, opened 2017-11-17
fix(s3): Fix EAI_AGAIN error at startup with no internet connection

lurch:
Change-type: patch


lurch:
I found this problem when trying to start Etcher without an active internet connection.

Issue #1856, opened 2017-11-17
webview success banner height is too small.


craig-mulligan:
- **Etcher version:**
- **Operating system and architecture:**
- **Image flashed:**
- **Do you see any meaningful error information in the DevTools?**

lurch:
What do you mean by "too small"?

craig-mulligan:
It's height is set to 255px and doesn't take advantage of the height of the default launch window.

lurch:
ping @Shou and @konmouz

craig-mulligan:
We realised there is was a burn again banner that I wasn't seeing because it was rending incorrectly so we actually are using most of the vertical space available - closing.

PR #1857, opened 2017-11-17
fix(gui): Increase height of webview

craig-mulligan:
connects to #1856

Change-Type: patch

### DIFF
--- a/lib/gui/components/safe-webview.js
+++ b/lib/gui/components/safe-webview.js
@@ -67,7 +67,7 @@ const API_VERSION_PARAM = 'api-version'
  * This version number is exposed to the banner such that it can determine what
  * features are safe to utilize.
  */
-const API_VERSION = 1
+const API_VERSION = 2
 
 /**
  * @summary Webviews that hide/show depending on the HTTP status returned
--- a/lib/gui/css/main.css
+++ b/lib/gui/css/main.css
@@ -6725,7 +6725,7 @@ body {
   bottom: 0;
   left: 0;
   width: 100%;
-  height: 255px; }
+  height: 340px; }
 
 .wrapper {
   height: 100%;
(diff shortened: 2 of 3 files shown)


jhermsmeier:
Can you add before & after screenshots for a quick visual impression?

jhermsmeier:
Ci is failing because the sass changes aren't committed ;)

craig-mulligan:
The sass is committed? Should I build em too?

craig-mulligan:
screen shot 2017-11-17 at 13 34 52
screen shot 2017-11-17 at 13 37 10

lurch:
That second screenshot isn't what Etcher currently shows on success, which is actually this:
image

jhermsmeier:
> The sass is committed? Should I build the css?

@craig-mulligan yeah, sorry – that's what I meant

craig-mulligan:
Oh @lurch - okay that `flash again` doesn't come up on my screen, I'm testing manually be setting the `success` state. I can't test it properly because etcher fails on my machine. (looks like elevation issues). So maybe there isn't actually more room to be used. @lurch do you mind pulling and testing this on your machine?

craig-mulligan:
If not, don't worry about it. I don't think this is v urgent.

craig-mulligan:
Yea defo's looks like the height used curbed to maintain that flash complete status. - closing.

jhermsmeier:
@craig-mulligan what kind of elevation issues are you running into? Are you on current master?

craig-mulligan:
@jhermsmeier yea, but it's happening with my standard etcher `1.1.2` as well. It happened a couple weeks ago, assumed it was the drive but looks like it's my machine. In the way of errors nothing throws, it starts and then just logs elevation cancelled as second later.

jhermsmeier:
@craig-mulligan can you pull again & run `make electron-develop` before starting it? – This should have been fixed on current master, but your dependencies might be stale.

craig-mulligan:
@jhermsmeier I get the same result. I'll open an separate issue for this.

PR #1858, opened 2017-11-17
fix(writer): Don't calcluate checksum before block-stream

jhermsmeier:
The checksum-stream being situated in front of the block-stream, which
ensures block size alignment to multiples of 512, and pads the last block,
caused the checksum to be incorrectly calculated for images where the last
block needed to be padded.

Change-Type: patch


jviotti:
There is a typo in the commit message: `calcluate` -> `calculate`, but we can probably fix it when merging from GitHub's interface.

PR #1859, opened 2017-11-17
fix(writer): Emit checksum event when verifying bmaps

jhermsmeier:
Due to the Blockmap.FilterStream not emitting a "checksum"
event (as it individually verifies specified ranges), the
flashing process would get stuck on finish.

This emits a "checksum" event on "finish" when blockmapping,
averting this issue.

Change-Type: patch

### DIFF
--- a/lib/writer/index.js
+++ b/lib/writer/index.js
@@ -271,7 +271,15 @@ class ImageWriter extends EventEmitter {
     if (options.image.bmap) {
       debug('verify:bmap')
       const blockMap = BlockMap.parse(options.image.bmap)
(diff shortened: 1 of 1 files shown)


jhermsmeier:
This initially worked up until a recent change in how the verification process ends in the new writer;
Also the FilterStream should probably be mended to also emit a "checksum" event once all regions have been verified.

jviotti:
@jhermsmeier can we merge this?

lurch:
So is the fix in here redundant once https://github.com/resin-io-modules/blockmap/issues/40 has been implemented?

PR #1861, opened 2017-11-17
upgrade(package): Bump sudo-prompt 6.1.0 -> 8.0.0

jhermsmeier:
This updates `sudo-prompt` from v6.1.0 to v8.0.0

Changes:

- v8.0.0:
  - Breaking: Windows: Set code page of command batch script to UTF-8
- v7.1.1:
  - readme: explicitly mention that no child process is returned
- v7.1.0:
  - Detect when PowerShell fails to launch command
  - Escape ampersand and other characters special to PowerShell
- v7.0.0:
  - Add status code to error on Windows and macOS
- v6.2.0:
  - Rename OS X to macOS

Tested on:

- [x] Mac OS
- [x] Windows
- [x] Linux

Change-Type: patch

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9560,9 +9560,9 @@
       "dev": true
     },
     "sudo-prompt": {
-      "version": "6.1.0",
-      "from": "sudo-prompt@6.1.0",
-      "resolved": "https://registry.npmjs.org/sudo-prompt/-/sudo-prompt-6.1.0.tgz"
+      "version": "8.0.0",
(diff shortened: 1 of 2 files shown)


jviotti:
Maybe worth testing this in all OSes to ensure nothing is broken? Mainly on Windows, given that most of the changes described in `sudo-prompt`'s CHANGELOG affect that platform.

jhermsmeier:
Yeah, we also don't *need* this for the release, whereas it'd be good if we can ship with it

PR #1863, opened 2017-11-17
feat(writer): Display actual write speed

jhermsmeier:
As we've actually been displaying the read-speed in various
forms during the flashing process, this is a venture into
displaying the actual write-speed from the end of the pipeline.

Change-Type: minor
Changelog-Entry: Display actual write speed

### DIFF
--- a/lib/writer/block-write-stream.js
+++ b/lib/writer/block-write-stream.js
@@ -18,9 +18,11 @@
 
 const stream = require('readable-stream')
 const fs = require('fs')
+const speedometer = require('speedometer')
 const debug = require('debug')('block-write-stream')
 
 const CHUNK_SIZE = 64 * 1024
+const UPDATE_INTERVAL_MS = 500
 
 /**
  * @summary BlockWriteStream
@@ -60,6 +62,23 @@ class BlockWriteStream extends stream.Writable {
     this.blocksRead = 0
     this.bytesWritten = 0
     this.blocksWritten = 0
+    this.meter = speedometer()
+    this.delta = 0
+    this.speed = 0
+
+    this.clear = () => {
+      clearInterval(this.timer)
+    }
+
+    this.update = () => {
+      this.speed = this.meter(this.delta)
+      this.delta = 0
+    }
+
+    this.once('end', this.clear)
+    this.once('error', this.clear)
+
+    this.timer = setInterval(this.update, UPDATE_INTERVAL_MS)
 
     this.closed = false
     this.destroyed = false
@@ -116,6 +135,7 @@ class BlockWriteStream extends stream.Writable {
 
     fs.write(this.fd, chunk, 0, chunk.length, chunk.position, (error, bytesWritten) => {
       this.bytesWritten += bytesWritten
+      this.delta += bytesWritten
       this.blocksWritten += 1
       this.position += bytesWritten
       next(error)
--- a/lib/writer/index.js
+++ b/lib/writer/index.js
@@ -226,6 +226,7 @@ class ImageWriter extends EventEmitter {
     progressStream.on('progress', (state) => {
       state.device = options.path
       state.type = 'write'
+      state.speed = target.speed
       this.emit('progress', state)
     })
 


jviotti:
@jhermsmeier Should we add a changelog entry?

jhermsmeier:
Yes. I keep forgetting this.

jhermsmeier:
Added changelog entry

lurch:
The AppVeyor Windows x86 build has failed with:
```
1) Browser: ImageWriter ImageWriterService .flash() given a successful write should set flashing to false when done:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at done (C:\projects\etcher\node_modules\mocha\lib\runnable.js:292:13)
      at callFn (C:\projects\etcher\node_modules\mocha\lib\runnable.js:363:7)
      at Test.Runnable.run (C:\projects\etcher\node_modules\mocha\lib\runnable.js:337:7)
      at Runner.runTest (C:\projects\etcher\node_modules\mocha\lib\runner.js:444:10)
      at C:\projects\etcher\node_modules\mocha\lib\runner.js:550:12
      at next (C:\projects\etcher\node_modules\mocha\lib\runner.js:361:14)
      at C:\projects\etcher\node_modules\mocha\lib\runner.js:371:7
      at next (C:\projects\etcher\node_modules\mocha\lib\runner.js:295:14)
      at C:\projects\etcher\node_modules\mocha\lib\runner.js:334:7
      at done (C:\projects\etcher\node_modules\mocha\lib\runnable.js:295:5)
      at callFn (C:\projects\etcher\node_modules\mocha\lib\runnable.js:363:7)
      at Hook.Runnable.run (C:\projects\etcher\node_modules\mocha\lib\runnable.js:337:7)
      at next (C:\projects\etcher\node_modules\mocha\lib\runner.js:309:10)
      at Immediate.<anonymous> (C:\projects\etcher\node_modules\mocha\lib\runner.js:339:5)
      at runCallback (timers.js:672:20)
      at tryOnImmediate (timers.js:645:5)
      at processImmediate [as _immediateCallback] (timers.js:617:5)
```
I haven't restarted it though, in case this is a genuine error that you want to dig into @jhermsmeier . The other 3 CI builds all passed.

lib/writer/block-write-stream.js
```diff
@@ -60,6 +61,23 @@ class BlockWriteStream extends stream.Writable {
...
+
+    this.update = () => {
+      this.speed = this.meter(this.delta)
+      this.delta = 0
+    }
+
+    this.once('end', this.clear)
+    this.once('error', this.clear)
+
+    this.timer = setInterval(this.update, 500)
```
lurch: Should we have a named constant here e.g. `UPDATE_INTERVAL_MS` ?

lib/writer/index.js
```diff
@@ -226,6 +226,7 @@ class ImageWriter extends EventEmitter {
     progressStream.on('progress', (state) => {
       state.device = options.path
       state.type = 'write'
+      state.speed = target.speed
```
lurch: Hmmm, so is this simply 'overwriting' the speed that we used to be setting from the ReadStream (or something) ?  Does it make sense to still be recording the ReadStream speed if we're now not using it?

Does a similar change to what you've done here need to be made in `_createVerifyPipeline` ?

lurch: ...or perhaps it'd make sense for _all_ the streams you added with pipage to have their own speedometer, to make it easier in the code elsewhere to pick'n'choose which speed(s) to display? \*shrug\*

jhermsmeier: > used to be setting from the ReadStream (or something)

It's the progress-stream

> Does it make sense to still be recording the ReadStream speed if we're now not using it?

Yes, otherwise the progress-stream will not be able to calculate the ETA and other values, which are based on the source

> ...or perhaps it'd make sense for all the streams you added with pipage to have their own speedometer, to make it easier in the code elsewhere to pick'n'choose which speed(s) to display?

No, this is perf sensitive code, and speed metering does come with additional cost. Also the speed won't differ within the pipeline. The only reason the source & target speeds differ is when the source is compressed, and less data is read than is written.

>  Does a similar change to what you've done here need to be made in _createVerifyPipeline ?

No. The verify pipeline doesn't write to anywhere.

lurch: Thanks for the clarification :+1:  Some of the answers now seem obvious, post-hangover ;-)

Although if the performance is really that sensitive, I guess the speedometer on the original progress-stream can be removed?

lurch: Ah, nope, I've just found https://github.com/resin-io/etcher/blob/master/lib/writer/progress-stream.js#L76

I didn't realise the ETA was calculated like this by the progress-stream. (with the writer / writer-proxy / etc. and now with added pipage, it's easy to get confused about what part of the code is doing what :wink: )

lib/writer/block-write-stream.js
jviotti: Agreed

jviotti:
This is blowing up with compressed images. For example:

screen shot 2017-11-21 at 1 20 42 pm

jviotti:
@jhermsmeier I'm not sure if the speed is right with compressed images. Check this:

screen shot 2017-11-21 at 1 36 03 pm

I get 14 MB write speed, and the eta is 1.5m to write the last 27% of a 300 MB file?

The progress reported on `master` seems more reasonable:

screen shot 2017-11-21 at 1 37 51 pm

jviotti:
I'm wondering if the above is bzip2 related. I tried an xz image and I do get reasonable sizes/etas from this branch.

jhermsmeier:
It is compression related, yeah – the reason behind that is that the ETA can only be calculated for the read-stream side if the final size is an estimation, resulting in an over-estimation of the ETA for compressed sources.

jhermsmeier:
That effect is naturally more pronounced if the source has a high compression ratio (which is usually the case with bzipped images)

lib/writer/block-write-stream.js
jhermsmeier: Added

jhermsmeier:
@jviotti can you clarify that the "stream.unpipe is not a function" error isn't relevant here to avoid confusion in the future (when re-reading this thread, i.e.)?

lurch:
> the ETA can only be calculated for the read-stream side if the final size is an estimation, resulting in an over-estimation of the ETA for compressed sources.

Actually, I think it'll be an _under_-estimation of the ETA for compressed sources - which is why #1171 happens.

I once suggested (in #638 ) that a possible workaround for that (for non-streamed images) would be to spawn a separate 'parallel thread' which just decompresses the whole image (and discards the resultant data) to work out what the real final uncompressed size is, and then use that to update the ETA and progress bar. (which of course assumes that it's the card-writing rather than the data-decompression which is the speed bottleneck, but I think that'll always be the case).
I.e.
 1. User selects a bz2 image file and clicks "Flash"
 1. Etcher doesn't know the decompressed size, so the progress bar and ETA are calculated on the still-compressed data (and hence are inaccurate) as it gets piped through the reader -> decompressor part of the pipeline
 1. Etcher starts up the usual writer process which decompresses the data and starts writing it to card
    * However at the same time, Etcher also spawns a process to just decompress the data to work out its uncompressed size - if possible, this should be running with a lower priority than the primary thread, so as to not slow down the card-writing in the primary thread
 1. While the secondary thread is still decompressing, the first thread remains writing data to card (with the inaccurate ETA and progress bar)
 1. Once the secondary thread has finished decompressing the data (which will happen _way_ before the primary thread finishes), we now know the **real** final uncompressed size (we can now abort the writing thread early if it turns out the uncompressed data is larger than the card we're writing to, rather than waiting for the primary writing thread to "fall off the end")
 1. So the progress-bar and ETA can now be updated to be accurately calculated based on the decompressor -> writer part of the pipeline (which may well cause them to "jump backwards" a bit)

However I'm not sure it's worth adding that level of complexity in order to "fix" this problem of the ETA and progress bar being inaccurate for very compressed images. And of course it won't work for compressed images which are streamed directly from the network and onto an SD card. And it'll increase the CPU-usage of Etcher (for these compressed files where we can't get an accurate final size) on low-powered computers.

PR #1872, opened 2017-11-20
fix(child-writer): Reduce env vars passed during elevation

jhermsmeier:
As we're not escaping env vars properly in `permissions.getEnvironmentCommandPrefix()`,
passing the entire environment along can cause another instance of the "stuck at starting"
problem, thus we limit the vars explicitly passed along again, keeping PATH and DEBUG.

Tested on:

- [x] Windows
- [x] Mac OS
- [ ] Linux

Change-Type: patch
Connects To: #1851

### DIFF
--- a/lib/child-writer/writer-proxy.js
+++ b/lib/child-writer/writer-proxy.js
@@ -101,7 +101,10 @@ permissions.isElevated().then((elevated) => {
     // For debugging purposes
     console.log(`Running: ${commandArguments.join(' ')}`)
 
-    const commandEnv = _.assign({}, process.env, {
+    const commandEnv = {
+      PATH: process.env.PATH,
+      DEBUG: process.env.DEBUG,
+      XDG_RUNTIME_DIR: process.env.XDG_RUNTIME_DIR,
       ELECTRON_RUN_AS_NODE: 1,
       IPC_SERVER_ID: process.env.IPC_SERVER_ID,
       IPC_CLIENT_ID: process.env.IPC_CLIENT_ID,
@@ -110,8 +113,7 @@ permissions.isElevated().then((elevated) => {
       // desktop integration script from presenting the
       // "installation" dialog.
       SKIP: 1
-
-    })
+    }
 
     return permissions.elevateCommand(commandArguments, {
       applicationName: packageJSON.displayName,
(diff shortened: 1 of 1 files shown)


lib/child-writer/writer-proxy.js
```diff
@@ -176,11 +178,6 @@ permissions.isElevated().then((elevated) => {
       ipc.of[process.env.IPC_SERVER_ID].on('connect', () => {
         // Inherit the parent evnironment
         const childEnv = _.assign({}, process.env, {
-
-          // The CLI might call operating system utilities (like `diskutil`),
-          // so we must ensure the `PATH` is inherited.
-          PATH: process.env.PATH,
```
lurch: Why was this line removed?

lib/child-writer/writer-proxy.js
```diff
@@ -176,11 +178,6 @@ permissions.isElevated().then((elevated) => {
       ipc.of[process.env.IPC_SERVER_ID].on('connect', () => {
         // Inherit the parent evnironment
         const childEnv = _.assign({}, process.env, {
```
lurch: How come this still uses the "bulk" `process.env`, rather than only copying out specific keys as the change on line 104 does?

jhermsmeier: Because this is a "normal" child process spawn, whereas the elevation prompt gets the env prepended to the command, which causes the issue as mentioned in the description – hence we can just pump the entire env through here.

lib/child-writer/writer-proxy.js
jhermsmeier: See PR description ;)

lib/child-writer/writer-proxy.js
lurch: Right, read that a couple of times and now I understand ;-)
So we _can_ just pump the entire env through here, but _should_ we? Would it be "better" to only pass in the env-vars that we know we'll need?

lib/child-writer/writer-proxy.js
lurch: The PR description doesn't explain why this line was removed, but your comment above about "just pump the entire env through" does :stuck_out_tongue:

PR #1873, opened 2017-11-21
fix(child-writer): Kill child process chain on exit

jhermsmeier:
This aims to ensure that the child processes exit properly,
and exit at all, if Etcher is quit while the flashing is still
in progress.

Change-Type: patch
Changelog-Entry: Ensure the writer process exits when Etcher is quit

### DIFF
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -75,6 +75,7 @@ class ChildWriter extends EventEmitter {
     this.child = null
 
     this.on('close', () => {
+      debug('close')
       this.aborting = false
       ChildWriter.currentInstance = null
     })
@@ -90,6 +91,7 @@ class ChildWriter extends EventEmitter {
    * child.on('error', this.onError)
    */
   onError (error) {
+    debug('onError', error)
     this.abort()
     this.emit('error', error)
   }
@@ -105,9 +107,10 @@ class ChildWriter extends EventEmitter {
    * child.on('exit', this.onExit)
    */
   onExit (code, signal) {
-    this.terminateServer()
+    debug('onExit', code, signal)
 
     if (code === EXIT_CODES.CANCELLED) {
+      this.terminateServer()
       this.emit('done', {
         cancelled: true
       })
@@ -117,6 +120,7 @@ class ChildWriter extends EventEmitter {
     // We shouldn't emit the `done` event manually here
     // since the writer process will take care of it.
     if (code === EXIT_CODES.SUCCESS || code === EXIT_CODES.VALIDATION_ERROR) {
+      this.terminateServer()
       return
     }
 
--- a/lib/child-writer/index.js
+++ b/lib/child-writer/index.js
@@ -16,7 +16,7 @@
 
 'use strict'
 
-const EventEmitter = require('events').EventEmitter
+const EventEmitter = require('events')
 const _ = require('lodash')
 const childProcess = require('child_process')
 const ipc = require('node-ipc')
@@ -25,6 +25,7 @@ const cli = require('./cli')
 const CONSTANTS = require('./constants')
 const EXIT_CODES = require('../shared/exit-codes')
 const robot = require('../shared/robot')
+const debug = require('debug')('child-writer')
 
 // There might be multiple Etcher instances running at
 // the same time, therefore we must ensure each IPC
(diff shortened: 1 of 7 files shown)


lib/gui/etcher.js
```diff
@@ -19,20 +19,10 @@
...
-electron.app.on('window-all-closed', electron.app.quit)
+const debug = require('debug')('gui:etcher')
 
-// Sending a `SIGINT` (e.g: Ctrl-C) to an Electron app that registers
-// a `beforeunload` window event handler results in a disconnected white
-// browser window in GNU/Linux and macOS.
-// The `before-quit` Electron event is triggered in `SIGINT`, so we can
-// make use of it to ensure the browser window is completely destroyed.
-// See https://github.com/electron/electron/issues/5273
-electron.app.on('before-quit', () => {
```
jviotti: But Ctrl-C keeps working, right?

lib/child-writer/index.js
```diff
@@ -165,59 +172,222 @@ exports.write = (image, drive, options) => {
...
+ *
+ * child.on('done', () => {
+ *   console.log('Validation was successful!')
+ * })
+ */
+ChildWriter.write = (image, drive, options) => {
+  if (ChildWriter.currentInstance) {
+    throw new Error('Write already in progress')
+  }
+  const writer = new ChildWriter().write(image, drive, options)
```
jviotti: Not something we have to deal with right before the release, but why the extra complexity of making this a class, when it could be modeled as a set of mostly stateless functions? I see your subclassing EventEmitter, but looks like we can use composition instead of inheritance as well.

lib/gui/etcher.js
jhermsmeier: Yup, see a bit further down, at the `mainWindow.on('close')` handler

lib/child-writer/writer-proxy.js
```diff
@@ -163,16 +163,36 @@ permissions.isElevated().then((elevated) => {
         error: error.message,
         data: error.stack
       })
-      child && child.kill()
+      child && child.kill('SIGINT')
       reject(error)
     }
 
+    process.on('SIGINT', () => {
+      if (child) {
```
jviotti: I guess we can make a utility function that we can re-use in both signal handles (basically killing the child with the same signal if there is a child, or existing)

lib/gui/app.js
```diff
@@ -249,13 +252,17 @@ app.run(($window) => {
...
-        electron.remote.process.exit(EXIT_CODES.SUCCESS)
+        childWriter.abort(() => {
+          // This circumvents the 'beforeunload' event unlike
+          // electron.remote.app.quit() which does not.
+          debug('aborted')
+          electron.remote.app.exit(EXIT_CODES.SUCCESS)
+        })
       }
 
       analytics.logEvent('Close rejected while flashing')
```
jviotti: If `confirmed` is true, then chances are that this message will be shown (and sent to Mixpanel) before the application exits. Looks like we should wrap `.abort()` in a promise.

lib/child-writer/writer-proxy.js
jhermsmeier: Done.

lib/gui/app.js
jhermsmeier: A simple `else` branch does the trick just as well :)

jviotti:
Found a bug:

- Select an image
- Click Flash
- Cancel the elevation dialog
- Click flash again

I consistently get:

screen shot 2017-11-21 at 6 29 13 pm

jhermsmeier:
That's already been fixed, relevant addition: https://github.com/resin-io/etcher/pull/1873/files#diff-63e65246331b6569b0a65f87630dc334R77

jviotti:
Getting the following error in macOS when confirming the "application close" dialog after pressing the window manager red close icon:

screen shot 2017-11-21 at 6 48 56 pm

jviotti:
Same thing with Cmd+Q. I see the warning dialog, but confirming throws an error.

jviotti:
Force killing the application from the Activity Monitor on macOS makes the "close confirmation" dialog appear for a second or so, but the application then dies without allowing the user to confirm or cancel. Can we make the dialog not appear in this case?

Also, the child process remains alive:

screen shot 2017-11-21 at 6 52 59 pm

jviotti:
Killing the app using Ctrl-C doesn't seem to work on macOS anymore. This is what I see:

screen shot 2017-11-21 at 6 54 58 pm

jviotti:
It does work when I'm not flashing, but I still see the grey window for a second or so in that case.

jviotti:
This is what I see when sending `SIGKILL` to the child process while flashing:

screen shot 2017-11-21 at 6 57 45 pm
screen shot 2017-11-21 at 6 57 48 pm

jviotti:
If I kill the child process (using any method), and then attempt to close the application while the state is stuck in "Starting...", the "application close" confirmation button does nothing.

Then, in this case, using Cmd-Q bypasses the confirmation dialog, even though it works fine in other cases.

jviotti:
Oh, wait, Cmd-Q seems to randomly work, and randomly bypass the confirmation dialog. Smells like a racing condition.

lib/gui/etcher.js
jviotti: Fair enough. There is the grey window thing described below, but other than that it works :+1:

lib/child-writer/index.js
```diff
@@ -165,59 +177,222 @@ exports.write = (image, drive, options) => {
...
+      debug('abort:child:exit', code, signal)
+      this.emit('done', {
+        cancelled: true
+      })
+      this.terminateServer(callback)
+    })
 
-      return emitError(error)
+    ipc.server.broadcast('abort', {
+      signal: 'SIGINT'
```
jviotti: Can we use the constants from `os.constants.signals`?

jhermsmeier: subprocess.kill([signal]) expects signal to be a string, not a number (which os.constants.signals.SIGINT is), so that's not going to work I'm afraid

jviotti:
For the record: on Windows, when closing the app while flashing, I get to see the confirmation prompt, but the app exits automatically after one second or so.

jhermsmeier:
After the last update to this branch?

jviotti:
I killed the writer process on Windows by using the Task Manager, and the GUI app just froze at that specific point (keeps showing a fixed percentage, eta, and speed). It doesn't seem to realize the children is dead.

jviotti:
@jhermsmeier Not sure, I'm on the `release-1.2.0-test` branch. Is that updated?

jhermsmeier:
Hmm yeah, it should be updated – let me completely rebuild that branch, just to be safe

jviotti:
> After the last update to this branch?

Now that bug is fixed, but the flash process does not start on Windows. I always get stuck in Starting...

jviotti:
@jhermsmeier After the updates, closing the app from the Task Manager while there is a flash in process kills the app without any confirmation dialog and the app exits with code 1. I think DevTools is showing an error right before dying, but its too fast to be able to tell.

jviotti:
> I killed the writer process on Windows by using the Task Manager, and the GUI app just froze at that specific point (keeps showing a fixed percentage, eta, and speed). It doesn't seem to realize the children is dead.

Same problem running the latest stuff.

jviotti:
After the updates, killing the app with the Task Manager while there is a flash in process causes the confirmation dialog to be shown, but then the app is killed without requiring confirmation, and with exit code 1.

lib/child-writer/writer-proxy.js
```diff
@@ -163,16 +163,45 @@ permissions.isElevated().then((elevated) => {
...
+     * @returns {Function}
+     *
+     * @example
+     * process.on('SIGINT', signalHandler('SIGINT'))
+     */
+    const signalHandler = (signal) => {
+      return () => {
+        if (child) {
+          child.kill(signal)
+        } else {
```
jviotti: Not sure if I'm getting this right, but shouldn't we call `process.exit()` also if the child exists? Basically, it looks like this shouldn't be an `else`, but something that happens even if there was a child.

jhermsmeier: If the child exists, the child's exit event will cause this process to exit, after emitting the appropriate things

jhermsmeier:
Closing, as this bug and the underlying mechanics need an overhaul from the ground up, before we waste any more time on monkey-patching this around.

PR #1874, opened 2017-11-21
docs: add high level manual testing script

jviotti:

### DIFF
--- a/docs/MANUAL-TESTING.md
+++ b/docs/MANUAL-TESTING.md
@@ -1,0 +1,115 @@
+Manual Testing
+==============
+
+This document describes a high-level script of manual tests to check for. We
+should aim to replace items on this list with automated Spectron test cases.
+
+Image Selection
+---------------
+
+- [ ] Cancel image selection dialog
+- [ ] Select an unbootable image (without a partition table), and expect a
+  sensible warning
+- [ ] Attempt to select a ZIP archive with more than one image
+- [ ] Attempt to select a tar archive (with any compression method)
+- [ ] Change image selection
+- [ ] Select a Windows image, and expect a sensible warning
+
+Drive Selection
+---------------
+
+- [ ] Open the drive selection modal
+- [ ] Switch drive selection
+- [ ] Insert a single drive, and expect auto-selection
+- [ ] Insert more than one drive, and don't expect auto-selection
+- [ ] Insert a locked SD Card and expect a warning
+- [ ] Insert a too small drive and expect a warning
+- [ ] Put an image into a drive and attempt to flash the image to the drive
+  that contains it
+- [ ] Attempt to flash a compressed image (for which we can get the
+  uncompressed size) into a drive that is big enough to hold the compressed
+  image, but not big enough to hold the uncompressed version
+- [ ] Enable "Unsafe Mode" and attempt to select a system drive
+- [ ] Enable "Unsafe Mode", and if there is only one system drive (and no
+  removable ones), don't expect autoselection
+
+Image Support
+-------------
+
+Run the following tests with and without validation enabled:
+
+- [ ] Flash an uncompressed image
+- [ ] Flash a Bzip2 image
+- [ ] Flash a XZ image
+- [ ] Flash a ZIP image
+- [ ] Flash a GZ image
+- [ ] Flash a DMG image
+- [ ] Flash an image whose size is not a multiple of 512 bytes
+- [ ] Flash a compressed image whose size is not a multiple of 512 bytes
+- [ ] Flash an archive whose image size is not a multiple of 512 bytes
+- [ ] Flash an archive image containing a logo
+- [ ] Flash an archive image containing a blockmap file
+- [ ] Flash an archive image containing a manifest metadata file
+
+Flashing Process
+----------------
+
+- [ ] Unplug the drive during flash or validation
+- [ ] Click "Flash", cancel elevation dialog, and click "Flash" again
+- [ ] Start flashing an image, try to close Etcher, cancel the application
+  close warning dialog, and check that Etcher continues to flash the image
+
+### Child Writer
+
+- [ ] Kill the child writer process (i.e. with `SIGINT` or `SIGKILL`), and
+  check that the UI reacts appropriately
+- [ ] Close the application while flashing using the window manager close icon
+- [ ] Close the application while flashing using the OS keyboard shortcut
+- [ ] Close the application from the terminal using Ctrl-C while flashing
+- [ ] Force kill the application (using a process monitor tool, etc)
+
+In all these cases, the child writer process should not remain alive. Note that
+in some systems you need to open your process monitor tool of choice with extra
+permissions to see the elevated child writer process.
+
+GUI
+----
+
+- [ ] Close application from the terminal using Ctrl-C while the application is
+  idle
+- [ ] Click footer links that take you to an external website
+- [ ] Attempt to change image or drive selection while flashing
+- [ ] Go to the settings page while flashing and come back
+- [ ] Flash consecutive images without closing the application
+- [ ] Remove the selected drive right before clicking "Flash"
+- [ ] Minimize the application
+- [ ] Start the application given no internet connection
+
+Success Banner
+--------------
+
+- [ ] Click an external link on the success banner (with and without internet
+  connection)
+
+Elevation Prompt
+----------------
+
+- [ ] Flash an image as `root`/administrator
+- [ ] Reject elevation prompt
+- [ ] Put incorrect elevation prompt password
+- [ ] Unplug the drive during elevation
+
+Unmounting
+----------
+
+- [ ] Disable unmounting and flash an image
+- [ ] Flash an image with a file system that is readable by the host OS, and
+  check that is unmounted correctly
+
+Analytics
+---------
+
+- [ ] Disable analytics, open DevTools Network pane or a packet sniffer, and
+  check that no request is sent
+- [ ] **Disable analytics, refresh application from DevTools (using Cmd-R or
+  F5), and check that initial events are not sent to Mixpanel**


docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+- [ ] Open the drive selection modal
+
+Image Support
+-------------
+
+- [ ] Flash an uncompressed image
+- [ ] Flash a Bzip2 image
+- [ ] Flash a XZ image
+- [ ] Flash a ZIP image
```
jhermsmeier: Can we add gzipped & dmg images here as well?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,69 @@
...
+----------------
+
+- [ ] Close the application while flashing using the window manager close icon
+- [ ] Close the application while flashing using the OS keyboard shortcut
+- [ ] If applicable, close the application from the terminal using Ctrl-C while
+  ashing
+- [ ] Force kill the application (using a process monitor tool, etc)
+- [ ] Cancel the application close warning dialog
+
+In all these cases, the child writer process should not remain alive.
```
jhermsmeier: Might be worth adding a note that the child writer process is running as root / admin (as some tools like top/htop etc need to be run as root as well to see those processes)

docs/MANUAL-TESTING.md
jviotti: Yeah, will do.

jviotti: BTW, do you think its worth uploading some example files for this format to Google Drive, and linking them here?

docs/MANUAL-TESTING.md
jviotti: Is that really the case? I just checked and I can see processes belonging to other users.

jhermsmeier: That's odd I can't see the elevated writer process if I'm not sudo-ing

jviotti: Weird. I added a note just in case.

docs/MANUAL-TESTING.md
jhermsmeier: We have a bunch of them in `tests/image-stream/data` – might be worth mentioning those as a source

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+==============
+
+This document describes a high-level script of manual tests to check for. We
+should aim to replace items on this list with automated Spectron test cases.
+
+Image Selection
+---------------
+
+- [ ] Cancel image selection dialog
+- [ ] Select an unbootable image (without a partition table), and expect a
```
lurch: There's also the Windows-specific image-recognition test that Etcher does, which we should probably test

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+- [ ] Change image selection
+
+Drive Selection
+---------------
+
+- [ ] Open the drive selection modal
+- [ ] Remove all drives
+- [ ] Switch drive selection
+- [ ] Insert a single drive, and expect auto-selection
+- [ ] Insert one system drive, and don't expect auto-selection
```
lurch: We should also test that enabling and disabling unsafe mode in the settings screen affects the drive-selection list as expected

docs/MANUAL-TESTING.md
lurch: > We have a bunch of them in `tests/image-stream/data`

Which also contains some zip images containing metadata (icon etc.) and bmaps, which aren't mentioned in this list yet.

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+Image Support
+-------------
+
+- [ ] Flash an uncompressed image
+- [ ] Flash a Bzip2 image
+- [ ] Flash a XZ image
+- [ ] Flash a ZIP image
+- [ ] Flash a GZ image
+- [ ] Flash a DMG image
+- [ ] Flash an image whose size is not a multiple of 512 bytes
```
lurch: And based on the bugs found during the testing of the recent release, this should both test an uncompressed image which isn't a multiple of 512 bytes, and an archive which contains an image which isn't a multiple of 512 bytes.

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+  sensible warning
+- [ ] Attempt to select an unsupported image
+- [ ] Change image selection
+
+Drive Selection
+---------------
+
+- [ ] Open the drive selection modal
+- [ ] Remove all drives
+- [ ] Switch drive selection
```
lurch: Should we also check that the drive-selection list displays warnings for drives that are too small for the selected image?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+- [ ] Insert a single drive, and expect auto-selection
+- [ ] Insert one system drive, and don't expect auto-selection
+
+Image Support
+-------------
+
+- [ ] Flash an uncompressed image
```
lurch: Should we also run all these tests both with and without validation turned on?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+- [ ] **Kill the child writer process (i.e. with `SIGINT` or `SIGKILL`), and
+  check that the UI reacts appropriately**
+- [ ] **Close the application while flashing using the window manager close
+  icon**
+- [ ] **Close the application while flashing using the OS keyboard shortcut**
+- [ ] **Close the application from the terminal using Ctrl-C while flashing**
+- [ ] **Force kill the application (using a process monitor tool, etc)**
+- [ ] **Cancel the application close warning dialog**
+
+In all these cases, the child writer process should not remain alive. Note that
```
lurch: If we cancel the application close warning, shouldn't the child writer remain running?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+in some systems you need to open your process monitor tool of choice with extra
+permissions to see the elevated child writer process.
+
+GUI
+----
+
+- [ ] Close application from the terminal using Ctrl-C
+- [ ] Click footer links that take you to an external website
```
lurch: ...and also click the links that get displayed in the success-banner, in both online and offline mode

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+- [ ] **Close the application while flashing using the OS keyboard shortcut**
+- [ ] **Close the application from the terminal using Ctrl-C while flashing**
+- [ ] **Force kill the application (using a process monitor tool, etc)**
+- [ ] **Cancel the application close warning dialog**
+
+In all these cases, the child writer process should not remain alive. Note that
+in some systems you need to open your process monitor tool of choice with extra
+permissions to see the elevated child writer process.
+
+GUI
```
lurch: Check that Etcher starts up (and runs correctly) when the computer has no active internet connection

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+- [ ] Flash consecutive images without closing the application
+- [ ] Remove the selected drive right before clicking "Flash"
+
+Elevation Prompt
+----------------
+
+- [ ] Flash an image as `root`/administrator
+- [ ] Reject elevation prompt
+- [ ] Put incorrect elevation prompt password
+- [ ] Enable unsafe mode, and check drive selector dialog
```
lurch: Surely this doesn't come under the "Elevation Prompt" category?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
...
+- [ ] Disable unmounting and flash an image
+- [ ] Flash an image with a file system that is readable by the host OS, and
+  check that is unmounted correctly
+
+Analytics
+---------
+
+- [ ] Disable analytics, open DevTools Network pane or a packet sniffer, and
+  check that no request is sent
+- [ ] **Disable analytics, refresh application from DevTools, check that
```
lurch: What does "refresh application from DevTools" mean? Isn't it better to simply disable analytics, and then close and re-run Etcher?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,89 @@
+Manual Testing
+==============
+
+This document describes a high-level script of manual tests to check for. We
+should aim to replace items on this list with automated Spectron test cases.
```
lurch: ...and presumably these tests should also be run on all OS and Arch combinations, i.e. Windows 32-bit, Windows 64-bit, Linux 32-bit, Linux 64-bit, macOS 64-bit? Which probably means using virtual machines to properly check that the 32-bit builds don't accidentally contain any 64-bit code. (The VM images from http://modern.ie are very useful for this, which will also allow us to test Windows 7, Windows 8 and Windows 10. And I guess testing in a VM also allows us to test that Etcher runs on computers that don't already have NodeJS installed :wink: )

docs/MANUAL-TESTING.md
jviotti: Sounds great.

docs/MANUAL-TESTING.md
jviotti: Yep, the CI services builds should take care of this pretty well.

docs/MANUAL-TESTING.md
jviotti: Makes sense!

docs/MANUAL-TESTING.md
jviotti: :+1:

docs/MANUAL-TESTING.md
jviotti: Added!

docs/MANUAL-TESTING.md
jviotti: Oops, fixed

docs/MANUAL-TESTING.md
jviotti: :+1:

docs/MANUAL-TESTING.md
jviotti: :+1:

docs/MANUAL-TESTING.md
jviotti: Oops!

docs/MANUAL-TESTING.md
jviotti: It means hitting Cmd-R or F5 from DevTools. You can do that, but DevTools is not fast enough to log the initial requests. You can only see those by keeping DevTools opened while you perform a refresh or by using an external network sniffer.

docs/MANUAL-TESTING.md
jviotti: Clarified!

docs/MANUAL-TESTING.md
lurch: In fact, probably makes sense to check all the different warnings the drive-selector can display (system drive, drive too small, locked drive, etc.)

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,111 @@
...
+Run the following tests with and without validation enabled:
+
+- [ ] Flash an uncompressed image
+- [ ] Flash a Bzip2 image
+- [ ] Flash a XZ image
+- [ ] Flash a ZIP image
+- [ ] Flash a GZ image
+- [ ] Flash a DMG image
+- [ ] Flash an image whose size is not a multiple of 512 bytes
+- [ ] Flash a compressed image whose compressed size is not a multiple of 512
```
lurch: Heh, it'll be pretty hard to find an image where the compressed size _is_ a multiple of 512 bytes ;-)

jviotti: Haha, true. I'll remove this line.

docs/MANUAL-TESTING.md
jviotti: Added!

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,112 @@
...
+- [ ] Select a Windows image, and expect a sensible warning
+
+Drive Selection
+---------------
+
+- [ ] Open the drive selection modal
+- [ ] Remove all drives
+- [ ] Switch drive selection
+- [ ] Insert a single drive, and expect auto-selection
+- [ ] Insert one system drive, and don't expect auto-selection
```
lurch: Was the "Insert one system drive" here supposed to say "Insert another drive"?

jviotti: Clarified!

lurch: Hmm? Still says "Insert one system drive, and don't expect auto-selection" ??

jviotti: Is that really incorrect? We don't do auto-selection for system drives

lurch: But how exactly do you "insert one system drive"??

Perhaps I'm getting confused with what you mean, and there should also be a separate checkpoint which says "Insert two drives, and don't expect auto-selection"?

jviotti: Oops, sorry; I didn't push my local changes :)

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,112 @@
...
+- [ ] Insert a single drive, and expect auto-selection
+- [ ] Insert more than one drive, and don't expect auto-selection
+- [ ] Insert a locked SD Card and expect a warning
+- [ ] Insert a too small drive and expect a warning
+- [ ] Put an image into a removable drive and attempt to flash the image to the
+  drive that contains it
+- [ ] Attempt to flash a compressed image into a drive that is big enough to
+  hold the compressed image, but not big enough to hold the uncompressed
+  version
+- [ ] Enable "Unsafe Mode" and attempt to select a system drive
```
lurch: I guess we should also check that if unsafe mode is selected, and if only one system drive (and no removable drives) are present, we shouldn't expect auto-selection? (which obviously can't be tested on computers with more than one system drive)

jviotti: Added!

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- [ ] Flash an archive whose image size is not a multiple of 512 bytes
+- [ ] Flash an archive image containing a logo
+- [ ] Flash an archive image containing a blockmap file
+- [ ] Flash an archive image containing a manifest metadata file
+
+Flashing Process
+----------------
+
+- [ ] Disable validation and flash an image
+- [ ] Unplug the drive during flash or validation
```
Shou: I was going to suggest we could also add the unplug drive during the elevation dialog test, but it might not be so relevant once the `child-writer` re-architecture goes through

lurch: Might make sense to add it anyway - we can always remove it later.

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+This document describes a high-level script of manual tests to check for. We
+should aim to replace items on this list with automated Spectron test cases.
+
+Image Selection
+---------------
+
+- [ ] Cancel image selection dialog
+- [ ] Select an unbootable image (without a partition table), and expect a
+  sensible warning
+- [ ] Attempt to select an unsupported image
```
lurch: Given that Etcher already filters the file-selection dialog by file-extension, it might be worth clarifying what "unsupported image" you should try selecting.
Two things I can think of off the top of my head are zip files containing more than one file inside them, and .tar.bz2 files. (perhaps there's others - should we try selecting all of them?)

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- [ ] Change image selection
+- [ ] Select a Windows image, and expect a sensible warning
+
+Drive Selection
+---------------
+
+- [ ] Open the drive selection modal
+- [ ] Remove all drives
```
lurch: What does "Remove all drives" mean here?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- [ ] Switch drive selection
+- [ ] Insert a single drive, and expect auto-selection
+- [ ] Insert more than one drive, and don't expect auto-selection
+- [ ] Insert a locked SD Card and expect a warning
+- [ ] Insert a too small drive and expect a warning
+- [ ] Put an image into a removable drive and attempt to flash the image to the
```
lurch: I guess 'removable' here isn't strictly necessary?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+  drive that contains it
+- [ ] Attempt to flash a compressed image into a drive that is big enough to
+  hold the compressed image, but not big enough to hold the uncompressed
```
lurch: The behaviour here will be different for those compression formats where we _can_ get the final uncompressed size, and those compression formats where we can't get the final uncompressed size. I'm assuming this is about testing the latter, given that Etcher doesn't allow the former?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- [ ] Flash a compressed image whose size is not a multiple of 512 bytes
+- [ ] Flash an archive whose image size is not a multiple of 512 bytes
+- [ ] Flash an archive image containing a logo
+- [ ] Flash an archive image containing a blockmap file
+- [ ] Flash an archive image containing a manifest metadata file
+
+Flashing Process
+----------------
+
+- [ ] Disable validation and flash an image
```
lurch: I guess this is already covered by the "...with and without validation... Flash an uncompressed image" in the previous section ;)

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- [ ] Unplug the drive during flash or validation
+- [ ] Click "Flash", cancel elevation dialog, and click "Flash" again
+- [ ] **Cancel the application close warning dialog**
```
lurch: This probably needs to be clarified to something like "Start flashing an image, try to close Etcher, Cancel the application close warning dialog, check that Etcher continues to flash the image" ?

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- [ ] **Force kill the application (using a process monitor tool, etc)**
+
+In all these cases, the child writer process should not remain alive. Note that
+in some systems you need to open your process monitor tool of choice with extra
+permissions to see the elevated child writer process.
+
+GUI
+----
+
+- [ ] Close application from the terminal using Ctrl-C
```
lurch: Perhaps this should clarify "...while Etcher is idle" to avoid overlap with the previous section?

docs/MANUAL-TESTING.md
jviotti: Done

docs/MANUAL-TESTING.md
jviotti: Done!

docs/MANUAL-TESTING.md
jviotti: We can probably delete this one.

docs/MANUAL-TESTING.md
jviotti: True, fixed

docs/MANUAL-TESTING.md
jviotti: Done

docs/MANUAL-TESTING.md
jviotti: Done

docs/MANUAL-TESTING.md
jviotti: Done

docs/MANUAL-TESTING.md
jviotti: Done

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,115 @@
...
+
+- [ ] Open the drive selection modal
+- [ ] Switch drive selection
+- [ ] Insert a single drive, and expect auto-selection
+- [ ] Insert more than one drive, and don't expect auto-selection
+- [ ] Insert a locked SD Card and expect a warning
+- [ ] Insert a too small drive and expect a warning
+- [ ] Put an image into a drive and attempt to flash the image to the drive
+  that contains it
+- [ ] Attempt to flash a compressed image (for which we can get the
```
lurch: This should be "can't get" ;-)

docs/MANUAL-TESTING.md
```diff
@@ -0,0 +1,115 @@
...
+  uncompressed size) into a drive that is big enough to hold the compressed
```
lurch: I guess you could clarify "uncompressed size e.g. bz2)"

lurch:
@jhermsmeier Hmmm, I was expecting @jviotti to fix my last two nits before merging this :-/  Should I not have approved in that case?

jhermsmeier:
Oh sorry, I missed those :/ Yeah, better to request changes in that case

Issue #1523, opened 2017-06-16
Validate write on success reads data from cache instead of SD card


net147:
- **Etcher version:**
1.0.0, git revision 103a048e4d4b670ca16d473ed6d6074b17519f94
- **Operating system and architecture:**
Arch Linux x86_64
- **Image flashed:**
150995968 byte image file
- **Do you see any meaningful error information in the DevTools?**
It does not show any "Progress (check)" messages in Console like it does on Windows

Validate write finishes immediately without actually reading from the SD card.

net147:
If I run the following command in another terminal to keep flushing all the caches, I can Etcher reading from the SD card:
while true; do echo 3 > /proc/sys/vm/drop_caches; done

lurch:
If true, this should be regarded as a high-priority bug, as I guess it means validation isn't doing what we actually think it is :-/
Perhaps there's some extra options we can pass to the 'open' call to get it to avoid cached reads?

lurch:
Hmmm, just done some searching and found https://stackoverflow.com/questions/15266115/read-file-without-disk-caching-in-linux/15266168#15266168 and also found https://nodejs.org/api/fs.html#fs_fs_open_path_flags_mode_callback which says
```
'rs+' - Open file for reading and writing in synchronous mode. Instructs the operating system to bypass the local file system cache.
```
And `rs+` is indeed what we're using: https://github.com/resin-io/etcher/blob/master/lib/cli/writer.js#L76

:confused:

lurch:
...although https://github.com/nodejs/node/blob/master/test/parallel/test-fs-open-flags.js#L42 implies that `rs+` only sets `O_RDWR | O_SYNC` and we'll instead need to pass the `flags` to `open()` as a number instead of a string, if we want to use `O_DIRECT` https://nodejs.org/api/fs.html#fs_file_open_constants
And see also https://linux.die.net/man/2/open

But it looks like `O_DIRECT` is specific to Linux, and the equivalent on Windows is `FILE_FLAG_NO_BUFFERING` https://msdn.microsoft.com/en-us/library/windows/desktop/cc644950(v=vs.85).aspx
And these links are useful too https://stackoverflow.com/questions/924900/portability-of-open-o-direct-in-c and https://stackoverflow.com/questions/2299402/how-does-one-do-raw-io-on-mac-os-x-ie-equivalent-to-linuxs-o-direct-flag

[...]

PR #1875, opened 2017-11-22
fix(writer): Set flags manually to avoid verifying cache

jhermsmeier:
We change from using `rs+` to a composition of read/write,
exlusive, sync & direct i/o flags, in order to avoid reading
stale data from the cache during verification.

Tested on:

- [x] Linux
- [x] Windows
- [x] Mac OS

Connects To: https://github.com/resin-io/etcher/issues/1523
Change-Type: patch
Changelog-Entry: Fix verification step reading from the cache

### DIFF
--- a/lib/cli/writer.js
+++ b/lib/cli/writer.js
@@ -75,7 +75,15 @@ exports.writeImage = (imagePath, drive, options, onProgress) => {
   }).then(() => {
     return diskpart.clean(drive.device)
   }).then(() => {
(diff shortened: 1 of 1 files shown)


jhermsmeier:
NOTE: I am not sure whether this fixes it in all cases, but it does yield steady results for what I have available here. Not setting these causes reads from cache during verification on Linux 100% of the time for me.

PR #1876, opened 2017-11-23
doc(MAINTAINERS): Elaborate on the process of releasing

jhermsmeier:
This updates the maintainers' doc with a little more detail
regarding the release process, to avoid it staying tribal knowledge.

### DIFF
--- a/docs/MAINTAINERS.md
+++ b/docs/MAINTAINERS.md
@@ -1,52 +1,187 @@
 Maintaining Etcher
 ==================
 
-This document is meant to serve as a guide for maintainers to perform common
-tasks.
+This document is meant to serve as a guide for maintainers to perform common tasks.
 
-Preparing a new version
------------------------
+Releasing
+---------
 
+### Release Types
+
+- **snapshot** (default): A continues snapshot of current master, made by the CI services
+- **production**: Full releases
+
+### Flight Plan
+
+#### Preparation
+
+- [Prepare the new version](#preparing-a-new-version)
+- [Generate build artefacts](#generating-binaries) (binaries, archives, etc.)
+- [Draft a release on GitHub](https://github.com/resin-io/etcher/releases)
+    - Upload build artefacts to GitHub release draft
+
+#### Testing
+
+- Test the prepared release and build artefacts properly on **all supported operating systems** to prevent regressions that went uncaught by the CI tests (see [MANUAL-TESTING.md](MANUAL-TESTING.md))
+- If regressions or other issues arise, create issues on the repository for each one, and decide whether to fix them in this release (meaning repeating the process up until this point), or to follow up with a patch release
+
+#### Publishing
+
+- [Publish release draft on GitHub](https://github.com/resin-io/etcher/releases)
+- [Post release note to forums](https://forums.resin.io/c/etcher)
+- [Update the website](https://github.com/resin-io/etcher-homepage)
+- Wait 2-3 hours for analytics (Sentry, Mixpanel) to trickle in and check for elevated error rates, or regressions
+- If regressions arise; pull the release, and release a patched version, else:
+- [Upload deb & rpm packages to Bintray](#uploading-packages-to-bintray)
+- [Upload build artefacts to Amazon S3](#uploading-binaries-to-amazon-s3)
+- Post changelog with `#release-notes` tag on Flowdock
+- If this release packs noteworthy major changes:
+  - Write a blog post about it, and / or
+  - Write about it to the Etcher mailing list
+
+### Preparing a New Version
+
+- Create & hop onto a new release branch, i.e. `release-1.0.0`
 - Bump the version number in the `package.json`'s `version` property.
-
-- Bump the version number in the `npm-shrinkwrap.json`'s `version` property.
-
-- Add a new entry to `CHANGELOG.md` by running `make changelog`.
-
+- Bump the version number in the `npm-shrinkwrap.json`'s `version` property
+- Add a new entry to `CHANGELOG.md` by running `make changelog`
+- Manually revise the `CHANGELOG.md` versionist output
 - Update `screenshot.png` so it displays the latest version in the bottom
-right corner.
-
+right corner
 - Revise the `updates.semverRange` version in `package.json`
+- Commit the changes with the version number as the commit title, including the `v` prefix, to `master`. For example:
 
-- Commit the changes with the version number as the commit title, including the
-`v` prefix, to `master`. For example:
+**NOTE:** The version **MUST** be prefixed with a "v"
 
-```sh
+```bash
 git commit -m "v1.0.0" # not 1.0.0
 ```
 
-- Create an annotated tag for the new version. The commit title should equal
-the annotated tag name. For example:
+- Create an annotated tag for the new version. The commit title should equal the annotated tag name. For example:
 
-```sh
+```bash
 git tag -a v1.0.0 -m "v1.0.0"
 ```
 
 - Push the commit and the annotated tag.
 
-```sh
+```bash
 git push
 git push --tags
 ```
 
-Upgrading Electron
-------------------
+- Open a pull request against `master` titled "Release v1.0.0"
 
-- Upgrade the `electron` dependency version in `package.json` to an *exact
-  version* (no `~`, `^`, etc).
+### Generating binaries
 
-Dealing with a problematic release
-----------------------------------
+**Environment**
+
+Make sure to set the analytics tokens when generating production release binaries:
+
+```bash
+export ANALYTICS_SENTRY_TOKEN="xxxxxx"
+export ANALYTICS_MIXPANEL_TOKEN="xxxxxx"
+```
+
+#### Linux
+
+##### Clean dist folder
+
+**NOTE:** Make sure to adjust the path as necessary (here the Etcher repository has been cloned to `/home/$USER/code/etcher`)
+
+```bash
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make distclean"
+```
+
+##### Generating artefacts
+
+```bash
+# x64
+
+# Build Debian packages
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-debian"
+# Build RPM packages
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-redhat"
+# Build AppImages
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-appimage"
+# Build CLI
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production cli-installer-tar-gz"
+
+# x86
+
+# Build Debian packages
+./scripts/build/docker/run-command.sh -r x86 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-debian"
+# Build RPM packages
+./scripts/build/docker/run-command.sh -r x86 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-redhat"
+# Build AppImages
+./scripts/build/docker/run-command.sh -r x86 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-appimage"
+# Build CLI
+./scripts/build/docker/run-command.sh -r x86 -s . -c "make electron-develop && make RELEASE_TYPE=production cli-installer-tar-gz"
+```
+
+#### Mac OS
+
+**ATTENTION:** For production releases you'll need the code-signing key,
+and set `CSC_NAME` to generate signed binaries on Mac OS.
+
+**NOTE:** The CLI is not code-signed for either at this time.
+
+```bash
+make electron-develop
+
+# Build the CLI
+make RELEASE_TYPE=production cli-installer-tar-gz
+# Build the zip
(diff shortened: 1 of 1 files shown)


docs/MAINTAINERS.md
```diff
@@ -1,52 +1,138 @@
...
+```
+
+#### Linux
+
+##### Clean dist folder
+
+**NOTE:** Make sure to adjust the path as necessary (here the Etcher repository has been cloned to `/home/$USER/code/etcher`)
+
+```bash
+docker run -v "/home/$USER:/home/$USER" "busybox rm -rf /home/$USER/code/etcher/release /home/$USER/code/etcher/node_modules /home/$USER/code/etcher/build /home/$USER/code/etcher/dist"
```
lurch: Or `make distclean` ;-)  (and there's no need to use docker for this step either)

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,138 @@
...
+# x64
+
+# Build Debian packages
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-debian"
+# Build RPM packages
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-redhat"
+# Build AppImages
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production electron-installer-appimage"
+# Build CLI
+./scripts/build/docker/run-command.sh -r x64 -s . -c "make electron-develop && make RELEASE_TYPE=production cli-installer-tar-gz"
```
lurch: Once it's been run once, I don't think there's any need to run `make electron-develop` on each of the subsequent steps?

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,138 @@
...
+
+# x86
```
lurch: ...but of course you _do_ need to run `make electron-develop` once when switching to x86, as all the native modules need to be built for x86 instead of x64.

docs/MAINTAINERS.md
jhermsmeier: There is – when doing the builds on misc1, everything under `dist` will be owned by root, and `make distclean` will not work.

docs/MAINTAINERS.md
jhermsmeier: Each of these are individual containers...

lurch: Nope ;-)  There's only one container that's used for all the builds, and it simply maps `.` to `/etcher` inside the container https://github.com/resin-io/etcher/blob/master/scripts/build/docker/run-command.sh#L90
(which is why the files in `/home/$USER/code/etcher/dist` end up being owned by root)

docs/MAINTAINERS.md
lurch: Fair enough. You could still replace this with `./scripts/build/docker/run-command.sh -r x64 -s . -c "make distclean"` though.

docs/MAINTAINERS.md
lurch: I think the "proper Docker way of doing things" would be to copy all the sourcecode into the container (so that the local `etcher` directory would effectively be read-only), in which case you _would_ need to run `make electron-develop` each time, but for convenience and/or speed, the whole `etcher` directory is simply mounted as read-write inside the Docker container.

lurch: ...which probably also means that you should do `./scripts/build/docker/run-command.sh -r x86 -s . -c "make distclean"` between doing the 64-bit builds and the 32-bit builds?

jhermsmeier:
@jviotti do we have instructions / commands at the ready for building arm packages that I could include here?

jviotti:
See https://github.com/resin-io/etcher/pull/1482. I haven't build ARM for a while, but setting the correct TARGET_ARCH should do it.

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,138 @@
...
+#### Testing
+
+- Test the prepared release and build artefacts properly on **all supported operating systems** to prevent regressions that went uncaught by the CI tests
+- If regressions or other issues arise, create issues on the repository for each one, and decide whether to fix them in this release (meaning repeating the process up until this point), or to follow up with a patch release
 
+#### Publishing
+
+- Publish release draft on GitHub
+- Upload build artefacts to Bintray & Amazon S3
+- Post release note to forums
```
jviotti: We should add the `#release-notes` Flowdock hash-tag thing here.

lurch:
@jhermsmeier Could you take another look at the two `run-command.sh` comments I left 17 days ago?

Also, it's outside the scope of this PR, but just including the link here for future reference (so we don't lose it) - I started working on some shell scripts to automate parts of this https://gist.github.com/lurch/88879362a9a106cf5516a10ba5dbdbcd which it might be worth revisiting before we do the next release (or perhaps some of this would be useful for @jviotti 's Concourse stuff?)

docs/MAINTAINERS.md
jhermsmeier: Nope, that would delete the build artefacts in the process (and that's mildly unpractical when doing a release)

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
-Preparing a new version
------------------------
+Releasing
+---------
 
-- Bump the version number in the `package.json`'s `version` property.
+### Release Types
 
-- Bump the version number in the `npm-shrinkwrap.json`'s `version` property.
+- **snapshot** (default): Describe me!
```
jviotti: Should this be completed?

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
+#### Preparation
+
+- [Prepare the new version](#preparing-a-new-version)
+- [Generate build artefacts](#generating-binaries) (binaries, archives, etc.)
+- [Draft a release on GitHub](https://github.com/resin-io/etcher/releases)
+    - Upload build artefacts to GitHub release draft
+
+#### Testing
+
+- Test the prepared release and build artefacts properly on **all supported operating systems** to prevent regressions that went uncaught by the CI tests
```
jviotti: Maybe worth linking to my manual tests doc?

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
+
+#### Publishing
 
+- [Publish release draft on GitHub](https://github.com/resin-io/etcher/releases)
+- [Update the website](https://github.com/resin-io/etcher-homepage)
+- [Post release note to forums](https://forums.resin.io/c/etcher)
+- [Upload deb & rpm packages to Bintray](#uploading-packages-to-bintray)
+- [Upload build artefacts to Amazon S3](#uploading-binaries-to-amazon-s3)
+- Post changelog with `#release-notes` tag on Flowdock
+- If this release packs noteworthy major changes, maybe write a blog post about it as well
```
jviotti: Maybe worth mentioning the Etcher mailing list?

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
+
+### Preparing a New Version
+
+- Create & hop onto a new release branch, i.e. `release-1.0.0`
+- Bump the version number in the `package.json`'s `version` property.
+- Bump the version number in the `npm-shrinkwrap.json`'s `version` property
+- Add a new entry to `CHANGELOG.md` by running `make changelog`
```
jviotti: Should we add a step here to manually revise Versionist output, just in case?

docs/MAINTAINERS.md
jviotti: Yeah, I think @lurch's last comment should work.

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
+# Build CLI
+./scripts/build/docker/run-command.sh -r x86 -s . -c "make electron-develop && make RELEASE_TYPE=production cli-installer-tar-gz"
+```
 
-- Upgrade the `electron` dependency version in `package.json` to an *exact
-  version* (no `~`, `^`, etc).
+#### Mac OS
+
+**ATTENTION:** For production releases you'll need the code-signing key,
+and set `CSC_NAME`, `CSC_LINK`, and `CSC_KEY_PASSWORD` to generate signed binaries on Mac OS.
```
jviotti: We only need `CSC_NAME` for macOS

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
+# Build the zip
+make RELEASE_TYPE=production electron-installer-app-zip
+# Build the dmg
+make RELEASE_TYPE=production electron-installer-dmg
+```
+
+#### Windows
+
+**ATTENTION:** For production releases you'll need the code-signing key,
+and set `CSC_NAME`, `CSC_LINK`, and `CSC_KEY_PASSWORD` to generate signed binaries on Windows.
```
jviotti: We don't need `CSC_NAME` here, only the other two.

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
+
+**NOTE:** The CLI is not code-signed for either at this time.
+
+```bash
+make electron-develop
+
+# Build the CLI
+make RELEASE_TYPE=production cli-installer-zip
```
jviotti: You should clarify that `TARGET_ARCH` is needed to build x86 packages.

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,180 @@
...
+```
+
+### Uploading binaries to Amazon S3
+
+```bash
+export S3_KEY="..."
+```
+
+```bash
+./scripts/publish/aws-s3.sh -b "resin-production-downloads" -v "1.2.1" -p "etcher" -k $(date +"%Y-%m-%d") -f "dist/<filename>"
```
jviotti: The `-k` option is only used for snapshot builds

docs/MAINTAINERS.md
lurch: Yep, and this is already part of https://gist.github.com/lurch/88879362a9a106cf5516a10ba5dbdbcd#file-build_release_artefacts-sh-L60 ;-)

docs/MAINTAINERS.md
lurch: Yeah, my mistake - you only need to run `./scripts/build/docker/run-command.sh -r x86 -s . -c "make TARGET_ARCH=x86 RELEASE_TYPE=production electron-develop"` between doing the 64-bit builds and the 32-bit builds.
https://gist.github.com/lurch/88879362a9a106cf5516a10ba5dbdbcd#file-build_release_artefacts-sh-L99

jviotti:
Some things we should add from recent conversations about 1.3.0:

- First release to GitHub Releases and update the website
- Wait for a couple of hours and check Mixpanel to ensure everything is OK
- If everything is fine, update S3 and Bintray to trigger update notification dialogs
- Wait for a couple of hours and check Mixpanel to ensure everything is OK

Also lets put instructions on how to put down a release. Maybe we can collect a set of `aws` and `curl` commands to take down a GitHub Release, Bintray version, and a set of S3 entries?

alexandrosm:
maybe the pull down instructions can be automated into e.g. a hubot script
so that a non-technical user (or someone on the road) can run it?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lurch:
> so that a non-technical user (or someone on the road) can run it?

Or just add a note to the release-instructions to never release on a Friday? ;-)

docs/MAINTAINERS.md
lurch: It's just occurred to me that it might also be worth sending out a tweet announcing the new Etcher version on the https://twitter.com/resin_io Twitter account?

docs/MAINTAINERS.md
jhermsmeier: Yup, done

docs/MAINTAINERS.md
jhermsmeier: Added

docs/MAINTAINERS.md
jhermsmeier: Added

docs/MAINTAINERS.md
jhermsmeier: Fixed

docs/MAINTAINERS.md
jhermsmeier: Fixed

docs/MAINTAINERS.md
jhermsmeier: Done

docs/MAINTAINERS.md
jhermsmeier: Aye, removed!

docs/MAINTAINERS.md
jhermsmeier: Yup, added the mailing list

docs/MAINTAINERS.md
```diff
@@ -1,52 +1,187 @@
...
+```
 
-- Upgrade the `electron` dependency version in `package.json` to an *exact
-  version* (no `~`, `^`, etc).
+#### Mac OS
+
+**ATTENTION:** For production releases you'll need the code-signing key,
+and set `CSC_NAME` to generate signed binaries on Mac OS.
+
+**NOTE:** The CLI is not code-signed for either at this time.
```
jviotti: What does "either" mean here?

jhermsmeier: Oh, looks like there's a "platform" missing after that "either"

Issue #1772, opened 2017-10-14
Mixpanel tracking fires with disabled analytics


jhermsmeier:
- **Etcher version:** v1.1.2
- **Operating system and architecture:** any
- **Image flashed:** any
- **Do you see any meaningful error information in the DevTools?**

![screen shot 2017-10-14 at 18 02 59](https://user-images.githubusercontent.com/244907/31577170-26b162c6-b10a-11e7-97c5-f1e244af70f0.png)

As can be seen here, two events are tracked and sent to Mixpanel at application startup, even though the analytics setting is disabled.

Connects To: #1770

ValdikSS:
Not fixed in 1.2.0.

jviotti:
Hi @ValdikSS ,

Thanks for reviving this thread. I thought we sent a fix for this ticket already, and that we simply forgot to close the issue with the corresponding PR. @jhermsmeier do you remember where I can find that PR?

@ValdikSS I gave this another go using the v1.2.0 on my Mac but couldn't reproduce the issue anymore. I disabled analytics, opened Etcher, and no request was sent to `api.mixpanel.com`:

<img width="1082" alt="screen shot 2017-11-26 at 12 17 56 pm" src="https://user-images.githubusercontent.com/2192773/33241982-2b68db4e-d2a4-11e7-995c-cbe17f51dcfb.png">

How are you analyzing this? Could it be that the DevTools Network monitor is not fast enough to log certain requests you found somewhere else? Let me know so we can track this down!

jviotti:
Thanks for the clarification @ValdikSS. I was able to get to the same conclusion by using WireShark. I sent an initial PR (https://github.com/resin-io/etcher/pull/1879/files) that ensures that local settings are loaded before bootstrapping the application, but looks like the Mixpanel library still sends some initial events during its own startup that we have to find a way to shutdown. This will probably need to be done in Corvus.

jviotti:
I fixed one last mysterious `/decide/` call from the Mixpanel library here: https://github.com/resin-io-modules/resin-corvus/pull/35

jviotti:
I fixed the last remaining edge case: https://github.com/resin-io-modules/resin-corvus/pull/36.

Thanks for the heads up @ValdikSS! I definitely wouldn't have noticed that this was still the case without a packet sniffer or by refreshing the application through DevTools. 

This fix will land in the next release (probably v1.2.1) very soon, and apologies for not allocating more time to properly fix this!

[...]

PR #1879, opened 2017-11-26
fix(GUI): ensure settings are loaded before bootstrapping application

jviotti:
We load localStorage settings into the Redux store in an asynchronous
way. This means that user settings might not be loaded by the time the
application starts, resulting in Mixpanel sending a few tracking events
before Etcher realises that the user opted out from anonoymous analytics
and error reporting.

In order to fix that, we remove `ng-app` and we manually bootstrap the
Angular.js application *after* the local settings are loaded.

Change-Type: patch
Changelog-Entry: Don't send initial Mixpanel events before "Anonymous Tracking" settings are loaded.
Fixes: https://github.com/resin-io/etcher/issues/1772

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -102,77 +102,75 @@ app.run(() => {
     version: currentVersion
   })
 
-  settings.load().then(() => {
-    const shouldCheckForUpdates = updateNotifier.shouldCheckForUpdates({
-      currentVersion,
-      lastSleptUpdateNotifier: settings.get('lastSleptUpdateNotifier'),
-      lastSleptUpdateNotifierVersion: settings.get('lastSleptUpdateNotifierVersion')
+  const shouldCheckForUpdates = updateNotifier.shouldCheckForUpdates({
+    currentVersion,
+    lastSleptUpdateNotifier: settings.get('lastSleptUpdateNotifier'),
+    lastSleptUpdateNotifierVersion: settings.get('lastSleptUpdateNotifierVersion')
+  })
+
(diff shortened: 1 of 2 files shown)


lib/gui/app.js
```diff
@@ -102,77 +102,75 @@ app.run(() => {
...
+  const updatesEnabled = settings.get('updatesEnabled')
+
+  if (!shouldCheckForUpdates || !updatesEnabled) {
+    analytics.logEvent('Not checking for updates', {
+      shouldCheckForUpdates,
+      updatesEnabled,
+      stable: isStableRelease
     })
 
-    const isStableRelease = release.isStableRelease(currentVersion)
```
lurch: Sometimes the diffs that Github produces are so painful to decipher :-(

lib/gui/app.js
```diff
@@ -102,77 +102,75 @@ app.run(() => {
     version: currentVersion
   })
 
-  settings.load().then(() => {
```
lurch: Just out of curiosity, would moving the `analytics.logEvent('Application start'` line below the `settings.load().then(` line also have fixed the problem? Or does it go deeper than that?

jviotti: Yeah, that would have fixed the issue. This PR changes the whole Angular.js bootstrapping method so we can be sure everything will be ready before we execute anything, protecting us from forgetting to move analytic events call below a certain other instruction, etc.

lurch:
This _partially_ fixes #1772 - #1902 was the other half of the fix.

Issue #1870, opened 2017-11-20
Add integration tests


jhermsmeier:
Possibly utilizing spectron

lurch:
I made a start on this a while ago, (IIRC it worked with both AppImages and .deb versions of Etcher) but I got stuck trying to figure out how to navigate the DOM
https://gist.github.com/lurch/6dac70712beef0a071cb02f4fd0bfa67

When I came back to this recently (for running smoke-tests on different Linux distros using Vagrant), I found that I had to manually downgrade the version of `webdriverio` for `spectron` to work correctly; but perhaps that's been fixed by now?

jviotti:
@lurch Great example! I'll send a quick PR setting up Spectron on the repo, so we can then start adding more and more tests there.

PR #1884, opened 2017-11-28
chore: setup a Spectron integration test suite

jviotti:
- Add a `make test-spectron` target

- Install `spectron` and `mocha` (since we don't need to run the tests
  inside an Electron instance like in the case of `electron-mocha`)

- Add some example tests

Fixes: https://github.com/resin-io/etcher/issues/1870
Change-Type: patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ endif
 
 BUILD_TEMPORARY_DIRECTORY = $(BUILD_DIRECTORY)/.tmp
 
+# See https://github.com/electron/spectron/issues/127
+ETCHER_SPECTRON_ENTRYPOINT ?= $(shell node -e 'console.log(require("electron"))')
+
 # ---------------------------------------------------------------------
 # Operating system and architecture detection
 # ---------------------------------------------------------------------
@@ -376,6 +379,7 @@ TARGETS = \
 	lint-cpp \
 	lint-html \
 	lint-spell \
+	test-spectron \
 	test-gui \
 	test-sdk \
 	test \
@@ -543,18 +547,23 @@ lint-spell:
 
 lint: lint-js lint-sass lint-cpp lint-html lint-spell
 
-ELECTRON_MOCHA_OPTIONS=--recursive --reporter spec
+MOCHA_OPTIONS=--recursive --reporter spec
+
+test-spectron:
+	$(NPX) env \
+		ETCHER_SPECTRON_ENTRYPOINT=$(ETCHER_SPECTRON_ENTRYPOINT) \
+		mocha $(MOCHA_OPTIONS) tests/spectron
 
 test-gui:
-	$(NPX) electron-mocha $(ELECTRON_MOCHA_OPTIONS) --renderer tests/gui
+	$(NPX) electron-mocha $(MOCHA_OPTIONS) --renderer tests/gui
 
 test-sdk:
-	$(NPX) electron-mocha $(ELECTRON_MOCHA_OPTIONS) \
+	$(NPX) electron-mocha $(MOCHA_OPTIONS) \
 		tests/shared \
 		tests/child-writer \
 		tests/image-stream
 
-test: test-gui test-sdk
+test: test-gui test-sdk test-spectron
 
 help:
 	@echo "Available targets: $(TARGETS)"
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -241,6 +241,18 @@
       "from": "arch@>=2.1.0 <3.0.0",
       "resolved": "https://registry.npmjs.org/arch/-/arch-2.1.0.tgz"
     },
+    "archiver": {
+      "version": "2.1.0",
+      "from": "archiver@>=2.1.0 <2.2.0",
+      "resolved": "https://registry.npmjs.org/archiver/-/archiver-2.1.0.tgz",
+      "dev": true
+    },
+    "archiver-utils": {
+      "version": "1.3.0",
+      "from": "archiver-utils@>=1.3.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/archiver-utils/-/archiver-utils-1.3.0.tgz",
+      "dev": true
+    },
     "are-we-there-yet": {
       "version": "1.1.2",
       "from": "are-we-there-yet@>=1.1.2 <1.2.0",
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -541,18 +541,21 @@ lint-spell:
 
 lint: lint-js lint-sass lint-cpp lint-html lint-spell
 
-ELECTRON_MOCHA_OPTIONS=--recursive --reporter spec
+MOCHA_OPTIONS=--recursive --reporter spec
```
lurch: Should these be kept separate? Or is it guaranteed that all `mocha` options will always be the same as the `electron-mocha` options?

Makefile
```diff
@@ -541,18 +541,21 @@ lint-spell:
...
+
+test-spectron:
```
lurch: This should also be added to the `TARGETS` list on line 370

npm-shrinkwrap.json
```diff
@@ -4710,7 +4860,7 @@
     },
     "mocha": {
       "version": "3.2.0",
-      "from": "mocha@>=3.0.0 <4.0.0",
+      "from": "https://registry.npmjs.org/mocha/-/mocha-3.2.0.tgz",
```
lurch: Huh?!? Why has this `from` specifier changed from a version-range to a URL?

tests/spectron/runner.spec.js
```diff
@@ -0,0 +1,44 @@
...
+const Bluebird = require('bluebird')
+const spectron = require('spectron')
+const path = require('path')
+
+describe('Spectron', function () {
+  this.timeout(15000)
+
+  beforeEach(function () {
+    this.app = new spectron.Application({
+      path: path.join(__dirname, '..', '..', 'node_modules', '.bin', 'electron'),
```
lurch: Would there be any way to parameterise / override this path, so that these same Spectron tests could also be used to test the AppImage etc?  (and would this fit in better with the Makefile if it used `npx electron`, or even `npm start`?)

Makefile
jviotti: Its guaranteed. `electron-mocha` aims to remain 100% CLI compatible with `mocha`.

Makefile
jviotti: Added!

npm-shrinkwrap.json
jviotti: Weird. Fixed

tests/spectron/runner.spec.js
jviotti: I don't think we will ever be able to use Spectron to test an AppImage. The AppImage runtime will run the predefined binary as soon as we execute it, where with Spectron, we would need to run the Electron binary that is embedded inside the AppImage directly, bypassing the runtime.

lurch: I've just double-checked, and the test spectron script I posted in #1870 definitely works with both the .deb installed version of Etcher and also the AppImage version of Etcher.

jviotti: Oh, wow. I wonder how that works. Let me see how can easily parameterise this.

jviotti: Done!

Makefile
```diff
@@ -541,18 +542,23 @@ lint-spell:
...
+	$(NPX) env \
+		ETCHER_SPECTRON_ENTRYPOINT=$(BUILD_DIRECTORY_PARENT)/node_modules/.bin/electron \
```
lurch: Could we have
```
ETCHER_SPECTRON_ENTRYPOINT ?= $(BUILD_DIRECTORY_PARENT)/node_modules/.bin/electron
```
up near the top of the Makefile, and then change this line to:
```
ETCHER_SPECTRON_ENTRYPOINT=$(ETCHER_SPECTRON_ENTRYPOINT) \
```
?

That should allow us to then manually run e.g.
```
make ETCHER_SPECTRON_ENTRYPOINT=$(which etcher-electron) test-spectron
```
on the  command line, to run the spectron tests against the .deb installed version of Etcher :-)

jviotti: Sounds good, done!

jviotti:
Okay, I think it should be all green now. I made some small changes to the Appveyor configuration file, and to the Spectron's entry point path.

tests/spectron/runner.spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+if (!entrypoint) {
+  console.error('You need to properly configure ETCHER_SPECTRON_ENTRYPOINT')
+  process.exit(EXIT_CODES.GENERAL_ERROR)
+}
+
+describe('Spectron', function () {
+  // Mainly for CI jobs, Appveyor CI in particular
+  // times out even on >30 seconds timeouts.
+  this.timeout(120000)
```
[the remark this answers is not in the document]
jviotti: Any test. I reduced it a bit, since it now seems to work better after my other changes.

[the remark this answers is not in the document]
jhermsmeier: I just realized I was being a bit too vague here – what I meant is; Do all tests hit the timeout on appveyor (i.e. are all of them that slow), or is it just the startup that's so slow and might hit the timeout?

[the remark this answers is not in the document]
jviotti: Its random as far as I could see.

jviotti:
Its random as far as I could see

On Wed, Dec 13, 2017 at 07:41:35AM -0800, Jonas Hermsmeier wrote:
> jhermsmeier commented on this pull request.
> 
> 
> 
> > +'use strict'
> +
> +const Bluebird = require('bluebird')
> +const spectron = require('spectron')
> +const EXIT_CODES = require('../../lib/shared/exit-codes')
> +const entrypoint = process.env.ETCHER_SPECTRON_ENTRYPOINT
> +
> +if (!entrypoint) {
> +  console.error('You need to properly configure ETCHER_SPECTRON_ENTRYPOINT')
> +  process.exit(EXIT_CODES.GENERAL_ERROR)
> +}
> +
> +describe('Spectron', function () {
> +  // Mainly for CI jobs, Appveyor CI in particular
> +  // times out even on >30 seconds timeouts.
> +  this.timeout(120000)
> 
> I just realized I was being a bit too vague here – what I meant is; Do all tests hit the timeout on appveyor (i.e. are all of them that slow), or is it just the startup that's so slow and might hit the timeout?
> 
> -- 
> https://github.com/resin-io/etcher/pull/1884#discussion_r156695101

-- 
Juan Cruz Viotti
Software Engineer

Issue #1881, opened 2017-11-27
Fix handling of unplugged device, and transient I/O errors


jhermsmeier:
Fix the handling and population of EIO, ENXIO, ENOENT, EBUSY and UNKNOWN during writing.
See https://github.com/resin-io-modules/etcher-image-write/blob/master/lib/filesystem.js

PR #1885, opened 2017-11-28
fix(writer): Add missing write-retry handling

jhermsmeier:
This adds write-retries and handling of potentially temporary errors,
as well as errors due to device disconnection.

Connects To: https://github.com/resin-io/etcher/issues/1881
Change-Type: patch
Changelog-Entry: Fix handling of temporary write errors

### DIFF
--- a/lib/writer/block-write-stream.js
+++ b/lib/writer/block-write-stream.js
@@ -20,6 +20,7 @@ const stream = require('readable-stream')
 const fs = require('fs')
 const speedometer = require('speedometer')
 const debug = require('debug')('block-write-stream')
+const errors = require('./error-types')
 
 const CHUNK_SIZE = 64 * 1024
 const UPDATE_INTERVAL_MS = 500
@@ -30,13 +31,14 @@ const UPDATE_INTERVAL_MS = 500
  */
 class BlockWriteStream extends stream.Writable {
   /**
-   * @summary BlockReadStream constructor
+   * @summary BlockWriteStream constructor
    * @param {Object} [options] - options
    * @param {Number} [options.fd] - file descriptor
    * @param {String} [options.path] - file path
    * @param {String} [options.flags] - file open flags
    * @param {Number} [options.mode] - file mode
    * @param {Boolean} [options.autoClose] - automatically close the stream on end
+   * @param {Number} [options.maxRetries] - maximum number of retries per write
    * @example
    * new BlockWriteStream(options)
    */
@@ -56,12 +58,14 @@ class BlockWriteStream extends stream.Writable {
     this.flags = options.flags
     this.mode = options.mode
     this.autoClose = options.autoClose
+    this.maxRetries = options.maxRetries || 5
 
     this.position = 0
     this.bytesRead = 0
     this.blocksRead = 0
     this.bytesWritten = 0
     this.blocksWritten = 0
+    this.retries = 0
     this.meter = speedometer()
     this.delta = 0
     this.speed = 0
@@ -115,8 +119,10 @@ class BlockWriteStream extends stream.Writable {
       return
     }
 
-    this.bytesRead += chunk.length
-    this.blocksRead += 1
+    if (this.retries === 0) {
+      this.bytesRead += chunk.length
+      this.blocksRead += 1
+    }
 
     if (chunk.position == null) {
       chunk.position = this.position
@@ -134,10 +140,26 @@ class BlockWriteStream extends stream.Writable {
     }
 
     fs.write(this.fd, chunk, 0, chunk.length, chunk.position, (error, bytesWritten) => {
-      this.bytesWritten += bytesWritten
-      this.delta += bytesWritten
-      this.blocksWritten += 1
-      this.position += bytesWritten
+      if (!error) {
+        this.bytesWritten += bytesWritten
+        this.delta += bytesWritten
+        this.blocksWritten += 1
+        this.position += bytesWritten
+        this.retries = 0
+        next(error)
+        return
+      }
+
+      const isTransient = errors.isTransientError(error)
+
+      if (isTransient && (this.retries < this.maxRetries)) {
+        this.retries += 1
+        this._write(chunk, encoding, next)
+        return
+      } else if (isTransient) {
+        error.code = 'EUNPLUGGED'
+      }
+
       next(error)
     })
   }
--- a/lib/writer/block-read-stream.js
+++ b/lib/writer/block-read-stream.js
@@ -19,6 +19,7 @@
 const stream = require('readable-stream')
 const fs = require('fs')
 const debug = require('debug')('block-read-stream')
+const errors = require('./error-types')
 
 const CHUNK_SIZE = 64 * 1024
 const MIN_CHUNK_SIZE = 512
@@ -38,6 +39,7 @@ class BlockReadStream extends stream.Readable {
    * @param {Number} [options.start] - start offset in bytes
    * @param {Number} [options.end] - end offset in bytes
    * @param {Boolean} [options.autoClose] - automatically close the stream on end
+   * @param {Number} [options.maxRetries] - maximum number of retries per read
    * @example
    * new BlockReadStream()
    */
@@ -56,7 +58,9 @@ class BlockReadStream extends stream.Readable {
     this.mode = options.mode
     this.end = options.end || Infinity
     this.autoClose = options.autoClose
+    this.maxRetries = options.maxRetries || 5
 
+    this.retries = 0
     this.position = options.start || 0
     this.bytesRead = 0
 
@@ -83,14 +87,28 @@ class BlockReadStream extends stream.Readable {
       }
 
       if (error) {
+        const isTransient = errors.isTransientError(error)
+
+        if (isTransient && (this.retries < this.maxRetries)) {
+          this.retries += 1
+          this._read()
+          return
+        } else if (isTransient) {
+          error.code = 'EUNPLUGGED'
+        }
+
         if (this.autoClose) {
           this.destroy()
         }
+
         this.emit('error', error)
+
         return
       }
 
+      this.retries = 0
       this.bytesRead += bytesRead
+      this.position += buffer.length
       this.push(buffer)
     }
 
@@ -123,7 +141,6 @@ class BlockReadStream extends stream.Readable {
     const buffer = Buffer.alloc(length)
 
     this.fs.read(this.fd, buffer, 0, length, this.position, this._onRead)
-    this.position += length
   }
 
   /**
--- a/lib/writer/error-types.js
+++ b/lib/writer/error-types.js
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2017 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+module.exports = {
+
+  /**
+   * @summary Determine whether an error is considered a
+   * transient occurrence, and the operation should be retried
+   * Errors considered potentially temporary are:
+   *   - Mac OS: ENXIO, EBUSY
+   *   - Windows: ENOENT, UNKNOWN
+   *   - Linux: EIO
+   * @private
+   * @param {Error} error - Error
+   * @returns {Boolean}
+   * @example
+   * errors.isTransientError(error)
+   */
+  isTransientError (error) {
+    if (process.platform === 'darwin') {
+      return error.code === 'ENXIO' || error.code === 'EBUSY'
+    } else if (process.platform === 'linux') {
+      return error.code === 'EIO'
+    } else if (process.platform === 'win32') {
+      return error.code === 'ENOENT' || error.code === 'UNKNOWN'
+    }
+    return false
+  }
+
+}


lib/writer/block-write-stream.js
```diff
@@ -134,10 +157,24 @@ class BlockWriteStream extends stream.Writable {
     }
 
     fs.write(this.fd, chunk, 0, chunk.length, chunk.position, (error, bytesWritten) => {
```
jhermsmeier: If you fancy testing this, you can add the following here (also try EIO for a different error message):
```js
// Simulate a 80% failure rate
if (!error && (Math.random() > 0.2)) {
  error = new Error('Random I/O error')
  error.code = 'ENXIO'
}
```

lib/gui/pages/main/controllers/flash.js
```diff
@@ -92,7 +92,7 @@ module.exports = function (
       // messages on the first place.
       if (error.code === 'EVALIDATION') {
         FlashErrorModalService.show(messages.error.validation())
-      } else if (error.code === 'EUNPLUGGED') {
+      } else if (error.code === 'ENXIO' || error.code === 'ENOENT' || error.code === 'EBUSY' || error.code === 'UNKNOWN') {
```
jviotti: What worries me about this is that some of these errors codes can happen in other circumstances. The writer process may:

- Throw ENOENT on an unrelated `fs` call
- Throw EBUSY when unmounting
- Throw UNKNOWN on other weird Windows use cases

And in all these cases, the current implementation will think that the drive was unplugged.

jhermsmeier: Like what? At this point the only fs calls going on are writes and/or reads (during verification) – the source has already been opened, and unmount errors don't have `error.code` populated. The only way I can see this come from an unrelated part is if someone deletes the source image while it's flashing.

Interestingly enough, I've just deleted the image while flashing on Mac OS, and it just continues as if nothing happened. And yup, the contents are complete and correct from what I can tell. Windows won't let you delete a file that's being used by a running process, and no clue what Linux does.

lib/writer/block-write-stream.js
```diff
@@ -134,10 +157,24 @@ class BlockWriteStream extends stream.Writable {
...
+        this.retries = 0
+        next(error)
+        return
+      }
+
+      const isTransient = BlockWriteStream.isTransientError(error)
+
+      if (isTransient && (this.retries < this.maxRetries)) {
+        this.retries += 1
+        this._write(chunk, encoding, next)
```
lurch: Does "something else" also need to be done here, given that every call to `_write` does `this.bytesRead += chunk.length` ?

jhermsmeier: It only does that if there hasn't been an error

lib/writer/block-write-stream.js
```diff
@@ -95,6 +97,27 @@ class BlockWriteStream extends stream.Writable {
...
+   *   - Mac OS: ENXIO, EBUSY
+   *   - Windows: ENOENT, UNKNOWN
+   *   - Linux: EIO
+   * @private
+   * @param {Error} error - Error
+   * @returns {Boolean}
+   * @example
+   * BlockWriteStream.isTransientError(error)
+   */
+  static isTransientError (error) {
```
jviotti: Can we somehow re-use this function from the GUI side? Looks like we're basically re-copying this whole check.

lib/writer/block-write-stream.js
```diff
@@ -95,6 +97,27 @@ class BlockWriteStream extends stream.Writable {
     this.open()
   }
 
+  /**
+   * @summary Determine whether an error is considered a
+   * transient occurrence, and the operation should be retried
+   * Errors considered potentially temporary are:
+   *   - Mac OS: ENXIO, EBUSY
+   *   - Windows: ENOENT, UNKNOWN
+   *   - Linux: EIO
```
lurch: So this comment implies that each OS has unique error codes... Is that correct, or could Linux throw a `ENOENT` in a different situation, and we'd want to treat that as a non-transient error? (i.e. do we need to check both the error code **and** the current OS?)

lib/gui/pages/main/controllers/flash.js
jviotti: > Like what? At this point the only fs calls going on are writes and/or reads (during verification) 

That sounds like a risky assumption to make. Things can change that would make this misbehave. Unmounting should already be the case.

> Interestingly enough, I've just deleted the image while flashing on Mac OS, and it just continues as if nothing happened. And yup, the contents are complete and correct from what I can tell. Windows won't let you delete a file that's being used by a running process, and no clue what Linux does.

In UNIX systems, when you delete the file, the kernel schedules its actual deletion to the point where all file descriptors pointing to it are closed.

lib/writer/block-write-stream.js
```diff
@@ -95,6 +97,27 @@ class BlockWriteStream extends stream.Writable {
...
+   * @returns {Boolean}
+   * @example
+   * BlockWriteStream.isTransientError(error)
+   */
+  static isTransientError (error) {
+    return error && (error.code === 'EIO' ||
+      error.code === 'ENXIO' ||
+      error.code === 'ENOENT' ||
+      error.code === 'EBUSY' ||
+      error.code === 'UNKNOWN')
```
lurch: Would this be 'neater' if it used https://lodash.com/docs/4.17.4#find instead?

lib/writer/block-write-stream.js
jviotti: Correct. We should check for the platform as in https://github.com/resin-io-modules/etcher-image-write/blob/master/lib/filesystem.js#L75

jviotti:
Can we write tests, at lest for the EIO errors on the writer side?

lib/writer/block-write-stream.js
lurch: Are you sure? It's possible I'm not understanding the code properly, but from what I can see:
 *  the `_write (chunk, encoding, next) {` function starts on line 130
 * on line 141 you then do `this.bytesRead += chunk.length`
 * on line 159 you do the actual `fs.write(...`
 * on line 161 you do `this.bytesWritten += bytesWritten` if there wasn't an error
 * if the error was transient, you then do `this._write(chunk, encoding, next)` on line 174

So AFAICT if you get a bunch of transient errors then `this.bytesWritten` will be correct, but `this.bytesRead` will be incremented too much?

lib/writer/block-write-stream.js
jhermsmeier: I'd rather not have lodash anywhere near performance critical code, as our multipipage benchmarks revealed it was a source of allocations & slowness; a demonstration:

```
NANOBENCH version 2
> node error-code-check.js

# equals ⨉ 1000000
ok ~13 ms (0 s + 12832994 ns)

# branch equals ⨉ 1000000
ok ~23 ms (0 s + 23072749 ns)

# switch equals ⨉ 1000000
ok ~25 ms (0 s + 24974775 ns)

# branch / array includes ⨉ 1000000
ok ~35 ms (0 s + 35088306 ns)

# switch / array includes ⨉ 1000000
ok ~38 ms (0 s + 38391439 ns)

# lodash array includes ⨉ 1000000
ok ~180 ms (0 s + 180124552 ns)

# lodash branch array includes ⨉ 1000000
ok ~188 ms (0 s + 187953699 ns)

all benchmarks completed
ok ~502 ms (0 s + 502438514 ns)
```

Where the "branch equals" is the check we want here (platform dependant), and "equals" is what this initially is/was here.

Which makes use of lodash a slowdown of ~800% – even though minuscule at these time scales, the calls add up throughout and end up accounting for a noticeable slowdown.

lib/writer/block-write-stream.js
jhermsmeier: Oh, sorry I misread `bytesRead` – yes, you're right.

jhermsmeier: Fixed!

lib/writer/block-write-stream.js
jhermsmeier: Done

lib/writer/block-write-stream.js
jhermsmeier: Done

lib/writer/block-write-stream.js
lurch: Would it make sense to globally disable the 'prefer lodash' linter settting? Or is it still okay to use lodash, just not in performance critical code?

lib/gui/modules/image-writer.js
```diff
@@ -124,7 +125,7 @@ imageWriter.service('ImageWriterService', function ($q, $rootScope) {
 
       if (error.code === 'EVALIDATION') {
         analytics.logEvent('Validation error', analyticsData)
-      } else if (error.code === 'EUNPLUGGED') {
+      } else if (BlockWriteStream.isTransientError(error)) {
```
lurch: So, is a "transient error" always guaranteed to be a "drive unplugged" error?
Unplugging a drive doesn't sound very transient to me ;-)
I wonder if this `isTransientError` function should be moved outside of `BlockWriteStream`, given that a drive-unplugged event could also occur during validation, and not just during writing?

And apologies for the stupid question, but why do we have nearly identical error handling in both `image-writer.js` and `flash.js` ?

lib/writer/block-write-stream.js
```diff
@@ -56,12 +58,14 @@ class BlockWriteStream extends stream.Writable {
...
     this.mode = options.mode
     this.autoClose = options.autoClose
+    this.maxRetries = 5
 
     this.position = 0
     this.bytesRead = 0
     this.blocksRead = 0
     this.bytesWritten = 0
     this.blocksWritten = 0
+    this.retries = 0
```
lurch: Not that it really matters, but in other code within Etcher (I can't remember where now) the 'retry' was handled via an extra function parameter (which defaulted to 0), e.g.
```javascript
somefunc(params, tries=0) {
    doSomething()
    if (error)
        if (tries < maxRetries) {
            somefunc(params, tries + 1)
        } else {
            displayError()
        }
    }
}
```
whereas here you're storing the retry count on `this`.
Are there any pros or cons to either approach?

lib/writer/block-write-stream.js
```diff
@@ -56,12 +58,14 @@ class BlockWriteStream extends stream.Writable {
     this.flags = options.flags
     this.mode = options.mode
     this.autoClose = options.autoClose
+    this.maxRetries = 5
```
lurch: Does it make sense to store `maxRetries` on `this`? I can't imagine it ever changing, so perhaps a global `MAX_WRITE_RETRIES` const would be more appropriate?

lib/gui/modules/image-writer.js
jhermsmeier: Well, it's the same error codes you get if the device is unplugged, as they indicate a loss in connection (which is usually temporary, if the device hasn't been unplugged), hence transient.

jhermsmeier: > And apologies for the stupid question, but why do we have nearly identical error handling in both image-writer.js and flash.js ?

I don't know; it's been a topic in meetings, that we need to consolidate our error handling, because there's quite a bit of duplication and it's scattered about. Not in the scope of this PR though.

lurch: Ahhh, so I guess this is _effectively_ saying "if we had a transient error, and we still get the error even after 5 retries, then assume it means that the drive has been unplugged" ?

lib/writer/block-write-stream.js
```diff
@@ -134,10 +164,24 @@ class BlockWriteStream extends stream.Writable {
...
+        this.position += bytesWritten
+        this.retries = 0
+        next(error)
+        return
+      }
+
+      const isTransient = BlockWriteStream.isTransientError(error)
+
+      if (isTransient && (this.retries < this.maxRetries)) {
+        this.retries += 1
```
lurch: Perhaps I'm mis-remembering, or perhaps I'm getting confused with something else, but ISTR that when @jviotti added retry-handling somewhere, he actually put in a small delay between getting the error and retrying the operation, rather than retrying immediately? Would/could that be beneficial here?

lurch: Aha, found it :smiley: 
https://github.com/resin-io-modules/etcher-image-write/blob/master/lib/filesystem.js#L92

It actually increases the delay as the number of retries increases. And it also does up to 10 retries, rather than the up to 5 that you do here. I'll let you and @jviotti argue about the pros and cons of the different approaches... :wink:

lib/gui/modules/image-writer.js
jviotti: This is still not correct. Anything on the writer can throw these errors, and they would be incorrectly caught here. Instead of using a global handler for these generic errors, we should only check them in `fs.read` and `fs.write`. We should limit the scope of the error handler. That was the reasoning behind `EUNPLUGGED`.

jviotti:
We shouldn't merge this as it is. We're adding a massive error handler covering way more than we need to cover, potentially leading to non-sense error messages as a result of unrelated error messages that happen to have the same error code. We should limit the scope of the error handler.

lib/writer/block-write-stream.js
```diff
@@ -95,6 +98,30 @@ class BlockWriteStream extends stream.Writable {
...
+   *   - Mac OS: ENXIO, EBUSY
+   *   - Windows: ENOENT, UNKNOWN
+   *   - Linux: EIO
+   * @private
+   * @param {Error} error - Error
+   * @returns {Boolean}
+   * @example
+   * BlockWriteStream.isTransientError(error)
+   */
+  static isTransientError (error) {
```
jviotti: This function is a copy-paste of the `isTransientError` function in `BlockReadStream`. Can we move it somewhere else so that both streams can re-use the same function?

lib/writer/block-write-stream.js
```diff
@@ -134,10 +163,26 @@ class BlockWriteStream extends stream.Writable {
...
+      }
+
+      const isTransient = BlockWriteStream.isTransientError(error)
+
+      if (isTransient && (this.retries < this.maxRetries)) {
+        this.retries += 1
+        this._write(chunk, encoding, next)
+        return
+      } else if (isTransient) {
+        error.code = 'EUNPLUGGED'
```
jviotti: Should we make `EUNPLUGGED` a constant?

lib/writer/block-write-stream.js
jhermsmeier: I'd rather keep the streams self-contained, but can do – where'd you put it though?

lib/writer/block-write-stream.js
jhermsmeier: I don't see much sense in that, tbh – and we don't do that for other error codes either.

lib/writer/block-write-stream.js
jviotti: In some sort of `utils.js` file?

lib/writer/block-write-stream.js
lurch: Any thoughts @jhermsmeier re: adding a delay between retry attempts?

lib/writer/block-write-stream.js
jhermsmeier: Done

lib/writer/block-write-stream.js
```diff
@@ -134,10 +140,26 @@ class BlockWriteStream extends stream.Writable {
...
+        return
+      }
+
+      const isTransient = errors.isTransientError(error)
+
+      if (isTransient && (this.retries < this.maxRetries)) {
+        this.retries += 1
+        this._write(chunk, encoding, next)
+        return
+      } else if (isTransient) {
```
lurch: Minor nit, but this looks a bit weird. Could you instead do:
```
if (isTransient) {
    if (this.retries < this.maxRetries) {
    } else {
    }
}
```
?

lurch:
I added a comment a few days ago, but GitHub is now helpfully "auto-hiding" it, so I dunno if you saw it? https://github.com/resin-io/etcher/pull/1885#discussion_r154167695

lib/writer/block-write-stream.js
jhermsmeier: Excessive branching ain't good for perf; better to keep conditions as flat as possible

lib/writer/block-write-stream.js
jhermsmeier: Hmm yep

jhermsmeier: There'll be a delay of a few milliseconds between the retries by default, due to the I/O and it running in the next loop tick; I don't think we need to artificially delay it even more – @jviotti thoughts?

lib/writer/block-write-stream.js
lurch: Huh? I wouldn't expert there to be much performance difference (if any) between that and the:
```
if (isTransient && (this.retries < this.maxRetries)) {
    // stuff
} else if (isTransient) {
    // other stuff
}
```
that you're using currently? Note that in both versions there's two `if` statements and one `else` statement, they're just arranged slightly differently ;-)

But then I obviously have no idea how NodeJS's internal optimisations work, so perhaps there really is a difference?

Issue #1883, opened 2017-11-28
Fix OS window keyboard shortcuts


jhermsmeier:
Due to `electron.Menu.setApplicationMenu(null)`, native window shortcuts such as hiding or minimising the Window get removed.

lurch:
But Etcher is beautiful! Surely nobody would _ever_ want to hide it?

lurch:
In #1909 @joshlewis also requested:
⌘, - Open preferences (Command-comma)

Do the other platforms we support (Linux and Windows) have a similar keyboard shortcut for the same thing?

jhermsmeier:
> Do the other platforms we support (Linux and Windows) have a similar keyboard shortcut for the same thing?

Not that I'm aware of, but will double-check

lurch:
It occurred to me last night that `Ctrl` + `O` (open) might be a common keyboard shortcut that we could implement, to open the image-selection dialog?

lurch:
(sorry, fat-fingered the wrong button!)

jhermsmeier:
Closed in https://github.com/resin-io/etcher/pull/1888, released in v1.3.0

Issue #1840, opened 2017-11-11
Etcher will not go full screen


ghost:
- 1.1.2
- Mac OSX 64 bit
- Raspbian
- No

It will not resize to full screen when you click the + button.

jezmck:
The maximise button in Windows is greyed out, so I guess the interface isn't designed for it.

Any particular reason you want it full screen?

JBoyAdvance:
No, there is not a specific reason. I just was wondering why it did not go full screen. It'd be nice if it did, though.

JBoyAdvance:
No, there is not a specific reason. I just was wondering why it did not go full screen. It'd be nice if it did, though.

lurch:
Correct. The Etcher GUI is currently non-resizable, and therefore can't be maximised. On Linux the maximise button isn't shown either, so perhaps the maximise button getting shown on OSX is a bug in Electron? Or maybe OSX simply doesn't allow the maximise button to be disabled / removed? \*shrug\*

PR #1888, opened 2017-11-30
fix(gui): Re-enable application menu

jhermsmeier:
This re-enables the application menu to allow for OS native shortcuts
to work again (i.e. hide/minimize window), which also allows us to
get rid of the global-shortcuts hack to prevent window reloads.
Also adds links to the website, Etcher Pro and GitHub issues in the help menu.

Menu:

- App (Mac OS): Quit, hide / show window, preferences
- App (Windows, Linux): Quit, preferences
- View: Enable showing DevTools without that global shortcut hack, disable content zoom and reloads
- Edit: Don't disable native copy/paste & shortcuts (needed in DevTools and when we'll add an "Open URL" etc)
- Window: Native window management shortcuts, like minimize / maximise and friends (dep. on OS)

## Mac OS

screen shot 2017-12-11 at 20 37 29

etcher-mac-os-preferences

## Windows

Default state:

etcher-menu-default

When hitting <kbd>Alt</kbd>:

etcher-menu-open

## Linux

Default state:

screenshot from 2018-01-03 16-36-44

When hovering over the menu bar:

screenshot from 2018-01-03 16-39-20

Connects To: #1883 #1909
Change-Type: patch
Changelog-Entry: Fix disabled native OS window shortcuts

### DIFF
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -346,6 +346,13 @@ app.controller('StateController', function ($rootScope, $scope) {
   this.currentName = null
 })
 
+// Handle [Cmd] + [,] shortcut to open the settings
+app.run(($state) => {
+  electron.ipcRenderer.on('menu:preferences', () => {
+    $state.go('settings')
+  })
+})
+
 // Ensure user settings are loaded before
 // we bootstrap the Angular.js application
 angular.element(document).ready(() => {
--- a/lib/gui/app.js
+++ b/lib/gui/app.js
@@ -346,7 +346,7 @@ app.controller('StateController', function ($rootScope, $scope) {
   this.currentName = null
 })
 
-// Handle [Cmd] + [,] shortcut to open the settings
+// Handle keyboard shortcut to open the settings
 app.run(($state) => {
   electron.ipcRenderer.on('menu:preferences', () => {
     $state.go('settings')
--- a/lib/gui/etcher.js
+++ b/lib/gui/etcher.js
@@ -17,7 +17,6 @@
 'use strict'
 
 const electron = require('electron')
-const _ = require('lodash')
 const path = require('path')
 const EXIT_CODES = require('../shared/exit-codes')
 let mainWindow = null
@@ -35,8 +34,7 @@ electron.app.on('before-quit', () => {
 })
 
 electron.app.on('ready', () => {
-  // No menu bar
-  electron.Menu.setApplicationMenu(null)
+  electron.Menu.setApplicationMenu(require('./menu'))
 
   mainWindow = new electron.BrowserWindow({
     width: 800,
@@ -44,7 +42,10 @@ electron.app.on('ready', () => {
     useContentSize: true,
     show: false,
     resizable: false,
+    maximizable: false,
     fullscreen: false,
+    fullscreenable: false,
+    autoHideMenuBar: true,
     titleBarStyle: 'hidden-inset',
     icon: path.join(__dirname, '..', '..', 'assets', 'icon.png')
   })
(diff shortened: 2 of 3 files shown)


lib/gui/etcher.js
```diff
@@ -35,8 +35,13 @@ electron.app.on('before-quit', () => {
 })
 
 electron.app.on('ready', () => {
-  // No menu bar
-  electron.Menu.setApplicationMenu(null)
+
+  if (process.platform === 'win32') {
```
lurch: Why specifically different behaviour on Windows?

lurch:
Before this gets approved, IMHO you should include before and after screenshots from all 3 OSes, and we should get feedback from @konmouz and @alexandrosm 

I thought Etcher not having a menu was a "feature", which adds to its elegant interface and simple usability...?

lib/gui/etcher.js
jhermsmeier: Because the menu is displayed within the window on Windows, which we don't want – but I still need to check if the shortcuts etc. still work there.

lurch:
Does #1840 relate to this at all? Does re-enabling the menubar still prevent Etcher from being maxmised? (and to clarify, I believe we want to keep the behaviour of Etcher being non-maximisable?)

jhermsmeier:
@lurch no, the menubar doesn't magically override menu settings

jhermsmeier:
@Shou is there an easy way to programatically toggle the settings view?

lurch:
Why does the first menu item say `Electron` and not `Etcher` ? :confused:

jhermsmeier:
Because that's the appName when running `npm start`, as that runs Electron – the appName will only be set to "Etcher" once packaged

lurch:
Ahhh, might be worth re-taking the screenshots with packaged versions of Etcher then, to avoid confusion ;-)

What does the "about" option display?  (on Linux and Windows this is usually an option on the Help menu)
What "edit-menu" functionality does Etcher expose? (I can't think where copy'n'paste would come in handy)

jhermsmeier:
Added <kbd>⌘</kbd>+<kbd>,</kbd> to open / toggle preferences on Mac OS now. On Windows there doesn't seem to be a convention around that, other than having a "Settings" thing in the menu. Still need to dig into Linux.

lurch:
In other Mac apps, is the Preferences thing always a 'toggle'? I don't have any Mac experience myself.
In Linux apps, if there was a shortcut to display the preferences (I also haven't done any digging to see if there's a convention here or not) I'd expect that to display the preferences window, and for pressing it again to leave the preferences window open, rather than toggling it closed again.

Did you see my suggestion in #1909 about adding a `Ctrl`+`o` shortcut?  Could you update your GIF to quickly show what _each_ of the top-level menus displays?

lib/gui/app.js
```diff
@@ -346,6 +346,17 @@ app.controller('StateController', function ($rootScope, $scope) {
   this.currentName = null
 })
 
+// Handle [Cmd] + [,] shortcut to open the settings
```
lurch: Given that the ` [Cmd] + [,] shortcut` is encapsulated inside `lib/gui/menu.js`, perhaps this comment should just mention the `preferences:toggle` action? \*shrug\*

jhermsmeier:
> Maybe I'm missing something, but I don't understand why Etcher needs Edit, View or Window menus??

I mentioned this in last Fridays meeting, but should've added it to the description, so here goes (also added this to the PR description):

- App (Mac OS): Quit, hide / show window, preferences
- App (Windows, Linux): Quit, preferences
- View: Enable showing DevTools without that global shortcut hack, disable content zoom and reloads
- Edit: Don't disable native copy/paste & shortcuts (needed in DevTools and when we'll add an "Open URL" etc)
- Window: Native window management shortcuts, like minimize / maximise and friends (dep. on OS)

konmouz:
@jhermsmeier, in my opinion, we should avoid the menu if that is possible. If we can't completely remove the menu, hiding is fine. But in any case, the menu shouldn't provide any additional functionality that can't be found in the Etcher UI.

jhermsmeier:
@konmouz Good point; so far it doesn't add functionality that's not available through the UI – apart from native OS window management keyboard shortcuts, which shouldn't have been disabled in the first place – even the contents of the Help menu can be found in the UI.

jhermsmeier:
Finally added screenshots for Linux as well

Shou:
It works great, but I think lurch's comment above highlights a small UX issue:

> In other Mac apps, is the Preferences thing always a 'toggle'? I don't have any Mac experience myself.

I believe it makes more sense for it to always link to settings, so I'd say either do just that, or maybe change the label to 'main screen'?

jhermsmeier:
Ah yes, missed that – returned it to what'd be considered default; it only goes one way now.

lib/gui/app.js
jviotti: Yeah, I agree. This part of the code doesn't care what the shortcut is.

jhermsmeier: Ah yes, missed this comment before. Updated now!

PR #1892, opened 2017-12-01
feat(SDK): display Compute Modules even if host OS has no drivers

jviotti:
This is the first step towards full usbboot Windows support. The driver
selector dialog will now display disabled devices to represent Compute
Modules even when Windows drivers are not installed to act on them.
These drives will state "Missing drivers."

Change-Type: minor
Changelog-Entry: Display connected Compute Modules even if Windows doesn't have the necessary drivers to act on them.


lurch:
```
C:\projects\etcher\lib\shared\sdk\usbboot\usb.js
  79:5  error  Prefer _.map over a _.forEach with a push to an array inside  lodash/prefer-map
```
:-/

lib/shared/sdk/usbboot/usb.js
```diff
@@ -67,7 +68,26 @@ const USB_CONTROL_TRANSFER_TIMEOUT_MS = 0
...
+  const devices = _.map(usb.getDeviceList(), (device) => {
+    device.driver = true
+    return device
+  })
+
+  // Include driverless devices into the list of USB devices
+  if (os.platform() === 'win32') {
+    const winusbDriverGenerator = require('winusb-driver-generator')
+    _.each(winusbDriverGenerator.listDriverlessDevices(), (device) => {
+      devices.push({
```
lurch: I guess we probably want to filter this list of all DriverlessDevices to just those VIDs and PIDs that we're interested in :stuck_out_tongue_winking_eye:

lib/shared/sdk/usbboot/index.js
```diff
@@ -251,6 +251,21 @@ class USBBootAdapter extends EventEmitter {
     usb.listDevices().filter(isUsbBootCapableUSBDevice).map((device) => {
     /* eslint-enable lodash/prefer-lodash-method */
 
+      if (!device.driver) {
+        return {
+          device: `${usbIdToString(device.deviceDescriptor.idVendor)}:${usbIdToString(device.deviceDescriptor.idProduct)}`,
```
lurch: So, is this the first time that a "drivelist object" is coming directly from the SDK rather than from `drivelist` itself? I guess other SDK modues might also need to do something similar in future, so we should probably document some kind of "drivelist object spec" with what compulsory and optional properties (and their types) we expect this object to have?
Is there code in Etcher that emits a warning if a drivelist object is missing some of its compulsory properties? (and then presumably just ignores that incomplete drivelist object)

jviotti: @lurch We've been already doing so for a while in the usbboot implementation (look in this same file). The required properties should be defined by the SDK (@jhermsmeier is working on the spec). Drivelist happens to be an implementation that adheres to the current required properties.

> Is there code in Etcher that emits a warning if a drivelist object is missing some of its compulsory properties? (and then presumably just ignores that incomplete drivelist object)

Most of the functions have fallbacks in case a certain property is not defined. In any case, this will probably be discussed in much more detail in the SDK spec.

lib/shared/sdk/usbboot/usb.js
jviotti: We do the filtering on the usbboot side. This is a "usb" module that doesn't care about the things it lists.

jviotti:
I fixed the lint warning.

jviotti:
screen shot 2017-12-04 at 2 20 19 pm

This is how it looks like atm ^

lib/shared/sdk/usbboot/index.js
```diff
@@ -251,6 +251,21 @@ class USBBootAdapter extends EventEmitter {
     usb.listDevices().filter(isUsbBootCapableUSBDevice).map((device) => {
     /* eslint-enable lodash/prefer-lodash-method */
 
+      if (!device.accessible) {
+        return {
+          device: `${usbIdToString(device.deviceDescriptor.idVendor)}:${usbIdToString(device.deviceDescriptor.idProduct)}`,
+          displayName: 'Missing drivers',
+          description: 'Compute Module',
```
lurch: I guess one refinement here would be to display separate labels for `Compute Module 1` and `Compute Module 3` based on the `.idProduct` (since this will be at the stage from before usbboot is run, so the PIDs are still distinct)

lib/shared/sdk/usbboot/usb.js
lurch: Ah, my mistake! I initially saw:
```
      if (!device.accessible) {
        return {
          device: `${usbIdToString(device.deviceDescriptor.idVendor)}:${usbIdToString(device.deviceDescriptor.idProduct)}`,
          displayName: 'Missing drivers',
          description: 'Compute Module',
```
in `lib/shared/sdk/usbboot/index.js` and thought you were assuming that any driverless device was a 'Compute Module'. I'd missed that you're also doing a `filter(isUsbBootCapableUSBDevice)` just a few lines above ;-)

lurch:
Appveyor says
```
rebuild-install info begin Prebuild-install version 2.3.0
prebuild-install info looking for local prebuild @ prebuilds\winusb-driver-generator-v1.1.0-node-v48-win32-x64.tar.gz
prebuild-install info looking for cached prebuild @ C:\Users\appveyor\AppData\Roaming\npm-cache\_prebuilds\https-github.com-resin-io-modules-winusb-driver-generator-releases-download-v1.1.0-winusb-driver-generator-v1.1.0-node-v48-win32-x64.tar.gz
prebuild-install http request GET https://github.com/resin-io-modules/winusb-driver-generator/releases/download/v1.1.0/winusb-driver-generator-v1.1.0-node-v48-win32-x64.tar.gz
prebuild-install http 404 https://github.com/resin-io-modules/winusb-driver-generator/releases/download/v1.1.0/winusb-driver-generator-v1.1.0-node-v48-win32-x64.tar.gz
prebuild-install WARN install No prebuilt binaries found (target=6.1.0 runtime=node arch=x64 platform=win32)
make: *** [dist/Etcher-cli-1.2.0+6ce8fea-win32-x64-app] Error 1
```
:-/
https://github.com/resin-io-modules/winusb-driver-generator/releases seems to have `winusb-driver-generator-v1.1.0-electron-v54-win32-x64.tar.gz` but not the `winusb-driver-generator-v1.1.0-node-v48-win32-x64.tar.gz` that is being looked for here? Ping @jhermsmeier as he seems to be familiar with the 'prebuild' stuff :wink: 

TravisCI builds are passing on Linux, but failing on OSX, which is weird :-/
```
  1) Browser: driveScanner given windows given available drives should emit the non removable drives:
     Error: Module did not self-register.
```
?

lurch:
Might be worth updating that screenshot, now #1900 is merged ;-)

jviotti:
Regarding the first prebuild error, that's happening because I'm pre-building for Electron, but not for Node.js, which we use to package the CLI. I'll update the module to provide that package.

Regarding build errors on macOS but not on GNU/Linux, remember that we don't load usbboot on GNU/Linux yet and that's why only macOS exhibits the problem. I can locally reproduce so I'll fix it soon.

lurch:
Oh yeah, I hadn't even noticed that one of the prebuild filenames contains `electron` and the other contains `node`!  I'd only noticed the `v54` and `v48` difference :confused:

jviotti:
Here's the updated screenshot:

screen shot 2017-12-08 at 12 03 55 pm

jviotti:
Updated! Everything should be green now.

lib/gui/assets/warning.svg
```diff
@@ -0,0 +1,38 @@
...
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
```
jhermsmeier: Where are all these empty groups coming from? Can we remove them?

lib/shared/sdk/usbboot/index.js
jviotti: Done!

lib/gui/assets/warning.svg
jviotti: Removed!

lib/shared/sdk/usbboot/index.js
```diff
@@ -251,14 +261,31 @@ class USBBootAdapter extends EventEmitter {
     usb.listDevices().filter(isUsbBootCapableUSBDevice).map((device) => {
     /* eslint-enable lodash/prefer-lodash-method */
 
+      const description = COMPUTE_MODULE_DESCRIPTIONS[device.deviceDescriptor.idProduct] || 'Compute Module'
```
jhermsmeier: Should this maybe default to something more generic, like "USBBoot Device"? If that makes sense, that is

lib/gui/assets/warning.svg
lurch: I'd guess `Generator: Adobe Illustrator 16.0.0` is the culprit ;-)

lib/shared/sdk/usbboot/index.js
lurch: Nope, it's only the Raspberry Pi Compute Modules that implement this (proprietary) USB-boot protocol that we're dealing with here.

lurch:
Apologies for the delay, but I've finally got round to testing this PR on my old Win7 x64 laptop... and it seems to work fine :-)
Here's a screenshot with both a CM1 and CM3 plugged in at the same time (after I managed to get all their drivers uninstalled)
image

lib/shared/sdk/usbboot/index.js
```diff
@@ -251,14 +261,31 @@ class USBBootAdapter extends EventEmitter {
...
+
       // This is the only way we can unique identify devices
       device.raw = `${device.busNumber}:${device.deviceAddress}`
 
       const result = {
         device: device.raw,
         raw: device.raw,
         displayName: 'Initializing device',
-        description: 'Compute Module',
+        description,
```
lurch: Is this `description` here only shown before `bootcode.bin` has been written to the Pi, or also afterwards? If it's the latter, remember that after `bootcode.bin` has been written, both the CM1 and CM3 re-enumarate as the identical VID:PID combo of 0x0a5c:0x2764 which is why this previously said only "Compute Module" rather than being more specific.

Whereas in the `if (!device.accessible) {` case, we know `bootcode.bin` definitely hasn't been written to the device yet, so at that point it _is_ safe to differentiate by PID.

jviotti: Ah, true, let me fix that.

jviotti: Done, glad you caught this :D

lurch: Of course, once we add the USB-hardware-path stuff into Etcher (needed for Etcher Pro) then we'll be able to change all this yet again as we'll be able to uniquely track devices by which physical slot they're plugged into, so we'll then know what a device initially enumerated as, even after it re-enumerates with 0x0a5c:0x2764 :-)

lurch:
```
C:\projects\etcher\lib\shared\sdk\usbboot\index.js
  288:9  error  Expected line before comment  lines-around-comment
```

jviotti:
Fixed!

PR #1900, opened 2017-12-04
minifix: hide drive selector hyphen when no size info

Shou:
We hide the separator hyphen between the name and size when there is no
size information available, in the drive selector modal.

Change-Type: patch
Changelog-Entry: Hide the drive-selector separator hyphen when no drive
size is available.


Shou:
Screenshot

image

lurch:
When no drive-size is available, does Etcher always assume that _any_ size disk-image will fit onto that drive?

PR #1905, opened 2017-12-05
upgrade(package): Bump electron-builder 19.9.1 -> 19.40.0

jhermsmeier:
Change-Type: patch
Connects To: #1914


jhermsmeier:
This currently fails for deb packages due to

```
{:timestamp=>"2017-12-05T17:06:10.361199+0000", :message=>"Process failed: tar failed (exit code 2). Full command was:[\"tar\", \"-C\", \"/tmp/package-dir-staging-8abc4590f5fcbb8ef993da8fef97ba2aa9169d2c54a24f906c4da425a75a\", \"-I'/etcher/node_modules/7zip-bin/7x.sh'\", \"--numeric-owner\", \"--owner\", \"0\", \"--numeric-owner\", \"--group\", \"0\", \"-cf\", \"/tmp/package-deb-build-f194c570b6596c408446b22df37210ba51e03b9d6a4b2d491ca71786153b/data.tar.xz\", \".\"]", :level=>:error}
```

lurch:
This introduces a massive chunk of changes to `npm-shrinkwrap.json` - perhaps I'm being overly conservative, but does that mean we ought to do a bit of extra testing before release, just to make sure this doesn't accidentally break anything?

jhermsmeier:
This isn't actually even working for deb packages atm; and I haven't gotten it to work so far, let alone figure out what exactly is causing fpm to explode

scripts/build/docker/Dockerfile.template
```diff
@@ -12,6 +12,10 @@ RUN echo "deb http://archive.ubuntu.com/ubuntu precise-backports main restricted
 RUN echo "deb http://ftp.debian.org/debian jessie-backports main" >> /etc/apt/sources.list
 <% } %>
 
+# Workaround: Install a newer version of `tar` to make fpm in electron-builder work again
+RUN echo "deb http://ftp.de.debian.org/debian wheezy main contrib non-free" >> /etc/apt/sources.list
```
lurch: Should we just use `http://ftp.debian.org/` as in the previous lines, rather than selecting the `de`-specific subdomain?

scripts/build/docker/Dockerfile.template
```diff
@@ -43,6 +47,10 @@ RUN apt-get update \
     zip \
     rpm
 
+# Workaround: Install a newer version of `tar` to make fpm in electron-builder work again
```
lurch: IMHO these two 'Workaround' blocks should only be run in the `architecture == 'i686'` and `architecture == 'x86_64'` Dockerfiles - they're not needed in the `architecture == 'armv7hf'` Dockerfile as that uses a newer base-image and so the workaroud would actually _downgrade_ the version of `tar`.

scripts/build/docker/Dockerfile.template
```diff
@@ -14,7 +14,7 @@ RUN echo "deb http://ftp.debian.org/debian jessie-backports main" >> /etc/apt/so
 
 # Install dependencies
 RUN apt-get update \
-  && apt-get install -y \
+  && apt-get install --force-yes -y \
```
jviotti: Is this really needed at this point?

scripts/build/docker/Dockerfile.template
```diff
@@ -43,6 +43,12 @@ RUN apt-get update \
     zip \
     rpm
 
+# Workaround: Install a newer version of `tar` to make fpm in electron-builder work again
+RUN echo "deb http://ftp.de.debian.org/debian wheezy main contrib non-free" >> /etc/apt/sources.list
```
jviotti: I guess we don't need the `non-free`. We can also try to remove `contrib`

scripts/build/docker/Dockerfile.template
jviotti: Something like 

```
<% if (architecture == 'i686' || architecture == 'x86_64') { %>
...
<% } %>
```

should do.

lurch:
Is the new `assets/icons/256x256.png` just a duplicate of the existing `assets/icon.png`? If that's the case, perhaps a neater solution would be to remove this new file, and instead add:
```
assets/icons/256x256.png: assets/icon.png
    cp $< $@
```
to the Makefile, and then list `assets/icons/256x256.png` as a dependency of those electron-builder rules that actually need it? (and that file should also then be deleted by the `clean` rule, and be added to `.gitignore` - and in fact these last two steps should probably be added for `assets/osx/installer.tiff` too)

scripts/build/docker/Dockerfile.template
jviotti: Yeah, `libacl1` is on `main`

scripts/build/docker/Dockerfile.template
jhermsmeier: Nope not anymore, removed

scripts/build/docker/Dockerfile.template
jhermsmeier: Done

scripts/build/docker/Dockerfile.template
jhermsmeier: Done

scripts/build/docker/Dockerfile.template
jhermsmeier: Done

scripts/build/docker/Dockerfile.template
```diff
@@ -44,10 +44,18 @@ RUN apt-get update \
     rpm
 
 <% if (architecture != 'armv7hf') { %>
+
+# Workaround: Install a newer version of `tar` to make fpm in electron-builder work again
+RUN echo "deb http://ftp.debian.org/debian wheezy main" >> /etc/apt/sources.list
+RUN echo "deb http://ftp.debian.org/debian wheezy-backports main" >> /etc/apt/sources.list
+RUN apt-get install --force-yes -y -t wheezy libacl1 && \
```
lurch: Ah, you need an `apt-get update` between modify `sources.list` and running `apt-get install`

lurch: ...and perhaps this should be done in a single-step `RUN` command (just like the compiler-install step below), rather than as separate `RUN` commands?

jhermsmeier: Fixed

lurch: I was thinking more:
```
RUN echo -e "deb http://ftp.debian.org/debian wheezy main\ndeb http://ftp.debian.org/debian wheezy-backports main" >> /etc/apt/sources.list && \
  apt-get update && apt-get install --force-yes -y -t wheezy libacl1 && \
  apt-get install --force-yes -y -t wheezy-backports tar
```
But it probably makes little difference here.

lurch:
Is it worth adding a note to the commit message about _why_ we've put so much effort into upgrading `electron-builder`? (IIRC it's so that the hdiutil still works on the latest versions of OSX?)

jviotti:
@jhermsmeier Even though DMGs build fine, I get weird ABI issues when running this particular `electron-builder` version. I started bisecting it a bit, and found that 19.40.0 works like a charm. Can we upgrade to that instead?

lurch:
@jviotti out of curiosity, did you also try `19.49.0` ?

jviotti:
I'm getting the same weird ABI error on 19.49.0.

On Fri, Dec 15, 2017 at 05:49:21PM -0800, Andrew Scheller wrote:
> @jviotti out of curiosity, did you also try `19.49.0` ?
> 
> -- 
> https://github.com/resin-io/etcher/pull/1905#issuecomment-352151787

-- 
Juan Cruz Viotti
Software Engineer

lurch:
Have you filed an upstream bug? ;)

jhermsmeier:
@jviotti can you post the relevant log of that ABI error?

jviotti:
@jhermsmeier The CI services are showing the exact error I get: https://travis-ci.org/resin-io/etcher/jobs/312464338#L2132

jviotti:
@lurch See https://github.com/electron-userland/electron-builder/issues/2260. Apparently the solution is:

> To avoid such errors in the future, we will get rid of NodeJS and this logic will reimplemented using Go.

jhermsmeier:
Oh, I see – that's because the module's been compiled against Electron's ABI, while being invoked by Node during the build step. One way to get around this would be to use the Node version corresponding to Electron's Node (v7.9.0 in Electron v1.7.9), but that's not something we want to do I think (our build process should ideally be fairly independent of Node version). I'll downgrade electron-builder to 19.40.0 for now then.

lurch:
> See electron-userland/electron-builder#2260

...which says "The rabin dependency was added as part of commit 1dc2e49, so 19.36.0 onward exhibit this issue. (19.35.1 works)" ??

jviotti:
@lurch Yeah, but that version hangs on High Sierra with 100% CPU usage when building the DMG :( 19.40.0 works like a charm from all sides. I say lets upgrade to that, and let `electron-builder` sort out the ABI problem. We will try again once they have a fix. @jhermsmeier what do you think? Lets try to get this merged soon so I can integrate Concourse without hanging all the jobs

jviotti:
> I'll downgrade electron-builder to 19.40.0 for now then.

Ah, I missed that comment :D

PR #1910, opened 2017-12-07
refactor(GUI): remove the intermediate child writer proxy process

jviotti:
Etcher currently elevates a child writer proxy that itself spawns the
Etcher CLI in robot mode, parses the output, and proxies those messages
to the GUI application over IPC.

After these set of changes, Etcher elevates a single child writer
process that directly communicates back with the GUI using IPC. The main
purpose behind these changes is to simplify the overall architecture and
fix various issues caused by the current complex child process tree.

Here's a summary of the changes:

- Stop wrapping the Etcher CLI to perform writing
- Remove the robot option from the Etcher CLI (along with related
  documentation)
- Elevate a new `child-write.js` standalone executable
- Move the relevant bits of `lib/child-writer` to the `image-writer` GUI
  module
- Remove the `lib/child-writer` directory
- Add a new "Child died unexpectedly" Mixpanel event
- Floor state percentage in the flash state model

The above changes made is possible to tackle all the remaining issues
where the writer process would remain alive even if the parent died.

Change-Type: patch
Changelog-Entry: Ensure the writer process dies when the GUI application is killed.
See: https://github.com/resin-io/etcher/pull/1873
See: https://github.com/resin-io/etcher/pull/1843


jviotti:
I tested this on macOS and GNU/Linux, both locally and packaged, and I went through all the flashing/elevation related items in the manual testing document on macOS.

.eslintrc.yml
```diff
@@ -222,8 +222,7 @@ rules:
     - error
     - max: 1
   multiline-ternary:
-    - error
-    - never
+    - off
```
jviotti: If you agree, of course :) Ternary operations are very handy, but the lines can get long very quickly,

jhermsmeier: Wait, does this enable the use of multiline ternaries? Hell yes! 👍

lib/gui/modules/child-writer.js
```diff
@@ -0,0 +1,130 @@
...
+
+    // These come as strings containing 0 or 1. We need to convert
+    // them to numbers first, and then to booleans, otherwise something
+    // like `Boolean('0')` will be `true`
+    const unmountOnSuccess = Boolean(Number(process.env.OPTION_UNMOUNT))
+    const validateWriteOnSuccess = Boolean(Number(process.env.OPTION_VALIDATE))
+    log(`Umount on success: ${unmountOnSuccess}`)
+    log(`Validate on success: ${validateWriteOnSuccess}`)
+
+    writer.writeImage(process.env.OPTION_IMAGE, process.env.OPTION_DEVICE, {
```
Shou: Should we pass this over IPC instead of through env vars? If we want to add persistence and stopping it might make more sense.

lib/start.js
```diff
@@ -27,7 +27,7 @@
 // an older equivalent of `ELECTRON_RUN_AS_NODE` that still gets set when
 // using `child_process.fork()`.
 if (process.env.ELECTRON_RUN_AS_NODE || process.env.ATOM_SHELL_INTERNAL_RUN_AS_NODE) {
-  require('./cli/etcher')
+  require('./gui/modules/child-writer')
```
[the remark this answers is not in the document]
jviotti: You call its entry point directly (ie `node lib/cli/etcher.js`)

lib/gui/modules/child-writer.js
jviotti: Yeah, that's next step (otherwise this PR will keep growing). I want to get most of this merged, and will send and follow up PR that elevates on long-living process at startup that does scanning and flashing using IPC commands.

jviotti:
Ping @jhermsmeier @Shou @lurch

jviotti:
I updated the PR. It should work with these additional PRs:

- https://github.com/resin-io/etcher/pull/1949
- https://github.com/resin-io/etcher/pull/1950

On Wed, Jan 03, 2018 at 09:26:53AM -0800, Jonas Hermsmeier wrote:
> jhermsmeier requested changes on this pull request.
> 
> @jviotti, this is not sadly working for me on Windows. Looks like a combination of an issue with the elevation & the child exiting unexpectedly (prob. due to lack of permissions)
> 
> 
> 
> -- 
> https://github.com/resin-io/etcher/pull/1910#pullrequestreview-86436069

-- 
Juan Cruz Viotti
Software Engineer

jviotti:
Rebased! It should work fine now.

lib/gui/modules/image-writer.js
```diff
@@ -48,14 +76,134 @@ const analytics = require('../modules/analytics')
...
+        }
+      }).then((results) => {
+        flashResults.cancelled = results.cancelled
+        console.log('Flash results', flashResults)
+
+        // This likely means the child died halfway through
+        if (!flashResults.cancelled && !flashResults.bytesWritten) {
+          throw errors.createUserError({
+            title: 'The writer process ended unexpectedly',
+            description: 'Please try again, and contact the Etcher team if the problem persists',
```
lurch: Sorry if this is a stupid question, but why does the identical string appear both here and in `lib/shared/messages.js` ?

lurch:
I guess `docs/ARCHITECTURE.md` now needs a small tweak? @jviotti

PR #1917, opened 2017-12-12
feat(GUI): move drive selector warning to flash step

Shou:
We move the drive selector warning to the flash step, and concatenate
warning messages when more than one needs to be displayed at once.

Change-Type: patch
Changelog-Entry: Move the drive selector warning dialog to the flash step.

### DIFF
--- a/lib/gui/app/pages/main/controllers/flash.js
+++ b/lib/gui/app/pages/main/controllers/flash.js
@@ -16,6 +16,7 @@
 
 'use strict'
 
+const _ = require('lodash')
 const messages = require('../../../../../shared/messages')
 const flashState = require('../../../../../shared/models/flash-state')
 const driveScanner = require('../../../modules/drive-scanner')
@@ -25,19 +26,88 @@ const exceptionReporter = require('../../../modules/exception-reporter')
 const imageWriter = require('../../../modules/image-writer')
 const path = require('path')
 const store = require('../../../../../shared/store')
+const constraints = require('../../../../../shared/drive-constraints')
+const availableDrives = require('../../../../../shared/models/available-drives')
 
 module.exports = function (
+  $q,
   $state,
   $timeout,
-  FlashErrorModalService
+  FlashErrorModalService,
+  WarningModalService,
+  DriveSelectorService
 ) {
+  /**
+   * @summary Spawn a confirmation warning modal
+   * @function
+   * @public
+   *
+   * @param {Array<String>} warningMessages - warning messages
+   * @returns {Promise} warning modal promise
+   *
+   * @example
+   * confirmationWarningModal([ 'Hello, World!' ])
+   */
+  const confirmationWarningModal = (warningMessages) => {
+    return WarningModalService.display({
+      confirmationLabel: 'Continue',
+      rejectionLabel: 'Change',
+      description: [
+        warningMessages.join('\n\n'),
+        'Are you sure you want to continue?'
+      ].join(' ')
+    })
+  }
+
+  /**
+   * @summary Display warning tailored to the warning of the current drives-image pair
+   * @function
+   * @public
+   *
+   * @param {Array<Object>} drives - list of drive objects
+   * @param {Object} image - image object
+   * @returns {Promise}
+   *
+   * @example
+   * displayTailoredWarning(drives, image).then(() => {
+   *   console.log('Continue pressed')
+   * }).catch(() => {
+   *   console.log('Change pressed')
+   * })
+   */
+  const displayTailoredWarning = (drives, image) => {
+    const warningMessages = _.reduce(drives, (accumMessages, drive) => {
+      if (constraints.isDriveSizeLarge(drive)) {
+        return accumMessages.concat(messages.warning.largeDriveSize(drive))
+      } else if (!constraints.isDriveSizeRecommended(drive, image)) {
+        return accumMessages.concat(messages.warning.unrecommendedDriveSize(image, drive))
+      }
+
+      return accumMessages
+    }, [])
+
+    if (!warningMessages.length) {
+      // TODO(Shou): we should consider adding the same warning dialog for system drives and remove unsafe mode
+      return $q.resolve()
+    }
+
+    return confirmationWarningModal(warningMessages).then((value) => {
+      if (!value) {
+        DriveSelectorService.open()
+        return $q.reject()
+      }
+
+      return $q.resolve()
+    })
+  }
+
   /**
    * @summary Flash image to drives
    * @function
    * @public
    *
    * @param {Object} image - image
-   * @param {Array<Object>} drives - drives
+   * @param {Array<String>} devices - list of drive device paths
    *
    * @example
    * FlashController.flashImageToDrive({
@@ -49,22 +119,23 @@ module.exports = function (
    *       value: 1000000000
    *     }
    *   }
-   * }, {
-   *   device: '/dev/disk2',
-   *   description: 'Foo',
-   *   size: 99999,
-   *   mountpoint: '/mnt/foo',
-   *   system: false
-   * })
+   * }, [
+   *   '/dev/disk2',
+   *   '/dev/disk5'
+   * ])
    */
-  this.flashImageToDrive = (image, drives) => {
+  this.flashImageToDrive = (image, devices) => {
     if (flashState.isFlashing()) {
       return
     }
 
-    // Stop scanning drives when flashing
-    // otherwise Windows throws EPERM
-    driveScanner.stop()
+    const drives = _.filter(availableDrives.getDrives(), (drive) => {
+      return _.includes(devices, drive.device)
+    })
+
+    const hasDangerStatus = constraints.hasListDriveImageCompatibilityStatus(drives, image)
+
+    const userConfirm = hasDangerStatus ? _.partial(displayTailoredWarning, drives, image) : $q.resolve
 
     // Trigger Angular digests along with store updates, as the flash state
     // updates. Without this there is essentially no progress to watch.
@@ -72,7 +143,13 @@ module.exports = function (
 
     const iconPath = '../../../assets/icon.png'
 
-    imageWriter.flash(image.path, drives).then(() => {
+    userConfirm().then(() => {
+      // Stop scanning drives when flashing
+      // otherwise Windows throws EPERM
+      driveScanner.stop()
+
+      return imageWriter.flash(image.path, devices)
+    }).then(() => {
       if (!flashState.wasLastFlashCancelled()) {
         notification.send('Success!', {
           body: messages.info.flashComplete(path.basename(image.path), drives),
@@ -81,6 +158,11 @@ module.exports = function (
         $state.go('success')
       }
     }).catch((error) => {
+      // When flashing is cancelled before starting above there is no error
+      if (!error) {
+        return
+      }
+
       notification.send('Oops! Looks like the flash failed.', {
         body: messages.error.flashFailure(path.basename(image.path), drives),
         icon: iconPath
--- a/lib/gui/app/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/app/components/drive-selector/controllers/drive-selector.js
@@ -18,7 +18,6 @@
 
 const angular = require('angular')
 const _ = require('lodash')
-const messages = require('../../../../../shared/messages')
 const constraints = require('../../../../../shared/drive-constraints')
 const analytics = require('../../../modules/analytics')
 const availableDrives = require('../../../../../shared/models/available-drives')
@@ -27,8 +26,7 @@ const utils = require('../../../../../shared/utils')
 
 module.exports = function (
   $q,
-  $uibModalInstance,
-  WarningModalService
+  $uibModalInstance
 ) {
   /**
    * @summary The drive selector state
@@ -72,28 +70,7 @@ module.exports = function (
    *    });
    */
   const shouldChangeDriveSelectionState = (drive) => {
-    if (!constraints.isDriveValid(drive, selectionState.getImage())) {
-      return $q.resolve(false)
-    }
-
-    if (!constraints.isDriveSizeRecommended(drive, selectionState.getImage())) {
-      return WarningModalService.display({
-        confirmationLabel: 'Yes, continue',
-        description: [
-          messages.warning.unrecommendedDriveSize(selectionState.getImage(), drive),
-          'Are you sure you want to continue?'
-        ].join(' ')
-      })
-    }
-
-    if (constraints.isDriveSizeLarge(drive)) {
-      return WarningModalService.display({
-        confirmationLabel: 'Yes, continue',
-        description: messages.warning.largeDriveSize(drive)
-      })
-    }
-
-    return $q.resolve(true)
+    return $q.resolve(constraints.isDriveValid(drive, selectionState.getImage()))
   }
 
   /**
--- a/lib/gui/app/components/drive-selector/drive-selector.js
+++ b/lib/gui/app/components/drive-selector/drive-selector.js
@@ -24,7 +24,6 @@ const angular = require('angular')
 const MODULE_NAME = 'Etcher.Components.DriveSelector'
 const DriveSelector = angular.module(MODULE_NAME, [
   require('../modal/modal'),
-  require('../warning-modal/warning-modal'),
   require('../../utils/byte-size/byte-size')
 ])
 
--- a/lib/gui/app/components/warning-modal/styles/_warning-modal.scss
+++ b/lib/gui/app/components/warning-modal/styles/_warning-modal.scss
@@ -21,3 +21,8 @@
 .modal-warning-modal .modal-title .glyphicon {
   color: $palette-theme-danger-background;
 }
+
+.modal-warning-modal .modal-body {
+  max-height: 200px;
+  overflow-y: auto;
+}
--- a/lib/gui/app/pages/main/main.js
+++ b/lib/gui/app/pages/main/main.js
@@ -37,6 +37,7 @@ const MainPage = angular.module(MODULE_NAME, [
   require('../../components/tooltip-modal/tooltip-modal'),
   require('../../components/flash-error-modal/flash-error-modal'),
   require('../../components/progress-button/progress-button'),
+  require('../../components/warning-modal/warning-modal'),
 
   require('../../os/open-external/open-external'),
   require('../../os/dropzone/dropzone'),
--- a/lib/gui/css/main.css
+++ b/lib/gui/css/main.css
@@ -6483,6 +6483,10 @@ svg-icon {
 .modal-warning-modal .modal-title .glyphicon, .modal-warning-modal .modal-title .tick {
   color: #d9534f; }
 
+.modal-warning-modal .modal-body {
+  max-height: 200px;
+  overflow-y: auto; }
+
 /*
  * Copyright 2016 resin.io
  *


lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -56,6 +57,20 @@ module.exports = function (
    */
   this.drives = availableDrives
 
+  /**
+   * @summary
```
lurch: Should this have more complete docstrings?

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -75,20 +90,38 @@ module.exports = function (
       return $q.resolve(false)
     }
 
-    if (constraints.isDriveSizeRecommended(drive, selectionState.getImage())) {
-      return $q.resolve(true)
+    const isLargeDrive = constraints.isDriveSizeLarge(drive)
+    const isRecommendedDrive = constraints.isDriveSizeRecommended(drive, selectionState.getImage())
```
lurch: Minor nit, but could we rename this variable to `isRecommendedSizeDrive` ?
To me, a variable named `isRecommendedDrive` implies that this drive is recommended above all other drives :)

Shou:
Screenshot of concatenated warnings

image

There's a lot of space above when the text is long enough, looking into fixing that.

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -75,20 +90,38 @@ module.exports = function (
...
+        [
+          !isRecommendedDrive,
+          messages.warning.unrecommendedDriveSize({
+            drive,
+            image: selectionState.getImage()
+          })
+        ]
+      ]
+
+      const concatenatedMessage = _.reduce(messageTuples, (accumulatedMessage, messageTuple) => {
```
lurch: Do we really need to concatenate the messages? Seems _extremely_ unlikely that a drive will be both 'large' and 'smaller than the recommended size' ;-)

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -75,20 +90,38 @@ module.exports = function (
...
+          if (messageTuple[0]) {
+            return `${accumulatedMessage}\n\n${messageTuple[1]}`
+          } else {
+            return accumulatedMessage
+          }
+        }, '')
```
lurch: Why does this need to use `reduce` rather than just a straightforward `join`?

lib/shared/store.js
```diff
@@ -144,6 +144,9 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
           constraints.isDriveValid(drive, image),
           constraints.isDriveSizeRecommended(drive, image),
 
+          // We don't want to auto-select large drives
+          !constraints.isDriveSizeLarge(drive),
+
```
lurch: Should the unit-tests be added to? ;-)

lib/shared/messages.js
```diff
@@ -67,6 +67,11 @@ module.exports = {
       'It looks like this is not a bootable image.\n\n',
       'The image does not appear to contain a partition table,',
       'and might not be recognized or bootable by your device.'
+    ].join(' ')),
+
+    largeDriveSize: _.template([
+      'It looks like you\'re trying to flash a large drive of size <%= humanDriveSize %>,',
+      'this could potentially be a backup drive containing important data.',
```
lurch: Hmmm, not sure if we specifically want to mention "backup drive" here? Although I like the "important data" phrasing.
How about something like:
"It looks like you're trying to flash a large drive with size 120 GB, are you sure this disk doesn't contain important data? Etcher will overwrite everything."
?
I suspect @alexandrosm will want some input here :)

lib/shared/drive-constraints.js
```diff
@@ -278,6 +278,23 @@ exports.isDriveSizeRecommended = (drive, image) => {
   return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'recommendedDriveSize' ], UNKNOWN_SIZE)
 }
 
+/**
+ * @summary 64GB
+ */
+const LARGE_DRIVE_SIZE = 64000000000
```
lurch: I wonder if this should be base-2 units i.e. 68719476736 ?

jhermsmeier: Storage devices usually have sizes in base-10, so base-2 wouldn't trigger on a 64GB SD-card, for example. Guess that depends on when we want that to happen.

lib/shared/messages.js
jviotti: We can simplify this to:

> Because of its size, this drive doesn't look like a removable drive. Are you sure you want to flash it?

jviotti: I don't think we need to mention that 64 GB is the limit that triggers the heuristic

jviotti:
I think the "recommended size" part of the message can go away. We should try to be brief rather than throwing to paragraphs at the user.

lib/shared/messages.js
Shou: I think we should mention the size too, and we could also add the drive `description`. It's very helpful to the user in identifying which drive they've chosen just from that message since it covers the drive selector. What about:

> Because of its size of <%= humanDriveSize %>, <%= driveDescription %> doesn't look like a removable drive. Are you sure you want to flash it?

lib/gui/components/drive-selector/controllers/drive-selector.js
Shou: I guess we could `filter` and then `join` but that's two loops, but the reason we use `reduce` here is so we can pick which strings to add to the dialog in one recursion over the list rather than two since JavaScript doesn't do stream-fusion. This doesn't matter anymore now that we're not using this though!

Shou:
Let me know what you think now

image

lib/gui/components/drive-selector/controllers/drive-selector.js
lurch: Ahhh, I'd misunderstood what `messageTuples` actually contains - didn't realise it contained a boolean condition flag in addition to the message itself.

lib/shared/messages.js
lurch: I think the "removable drive" language isn't what we want either though, because somebody was complaining about accidentally overwriting their portable backup drive using Etcher - and in this case it actually _is_ a removable drive.

lib/shared/drive-constraints.js
lurch: Yeah, I mentioned in flowdock that we probably _don't_ want this triggering on 64 GB SD cards, as I suspect they'll continue to become more prevalent as their price continues to drop. In my extensive experience, SD cards are always some-number-of-megabytes smaller than their stated capacity, so no 64GB card has a capacity of 65,536 MB.
For the record, the "64 GB" SD cards I have have exact byte-capacities of 62587404288, 62730010624, 62881005568, 63416827904, 64289243136. (even the biggest one has a capacity of 'only' 61,331 MB)

And I've never seen a 64GB hard drive, they always seem to be a "round" number of units: 80GB, 100GB, 2TB, etc.

lurch: OTOH if we _do_ want this to trigger for 64 GB SD cards too, then it would probably make sense (for similar reasons as above) to set this to the base-2 32GB size, i.e. 34359738368 and this would then also catch 40 GB hard drives but no "32 GB" SD cards.

lurch:
Do you think it's obvious enough that clicking on the grey 'X' in the top-right of the "Attention" dialog is how a (possibly nervous?) user answers "No" to the question being asked of them?  Also looks a bit weird with Etcher displaying two dialogs at once, which I guess @konmouz might want to comment on?

lib/shared/drive-constraints.js
```diff
@@ -288,8 +290,12 @@ const LARGE_DRIVE_SIZE = 64000000000
  * @public
  *
  * @param {Object} drive - drive
- *
  * @returns {Boolean} whether drive size is large
+ *
+ * @example
+ * if (constraints.isDriveSizeLarge(drive)) {
+ *   console.log('Impressive')
```
lurch: :laughing:

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -90,35 +97,18 @@ module.exports = function (
...
-      const concatenatedMessage = _.reduce(messageTuples, (accumulatedMessage, messageTuple) => {
-          if (messageTuple[0]) {
-            return `${accumulatedMessage}\n\n${messageTuple[1]}`
-          } else {
-            return accumulatedMessage
-          }
-        }, '')
+    if (constraints.isDriveSizeLarge(drive)) {
+      return confirmationWarningModal(messages.warning.largeDriveSize({
+        humanDriveSize: units.bytesToClosestUnit(drive.size),
```
lurch: Hmmm, borrowing the 'pattern' from the display-units PR, I wonder if it _might_ make sense for Etcher to automatically add a `humanSize` property to every drivelist object, to prevent this `units.bytesToClosestUnit` being done multiple times in different places?

lurch:
Oh, looking at the code in more detail, it seems that this warning dialog would actually pop up **as soon as** they click on the drive in the drive-selection dialog? Is that the UX that we want? I wonder if it might be better to only display the dialog when they actually click on the "Continue" button on the drive-selection dialog, so that the drive-dialog has now disappeared? And if the "large drive dialog" then had a "No" button, clicking that could unselect the drive that they'd chosen. (IYSWIM)

Sorry if I'm sounding a bit picky, but I don't want us to make a repeat of the accessibility-with-orange-borders UX fiasco ;-)

Shou:
@lurch if having several visible modals open is undesirable we don't necessarily have to close the drive selector to hide it while the other warning modal is open – we could use CSS to hide it in a generic way such that only the most recent modal is shown at a time. I believe this would work:

```scss
.modal[aria-hidden="true"] {
  display: none;
}
```

lib/shared/messages.js
lurch: What about something like:

> <%= driveDescription %> is quite large (<%= humanDriveSize %>). Please check it doesn't contain any important data before Etcher overwrites it.

lurch:
I'm still not sure if displaying the modal automatically as soon as they click on the drive is desirable though? Feels a bit "in your face" to me?
This is obviously a UX question for @konmouz

jviotti:
I was thinking we should show it when the user clicks "Flash!", as a last opportunity to halt the process.

lurch:
> as a last opportunity to halt the process.

If we're doing that, then IMHO we _definitely_ need a far more obvious "No" button ;-)

lurch:
And BTW GitHub is now annoyingly auto-hiding the conversation (on `lib/shared/messages.js`) about what the warning message actually _is_ - I still think it needs some tweaks.

Shou:
This is definitely a good change for when multi-selection and writes arrive as well. If the user cancels the warning dialog we should probably re-open the drive selector as that's the obvious step anyway?

konmouz:
I agree with @lurch, it would be better to have 2 CTA's, one for 'continue' and one for 'cancel'. We don't need the 'x' button in this attention modal. Also, WRT the warning on large drive size selection, we can follow the flow of a system drive. The suspicious drive can get a label, and then when selected the 'continue' button becomes red and also gets an alert just above it. Finally, you should get another attention/confirmation modal after clicking the 'flash' CTA.

Shou:
So we want to move the warnings (not recommended + large drive) to appear when clicking flash in this PR?

jviotti:
I think we should try to open the least amount of modals as possible, and treat this as an exceptional case to protect the user from OSes that think that external HDDs are not fixed drives.

> If the user cancels the warning dialog we should probably re-open the drive selector as that's the obvious step anyway, i.e. they want to change the selected drive(s)?

Sounds reasonable. So the plan is the following:

- User selects a drive larger than 64 GB
- User clicks Flash
- Since the drive is larger than 64 GB, we show the warning modal
- If the user agrees, then everything carries on
- If the user doesn't agree, then the warning modal closes and the drive selector modal opens instead

konmouz:
So this is the behaviour when selecting a system drive:

screen shot 2017-12-14 at 10 43 26
screen shot 2017-12-14 at 10 43 35
screen shot 2017-12-14 at 10 43 54

lurch:
@konmouz There's two separate issues here. One is system drives (i.e. internal fixed drives, which we only display when 'unsafe mode' is enabled) and the other is very large (i.e. over 64GB) external drives.
Based on @konmouz 's comments above, do we also want a `LARGE DRIVE` warning-label displayed in the drive-selector window?

konmouz:
@lurch, I believe so. `LARGE DRIVE`s should follow the `SYSTEM DRIVE`s behaviour.

Shou:
Let me know what you think now

image
image
image

lurch:
Looks good, but I still think the message needs tweaking. In some earlier comments (which GitHub is now auto-hiding), I said:

I think the "removable drive" language isn't what we want either though, because somebody was complaining about accidentally overwriting their portable backup drive using Etcher - and in this case it actually _is_ a removable drive.

What about something like:

> <%= driveDescription %> is very large (<%= humanDriveSize %>). Please check it doesn't contain any important data before Etcher overwrites it.

lurch:
Looks like you need to rebase / update your shrinkwrap file too.

lib/shared/drive-constraints.js
```diff
@@ -428,3 +465,7 @@ exports.getDriveImageCompatibilityStatuses = (drive, image) => {
 
   return statusList
 }
+
+exports.hasDriveImageCompatibilityStatus = (drive, image) => {
```
jviotti: Missing JSDoc?

alexandrosm:
How about "Drive <...> is unusually large for an SD card or USB stick. Are
you sure you want to continue?"

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

On Fri, Dec 15, 2017 at 12:25 PM, Juan Cruz Viotti <notifications@github.com

lib/gui/pages/main/controllers/flash.js
```diff
@@ -16,19 +16,98 @@
...
+   * displayTailoredWarning(drive, image).then(() => {
+   *   console.log('Continue pressed')
+   * }).catch(() => {
+   *   console.log('Change pressed')
+   * })
+   */
+  const displayTailoredWarning = (drive, image) => {
+    const messageKeys = {
+      largeDriveSize: constraints.isDriveSizeLarge(drive),
+      unrecommendedDriveSize: !constraints.isDriveSizeRecommended(drive, image)
```
lurch: Hmm, I'm not sure if unrecommendedDriveSize is a serious-enough-problem to warrant displaying an extra warning modal to the user?

lib/gui/pages/main/controllers/flash.js
```diff
@@ -16,19 +16,98 @@
...
+        humanDriveSize: units.bytesToClosestUnit(drive.size),
+        drive
+      },
+      unrecommendedDriveSize: {
+        image,
+        drive
+      }
+    }
+    const messageKey = _.findKey(messageKeys, _.identity)
+    const message = messages.warning[messageKey](messageArgs[messageKey])
```
lurch: IMHO this line looks a little funny, with 4 different variables with `message` in the name, and one of the variables being used in two different places to index into two different objects.

I _think_ I see what you're doing with using `const` and lodash all over the place, but I wonder if something like this might be a little bit more readable?
```js
let message = null
if (constraints.isDriveSizeLarge(drive)) {
    message = messages.warning[largeDriveSize]({humanDriveSize: units.bytesToClosestUnit(drive.size), drive})
} else if (!constraints.isDriveSizeRecommended(drive, image)) {
    message = messages.warning[unrecommendedDriveSize]({image, drive})
}
```
?

Or if you really wanted to keep the const-ness I guess you _could_ do:
```js
const message = constraints.isDriveSizeLarge(drive)
    ? messages.warning[largeDriveSize]({humanDriveSize: units.bytesToClosestUnit(drive.size), drive})
    : (
        !constraints.isDriveSizeRecommended(drive, image)
        ? messages.warning[unrecommendedDriveSize]({image, drive})
        : null
    )
```
but that obviously starts heading into the unreadable-code category ;-)

Or perhaps you could do something like:
```js
const messageValues = [
    [constraints.isDriveSizeLarge(drive), messages.warning[largeDriveSize]({humanDriveSize: units.bytesToClosestUnit(drive.size), drive})],
    [!constraints.isDriveSizeRecommended(drive, image), messages.warning[unrecommendedDriveSize]({image, drive})]
}
// and then use `_.find` etc.
```
which is more similar to the approach you're already using, but gets rid of the separate `messageArgs` variable?

\*shrug\* I dunno really, @jviotti can probably offer better advice when it comes to Javascript best-practices ;-)  Feel free to ignore me.

lib/shared/drive-constraints.js
```diff
@@ -405,6 +437,11 @@ exports.getDriveImageCompatibilityStatuses = (drive, image) => {
       type: exports.COMPATIBILITY_STATUS_TYPES.ERROR,
       message: exports.COMPATIBILITY_STATUS_MESSAGES.LOCKED
     })
+  } else if (exports.isDriveSizeLarge(drive)) {
```
lurch: Should the unit-tests that deal with `getDriveImageCompatibilityStatuses` be updated too?

lib/shared/drive-constraints.js
```diff
@@ -405,6 +437,11 @@ exports.getDriveImageCompatibilityStatuses = (drive, image) => {
...
+    statusList.push({
+      type: exports.COMPATIBILITY_STATUS_TYPES.WARNING,
+      message: exports.COMPATIBILITY_STATUS_MESSAGES.LARGE_DRIVE
+    })
   } else if (!_.isNil(drive) && !_.isNil(drive.size) && !exports.isDriveLargeEnough(drive, image)) {
```
lurch: Hmmm, should `isDriveSizeLarge` (which is a warning) be checked for _after_ `isDriveLargeEnough` (which is an error) ?

Hmmm number 2, if we have a drive which is both 'large' and 'system', I wonder if we should display both warning labels?

lib/gui/pages/main/controllers/flash.js
Shou: We could remove it but it's the current behaviour, so if we want to remove it I think we should open a separate issue.

lib/gui/pages/main/controllers/flash.js
Shou: I noticed a bug with my current implementation here and now I think the best option is to just go for the if-then-else

lib/shared/drive-constraints.js
Shou: Done

lib/shared/drive-constraints.js
Shou: That's true, we should considering we do the same with unrecommended drive/image size.

Shou:
I've rebased and also added [some changes][1]. Let me know what you think now. 

[1]: https://github.com/resin-io/etcher/compare/5ea1c1eac11cb8521af949247b56bafdbaaca1c9...warn-large-drives

jhermsmeier:
Conflicts need resolving

jhermsmeier:
@Shou can you add an updated screenshot to the PR description?

Shou:
@jhermsmeier I've split off part of this PR into https://github.com/resin-io/etcher/pull/2014 and will probably split off some more soon

Shou:
Split off into https://github.com/resin-io/etcher/pull/2045 – this PR will be about moving the warning dialog to the flash step instead, because they don't depend on each other so it'll be faster to get the PRs through independently/concurrently.

Shou:
Current screenshot

image

lib/gui/app/pages/main/controllers/flash.js
```diff
@@ -81,6 +158,11 @@ module.exports = function (
         $state.go('success')
       }
     }).catch((error) => {
+      // When flashing is cancelled before starting above there is no error
+      if (!error) {
```
jhermsmeier: I'm going to add this to the list of problems one wouldn't have without Promises ;P

lib/gui/app/pages/main/controllers/flash.js
```diff
@@ -25,19 +26,88 @@ const exceptionReporter = require('../../../modules/exception-reporter')
...
+  }
+
   /**
    * @summary Flash image to drives
    * @function
    * @public
    *
    * @param {Object} image - image
-   * @param {Array<Object>} drives - drives
+   * @param {Array<Object>} devices - drive devices
```
jhermsmeier: The type doc here needs updating to `Array<String>` too I think

lib/gui/app/pages/main/controllers/flash.js
```diff
@@ -25,19 +26,88 @@ const exceptionReporter = require('../../../modules/exception-reporter')
...
+      ].join(' ')
+    })
+  }
+
+  /**
+   * @summary Display warning tailored to the warning of the current drives-image pair
+   * @function
+   * @public
+   *
+   * @param {Array<Object>} drives - list of drive objects
```
jhermsmeier: The type doc here needs updating to `Array<String>` I believe

Shou: It requires `Object` in order for the `constraints` predicates to work

lib/gui/app/pages/main/controllers/flash.js
Shou: Done that now

lib/gui/app/pages/main/controllers/flash.js
jhermsmeier: Oh, right - missed that

PR #1918, opened 2017-12-12
fix(writer): Add EBUSY to transient errors on Linux

jhermsmeier:
Change-Type: patch

### DIFF
--- a/lib/writer/error-types.js
+++ b/lib/writer/error-types.js
@@ -24,7 +24,7 @@ module.exports = {
    * Errors considered potentially temporary are:
    *   - Mac OS: ENXIO, EBUSY
    *   - Windows: ENOENT, UNKNOWN
-   *   - Linux: EIO
+   *   - Linux: EIO, EBUSY
    * @private
    * @param {Error} error - Error
    * @returns {Boolean}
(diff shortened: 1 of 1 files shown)


lurch:
Are we sure this is the right fix? The previous implementation in `etcher-image-write` didn't seem to ignore EBUSY errors on Linux https://github.com/resin-io-modules/etcher-image-write/blob/master/tests/filesystem.spec.js#L399 ?

lib/writer/error-types.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
     if (process.platform === 'darwin') {
       return error.code === 'ENXIO' || error.code === 'EBUSY'
     } else if (process.platform === 'linux') {
-      return error.code === 'EIO'
+      return error.code === 'EIO' || error.code === 'EBUSY'
```
lurch: If this **is** what we want to do, then the comment on line 27 needs updating ;-)

jviotti:
@jhermsmeier Did you study some of the Sentry instances of this error? Were you able to reproduce? Do we know for sure EBUSY means that we lost access to the drive?

lib/writer/error-types.js
lurch: bump ;0

jhermsmeier: Fixed

PR #1921, opened 2017-12-13
feat(GUI): link to drivers when clicking a driverless usbboot device

jviotti:
Step 2 until we support installing the drivers from within Etcher. This
also introduces an "Open drive link" Mixpanel event.

See: https://www.raspberrypi.org/documentation/hardware/computemodule/cm-emmc-flashing.md
See: https://github.com/resin-io/etcher/pull/1892
Change-Type: patch
Changelog-Entry: Download usbboot drivers installer when clicking a driverless usbboot device on Windows.

### DIFF
--- a/lib/gui/app/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/app/components/drive-selector/controllers/drive-selector.js
@@ -18,6 +18,7 @@
 
 const angular = require('angular')
 const _ = require('lodash')
+const Bluebird = require('bluebird')
 const constraints = require('../../../../../shared/drive-constraints')
 const analytics = require('../../../modules/analytics')
 const availableDrives = require('../../../../../shared/models/available-drives')
@@ -26,7 +27,9 @@ const utils = require('../../../../../shared/utils')
 
 module.exports = function (
   $q,
-  $uibModalInstance
+  $uibModalInstance,
+  WarningModalService,
+  OSOpenExternalService
 ) {
   /**
    * @summary The drive selector state
@@ -97,7 +100,23 @@ module.exports = function (
         })
 
         selectionState.toggleDrive(drive.device)
+      } else if (drive.link) {
+        analytics.logEvent('Open drive link modal', {
+          url: drive.link
+        })
+
+        const message = drive.message || `Etcher will open ${drive.link} in your browser`
+        return WarningModalService.display({
+          confirmationLabel: 'Yes, continue',
+          description: `${message}. Are you sure you want to continue?`
+        }).then((answer) => {
+          if (answer) {
+            OSOpenExternalService.open(drive.link)
+          }
+        })
       }
+
+      return Bluebird.resolve()
     })
   }
 
--- a/lib/gui/app/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/app/components/drive-selector/controllers/drive-selector.js
@@ -28,7 +28,7 @@ const utils = require('../../../../../shared/utils')
 module.exports = function (
   $q,
   $uibModalInstance,
-  WarningModalService,
+  ConfirmModalService,
   OSOpenExternalService
 ) {
   /**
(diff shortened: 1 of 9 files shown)


jviotti:
If you want to test this, you can mock a driverless usbboot device on any platform with the following patch:

```diff
diff --git a/lib/shared/sdk/usbboot/usb.js b/lib/shared/sdk/usbboot/usb.js
index d0e2d98f..56849527 100644
--- a/lib/shared/sdk/usbboot/usb.js
+++ b/lib/shared/sdk/usbboot/usb.js
@@ -72,6 +72,14 @@ exports.listDevices = () => {
     return device
   })

+  devices.push({
+    accessible: false,
+    deviceDescriptor: {
+      idVendor: 0x0a5c,
+      idProduct: 0x2764
+    }
+  })
+
   // Include driverless devices into the list of USB devices.
   if (process.platform === 'win32') {
     const winusbDriverGenerator = require('winusb-driver-generator')
```

lurch:
Could you include a screenshot of what this looks like please?

lib/gui/components/drive-selector/controllers/drive-selector.js
```diff
@@ -27,7 +27,8 @@ const selectionState = require('../../../../shared/models/selection-state')
 module.exports = function (
   $q,
   $uibModalInstance,
-  WarningModalService
+  WarningModalService,
```
Shou: Should we remove this? Or are there plans for a warning modal?

jviotti: There are other functions in this controller that use this service.

jviotti:
> Could you include a screenshot of what this looks like please?

Nothing really to screenshot. When you click on the device that said "Missing drivers", then your browser opens up.

jviotti:
> Looks good otherwise, but maybe we should use a modal with the link instead of downloading it on click? Might surprise the user e.g. they can think it's a virus.

Hm, maybe link to a landing page that links to the drivers instead?

lurch:
Perhaps we could underline the link and display it in blue, otherwise I guess there's no CTA that it's something that can be clicked on?

lurch:
> Hm, maybe link to a landing page that links to the drivers instead?

Or I guess we could display a dialog which says something like "We've detected that you've plugged in a Compute Module but you don't have the necessary driver installed. Click yes below to download the driver". Just so that the user gets a chance to know what's going on before their browser starts downloading an unknown executable file (while also giving them the chance to say No if they _don't_ want to download the drivers at this point).

jviotti:
Alright. I updated the PR so the "Missing drivers" entry is not disabled (so the user knows he can click it), plus added a modal before opening the link in the user's browser:

screen shot 2018-01-04 at 2 55 16 pm
screen shot 2018-01-04 at 2 55 19 pm

alexandrosm:
Hey @jviotti -- this UX is still really scary, and users don't really know
why to click. Can you bring this to a product call pls?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

konmouz:
@jviotti, I was thinking if it makes sense to have a download icon or even icon+text CTA so the user can download the relevant drivers. Also red feels a bit too much, we can have a smaller + message in orange.

jhermsmeier:
Rebased, and updated according to @konmouz' feedback;

- Added friendlier modal, leading to the driver download page
- Added CTA link next to "Missing Drivers" note on CM in drive list

screen shot 2018-05-02 at 23 24 31

screen shot 2018-05-02 at 23 24 35

lib/gui/app/components/drive-selector/controllers/drive-selector.js
```diff
@@ -98,9 +101,51 @@ module.exports = function (
...
+   *
+   * @example
+   * DriveSelectorController.installMissingDrivers({
+   *   linkTitle: 'Go to example.com',
+   *   linkMessage: 'Examples are great, right?',
+   *   linkCTA: 'Call To Action',
+   *   link: 'https://example.com'
+   * });
+   */
+  this.installMissingDrivers = (drive) => {
```
lurch: Is `drive` really a suitable name for this parameter??

lib/gui/app/components/drive-selector/controllers/drive-selector.js
```diff
@@ -98,9 +101,51 @@ module.exports = function (
...
+  /**
+   * @summary Prompt the user to install missing usbboot drivers
+   * @function
+   * @public
+   *
+   * @param {Object} drive - drive
+   * @returns {Promise} - resolved promise
+   *
+   * @example
+   * DriveSelectorController.installMissingDrivers({
```
lurch: Perhaps `downloadMissingDrivers` is a better name?

lib/gui/app/components/drive-selector/controllers/drive-selector.js
```diff
@@ -98,9 +101,51 @@ module.exports = function (
...
+   *   linkTitle: 'Go to example.com',
+   *   linkMessage: 'Examples are great, right?',
+   *   linkCTA: 'Call To Action',
```
lurch: Strictly speaking, `linkCTA` isn't used in this function.

lurch:
Looks like a nice UX, but presumably this "Install" link is only shown in the drive-selector when Etcher is running on Windows? (given that the dialog mentions "Window installer")

lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
```diff
@@ -23,7 +23,8 @@ <h4 class="list-group-item-heading">{{ drive.description }}
             <span class="word-keep"
               ng-show="drive.size"> - {{ drive.size | closestUnit }}</span>
           </h4>
-          <p class="list-group-item-text">{{ drive.displayName }}</p>
+          <p class="list-group-item-text" ng-if="!drive.link">{{ drive.displayName }}</p>
+          <p class="list-group-item-text" ng-if="drive.link">{{ drive.displayName }} - <b><a ng-click="modal.installMissingDrivers(drive)">{{ drive.linkCTA }}</a></b></p>
```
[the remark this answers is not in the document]
Shou: It won't be shown on Linux so we should be fine, but ideally it'd be hidden if that env var is set

PR #1924, opened 2017-12-15
fix(package): Fix noodled merge of shrinkwrap file

jhermsmeier:
Out of order squash merging resulted in some from-lines
not being removed in the npm-shrinkwrap.

Change-Type: patch


lurch:
Yay, all green on `master` again :-D

PR #1929, opened 2017-12-18
chore: get rid of npx

jviotti:
We can live without it with a simple `PATH` trick at the top of the
Makefile (thanks @lurch!).

Change-Type: patch


lurch:
Looks like you need to rebase before Appveyor can proceed?

lurch:
Apart from the rebase, looks like all the CI tests passed! :tada: :tada:

Makefile
```diff
@@ -3,6 +3,7 @@
 # ---------------------------------------------------------------------
 
 NPX = ./node_modules/.bin/npx
```
lurch: This line should be removed now ;-P

jviotti:
All set!

jviotti:
@lurch Sadly this doesn't seem to be working :/ The CI services pass because they add `./node_modules/.bin` to the `PATH` for all builds out of the box. This PR fails locally.

jviotti:
@lurch I found a workaround; see my recent changes. Updates to `PATH` persist if you set `SHELL` (?)

lurch:
> Updates to PATH persist if you set SHELL (?)

Yeah, I'd also seen comments regarding this inconsistent `make` behaviour on OSX on stackoverflow when I was trying this initially, which is why I was so surprised when it worked the first time on all the CI builds ;-)

Unfortunately the latest changes are causing Spectron to fail on Appveyor :-/

jviotti:
Looks like it was a quoting issue. Appveyor should be green now.

On Tue, Dec 19, 2017 at 12:15:31AM +0000, Andrew Scheller wrote:
> > Updates to PATH persist if you set SHELL (?)
> 
> Yeah, I'd also seen comments regarding this inconsistent `make` behaviour on OSX on stackoverflow when I was trying this initially, which is why I was so surprised when it worked the first time on all the CI builds ;-)
> 
> Unfortunately the latest changes are causing Spectron to fail on Appveyor :-/
> 
> -- 
> https://github.com/resin-io/etcher/pull/1929#issuecomment-352599170

-- 
Juan Cruz Viotti
Software Engineer

lurch:
Fantastic :-)

PR #1938, opened 2017-12-27
chore: declare Concourse dependencies in package.json

jviotti:
Resin Concourse will make sure to provide these during build time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,15 @@
     "7zip-bin-win",
     "7zip-bin-linux"
   ],
+  "concourse": {
+    "dependencies": {
+      "linux": [
+        "libudev-dev",
+        "libusb-1.0-0-dev",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -35,6 +35,13 @@
     "7zip-bin-win",
     "7zip-bin-linux"
   ],
+  "concourse": {
+    "dependencies": [
+      "libudev-dev",
+      "libusb-1.0-0-dev",
+      "libyaml-dev"
+    ]
```
jhermsmeier: How'd this handle different dependencies on different platforms?

jviotti: Good point. Divided the object per operating system.

jviotti:
@Shou @jhermsmeier Please rebase any branch so Concourse is happy again.

PR #1939, opened 2017-12-28
chore: move updates disabling logic out of the Makefile

jviotti:
The Makefile current has logic to disable updates when building deb or
rpm packages. To make the Concourse pipeline transition easier, the
logic that disables updates on deb and rpm has been moved to the main
application code.

Change-Type: patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,17 +168,6 @@ endif
 ELECTRON_BUILDER_OPTIONS = --$(TARGET_ARCH_ELECTRON_BUILDER)
 
 # ---------------------------------------------------------------------
-# Updates
-# ---------------------------------------------------------------------
-
-DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS = --extraMetadata.analytics.updates.enabled=false
-
-ifdef DISABLE_UPDATES
-$(warning Update notification dialog has been disabled (DISABLE_UPDATES is set))
-ELECTRON_BUILDER_OPTIONS += $(DISABLE_UPDATES_ELECTRON_BUILDER_OPTIONS)
-endif
-
-# ---------------------------------------------------------------------
 # Analytics
 # ---------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


lurch:
Hmmm, why is this necessary? This change also means you'll no longer be able to do e.g. https://github.com/resin-io/etcher/pull/1501#issuecomment-307154645 by setting the `DISABLE_UPDATES` variable when running `make`.

jviotti:
@lurch The packaging code (most of the Makefile) will be moved to the Resin Concourse Electron pipeline, which means that we won't be able to "pass an environment variable" to disable this anymore, therefore the application need to be able to sort this out by itself.

lurch:
Hmmm, so is the Makefile going to be deprecated, or will it still be used for local (non-Concourse) builds? (which is the only time that the _optional_ `DISABLE_UPDATES` variable got used anyway)
The way the Makefile works currently, the deb and rpm builds work without relying on any environmental variables (by passing the `--extraMetadata.analytics.updates.enabled=false` option to electron-builder, which makes `packageJSON.updates.enabled` false).

IMHO the `packageJSON.packageType` should remain as purely an informational / debug property, and I don't think that any Etcher behaviour should depend on it.

jviotti:
> Hmmm, so is the Makefile going to be deprecated, or will it still be used for local (non-Concourse) builds? (which is the only time that the optional DISABLE_UPDATES variable got used anyway)

We will still keep the Makefile for local debugging purposes, which will now refer to scripts inside the `resin-concourse` repo, probably as an optional git submodule.

> The way the Makefile works currently, the deb and rpm builds work without relying on any environmental variables (by passing the --extraMetadata.analytics.updates.enabled=false option to electron-builder, which makes packageJSON.updates.enabled false).

I guess the advantage is that Concourse will not be the one deciding what types of packages should or should not check for updates, but we let the client decide. Also, we don't have to document that clients should pay attention to a use-case specific variable in package.json; users can do whatever they want based on simply the package type, which has more flexibility in the long run.

lurch:
Ahhh, I think it sounds you're setting up resin-concourse as a 'generic' system then (is that what you mean by "let the client decide"?), whereas I'd been assuming that you were setting up Concourse just to build Etcher, and just from the Makefile (much like we'd done with Travis and Appveyor). My apologies for the lack of context - I've clearly got a lot (more) to learn! ;-)
(If we're having two separate build systems, hopefully it won't be too difficult to keep changes to the Makefile (for local builds) in sync with changes to Concourse (for CI builds))

..... Although even if the Makefile _is_ only being kept for local debugging purposes, maybe it'd be worth keeping the `ifdef DISABLE_UPDATES` section, merely as an easy way for a local user to set `packageJSON.updates.enabled` to `false`? (i.e. without having to edit `package.json` directly) Or possibly that's simply not worth it.

jviotti:
@lurch Yeah, exactly. We will keep the Makefile, but it will call code that is inside the `resin-concourse` repo (as a git submodule), so there will only be one true build system.

PR #1941, opened 2017-12-29
chore: re-organize files inside assets/

jviotti:
This is the file convention that Resin Concourse will use. I flattened
the directory for simplicity.

Change-Type: patch


jhermsmeier:
Sorry, I didn't get to this before. There was actually a reason behind having the 64x64.png icon images in an `iconset` subdirectory – for one it's how Mac OS CLI tools for .ico sets take input & output, and it also avoids the iconset utils unintentionally incorporating other images. In that regard, I think a non-flattened directory structure is actually simpler in the long run.

lurch:
Does Resin Concourse have a problem with subdirectories or something? :confused:

jviotti:
The reasoning behind this is that is easier to tell someone else (remember that the Electron pipeline might be available to the public at some point) to put a certain list of files inside an `assets` directory rather than telling them, put some of them in this directory, the these ones in the top level, etc.

What directory structure do you propose?

jhermsmeier:
I'd just keep the 16x16.png etc. icon size images in the iconset folder, and move the `dmg-installer.*` to `darwin/volume-background.*`, I think.

jviotti:
See https://github.com/resin-io/etcher/pull/1952

On Thu, Jan 04, 2018 at 11:45:13AM +0000, Jonas Hermsmeier wrote:
> I'd just keep the 16x16.png etc. icon size images in the iconset folder, and move the `dmg-installer.*` to `darwin/volume-background.*`, I think.
> 
> -- 
> You are receiving this because you modified the open/close state.
> https://github.com/resin-io/etcher/pull/1941#issuecomment-355262614

-- 
Juan Cruz Viotti
Software Engineer

PR #1943, opened 2018-01-02
feat(GUI): bundle the GUI code with Webpack 

jviotti:
This commit introduces a Webpack configuration file that bundles the GUI
code along with its dependencies (except the Etcher SDK and its own
dependencies), and uses Babel to add support for JSX (required by the
Rendition library).

The GUI code that goes into the bundle was moved to `lib/gui/app` so we
can easily ignore the whole subdirectory when creating production
distributable packages.

We now have a new make target called `webpack` that can be used to
create the GUI bundle. Such target will be called everytime a package is
generated.

Change-Type: patch


Makefile
```diff
@@ -399,6 +407,8 @@ TARGETS = \
 changelog:
 	versionist
 
+webpack: build/js/gui.js
```
lurch: Feels a little odd having a `webpack` target which depends on a `build/js/gui.js` file, when that file itself is built by running a `webpack` tool. Perhaps the Makefile target could be given a different name to avoid confusion? Perhaps something like `build-webpack`? (and then perhaps the existing `sass` target could be renamed to `build-sass` at the same time? And you could have a `build` target which runs both `build-sass` and `build-webpack`, in a similar style to the `lint` and `test` targets.)

lib/gui/app/components/modal/services/modal.js
```diff
@@ -43,13 +43,11 @@ module.exports = function ($uibModal, $q) {
       size: 'sm'
     })
 
-    analytics.logEvent('Open modal', {
-      template: options.template
-    })
+    analytics.logEvent('Open modal')
```
lurch: Hmmm, so this logEvent will no longer show _which_ modal was opened?

jviotti: `options.template` used to be a local URL, but its now inline HTML

Makefile
jviotti: Lets handle that in a separate PR (renaming some targets to include a `build` prefix)

jviotti:
Updated the PR description

lib/gui/app/components/modal/services/modal.js
lurch: Ahh, I see. I'm wondering how useful a `'Open modal'` logEvent will be if we can't even tell which modal was opened? :-/

jviotti: Fair enough. I added a `name` option to the `ModalService`, so we can track what it is.

Makefile
lurch: So, how does the 'sass' and css stuff fit into all of this? It's causing the CI builds to fail, and when I run `make sass` on a local copy of this PR, it complains that "An output directory must be specified when compiling a directory". I can fix that error by editing the Makefile rule to `node-sass lib/gui/app/scss/main.scss > lib/gui/css/main.css` but it then complains about a bunch of incorrect paths.

lib/gui/app/components/drive-selector/services/drive-selector.js
```diff
@@ -34,7 +34,8 @@ module.exports = function (ModalService, $q) {
    */
   this.open = () => {
     modal = ModalService.open({
-      template: './components/drive-selector/templates/drive-selector-modal.tpl.html',
+      name: 'drive selector',
```
lurch: Purely for consistency purposes, perhaps this should be `name: 'drive-selector',` ?

jviotti:
Fixed!

On Wed, Jan 03, 2018 at 07:54:19PM +0000, Andrew Scheller wrote:
> lurch commented on this pull request.
> 
> 
> 
> > @@ -34,7 +34,8 @@ module.exports = function (ModalService, $q) {
>     */
>    this.open = () => {
>      modal = ModalService.open({
> -      template: './components/drive-selector/templates/drive-selector-modal.tpl.html',
> +      name: 'drive selector',
> 
> Purely for consistency purposes, perhaps this should be `name: 'drive-selector',` ?
> 
> -- 
> https://github.com/resin-io/etcher/pull/1943#pullrequestreview-86474100

-- 
Juan Cruz Viotti
Software Engineer

jviotti:
Yeah, there were some issues with sass code. I fixed them all and
updated the PR.

> So, how does the 'sass' and css stuff fit into all of this?

I plan to also move this to the same generated directory as a separate
PR.

On Wed, Jan 03, 2018 at 11:51:15AM -0800, Andrew Scheller wrote:
> lurch commented on this pull request.
> 
> 
> 
> > @@ -399,6 +407,8 @@ TARGETS = \
>  changelog:
>  	versionist
>  
> +webpack: build/js/gui.js
> 
> So, how does the 'sass' and css stuff fit into all of this? It's causing the CI builds to fail, and when I run `make sass` on a local copy of this PR, it complains that "An output directory must be specified when compiling a directory". I can fix that error by editing the Makefile rule to `node-sass lib/gui/app/scss/main.scss > lib/gui/css/main.css` but it then complains about a bunch of incorrect paths.
> 
> -- 
> https://github.com/resin-io/etcher/pull/1943#discussion_r159512280

-- 
Juan Cruz Viotti
Software Engineer

scripts/ci/ensure-all-node-requirements-available.sh
```diff
@@ -62,6 +63,9 @@ git ls-tree -r HEAD | while IFS='' read line; do
               requirement_found=1
             elif [[ -f "$localpath.js" ]] || [[ -f "$localpath/index.js" ]]; then
               requirement_found=1
+            # Webpack HTML loader
+            elif [[ "$localpath" =~ $HTML_REGEX ]]; then
+              requirement_found=1
```
lurch: IMHO it would be better if this actually checked if the relevant HTML file exists. I did a bit of local experimentation, and the below diff adds this feature:
```diff
--- a/scripts/ci/ensure-all-node-requirements-available.sh
+++ b/scripts/ci/ensure-all-node-requirements-available.sh
@@ -49,7 +49,7 @@ git ls-tree -r HEAD | while IFS='' read line; do
     extension=${filename##*.}
     if [[ "$extension" == "js" ]]; then
       # 'grep -v' to filter out any comment-blocks
-      grep 'require(' "$fullpath" | grep -v "^ \* " | while IFS='' read line; do
+      grep 'require(' "$fullpath" | grep -v -E "^ +\* " | while IFS='' read line; do
         if [[ "$line" =~ $REQUIRE_REGEX ]]; then
           required=${BASH_REMATCH[1]}
         fi
@@ -64,7 +64,7 @@ git ls-tree -r HEAD | while IFS='' read line; do
             elif [[ -f "$localpath.js" ]] || [[ -f "$localpath/index.js" ]]; then
               requirement_found=1
             # Webpack HTML loader
-            elif [[ "$localpath" =~ $HTML_REGEX ]]; then
+            elif [[ "$localpath" =~ $HTML_REGEX ]] && [[ -f "$localpath" ]]; then
               requirement_found=1
             fi
           else
```
:smiley:

jviotti: Cool, thanks! I updated the PR with your fix :)

Shou:
Looks like Webpack assumes React is imported as `'React'` and the text casing matters, not a big deal but something to be aware of.

jhermsmeier:
@LucianBuzzo can you give me a hand looking at this PR?

jviotti:
@Shou Just a heads up, I rebase this branch to include the changes from the recently merged PR, just to check how big it still is

jviotti:
I was testing this today and found some path related issues. I updated the PR with some small changes. Everything works both locally and on a DMG on my mac now.

Shou:
Just posting to confirm it Works On My Machine™ with these changes – all of `npm start`, AppImages, and Deb packages.

jviotti:
Very nice! Thanks for the testing @Shou !

jviotti:
Lets see if the CI services agree :P

jhermsmeier:
The commits that don't adhere to our commit guidelines will need squashing / amending before merging @jviotti

PR #1946, opened 2018-01-03
upgrade(package): Bump mountutils 1.2.2 -> 1.3.8

jhermsmeier:
Change-Type: patch

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4238,12 +4238,24 @@
       "resolved": "https://registry.npmjs.org/moment-duration-format/-/moment-duration-format-1.3.0.tgz"
     },
     "mountutils": {
-      "version": "1.2.2",
-      "resolved": "https://registry.npmjs.org/mountutils/-/mountutils-1.2.2.tgz",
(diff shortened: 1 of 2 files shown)


lurch:
Not **directly** related to this PR, but looking at https://github.com/resin-io-modules/mountutils/releases it seems that it's still only prebuilding x64 modules, and not prebuilding x86 modules?
(I just did some digging and found an old comment [here](https://github.com/resin-io-modules/mountutils/pull/58#issuecomment-338050404) )

jhermsmeier:
@lurch then open an issue on mountutils instead of commenting off topic, please

lurch:
Sorry Jonas, done that now https://github.com/resin-io-modules/mountutils/issues/62

jhermsmeier:
@lurch is this good to merge though?

jhermsmeier:
@jviotti there hasn't really been a change that impacts Etcher in a noticeable way, so I figured a changelog entry would be excess, but can add one

jviotti:
Ah, alright, just making sure we didn't miss any important one.

PR #1949, opened 2018-01-04
fix(shared): trailing space in Windows elevation env vars

jviotti:
This is a strange one. On Windows, putting a space before the
double-ampersand command concatenator makes the environment variable
value contain a trailing space. So for something like `set foo=bar &&
...` the variable `foo` will be `'bar '` instead of `'bar'`.

Change-Type: patch
Changelog-Entry: Fix trailing space in environment variables during Windows elevation.

### DIFF
--- a/lib/shared/permissions.js
+++ b/lib/shared/permissions.js
@@ -99,7 +99,12 @@ exports.getEnvironmentCommandPrefix = (environment) => {
     }
 
     if (isWindows) {
-      return [ 'set', `${key}=${value}`, '&&' ]
+      // Doing something like `set foo=bar &&` (notice
+      // the space before the first ampersand) would
+      // cause the environment variable's value to
+      // contain a trailing space.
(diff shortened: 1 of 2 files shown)


lurch:
Just out of curiosity, how come we never spotted this before?

jviotti:
We were spawning an elevated process that only got passed "boolean"
environment variables. This elevated process then elevated the CLI and
passed the other important env vars, but we never spot this issue since
the CLI was spawned using the built-in `.spawn()` function (since the
parent process was already elevated).

On Mon, Jan 08, 2018 at 01:13:32PM +0000, Andrew Scheller wrote:
> Just out of curiosity, how come we never spotted this before?
> 
> -- 
> https://github.com/resin-io/etcher/pull/1949#issuecomment-355962795

-- 
Juan Cruz Viotti
Software Engineer

Issue #1915, opened 2017-12-10
Couldn't scan the drives: An unknown error occurred


lurch:
- **Etcher version:** 1.2.1
- **Operating system and architecture:** Windows7 x64
- **Image flashed:** _any_
- **Do you see any meaningful error information in the DevTools?** Yes... ;-)
```
[snipped the usual stuff]
Sun Dec 10 2017 22:39:06 GMT+0000 (GMT Standard Time) Progress (check): 96% at 11.51 MB/s (eta 0.2619589977220957s)
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 Sun Dec 10 2017 22:39:06 GMT+0000 (GMT Standard Time) Progress (check): 100% at 11.02 MB/s (eta 0s)
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] \\.\PHYSICALDRIVE1
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] Checking whether drive is ejectable: C
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] The drive is fixed
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] Opening drive letter handle: C
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] Closing drive letter handle
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] Continuing with the next available letter
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] Checking whether drive is ejectable: D
C:\Users\Andrew\AppData\Local\Temp\nsm31D5.tmp\app\resources\app.asar\node_modules\raven-js\dist\ra…:55 [mountutils] The drive is fixed
[...]

Issue #1867, opened 2017-11-19
Couldn't scan the drives: An unknown error occurred (With a twist)


Arffeh:
- **Etcher version:**
1.1.2

- **Operating system and architecture:**
Win7 Ultimate 64 Bit

- **Image flashed:**
N/A

- **Do you see any meaningful error information in the DevTools?**

{"stack":"Error: Couldn't scan the drives: An unknown error occurred","message":"Couldn't scan the drives: An unknown error occurred"}

So, turns out Etcher spits the dummy if there is a bitlocker volume not yet decrypted.

Confirmed by having the volume decrypted, then proceeding to lock the volume again. the instant I do I get the error again.
Etcher needs to gracefully handle bitlocker volumes.
Was pulling my hair out trying to figure it out.

jhermsmeier:
Hey, this should've been fixed in #1692 and will ship in the next release

jhermsmeier:
Duplicate of #1687

Issue #1782, opened 2017-10-16
Couldn't scan the drives: An unknown error occurred


liwanglin12:
- **Etcher version:**
v1.1.2

- **Operating system and architecture:**
Windows 10

- **Image flashed:** 
Any

Error on start

[Window Title]
Error

[Main Instruction]
Couldn't scan the drives: An unknown error occurred

[Content]
Error: Couldn't scan the drives: An unknown error occurred

jhermsmeier:
Duplicate of #1646

liwanglin12:
After unlocking a Bitlocker locked disk, the problem is lifted.

Hope to improve error handle! lol

Issue #1757, opened 2017-10-09
"Couldn't scan the drives: An unknown error occurred"


Rattle22:
Version 1.1.2, running on Windows 10 x64.

While a similar error has already been reported and fixed, this one differs in that the drivelist debug output spits out the error code 0x80070456.  I tried running etcher as administrator, but that did not help at all.

According to what I found online, this code usually appears when the Windows Media Creation tool encounters some weird error. However, none of the fixes I found online worked for me (specifically those telling me to delete temporary files). I could not find any other occurence of this error code.

While this seems related to some mysterious PHSICALDRIVE5 that has no logical drive equivalent (and also no physical object equivalent that I am aware of) and lists no size, I could not find anything on how deal with that thing either.

EDIT: A simple restart fixed the issue (yeah yeah, I know, should've done this earlier).

PR #1953, opened 2018-01-04
upgrade(package): Update drivelist 5.2.12 -> 6.0.0

jhermsmeier:
This updates `drivelist` to incorporate the new native
bindings rewrite on Windows, fixing scanning errors.

**TODO:**
- [x] Merge https://github.com/resin-io-modules/drivelist/pull/243
- [x] Update package.json & npm-shrinkwrap.json to the correct dependency

Change-Type: patch
Changelog Entry: Fix "Couldn't scan drives" error
Connects To: #1646, #1660, #1665, #1782, #1915, #1867, #1757


lib/shared/sdk/standard/index.js
```diff
@@ -71,15 +71,21 @@ class StandardAdapter extends EventEmitter {
...
 
       return drive
     }).catch((error) => {
       this.emit('error', error)
       callback && callback(error)
     }).filter((drive) => {
-      return options.includeSystemDrives || !drive.system
+      return !drive.error && (options.includeSystemDrives || !drive.isSystem)
+    }).map((drive) => {
+      drive.displayName = drive.device
```
jviotti: Can we move the displayName to `drivelist`, given we're likely going to use the same thing on the CLI? Of course doesn't need to be done right now

tests/shared/models/available-drives.spec.js
```diff
@@ -354,9 +386,11 @@ describe('Model: availableDrives', function () {
               device: '/dev/sdc',
               name: 'USB Drive',
               size: 9999999,
-              mountpoint: '/mnt/bar',
-              system: false,
-              protected: false
+              mountpoints: [ {
```
jviotti: I have the impression the linter will complain about this.

jhermsmeier: Quite the opposite; it wants a space there... :/

lib/shared/sdk/standard/index.js
jhermsmeier: I was thinking that this logic is very application specific – in general the `description` is the best display name, imo, so the CLI might just want to switch to that?
I just recreated the displayName here, as we've got this fairly specific difference in handling them on Win in contrast to Linux / Mac in Etcher.
What do you think? Of course we can add it back to drivelist also, but I feel that it shouldn't be concerned with these UI specific things.

jviotti:
We should unify how drives are displayed in both the GUI and the CLI
version. If the description is the way to go (and we determine so in ie
a product call), then we should use the description everywhere.
Otherwise, we rather put the logic in one place and re-use a new propery
in both UIs.

I think providing a "display name" property in drivelist is fine, and
can be ignored if people want to use something else.

On Fri, Jan 05, 2018 at 03:56:50PM +0000, Jonas Hermsmeier wrote:
> jhermsmeier commented on this pull request.
> 
> 
> 
> >        }
>  
>        return drive
>      }).catch((error) => {
>        this.emit('error', error)
>        callback && callback(error)
>      }).filter((drive) => {
> -      return options.includeSystemDrives || !drive.system
> +      return !drive.error && (options.includeSystemDrives || !drive.isSystem)
> +    }).map((drive) => {
> +      drive.displayName = drive.device
> 
> I was thinking that this logic is very application specific – in general the `description` is the best display name, imo, so the CLI might just want to switch to that?
> I just recreated the displayName here, as we've got this fairly specific difference in handling them on Win in contrast to Linux / Mac in Etcher.
> What do you think? Of course we can add it back to drivelist also, but I feel that it shouldn't be concerned with these UI specific things.
> 
> -- 
> You are receiving this because your review was requested.
> https://github.com/resin-io/etcher/pull/1953#discussion_r159909635

-- 
Juan Cruz Viotti
Software Engineer

tests/gui/modules/drive-scanner.spec.js
```diff
@@ -332,7 +367,8 @@ describe('Browser: driveScanner', function () {
                 path: 'H:'
               }
             ],
-            system: false
+            isSystem: false,
+            isRemovable: true
```
lurch: I'm wondering why you've added `isRemovable` to all these test-cases, given that nothing else even looks at this property?

lurch:
I can see both points of view, and from a "purist" PoV `displayName` doesn't belong in `drivelist`; but I think I tend to agree with @jviotti 's pragmatist PoV that it's such a generally-useful property that it makes sense to add it back to `drivelist`.

Issue #1955, opened 2018-01-05
Error: stdout maxBuffer exceeded


jhermsmeier:
- **Etcher version:** 1.3.0
- **Operating system and architecture:** Linux
- **Error information:** https://sentry.io/resinio-57/etcher-v130/issues/429230586/

bdfoster:
Reproduced on Ubuntu 16.04.2

potatowave:
Reproduced on Ubuntu 16.04.3

Seems to happen at 73% copying a 1.16GB ISO onto a 16GB USB stick.

nicklassundin:
It seems to happen randomly within spectrum of 30 - 70% into the process
4.9 GB .img to 16 GB Micro-SDHC
Ubuntu 16.04 64-bit

Cyb3rpunk-76:
Reproduced on Mint 18.2 Sonya, always on 65%, 32 Gb Micro-SDHC.

Timodz:
Reproduced on Mint 18.2 Sonya

krisYanachkov:
Etcher version: 1.3.0
Operating System: Mint 18.1 64-bit
SDHC: 8GB

SLiX69:
Get this too on 17.10 with Etcher 1.3

JGuillaumin:
Same error while flashing Raspian on a SD card, from Ubuntu 16.04
Finally I used `dd` command.

Lewiscowles1986:
Ooh

![dodgy sandisk](https://user-images.githubusercontent.com/2605791/35100153-995a54c0-fc52-11e7-931a-f59e6713e47b.png)

Looks like this is still the case in xenial repos. I get it, there are 4 other issues pending fix

![image](https://user-images.githubusercontent.com/2605791/35100177-b0531d06-fc52-11e7-8085-efb065b7fea4.png)
![image](https://user-images.githubusercontent.com/2605791/35100200-cf8e3c50-fc52-11e7-87af-540c6c33decb.png)

lurch:
We updated the bintray repos so that 1.2.1 is the latest version https://bintray.com/resin-io/debian/etcher
but I guess apt / dpkg must not downgrade packages if a newer version of the package is already installed :-/

We're hoping to release Etcher 1.3.1 soon, but as a temporary workaround you can do:
```
sudo apt-get remove etcher-electron
sudo apt-get update
sudo apt-get install etcher-electron
```
to manually downgrade to 1.2.1

manuti:
[...]

PR #1961, opened 2018-01-08
fix(writer): Silence IPC output on stdout

jhermsmeier:
This was causing the stdout maxBuffer size to be exceeded
when flashing larger images (or having flashes that took a while).

Change-Type: patch
Changlog Entry: Fix "stdout maxBuffer exceeded" error on Linux
Connects To: #1955

### DIFF
--- a/lib/gui/modules/child-writer.js
+++ b/lib/gui/modules/child-writer.js
@@ -23,7 +23,10 @@ const writer = require('../../cli/writer')
 
 ipc.config.id = process.env.IPC_CLIENT_ID
 ipc.config.socketRoot = process.env.IPC_SOCKET_ROOT
-ipc.config.silent = false
+
+// NOTE: Ensure this isn't disabled, as it will cause
+// the stdout maxBuffer size to be exceeded when flashing
+ipc.config.silent = true
 
 // > If set to 0, the client will NOT try to reconnect.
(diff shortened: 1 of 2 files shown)


jviotti:
Can we research why this only happens on Linux? Is it because that platform has a much lower max buffer limit? What are the default limits in the other OSes? Are we sure that we can't ever reach that limit even with a very big image or very slow drive? What things are printed to stdout by the child writer?

Shou:
I made a new issue where we can explore the underlying problem in case it takes a while so we don't block this PR

jhermsmeier:
Just looked into whether there's anything else on stdout/stderr; but both remain blank when `ipc.config.silent = true`, so this will definitely fix it for good.

jhermsmeier:
@jviotti regarding limits on OSs; the limit is 200KB by default as per Node's [child_process docs](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) – I think the symptoms showing might be different on Mac OS, and Windows isn't hitting it, because we're not using .exec() there

jviotti:
Got it. I created an issue on `sudo-prompt` with a potential solution to avoid this error even if the child logs a lot of stuff, since it can be hard to prevent our future-selves from making a similar mistake again: https://github.com/jorangreef/sudo-prompt/issues/66

PR #1963, opened 2018-01-08
feat: test that ipc verbosity is off

Shou:
We test that `ipc.config.silent` is set to `true` so that it doesn't
spam `stdout` for users.

Changelog-Entry: Test that IPC verbosity is off
Change-Type: patch
Depends on: https://github.com/resin-io/etcher/pull/1961


jhermsmeier:
@Shou this is having lint errors (due to the required modules not being used)

Shou:
@jhermsmeier here we go

tests/gui/modules/image-writer.js
```diff
@@ -0,0 +1,28 @@
...
+'use strict'
+
+const m = require('mochainon')
+const ipc = require('node-ipc')
+// eslint-disable-next-line no-unused-vars
+const imageWriter = require('../../../lib/gui/modules/image-writer')
+
+describe('Browser: imageWriter', function () {
+  it('should have the ipc config set to silent', function () {
+    m.chai.expect(ipc.config.silent).to.be.true
```
lurch: Unlike `childWriter` (which modifies it at the global scope), `imageWriter` only modifies `ipc.config.silent` inside the `.performWrite` function?  Which, unless I'm missing something, means that this test isn't checking what we expect it to?

lurch:
(and could you rebase this on `master`?)

tests/gui/modules/image-writer.js
Shou: Good catch, I've fixed this now.

tests/gui/modules/image-writer.js
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const m = require('mochainon')
+const ipc = require('node-ipc')
+// eslint-disable-next-line no-unused-vars
```
lurch: I guess this eslint pragma can be removed now :)

Shou: Done

lurch:
I've just noticed that we already have a `tests/gui/modules/image-writer.spec.js` - why does this PR add a separate `tests/gui/modules/image-writer.js` ?

Shou:
Can't believe I missed that, thanks – also infixed `.spec` to the `child-writer` test filename. Maybe I should also add a test that the filenames in `tests/` all match the pattern in a separate PR...

tests/gui/modules/child-writer.spec.js
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const m = require('mochainon')
+const ipc = require('node-ipc')
+// eslint-disable-next-line no-unused-vars
+const childWriter = require('../../../lib/gui/modules/child-writer')
```
jviotti: Why are we requiring something we don't use? I think you can keep the linter happy by doing:

```js
require('../../../lib/gui/modules/child-writer')
```

Notice that by running the child writer, then we're running code that tries to connect to an IPC server that doesn't exist. Are no errors throws?

Shou: Yup, forgot about that – fixed now! Yeah, there are no errors. `IPC_SERVER_ID` is `undefined` in `ipc.connectTo(IPC_SERVER_ID, () => { ...` which just exits silently as long as the verbosity is set to silent <sup>[1] [2]</sup>.

[1]: https://github.com/RIAEvangelist/node-ipc/blob/master/services/IPC.js#L226
[2]: https://github.com/RIAEvangelist/node-ipc/blob/master/services/IPC.js#L64

PR #1965, opened 2018-01-08
chore: add concourse.builder package.json configuration object

jviotti:
This object will eventually replace the `electron-builder.yml` file
that's currently present in the root of the project.

For now, it contains the `electron-builder` options that are project
specific (all the generic bits live in the Electron Concourse pipeline),
but in the future we might want to decouple how users configure packages
from the `electron-builder` project, instead making the user provide
Resin Concourse specific options that are then translated to
`electron-builder` (or any other packaging technology we might decide to
use) under the hood.

Change-Type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,44 @@
         "libusb-1.0-0-dev",
         "libyaml-dev"
       ]
+    },
+    "builder": {
+      "appId": "io.resin.etcher",
+      "copyright": "Copyright 2016 Resinio Ltd",
+      "productName": "Etcher",
+      "mac": {
+        "category": "public.app-category.developer-tools"
+      },
+      "dmg": {
+        "iconSize": 110,
+        "contents": [
+          {
+            "x": 140,
+            "y": 225
+          },
+          {
+            "x": 415,
+            "y": 225,
+            "type": "link",
+            "path": "/Applications"
+          }
+        ],
+        "window": {
+          "width": 540,
+          "height": 405
+        }
+      },
+      "linux": {
+        "category": "Utility",
+        "packageCategory": "utils",
+        "synopsis": "Etcher is a powerful OS image flasher built with web technologies to ensure flashing an SDCard or USB drive is a pleasant and safe experience. It protects you from accidentally writing to your hard-drives, ensures every byte of data was written correctly and much more."
+      },
+      "deb": {
+        "priority": "optional",
+        "depends": [
+          "polkit-1-auth-agent | policykit-1-gnome | polkit-kde-1"
+        ]
+      }
     }
   },
   "dependencies": {


jhermsmeier:
I think it'd be better to avoid polluting the `package.json` with data that's not relevant for npm; for one to avoid potential future conflicts, and also because of separation of concerns; what do you think about putting this into a `.concourse.json` file instead, for example?

package.json
```diff
@@ -42,6 +42,44 @@
...
+          "width": 540,
+          "height": 405
+        }
+      },
+      "linux": {
+        "category": "Utility",
+        "packageCategory": "utils",
+        "synopsis": "Etcher is a powerful OS image flasher built with web technologies to ensure flashing an SDCard or USB drive is a pleasant and safe experience. It protects you from accidentally writing to your hard-drives, ensures every byte of data was written correctly and much more."
+      },
+      "deb": {
```
lurch: Should there also be an `rpm` section?

lurch:
> what do you think about putting this into a `.concourse.json` file instead, for example?

Or something more platform-neutral like `.builder.json` in case we switch away from Concourse in future?

jhermsmeier:
If we'd switch to something else, it'd have a different config, plus having the file carry the name of what it's configuring means it doesn't require checking the docs on what it is for etc.

jviotti:
Sounds good. I named the file `.resinci.json`, since we're likely moving to Resin CI as a name.

package.json
jviotti: There were some known issues with Etcher's RPM dependencies, but I can't remember what it was. I think RPM repositories didn't support any elevation package we used, or something like that. Maybe worth trying to add it back in a new PR?

lurch: > I think RPM repositories didn't support any elevation package we used, or something like that.

I think that was the "elevate at startup" PR that @Shou was working on, which didn't get merged in the end? The current Etcher (Makefile & electron-builder) setup builds both deb and rpm files, so IMHO we shouldn't drop rpm files when switching to concourse.

.resinci.json
```diff
@@ -0,0 +1,47 @@
...
+      }
+    },
+    "linux": {
+      "category": "Utility",
+      "packageCategory": "utils",
+      "synopsis": "Etcher is a powerful OS image flasher built with web technologies to ensure flashing an SDCard or USB drive is a pleasant and safe experience. It protects you from accidentally writing to your hard-drives, ensures every byte of data was written correctly and much more."
+    },
+    "deb": {
+      "priority": "optional",
+      "depends": [
```
lurch: Why is the list of dependencies here so much shorter than the equivalent list at https://github.com/resin-io/etcher/blob/master/electron-builder.yml#L50 ?

.resinci.json
```diff
@@ -0,0 +1,47 @@
+{
+  "dependencies": {
+    "linux": [
+      "libudev-dev",
+      "libusb-1.0-0-dev",
+      "libyaml-dev"
```
lurch: I'm obviously not aware (yet) of how this file is going to be used, but it seems a bit strange that Windows and macOS won't also need YAML libraries?

.resinci.json
```diff
@@ -0,0 +1,47 @@
...
+    ]
+  },
+  "builder": {
+    "appId": "io.resin.etcher",
+    "copyright": "Copyright 2016 Resinio Ltd",
+    "productName": "Etcher",
+    "mac": {
+      "category": "public.app-category.developer-tools"
+    },
+    "dmg": {
```
lurch: Should "dmg" be listed _inside_ "mac"?

.resinci.json
```diff
@@ -0,0 +1,47 @@
...
+        "width": 540,
+        "height": 405
+      }
+    },
+    "linux": {
+      "category": "Utility",
+      "packageCategory": "utils",
+      "synopsis": "Etcher is a powerful OS image flasher built with web technologies to ensure flashing an SDCard or USB drive is a pleasant and safe experience. It protects you from accidentally writing to your hard-drives, ensures every byte of data was written correctly and much more."
+    },
+    "deb": {
```
lurch: ...and "deb" listed inside "linux" ?

.resinci.json
jviotti: This is passed as-is to electron-builder, which expects mac to be outside dmg (not up to us). We will put a layer on top of electron-builder at some point, though, so we will address this soon

.resinci.json
jviotti: Its a build-time dependency of a certain native add-on on Linux. I didn't dig into it, but maybe the module is Linux-only, or does things differently in other OSes?

.resinci.json
jviotti: That big list contains dependencies for every Electron application. The Resin Concourse Electron pipeline keeps track of those, and requires projects to only declare the ones that are specific to the application.

lurch: Ahhh, I didn't realise this was specifically a "Concourse Electron pipeline" and not just a "Concourse pipeline"! Perhaps it'd be worth specifying that in `.resinci.json` in case we add other non-Electron pipelines to Concourse in future?

jviotti: Yeah, I think that's a good point. I'll do it in a minute.

jviotti: Done!

.resinci.json
```diff
@@ -0,0 +1,49 @@
...
+      },
+      "linux": {
+        "category": "Utility",
+        "packageCategory": "utils",
+        "synopsis": "Etcher is a powerful OS image flasher built with web technologies to ensure flashing an SDCard or USB drive is a pleasant and safe experience. It protects you from accidentally writing to your hard-drives, ensures every byte of data was written correctly and much more."
+      },
+      "deb": {
+        "priority": "optional",
+        "depends": [
+          "polkit-1-auth-agent | policykit-1-gnome | polkit-kde-1"
```
jhermsmeier: Why aren't these individual elements in an array, like the following?

```js
"depends": [
  "polkit-1-auth-agent",
  "policykit-1-gnome",
  "polkit-kde-1"
]
```

Also, why is the key here "depends", rather than "dependencies"?

jviotti: I'm not choosing the names. These are passed as-is to `electron-builder`.

lurch: So presumably this'll also be "cleaned up" in future like https://github.com/resin-io/etcher/pull/1965#discussion_r160797221 ?

jhermsmeier: Oh, I see, thanks for the clarification!

PR #1968, opened 2018-01-10
fix(shrinkwrap): Update unbzip2-stream branch commit

jhermsmeier:
The shrinkwrap still contained the commit hash of a commit
previous to an npm install bugfix

Change-Type: patch


lurch:
To clarify, this is the new commit that this update includes https://github.com/resin-io-modules/unbzip2-stream/commit/942fc218013c14adab01cf693b0500cf6ac83193

Issue #1956, opened 2018-01-05
Error: *.tmp.node module not found


ben-h253:
- **Etcher version:** 1.3.0
![error capture](https://user-images.githubusercontent.com/35154374/34632840-c53d606a-f26f-11e7-9b44-8081a751d4cf.PNG)

- **Operating system and architecture:** Windows 8
- **Image flashed:** ?
- **Do you see any meaningful error information in the DevTools?** n.a.

vtabary:
Same here with Windows 10.
So I'll use the 1.2.1 for now : https://github.com/resin-io/etcher/releases/

jviotti:
Hi @ben-h253 and @vtabary ,

Were you running the 64-bit or 32-bit versions? Also, were you using the installer or the portable versions? Do you have any anti-virus installed?

vtabary:
Hi @jviotti ,

I'm running the 64-bits version of winfodows 10. And yes, I use Avast antivirus.

jhermsmeier:
We've discovered that this is caused by the lack of certain C runtime libraries on some Windows systems that the WinUSB driver generator requires to function.

PR #1970, opened 2018-01-11
fix(usb): Ignore errors if winusb doesn't load

jhermsmeier:
Due to some Windows systems missing certain C runtime libraries
(Visual C/C++ 2012 / 2015 Redistributables), we ignore errors when loading
this module until we can ensure distribution of those along with it.

Change-Type: patch
Changelog-Entry: Fix "The specified module could not be found" on Windows
Connects To: #1956

### DIFF
--- a/lib/shared/sdk/usbboot/usb.js
+++ b/lib/shared/sdk/usbboot/usb.js
@@ -74,15 +74,22 @@ exports.listDevices = () => {
 
   // Include driverless devices into the list of USB devices.
   if (process.platform === 'win32') {
-    const winusbDriverGenerator = require('winusb-driver-generator')
-    for (const device of winusbDriverGenerator.listDriverlessDevices()) {
-      devices.push({
-        accessible: false,
-        deviceDescriptor: {
-          idVendor: device.vid,
-          idProduct: device.pid
-        }
-      })
+    // NOTE: Temporarily ignore errors when loading winusb-driver-generator,
+    // due to C Runtime issues on Windows;
+    // see https://github.com/resin-io/etcher/issues/1956
+    try {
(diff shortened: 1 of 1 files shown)


lurch:
Some of the CI builds failed because of the earlier GitHub issues, so I restarted them for you :)

jhermsmeier:
@jviotti I've thought about this a little more, and now wrapped that in a try / catch block, so that it still works for people not missing those runtimes – what do you think?

lurch:
So this definitely won't spam Sentry with errors from people who don't have the appropriate CRT installed? (and those users won't get an on-screen error dialog?)

I'm not sure if it would make more sense for Etcher to behave consistently everywhere, rather than working differently for some people than others? But I'll leave that as a call for you to make. \*shrug\*

jhermsmeier:
@lurch updated!

> So this definitely won't spam Sentry with errors from people who don't have the appropriate CRT installed? (and those users won't get an on-screen error dialog?)

No, the catch swallows the error – it's not going anywhere.

> I'm not sure if it would make more sense for Etcher to behave consistently everywhere, rather than working differently for some people than others? But I'll leave that as a call for you to make. 

That's already the case with usbboot anyways; so this won't change things about that too much

lurch:
> @lurch updated!

Cheers!
Thinking about this a bit more, even when we _do_ bundle the necessary CRT files with `winusb-driver-generator`, perhaps we'd want to leave this try/catch in place anyway, so perhaps it's not "temporary" after all? ;-)

And I'm sorry to be pedantic, but to me "Temporarily bail out if ..." doesn't mean very much - perhaps "Ignore errors if ..." would be better?

Issue #1926, opened 2017-12-16
Etcher cannot differentiate folder and iso image when folder is named with .iso ending


lowks:
- **Etcher version: 1.2.1**
- **Operating system and architecture: Mac Mini 2012 running High Sierra**
- **Image flashed: openSuSE leap 42.3**
- **Do you see any meaningful error information in the DevTools? No**

When downloading OpenSuSE 42.3 image it will dump it in a folder called: `openSUSE-Leap-42.3-DVD-x86_64.iso`.  When using Etcher to pick the iso image it wrong assumes that the folder is the iso image and complains that the image is not valid.

lurch:
I'm not able to reproduce this on Linux - in fact the browse-file dialog doesn't allow me to select just a folder. I'll test on Windows later.
@jviotti Could you test this on your Mac to see if you can reproduce?

jhermsmeier:
This seems to be a bug in the Mac OS file open dialog (?), or more likely configuration thereof.

This is what it looks like in Finder (normal folder expansion):
![screen shot 2017-12-21 at 18 48 32](https://user-images.githubusercontent.com/244907/34268095-ba4231a0-e67f-11e7-8657-9f6af841105e.png)

While the open dialog doesn't allow expansion of the folder, due to the `.iso`:
![screen shot 2017-12-21 at 18 48 08](https://user-images.githubusercontent.com/244907/34268096-ba6398c2-e67f-11e7-88ea-5a1d53b2331c.png)

jviotti:
I confirm I can reproduce on macOS. Ping @jhermsmeier. It should be a
quick fix.

On Thu, Dec 21, 2017 at 09:31:26AM -0800, Andrew Scheller wrote:
> I'm not able to reproduce this on Linux - in fact the browse-file dialog doesn't allow me to select just a folder. I'll test on Windows later.
> @jviotti Could you test this on your Mac to see if you can reproduce?
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/resin-io/etcher/issues/1926#issuecomment-353410087

-- 
Juan Cruz Viotti
Software Engineer

jhermsmeier:
Fixed in #1976

PR #1976, opened 2018-01-18
fix(gui): Allow selection of images in folders with extension

jhermsmeier:
This fixes selection of images contained in directories with a file extension
(i.e. "openSUSE-Leap-42.3-DVD-x86_64.iso") in the open file dialog.

screen shot 2018-01-18 at 20 00 25

Change-Type: patch
Changelog-Entry: Fix selection of images in folders with file extension
Connects To: https://github.com/resin-io/etcher/issues/1926


lurch:
Minor nit, but could you edit the commit-message / changelog-entry to clarify that this bug only affected macOS ?

jhermsmeier:
Updated the changelog entry @lurch – couldn't fit it in the title though

PR #1986, opened 2018-01-23
fix(lib): Fix readonly property typo

jhermsmeier:
This fixes the camelcasing of the `.isReadOnly` property
of detected storage devices.

Change-Type: patch

### DIFF
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -43,13 +43,13 @@ const UNKNOWN_SIZE = 0
  *   device: '/dev/disk2',
  *   name: 'My Drive',
  *   size: 123456789,
- *   isReadonly: true
+ *   isReadOnly: true
  * })) {
  *   console.log('This drive is locked (e.g: write-protected)');
  * }
  */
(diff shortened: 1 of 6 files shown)


lib/shared/drive-constraints.js
```diff
@@ -231,12 +231,10 @@ exports.isDriveDisabled = (drive) => {
  * }
  */
 exports.isDriveValid = (drive, image) => {
-  return _.every([
-    !this.isDriveLocked(drive),
-    this.isDriveLargeEnough(drive, image),
-    !this.isSourceDrive(drive, image),
+  return !this.isDriveLocked(drive) &&
```
lurch: ?

jhermsmeier: Oh, that was due to testing, but it does the same – this way faster & more efficient though, as it doesn't waste-allocate an array on every call.

lurch: ...as long as the linter doesn't complain at you for not using lodash ;-)

jhermsmeier: Nah, linter is not complaining (thankfully)

PR #1987, opened 2018-01-23
chore: encode complete paths to patched file in patches/

jviotti:
So that the build system doesn't need to know in advance at which
directory the paths needs to be applied.

This will make it easier to add patches support to the Concourse
pipelines.

Change-Type: patch

### DIFF
--- a/patches/lzma-native-index-static-addon-require.patch
+++ b/patches/lzma-native-index-static-addon-require.patch
@@ -1,7 +1,5 @@
-diff --git a/index.js b/index.js
-index 1e7c766..e42c513 100644
---- a/index.js
-+++ b/index.js
+--- b/node_modules/lzma-native/index.js	2018-01-23 14:37:50.000000000 -0400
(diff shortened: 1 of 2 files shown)


patches/lzma-native-index-static-addon-require.patch
```diff
@@ -1,17 +1,15 @@
...
-index 1e7c766..e42c513 100644
---- a/index.js
-+++ b/index.js
+--- b/node_modules/lzma-native/index.js	2018-01-23 14:37:50.000000000 -0400
++++ a/node_modules/lzma-native/index.js	2018-01-23 14:37:00.000000000 -0400
 @@ -7,11 +7,8 @@ var extend = require('util-extend');
  var assert = require('assert');
  var fs = require('fs');
-
+
```
jhermsmeier: What's with the whitespace here? Is that a space or just a newline?

jviotti: They came from `lzma-native`. I was able to make the patch apply without them by using `--ignore-whitespace`

PR #1996, opened 2018-01-24
fix(gui): Fix menu's application name

jhermsmeier:
This replaces use of `electron.app.getName()` with the package.json's `.displayName`
property to ensure the correct application name is displayed when packaged.

Change-Type: patch

### DIFF
--- a/lib/gui/menu.js
+++ b/lib/gui/menu.js
@@ -17,6 +17,7 @@
 'use strict'
 
 const electron = require('electron')
+const packageJson = require('../../package.json')
 
 const menuTemplate = [
   {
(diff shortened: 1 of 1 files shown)


lurch:
Only a minor nit, but would it make sense to retrieve the `packageJson.displayName` in just one place, rather than two? (in case we ever need to change this again in future)

jhermsmeier:
I'm not sure I understand what you mean @lurch?

lurch:
I meant doing something like:
```
const applicationMenuName = packageJson.displayName
```
and then using `applicationMenuName` on lines 72 and 102, rather than using `packageJson.displayName` in two separate places. So that if we ever need to change what the `applicationMenuName` displays, we only need to do it in one place rather than two.

jhermsmeier:
@lurch that doesn't make much sense to me, tbh – it's only two references, assigning every prop ref to a var because of multiple use is kinda defeating the purpose of a programming language – should it become more complex for some awkward reason we can still abstract it away

lurch:
Fair enough, that's why I said it was only a minor not :)

PR #1997, opened 2018-01-24
fix(gui): Fix DevTools opening in docked mode

jhermsmeier:
This fixes the Developer Tools opening in docked mode by default.

Change-Type: patch

### DIFF
--- a/lib/gui/etcher.js
+++ b/lib/gui/etcher.js
@@ -19,6 +19,7 @@
 const electron = require('electron')
 const path = require('path')
 const EXIT_CODES = require('../shared/exit-codes')
+const buildWindowMenu = require('./menu')
 let mainWindow = null
 
 electron.app.on('window-all-closed', electron.app.quit)
@@ -34,8 +35,6 @@ electron.app.on('before-quit', () => {
 })
 
 electron.app.on('ready', () => {
-  electron.Menu.setApplicationMenu(require('./menu'))
-
   mainWindow = new electron.BrowserWindow({
     width: 800,
     height: 380,
@@ -49,6 +48,8 @@ electron.app.on('ready', () => {
     titleBarStyle: 'hidden-inset',
     icon: path.join(__dirname, '..', '..', 'assets', 'icon.png')
   })
+
+  buildWindowMenu(mainWindow)
 
   // Prevent flash of white when starting the application
   mainWindow.on('ready-to-show', mainWindow.show)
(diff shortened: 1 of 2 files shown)


lib/gui/menu.js
```diff
@@ -18,99 +18,139 @@
...
+   * @example
+   * toggleDevTools()
+   */
+  const toggleDevTools = () => {
+    if (!window) {
+      return
+    }
+
+    // NOTE: We can't use `webContents.toggleDevTools()` here,
+    // as we need to force detached mode
```
lurch: Would it also be possible to force it to open on the "Console" tab? :)

jhermsmeier: Nope, there's no option for that

lib/gui/etcher.js
```diff
@@ -50,6 +49,8 @@ electron.app.on('ready', () => {
     icon: path.join(__dirname, '..', '..', 'assets', 'icon.png')
   })
 
+  buildWindowMenu(mainWindow)
```
lurch: I'm not sure why, but I think it feels 'cleaner' to me for `./menu` to (somehow) return the menu structure / object, and then still leave it up to `etcher.js` to be responsible for actually calling `electron.Menu.setApplicationMenu` ? \*shrug\*
Although I also prefer the way that `./menu` now just returns an exported function, rather then running code automatically when you `require` it ;-)  (but again, this is just IMHO!)

jhermsmeier: Yeah, had to do this to be able to ref the mainWindow, because the previous .getFocusedWindow() could return the devtools window when open, and thus prevent toggling it on/off

lib/gui/menu.js
```diff
@@ -18,99 +18,139 @@
...
+        role: 'unhide'
+      }, {
+        type: 'separator'
+      }, {
+        role: 'quit'
+      } ]
+    })
+    menuTemplate.splice(2, 0, {
+      label: 'View',
+      submenu: [
```
jviotti: This same thing is added regardless of the platform. Can we declare it once? You can fix the `Command` and `Ctrl` differences by using an accelerator that join both (`CtrlOrCmd` or something like that) as described in the docs

jhermsmeier: Good point, the accelerator is completely different though; on Linux / Windows it's using the <kbd>Shift</kbd> modifier, whereas on Mac OS it's <kbd>Alt</kbd>

lurch: See also #2001 :wink:

jhermsmeier: @jviotti de-duplicated that submenu now

PR #2003, opened 2018-01-26
fix: specify utf-8 encoding with meta tag

Shou:
We specify the encoding to be UTF-8 with a meta tag such that Electron
won't get confused and try any other encodings.

Change-Type: patch
Changelog-Entry: Specify UTF-8 encoding with meta tag.


Shou:
This is something we got hit by recently in the Webpack PR and is common behaviour with basically any website (including etcher.io and resin.io).

Issue #2001, opened 2018-01-26
Instructions in .github/ISSUE_TEMPLATE.md need updating


lurch:
The instructions in `.github/ISSUE_TEMPLATE.md` for displaying the DevTools console only work for Etcher versions <= 1.2.1, in newer Etcher versions the key-combo has changed (but in newer versions the DevTools are _also_ accessible via the (hidden) menu).

lurch:
Fixed by #2008

PR #2008, opened 2018-01-29
doc(github): Update instructions in ISSUE_TEMPLATE

jhermsmeier:
This updates the instructions to open the Developer Tools in the issue template,
as the keyboard shortcuts have changed to their defaults on Linux & Windows
from <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>I</kbd> to <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd>.

Further, the editor config is updated to allow trailing spaces in Markdown
files to add trailing spaces to the list items in the issue template, in
order to avoid people not putting whitespace in between, causing the formatting
to not be parsed properly.

Change-Type: patch
Connects To: https://github.com/resin-io/etcher/issues/2001
See also: https://github.com/resin-io/etcher/pull/1997

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,5 +9,8 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[*.md]
+trim_trailing_whitespace = false
+
 [Makefile]
 indent_style = tab
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,6 @@
-- **Etcher version:**
-- **Operating system and architecture:**
-- **Image flashed:**
-- **Do you see any meaningful error information in the DevTools?**
+- **Etcher version:** 
+- **Operating system and architecture:** 
+- **Image flashed:** 
+- **Do you see any meaningful error information in the DevTools?** 
 
-<!-- You can open DevTools by pressing `Ctrl+Alt+I`, or `Cmd+Alt+I` if you're running OS X. -->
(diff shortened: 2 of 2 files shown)


lurch:
But this now means the instructions will be wrong for users of Etcher <= v1.2.1 - do we want that? What does the mixpanel data say about the spread of versions being used?
(If it's not too confusing, IMHO both key-combos should be documented)

> Further, the editor config is updated to allow trailing spaces in Markdown files

Given how infrequently `.github/ISSUE_TEMPLATE.md` gets edited, I'm not sure that it's worth changing `.editorconfig` for? (alternatively, could you modify `.editorconfig` to apply this exception _only_ for `.github/ISSUE_TEMPLATE.md` and not for all MarkDown files?)

jhermsmeier:
> alternatively, could you modify .editorconfig to apply this exception only for .github/ISSUE_TEMPLATE.md and not for all MarkDown files?

Trailing whitespace shouldn't be on for Markdown anyways (because it uses two trailing spaces to denote a hard line break)

jhermsmeier:
> But this now means the instructions will be wrong for users of Etcher <= v1.2.1 - do we want that?

Right, fixed!

jhermsmeier:
> What does the mixpanel data say about the spread of versions being used?

Use of v1.2.x is still significant, for some reason – makes me wonder if the update notifications worked as intended, or if people are just hesitant to update after v1.3.0...

lurch:
> Trailing whitespace shouldn't be on for Markdown anyways (because it uses two trailing spaces to denote a hard line break)

Ah yes, now that you mention that I remember asking you about the same thing on some other repo previously ;-)

PR #2010, opened 2018-01-30
feat(gui): Add ref params to homepage links in menu

jhermsmeier:
This adds a `ref` param to the URLs in the menu,
in order to see where page views are coming from.

Change-Type: patch

### DIFF
--- a/lib/gui/menu.js
+++ b/lib/gui/menu.js
@@ -84,13 +84,13 @@ const buildWindowMenu = (window) => {
         {
           label: 'Etcher Pro',
           click () {
-            electron.shell.openExternal('https://etcher.io/pro')
+            electron.shell.openExternal('https://etcher.io/pro?utm_source=etcher_menu&ref=etcher_menu')
           }
         },
         {
           label: 'Etcher Website',
           click () {
-            electron.shell.openExternal('https://etcher.io')
+            electron.shell.openExternal('https://etcher.io?ref=etcher_menu')
           }
         },
         {


lib/gui/menu.js
```diff
@@ -84,13 +84,13 @@ const buildWindowMenu = (window) => {
         {
           label: 'Etcher Pro',
           click () {
-            electron.shell.openExternal('https://etcher.io/pro')
+            electron.shell.openExternal('https://etcher.io/pro?ref=etcher_menu')
```
jhermsmeier: @tzikis what would be preferable here? `utm_source=etcher_menu`?

lurch:
See also https://github.com/resin-io/etcher-homepage/pull/89

lib/gui/menu.js
tzikis: @jhermsmeier as discussed in flowdock, i need `?utm_source=` for Mailchimp, but we can keep `?ref=` if we wanna use it on mixpanel

jhermsmeier: @tzikis aye, I've added `utm_source` now, while also keeping the `ref` for Etcher Pro

lurch:
Looks like the lint tests are causing the CI builds to fail?
```
node scripts/html-lint.js
Scanning...
3 errors at lib/gui/app/index.html
  [17:69] An element with the attribute “tabindex” must not appear as a descendant of the “button” element.
  [24:59] An element with the attribute “tabindex” must not appear as a descendant of the “button” element.
  [31:68] An element with the attribute “tabindex” must not appear as a descendant of the “button” element.
Total: 7/8
make: *** [lint-html] Error 1
```

PR #2012, opened 2018-02-01
feat: replace template paths with contents

Shou:
We replace the `templateUrl` fields with `template` fields and thus
switch from template paths to template contents in preparation for the
Webpack PR.

Changelog-Entry: Replace template paths with template contents.
Change-Type: patch

### DIFF
--- a/tests/gui/pages/main.spec.js
+++ b/tests/gui/pages/main.spec.js
@@ -18,6 +18,7 @@
 
 const m = require('mochainon')
 const _ = require('lodash')
+const fs = require('fs')
 const path = require('path')
 const supportedFormats = require('../../../lib/shared/supported-formats')
 const angular = require('angular')
@@ -25,6 +26,14 @@ const flashState = require('../../../lib/shared/models/flash-state')
 const availableDrives = require('../../../lib/shared/models/available-drives')
 const selectionState = require('../../../lib/shared/models/selection-state')
 require('angular-mocks')
+
+// Mock HTML requires by reading from the file-system
+// eslint-disable-next-line node/no-deprecated-api
+require.extensions['.html'] = (module, filename) => {
+  module.exports = fs.readFileSync(filename, {
+    encoding: 'utf8'
+  })
+}
 
 describe('Browser: MainPage', function () {
   beforeEach(angular.mock.module(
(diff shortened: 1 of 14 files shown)


lurch:
Picking up the changes from https://github.com/resin-io/etcher/pull/1943/files#diff-be8e7cbae4a85d432ac97a2164bcef95 might allow the CI tests to pass ;-)

tests/gui/pages/main.spec.js
```diff
@@ -241,4 +250,22 @@ describe('Browser: MainPage', function () {
...
   })
+
+  describe('page template', function () {
+    let $state
+
+    beforeEach(angular.mock.inject(function (_$state_) {
+      $state = _$state_
+    }))
+
+    it('should match the file contents', function () {
```
lurch: Is this also a temporary thing, or will it remain after webpack has been integrated?

Shou: What's in the test file should remain as-is with Webpack. We will only remove the custom HTML `request` handler outside `tests/` as Webpack takes care of that.

lurch: In that case is it also worth testing the other template files? Or is it safe to assume that as long as one got loaded okay, then all the others will have got loaded okay?

Shou:
@lurch thanks! Here it is with the sanity script changes.

tests/gui/pages/main.spec.js
Shou: Yeah that's a good idea, we could certainly test the other pages in the same manner, we just need to create test files for them. Hopefully there's a nice way to test the modals too.

Shou:
I've added tests for other pages, unfortunately I can't find a nice way to test the modal(s), but I did add some tests related to the template in any case. Let me know what you think.

Shou:
Tests pass here now, let me know what you think.

tests/gui/pages/main.spec.js
```diff
@@ -26,6 +27,14 @@ const availableDrives = require('../../../lib/shared/models/available-drives')
 const selectionState = require('../../../lib/shared/models/selection-state')
 require('angular-mocks')
 
+// Mock HTML requires by reading from the file-system
+// eslint-disable-next-line node/no-deprecated-api
+require.extensions['.html'] = (module, filename) => {
```
lurch: Out of curiosity, why is this snippet needed in `main.spec.js` but not in `finish.spec.js` or `settings.spec.js`?

PR #2013, opened 2018-02-01
feat: log the event status of the banner

Shou:
We add log the banner HTTP load event object to the analtytics.

Change-Type: patch
Changelog-Entry: Log the banner load event to analytics.

### DIFF
--- a/lib/gui/app/components/safe-webview.js
+++ b/lib/gui/app/components/safe-webview.js
@@ -194,12 +194,11 @@ class SafeWebview extends react.PureComponent {
     // only care about this event if it's a request for the main frame
     if (event.resourceType === 'mainFrame') {
       const HTTP_OK = 200
-      const HTTP_ERR = 400
 
       analytics.logEvent(event)
 
       this.setState({
-        shouldShow: event.httpResponseCode >= HTTP_OK && event.httpResponseCode < HTTP_ERR
+        shouldShow: event.httpResponseCode === HTTP_OK
       })
     }
   }
--- a/lib/gui/app/components/safe-webview.js
+++ b/lib/gui/app/components/safe-webview.js
@@ -190,12 +190,18 @@ class SafeWebview extends react.PureComponent {
    * @param {Event} event - Event object
    */
   didGetResponseDetails (event) {
-    const HTTP_OK = 200
-    const HTTP_ERR = 400
+    // This seems to pick up all requests related to the webview,
+    // only care about this event if it's a request for the main frame
+    if (event.resourceType === 'mainFrame') {
+      const HTTP_OK = 200
+      const HTTP_ERR = 400
 
-    this.setState({
-      shouldShow: event.httpResponseCode >= HTTP_OK && event.httpResponseCode < HTTP_ERR
-    })
+      analytics.logEvent(event)
+
+      this.setState({
+        shouldShow: event.httpResponseCode >= HTTP_OK && event.httpResponseCode < HTTP_ERR
+      })
+    }
   }
 
   /**


Shou:
Here's an example of what will get logged with `analytics.logEvent`, maybe we should pick specific fields to log?

```JSON
{
  "isTrusted": false,
  "status": false,
  "newURL": "https://etcher.io/success-banner/?etcher-version=1.3.1&api-version=1",
  "originalURL": "https://etcher.io/success-banner/?etcher-version=1.3.1&api-version=1",
  "httpResponseCode": 200,
  "requestMethod": "GET",
  "referrer": "",
  "headers": {
    "access-control-allow-origin": [
      "*"
    ],
    "cache-control": [
      "max-age=600"
    ],
    "cf-ray": [
      "3e65ae39ebae06be-LHR"
    ],
    "content-encoding": [
      "gzip"
    ],
    "content-type": [
      "text/html; charset=utf-8"
    ],
    "date": [
      "Thu, 01 Feb 2018 14:43:00 GMT"
    ],
    "expect-ct": [
      "max-age=604800, report-uri=\"https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct\""
    ],
    "expires": [
      "Mon, 29 Jan 2018 13:01:43 GMT"
    ],
    "last-modified": [
      "Tue, 23 Jan 2018 19:53:02 GMT"
    ],
    "server": [
      "cloudflare"
    ],
    "status": [
      "200"
    ],
    "vary": [
      "Accept-Encoding"
    ],
    "x-github-request-id": [
      "8909:4159:2332F1:31DDE1:5A732774"
    ]
  },
  "resourceType": "mainFrame"
}
```

lib/gui/app/components/safe-webview.js
```diff
@@ -191,12 +191,18 @@ class SafeWebview extends react.PureComponent {
    * @param {Event} event - Event object
    */
   didGetResponseDetails (event) {
-    const HTTP_OK = 200
-    const HTTP_ERR = 400
+    // This seems to pick up all requests related to the webview,
+    // only care about this event if it's a request for the main frame
```
jhermsmeier: Does this trigger for all requests? Like also loading of HTML / CSS / images etc.?

Shou: Yup, that's what I noticed now... I wonder if we should keep that for core requests (stylesheet, important scripts) to guarantee a good banner experience.

lurch: @Shou Do you mean turn on the online banner if the HTML page loads okay, but then turn off the online banner if any of the CSS or image files fail to load? Maybe it's better to leave it always showing as soon as the HTML has loaded, to prevent any "flickering" if a user is on a slow connection and the last resource fails to load? \*shrug\*

lurch: Although AFAICT all the resources on the HTML page are hosted on the same etcher.io domain anyway, so the chance of the HTML loading but any of the resources failing to load seems _very_ unlikely?

Shou: I looked at all the requests and it seems we should be safe with assuming if the banner HTML loads, so will the other content. The things that might _not_ load are analytics related and as such not something that impacts users directly, so we should show the banner.

Shou:
Let me know if the fields picked are good if we want more for the analytics

lib/gui/app/components/safe-webview.js
```diff
@@ -191,12 +191,25 @@ class SafeWebview extends react.PureComponent {
...
+    if (event.resourceType === 'mainFrame') {
+      const HTTP_OK = 200
+      const HTTP_ERR = 400
+
+      const pickedHeaders = _.pick(event.headers, [
+        'status'
+      ])
+      const pickedEvent = _.pick(_.assign({}, event, pickedHeaders), [
+        'httpResponseCode',
+        'requestMethod'
```
jhermsmeier: If I'm not mistaken, this will exclude the previously picked "status" header (which would make it redundant)? Also, what does the status header contain?

Shou:
Reverted the `_.pick` commit now, instead we log the whole event object. Let me know what you think.

lib/gui/app/components/safe-webview.js
```diff
@@ -191,12 +191,18 @@ class SafeWebview extends react.PureComponent {
...
+      const HTTP_OK = 200
+      const HTTP_ERR = 400
 
-    this.setState({
-      shouldShow: event.httpResponseCode >= HTTP_OK && event.httpResponseCode < HTTP_ERR
-    })
+      analytics.logEvent(event)
+
+      this.setState({
+        shouldShow: event.httpResponseCode >= HTTP_OK && event.httpResponseCode < HTTP_ERR
```
jhermsmeier: We should only show the banner if the HTTP status code is 200, and no other – also, @Shou did you check the Cloudflare docs for the error states when the backing server is unavailable?

PR #2014, opened 2018-02-01
feat: add icon next to drive on warnings

Shou:
We add an icon next to the drive size that is displayed when there is a
drive-image compatibility status message available. We display the first
one in the list and importance is then enforced by the order they are
added to the list in `drive-constraints`.

Change-Type: patch
Changelog-Entry: Add icon next to drive size when compatibility warnings exist.

### DIFF
--- a/lib/gui/app/pages/main/templates/main.tpl.html
+++ b/lib/gui/app/pages/main/templates/main.tpl.html
@@ -64,13 +64,16 @@
           <div class="step-selection-text"
             ng-class="{
               'text-disabled': main.shouldDriveStepBeDisabled()
-            }"
-            uib-tooltip="{{ main.selection.getDrive().description }} ({{ main.selection.getDrive().displayName }})">
-            <span class="step-drive step-name">
+            }">
+            <span class="drive-step step-name"
+              uib-tooltip="{{ main.selection.getDrive().description }} ({{ main.selection.getDrive().displayName }})">
               <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
               {{ (main.selection.getDrive().description || "") | middleEllipses:11 }}
             </span>
             <span class="step-drive step-size">{{ main.selection.getDrive().size | closestUnit }}</span>
+            <span class="step-drive step-warning glyphicon glyphicon-exclamation-sign"
+              uib-tooltip="{{ main.constraints.getDriveImageCompatibilityStatuses(main.selection.getDrive(), main.selection.getImage())[0].message }}"
+              ng-show="main.constraints.hasDriveImageCompatibilityStatus(main.selection.getDrive(), main.selection.getImage())"></span>
           </div>
           <button class="button button-link step-footer"
             tabindex="{{ main.selection.hasDrive() ? 2 : -1 }}"
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -294,7 +294,7 @@ exports.COMPATIBILITY_STATUS_MESSAGES = {
    * The image and drive compatibility is not recommended; happens when the
    * actual drive size is smaller than the image's recommended drive size.
    */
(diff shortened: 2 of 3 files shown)


Shou:
image

Shou:
@konmouz how do you feel about the warning icon used here? I know your mockup uses a triangle instead, but the Bootstrap triangle icon isn't that pretty (see below). Is this fine, or should we use the icon you used in the mockup?

image

Shou:
Here's the mockup @konmouz posted

image

konmouz:
@Shou, this is fine with me

lib/shared/drive-constraints.js
```diff
@@ -294,7 +294,7 @@ exports.COMPATIBILITY_STATUS_MESSAGES = {
    * The image and drive compatibility is not recommended; happens when the
    * actual drive size is smaller than the image's recommended drive size.
    */
-  SIZE_NOT_RECOMMENDED: 'Not Recommended',
+  SIZE_NOT_RECOMMENDED: 'Unrecommended Size',
```
jhermsmeier: The wording here sounds a bit weird to me, and it also doesn't convey the reason it's not recommended. Can we change this to something with more detail?

lurch: @jhermsmeier AFAIK this warning is only ever displayed for images that have a 'recommended size' set in their metadata manifest file (which we used to document, but now don't, because we decided that the "online catalog" was going to be storing all that info instead), and the _only_ images I'm aware of that use the manifest file are the images from http://www.ev3dev.org/downloads/ (they were an "early adopter" of Etcher). I've just checked a freshly downloaded ev3dev image, and the only fields they set in their manifest are `name`, `version`, `url` & `supportUrl` (and they also provide a logo and a blockmap).

So while I agree that 'Unrecommended Size' sounds weird, I don't think anybody will ever actually see it ;-)
(Just a little bit of historical context for you :slightly_smiling_face: )

jhermsmeier: I'm aware of the history; but still – what's the point if this is not going to be used? Also, I think we should do it nicely, otherwise we'll forget about changing this later.

lurch: I think it's something that @Shou suggested before, but given that this has/will never be used, what if this "recommended size" stuff was just completely removed (obviously that'd need to be done in a separate PR to this one, so apologies for being OT, but this seems the best place ATM to discuss this).

Shou: I guess there's a larger discussion surrounding this so I've made this issue: https://github.com/resin-io/etcher/issues/2017. I'll leave the message alone in this PR by rolling back that change and we should get this through faster.

lib/gui/app/pages/main/templates/main.tpl.html
```diff
@@ -64,13 +64,16 @@
...
-            <span class="step-drive step-name">
+            }">
+            <span class="drive-step step-name"
+              uib-tooltip="{{ main.selection.getDrive().description }} ({{ main.selection.getDrive().displayName }})">
               <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
               {{ (main.selection.getDrive().description || "") | middleEllipses:11 }}
             </span>
             <span class="step-drive step-size">{{ main.selection.getDrive().size | closestUnit }}</span>
+            <span class="step-drive step-warning glyphicon glyphicon-exclamation-sign"
+              uib-tooltip="{{ main.constraints.getDriveImageCompatibilityStatuses(main.selection.getDrive(), main.selection.getImage())[0].message }}"
```
jhermsmeier: Will this code still be run if `ng-show` is `false`?

Shou: I checked this, and it does – luckily the fix is simple and I'll open a new PR for that.

Shou: https://github.com/resin-io/etcher/pull/2021

PR #2021, opened 2018-02-07
feat(GUI): replace ng-show and ng-hide with ng-if

Shou:
We replace `ng-show/hide` with `ng-if` on the main page in order to
remove unnecessary calls and become more efficient.

Changelog-Entry: Replace ng-show/hide with ng-if on main page.
Change-Type: patch

### DIFF
--- a/lib/gui/app/pages/main/templates/main.tpl.html
+++ b/lib/gui/app/pages/main/templates/main.tpl.html
@@ -7,7 +7,7 @@
       </div>
 
       <div class="space-vertical-large">
-        <div ng-hide="main.selection.hasImage()">
+        <div ng-if="!main.selection.hasImage()">
           <button
             class="button button-primary button-brick"
             tabindex="{{ main.selection.hasImage() ? -1 : 1 }}"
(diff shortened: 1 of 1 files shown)


lurch:
Should this same change also be made to the other pages that use `ng-show` and `ng-hide`?
(`drive-selector-modal.tpl.html`, `settings.tpl.html` & `index.html`)

Shou:
@lurch those don't have as potential of an impact as the ones on the main page so we don't have to worry, esp. considering they're generally rather transient pages/components.

jhermsmeier:
@Shou actually we do kinda have to worry – a lot of these expressions within a ng-show/hide block access properties on objects that would be undefined when the condition isn't true, potentially causing reference errors

Shou:
@jhermsmeier Angular ignores errors that happen in its templates and just returns an empty string, which isn't ideal, but React changes that behaviour, and we do have to be more careful when we rewrite things in React. The `ngHide/ngShow`s in the other templates are small and most of the benefit would be from restructuring the tags so that the attribute expressions would be inside the `ngHide/ngShow`s (rather than on the same tag) and we could turn them into `ngIf`s.

jhermsmeier:
@Shou okay, still something to keep in mind I guess

PR #2022, opened 2018-02-07
fix: move tabindex attrs to button from within

Shou:
We move the `tabindex` attributes to the button element directly from
elements contained within the button element – this is to satisfy the
HTML linter.

Changelog-Entry: Move tabindex attributes to button elements from
contained elements.
Change-Type: patch

### DIFF
--- a/lib/gui/app/index.html
+++ b/lib/gui/app/index.html
@@ -13,22 +13,22 @@
     <header class="section-header" ng-controller="HeaderController as header">
       <button class="button button-link"
         ng-click="header.openHelpPage()"
-        tabindex="-1">
-        <span tabindex="4" class="glyphicon glyphicon-question-sign"></span>
+        tabindex="4">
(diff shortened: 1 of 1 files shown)


lurch:
Yeah, from a logical point of view, I guess it makes more sense to have the tabindex on the buttons rather than on spans. Looking back through repo history, it seems these tab-indices were added with #1745 
The overly-strong outline-highlighting from that PR got removed in #1854 but does this PR change how the current tab-highlighting (if any) gets displayed within Etcher?

jhermsmeier:
> The overly-strong outline-highlighting from that PR got removed in #1854 but does this PR change how the current tab-highlighting (if any) gets displayed within Etcher?

Nope, still same old as far as I can tell

Shou:
Yup, the `:hover` and `:focus` effects still happen

Issue #1991, opened 2018-01-24
Error couldn't clean the drive...


cwsyt7:
- **Etcher version:**1.2.1
- **Operating system and architecture:** Windows 10 Home v. 1703, built-in SD card reader (Samsung 32 GB micro SD card in adapter)
- **Image  #flashed:** Retropie 4.3
- **Do you see any meaningful error information in the DevTools?**  Yes, although I'm not sure what it means.

I am trying to install the retropie onto my SD card, however Etcher keeps hanging on "starting" for a few seconds then erroring out (see below). So far I have tried to reinstall Etcher a couple times, tried to run Etcher as both administrator (I get the error) and not as the administrator (just hangs on "starting"). I tried cleaning the SD card with diskpart, however I keep getting this error. I can manually copy files onto the SD card from within windows, so the SD card isn't locked out.

Any advice that can be offered would be greatly appreciated.

Thank you.

Windows Error Message:

![image](https://user-images.githubusercontent.com/35754118/35313926-ff5214f6-0090-11e8-8668-5f9b08cccad5.png)

Etcher DevTools

![image](https://user-images.githubusercontent.com/35754118/35313942-12fbb566-0091-11e8-8855-fae49920599c.png)

jhermsmeier:
Hey, this is a known issue with Windows; the drive cleaning will be retried 5 times, but sometimes it takes more than that, or Windows won't do it at all (and would need a restart to do so). It's surprisingly fragile.

One way you can work around this is with Disk Management; first remove all volumes on the device there, and then retry flashing it.

jhermsmeier:
You also might want to try Etcher v1.3.1, which we've released yesterday

cwsyt7:
Thanks for the info, my google fu didn't turn up anything. So if I keep retrying it, eventually it will work?

cwsyt7:
Ok, I downloaded the updated version of Etcher, but now I'm getting a new error (same PC hardware, Retropie version).

Popup error window:

![image](https://user-images.githubusercontent.com/35754118/35370144-056e9f66-015a-11e8-90df-9dcd8a63284c.png)

Code:

![image](https://user-images.githubusercontent.com/35754118/35370162-19441688-015a-11e8-9811-d2c2676cfbe3.png)

lurch:
Thanks @cwsyt7 . Please could you switch to the "Console" tab of the DevTools window, and copy'n'paste the whole thing here please?

cwsyt7:
[...]

Issue #1793, opened 2017-10-19
Couldn't clean the drive, Command Failed: diskpart /s   --- WIn 8.1


MountainLogic:
- **Etcher version:** 1.1.2
- **Operating system and architecture:** WIn 8.1 on bootcamp using no-name external reader 
- **Image flashed:** Beagle bone black Ubuntu image
- **Note,** in Windows Explorer, the drive icon & lebel "boot (e:)" on left appeared and disappeared several times before error message.  Also Viper anti-virus scanned with each cycle.  
- **Note,** Attempted several times, including as admin with same results.  Also tried with anti-virus turned off with no success.  Also tried in unsafe mode with no success

- **Do you see any meaningful error information in the DevTools?**  
Couldn't clean the drive, Command Failed: diskpart

![image](https://user-images.githubusercontent.com/13784198/31747884-6b1d86a2-b424-11e7-9c51-e1b8fe84d4cf.png)

jhermsmeier:
ref #1054

jhermsmeier:
Thanks for the additional info on the issue – we thought we'd squashed it already

lurch:
@jhermsmeier It looks like #989 is a better duplicate for this than #1054 ;-)
#1054 is about the `.bat` file failing, whereas this PR and #989 are specifically about `diskpart /s` failing. But I guess this is another one of those problems that'll be fixed when we manage to get native drive-locking into place?

jhermsmeier:
Hmm, yeah, maybe – this doesn't fail due to a timeout or running out of retries – might be entirely new thing. But yes, this could possibly be fixed by drive-locking.

cybertonian:
I am getting the "Couldn't clean the drive..." (code: 2147942401) on two different drives.  The only difference between when this used to work and now is that I installed the latest patches to Windows10 (KB4043901).  Is there an issue with this MS update?

MountainLogic:
FWIW, I started to notice that Windows kept asking with a pop-up to format the drive and just before I saw the pop-up, is when I got the error in etcher.  Also saw an generic write error when using Win32DiskImager and Rufus.  Eventually, I kept thrashing and was able to get it flashed with Win32DiskImager.  Not sure how.

DuaneEllis-TI:
FYI - I had the exact same problem

In my case - this is what happened and how I fixed it.

In my case I am using  FULL SIZED sd card adapter
With a small micro sd card installed in the adapter.

[...]

PR #2023, opened 2018-02-07
feat(lib): Use win-drive-clean instead of diskpart

jhermsmeier:
This replaces shelling out to `diskpart` on Windows to clear
the partition table with `win-drive-clean`, which does so via DeviceIoControl.

**TODO:**
- [x] Changelog entry
- [x] Publish `win-drive-clean` to npm & update package.json & shrinkwrap
- [x] Evaluate whether this actually resolves the issues with diskpart on Windows

Change-Type: minor
Changelog-Entry: Use native code to clean drives on Windows
Connects To: #1793 #1991

### DIFF
--- a/lib/cli/diskpart.js
+++ b/lib/cli/diskpart.js
@@ -16,70 +16,13 @@
 
 'use strict'
 
-const os = require('os')
-const fs = require('fs')
-const path = require('path')
-const crypto = require('crypto')
-const childProcess = require('child_process')
 const debug = require('debug')('etcher:cli:diskpart')
 const Promise = require('bluebird')
 const retry = require('bluebird-retry')
+const driveClean = require('win-drive-clean')
 
-const TMP_RANDOM_BYTES = 6
 const DISKPART_DELAY = 2000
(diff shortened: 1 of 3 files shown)


lib/cli/diskpart.js
```diff
@@ -93,32 +36,23 @@ module.exports = {
...
+        driveClean(device, (error) => {
+          return error ? reject(error) : resolve()
+        })
+      }).delay(DISKPART_DELAY)
+    }, {
+      /* eslint-disable camelcase */
+      max_tries: DISKPART_RETRIES
+      /* eslint-enable camelcase */
+    }).catch((error) => {
+      throw error
```
zvin: This `catch` then `throw` looks useless.

jhermsmeier: Uhm... that's because it is – fixed!

jhermsmeier:
@lurch @jviotti can you guys give this a test run on your Windows machines if you have time? I did a bunch of flashes, and it works extremely well for me so far.

PR #2026, opened 2018-02-08
upgrade(package): Update eslint 3.19.0 -> 4.17.0

Shou:
We upgrade eslint for better support of the object-curly-newline
options, and change the quantity of elements required before the linter
enforces newlines on objects, in addition we apply `consistent` to avoid
errors when there are newlines in objects with less than 2 elements.

```js
const { field } = object

const {
  field
} = object

const {
  field1,
  field2
} = object
````

Change-Type: patch
Changelog-Entry: Upgrade eslint to use object-curly-newline options.

### CODE REVIEW

File: .eslintrc.yml [hunk 1]
```diff
@@ -254,7 +254,8 @@ rules:
     - allowAfterThis: false
   object-curly-newline:
     - error
-    - minProperties: 1
+    - minProperties: 2
```
jhermsmeier: Should we expand this to 3?

jhermsmeier:
I just noticed, this only updates to 4.16.0, not 4.17.0 – is that intentional?

.eslintrc.yml
Shou: Sounds good with me

Shou:
4.17.0 must've come out recently – yeah we might as well change it to that then

jhermsmeier:
It's been out since Feb 2 – how long have you had this branch sitting around? ;P

Shou:
@jhermsmeier looks npm is complaining about 4.17.0 – [the changelog][1] doesn't seem to have anything crucial for us right now in any case so I'm gonna go ahead and merge this once the tests pass.

```fish
<@etcher-alt>-<⎇ upgrade-eslint>-<*>-> npm install eslint@latest --dev                
npm WARN install Usage of the `--dev` option is deprecated. Use `--only=dev` instead. 
etcher@1.3.1 /home/benedict/Prog/Javascript/etcher-alt                                
└── eslint@4.17.0  invalid      
```

[1]: https://eslint.org/blog/2018/02/eslint-v4.17.0-released

jhermsmeier:
That's because you're not saving & shrinkwrapping it, also the --dev isn't what you're looking for; you'll want to do the following when updating deps in Etcher:

```
npm install --save-dev --save-exact eslint@4.17.0
```

Shou:
Ah, I got confused with npm's save by default but that doesn't apply to our version of npm. Thanks.

PR #2029, opened 2018-02-09
refactor(lib): Remove unused robot protocol

jhermsmeier:
This removes the remnants of the unused "robot protocol",
which was previously used by the CLI, but has been superseded by IPC.

Change-Type: minor
Changelog-Entry: Remove unused robot protocol


lib/gui/app/components/safe-webview.js
```diff
@@ -221,28 +220,10 @@ class SafeWebview extends react.PureComponent {
...
-    }
-
-    const message = robot.parseMessage(event.message)
-
-    if (robot.getCommand(message) === robot.COMMAND.LOG) {
-      analytics.logEvent(robot.getData(message))
-    } else if (robot.getCommand(message) === robot.COMMAND.ERROR) {
-      analytics.logException(robot.getData(message))
-    }
+    analytics.logEvent(event.message)
```
jhermsmeier: @Shou what was the use-case of a WebView sending robot commands? Do we still need to support something here?

lib/gui/app/components/safe-webview.js
```diff
@@ -221,27 +220,24 @@ class SafeWebview extends react.PureComponent {
...
+   * console.log('Good night!')
    */
   static consoleMessage (event) {
-    if (!robot.isMessage(event.message)) {
+    if (_.isNil(event.message)) {
       return
     }
 
-    const message = robot.parseMessage(event.message)
+    let message = event.message
```
Shou: Huh I thought this'd be a linter error but I guess you need `prefer-destructuring` for it to start enforcing `let { message } = event`

jhermsmeier: I don't think we'd want to force the use of destructuring – while it's useful in some situations, it doesn't have to be used everywhere just for the sake of it.

Shou: I don't think we should have any problems with enforcing it when we have renamings, e.g. `let { name: newName } = object`, but yeah I was just under the impression it was already on, I'm fine with it as-is.

lurch:
> previously used by the CLI, but has been superseded by IPC

LOL, replacing one TLA with another TLA, FTW! :slightly_smiling_face:

PR #2031, opened 2018-02-09
chore(package): Remove unnecessary deps

jhermsmeier:
This removes unnecessary or unused dependencies; `trackjs` hasn't been in use for a very long time, and `file-exists` was a single use dependency, which can be replaced with a single line of code.

### DIFF
--- a/tests/image-stream/tester.js
+++ b/tests/image-stream/tester.js
@@ -34,17 +34,17 @@ const doFilesContainTheSameData = (file1, file2) => {
 }
 
 const deleteIfExists = (file) => {
-  return Bluebird.try(() => {
+  return new Bluebird((resolve, reject) => {
     try {
       fs.accessSync(file)
       fs.unlinkSync(file)
     } catch (error) {
       if (error.code !== 'ENOENT') {
-        return Bluebird.reject(error)
+        return reject(error)
       }
     }
 
-    return Bluebird.resolve()
+    resolve()
   })
 }
 
--- a/tests/image-stream/tester.js
+++ b/tests/image-stream/tester.js
@@ -19,8 +19,7 @@
 const m = require('mochainon')
 const _ = require('lodash')
 const Bluebird = require('bluebird')
-const fileExists = require('file-exists')
-const fs = Bluebird.promisifyAll(require('fs'))
+const fs = require('fs')
 const os = require('os')
 const path = require('path')
 const imageStream = require('../../lib/image-stream/index')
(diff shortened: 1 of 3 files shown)


tests/image-stream/tester.js
```diff
@@ -35,9 +34,14 @@ const doFilesContainTheSameData = (file1, file2) => {
 }
 
 const deleteIfExists = (file) => {
```
zvin: Why should we test if the file exists?
Couldn't we just delete it in the try block?

jhermsmeier: I think the reasoning behind it was that we want it to fail should the file not have been written, or not been written to that location.

zvin:
```javascript
return fs.unlinkAsync(file)
.catchReturn(function(error) {
  return (error.code === 'ENOENT')
})
```

zvin: Otherwise somethig could happen to the file between `access` and `unlink`.

jhermsmeier: Not sure I follow; both the access & unlink are in the `try` block

zvin: You're right, this will work as expected.
It is just that the call to access is useless.
From the node fs.access docs:
>Using fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

jhermsmeier: True that, fixing

tests/image-stream/tester.js
```diff
@@ -35,12 +34,16 @@ const doFilesContainTheSameData = (file1, file2) => {
...
-  return Bluebird.try(function () {
-    if (fileExists(file)) {
-      return fs.unlinkAsync(file)
+  return new Bluebird((resolve, reject) => {
+    try {
+      fs.unlinkSync(file)
```
zvin: It seems strange to use `unlinkSync` and transform it to a `Promise` instead of `unlinkAsync`.

tests/image-stream/tester.js
zvin:
```javascript
return fs.unlinkAsync('klklklklklklklklklk')
.catch({ code: 'ENOENT' }, _.noop)
```

jhermsmeier: Fair enough; updated!

PR #2036, opened 2018-02-12
feat(GUI): add descriptive name to modals

Shou:
We add a descriptive name to modals for analytics and debugging.

Change-Type: patch
Changelog-Entry: Add descriptive name to modals for analytics and
debugging.

### DIFF
--- a/lib/gui/app/components/modal/services/modal.js
+++ b/lib/gui/app/components/modal/services/modal.js
@@ -34,6 +34,7 @@ module.exports = function ($uibModal, $q) {
    *
    * @example
    * ModalService.open({
+   *   name: 'my modal',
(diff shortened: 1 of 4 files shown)


lib/gui/app/components/modal/services/modal.js
```diff
@@ -44,7 +45,7 @@ module.exports = function ($uibModal, $q) {
     })
 
     analytics.logEvent('Open modal', {
-      template: options.template
+      name: options.name
```
jhermsmeier: Should we do the same for "close" events?

Shou: Yeah that's a good idea, done now

PR #2037, opened 2018-02-12
refactor: rename selection state and store functions

Shou:
We rename functions in `selection-state` and the `store`.

Change-Type: patch
Changelog-Entry: Rename selection-state and store functions.

### DIFF
--- a/lib/shared/models/selection-state.js
+++ b/lib/shared/models/selection-state.js
@@ -49,7 +49,7 @@ exports.selectDrive = (driveDevice) => {
  */
 exports.toggleDrive = (driveDevice) => {
   if (exports.isCurrentDrive(driveDevice)) {
-    exports.unselectDrive()
+    exports.deselectDrive()
   } else {
     exports.selectDrive(driveDevice)
   }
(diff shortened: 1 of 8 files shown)


lib/shared/models/selection-state.js
```diff
@@ -21,49 +21,49 @@ const store = require('../store')
...
  * @example
- * selectionState.toggleSetDrive('/dev/disk2');
+ * selectionState.toggleDrive('/dev/disk2');
  */
-exports.toggleSetDrive = (drive) => {
-  if (exports.isCurrentDrive(drive)) {
-    exports.removeDrive()
+exports.toggleDrive = (driveDevice) => {
+  if (exports.isCurrentDrive(driveDevice)) {
+    exports.unselectDrive()
```
jhermsmeier: Can we use "deselect" here ("unselect" isn't to be found in dictionaries)?

Shou: Done

Shou:
(isn't GH supposed to dismiss stale reviews automatically?)

PR #2042, opened 2018-02-14
fix(Makefile): Support arm64 / armv8 / aarch64

jhermsmeier:
This adds support for 64bit arm cpu architectures.

Change-Type: patch
Changelog-Entry: Support building Etcher on armv8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,15 @@ else
 		ifeq ($(shell uname -m),armv7l)
 			HOST_ARCH = armv7hf
 		endif
+		ifeq ($(shell uname -m),aarch64)
+			HOST_ARCH = aarch64
+		endif
+		ifeq ($(shell uname -m),armv8)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -51,6 +51,9 @@ else
 		ifeq ($(shell uname -m),armv7l)
 			HOST_ARCH = armv7hf
 		endif
+		ifeq ($(shell uname -m),aarch64)
```
jviotti: I think we can support arm64 builds, now that electron supports that

jhermsmeier: Updated!

jviotti:
@jhermsmeier Wait. I think you need to update the architecture convert shell script, otherwise it won't know how to handle arm64.

lurch:
Sweet! I'll have to get an ARM64 distro set up on one of my RPi3s and have a play... :wink:

PR #2043, opened 2018-02-14
refactor(GUI): remove selection-state clear options to simplify

Shou:
We refactor and simplify the selection-state `.clear()` by removing the
options argument.

Change-Type: patch
Changelog-Entry: Remove `selectionState.clear()` options argument to
simplify.

### DIFF
--- a/lib/shared/models/selection-state.js
+++ b/lib/shared/models/selection-state.js
@@ -304,25 +304,15 @@ exports.deselectImage = () => {
  * @function
  * @public
  *
- * @param {Object} options - options
- * @param {Boolean} [options.preserveImage] - preserve image
- *
  * @example
  * selectionState.clear();
  *
  * @example
  * selectionState.clear({ preserveImage: true });
  */
-exports.clear = (options = {}) => {
-  if (!options.preserveImage) {
-    store.dispatch({
-      type: store.Actions.DESELECT_IMAGE
-    })
-  }
-
-  store.dispatch({
-    type: store.Actions.DESELECT_DRIVE
-  })
+exports.clear = () => {
+  exports.deselectImage()
+  exports.deselectDrive()
 }
 
 /**
--- a/tests/shared/models/selection-state.spec.js
+++ b/tests/shared/models/selection-state.spec.js
@@ -853,7 +853,7 @@ describe('Model: selectionState', function () {
     })
   })
 
-  describe('given a drive', function () {
+  describe('given a drive and an image', function () {
     beforeEach(function () {
       availableDrives.setDrives([
         {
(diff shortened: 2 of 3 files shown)


tests/shared/models/selection-state.spec.js
```diff
@@ -890,39 +890,6 @@ describe('Model: selectionState', function () {
         m.chai.expect(selectionState.hasImage()).to.be.false
       })
     })
-
-    describe('given the preserveImage option', function () {
```
jviotti: So is this functionality not needed anymore?

Shou: The `preserveImage` option is only used in one instance – https://github.com/resin-io/etcher/pull/2043/files#diff-be4cad3a825aa7d1fd460dd90806eda2R48 – so I think it makes sense to just move the logic there where it's used instead.

jviotti: Alright. In that case, can we move these test cases to a `test/gui/pages/finish.spec.js` file?

jviotti: Ah, but its going to be tricky to test that given `$state` and `analytics` :( @jhermsmeier What do you think?

Shou: Maybe we can sinon spy on `selectionState.deselectImage` and `selectionState.deselectDrive`?

jviotti: I'd like to avoid spies as most as possible. If we really want to test this, then we should split `finish.restart()` into two functions and test the one that deselects stuff separately, but then its a simple enough logic that I think we should instead test with something like Spectron.

jhermsmeier: @Shou wdyt about what @jviotti said?

Shou: @jhermsmeier @jviotti what if we just rename the existing test block from `.clear()` to `.deselectImage() and .deselectDrive()` given the intention is to test that these two work together?

jhermsmeier: Sounds good to me

tests/shared/models/selection-state.spec.js
```diff
@@ -853,7 +853,7 @@ describe('Model: selectionState', function () {
     })
   })
 
-  describe('given a drive', function () {
+  describe('given a drive and an image', function () {
```
Shou: This was an existing typo, 'given a drive' already exists above.

Shou:
So it wouldn't make sense to re-use the `.clear()` tests – instead I have added cases ensuring that `.deselectImage()` doesn't touch drives and vice-versa for `.deselectDrive()`. This should cover the logic added here.

lib/shared/models/selection-state.js
```diff
@@ -304,25 +304,15 @@ exports.deselectImage = () => {
...
  * @public
  *
- * @param {Object} options - options
- * @param {Boolean} [options.preserveImage] - preserve image
- *
  * @example
  * selectionState.clear();
  *
  * @example
  * selectionState.clear({ preserveImage: true });
```
lurch: I guess this example needs updating too?

PR #2044, opened 2018-02-14
feat(GUI): warning makes drive-selector button orange

Shou:
We make the drive-selector button orange when there is a warning attached
to the image-drive pair.

Change-Type: patch
Changelog-Entry: Make the drive-selector button orange on warnings.

### DIFF
--- a/lib/gui/app/scss/components/_button.scss
+++ b/lib/gui/app/scss/components/_button.scss
@@ -69,6 +69,10 @@ $button-types-styles: (
   danger: (
     bg: $palette-theme-danger-background,
     color: $palette-theme-danger-foreground
+  ),
+  warning: (
+    bg: $palette-theme-warning-background,
+    color: $palette-theme-danger-foreground
   )
 );
 
--- a/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
+++ b/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
@@ -55,6 +55,9 @@
 <div class="modal-footer">
   <button class="button button-primary button-block"
     tabindex="{{ 15 + modal.getDrives().length }}"
+    ng-class="{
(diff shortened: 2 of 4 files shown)


Shou:
image

jhermsmeier:
This feels like it might be confusing, as there's no indication as to why the button has changed color, since the warning is hidden behind the modal – wdyt @konmouz?

Shou:
Maybe we should match the yellow warning colour instead? @konmouz

jhermsmeier:
Also, is this something that's been in @konmouz' designs?

Shou:
@jhermsmeier yeah, [here it is on Zeplin][1] – the difference is system drives aren't errors so they're orange. Also here's a screenshot with warning colors on the button instead, @konmouz let us know what you think.

image


[1]: https://app.zeplin.io/project/5857c829442942ea5086c374/screen/5857c8acac5507291ca386a1

jhermsmeier:
@Shou have you made the button orange now?

jhermsmeier:
So, have you changed the button to be orange already @Shou? If so, the commit message needs updating

Shou:
I've pushed the commit with orange now and rebased

lib/gui/app/scss/components/_button.scss
```diff
@@ -69,6 +69,10 @@ $button-types-styles: (
   danger: (
     bg: $palette-theme-danger-background,
     color: $palette-theme-danger-foreground
+  ),
+  warning: (
+    bg: $palette-theme-warning-background,
+    color: $palette-theme-danger-foreground
```
lurch: @Shou I'm confused - if you're using `bg: $palette-theme-warning-background`, wouldn't it also make sense to use `color: $palette-theme-warning-foreground` ? Is that what _was_ intended, given that you've added `$palette-theme-warning-foreground: #fff;` to `_theme.scss` ?

Issue #1916, opened 2017-12-12
Show a confirmation dialog when attempting to flash a drive larger than 64 GB


jviotti:
Just in case, given larger drives are usually system drives.

lurch:
Duplicate of #1728 ?

Issue #983, opened 2016-12-20
Do not allow automatic target selection for a retry operation


saylinglow:
- **Etcher version:** 1.0.0-beta.17
- **Operating system and architecture:**macOS Sierra v10.12.2 2.5GHz Intel Core i5 Mac Mini
My Mac Mini (Mid 2011) has a SD card reader, a SATA SSD as the Mac operating system disk and also a SATA drive as a data disk
- **Do you see any meaningful error information on DevTools?**

I was creating a Raspberry Pi boot image on a SD card plugged into the built-in SD card reader of my Mac Mini. The process completed with an error (I think it was something about not being able to unmount the SD card disk) and the 'Retry' button was in red. I clicked 'Retry' but unfortunately, the SD card had been unmounted and I didn't notice a different disk (my SATA data disk) had been selected as the target drive for the image copy. So I've spent the afternoon recovering the data for my SATA disk as Etcher had put the Raspberry Pi image on to my SATA drive.

I would suggest more checks are made in Etcher before allowing a retry operation; do not allow automatic target selection for retry; use the same unix device mount point for a retry, or get the user to actively choose the target disk if the previous target has been unmounted.

jviotti:
Ugh, that sucks! Intercepting the auto-selection logic if a previous drive was selected but don't currently exists sounds tricky and would result in unwanted state to be preserved in the application.

I still don't understand why Etcher auto-selected your SATA disk.

- Did you manually pull the SDCard out? An unmount operation shouldn't eject the drive, and it should still remain accessible
- Is your SATA drive external? Can you show us the output of `diskutil info /dev/diskN` where `N` corresponds to your SATA drive? Ideally, Etcher shouldn't be allowing you to write to it in any case, but the OS sometimes detects these external HD as exactly the same thing as an USB drive, and therefore we can't tell them apart without some weak heuristics

jviotti:
We should make sure "Retry" is not available if the same drive is no longer accessible though, or can you see a better UX solution @konmouz ?

saylinglow:
I didn’t pull out the SDcard.

Fearns% diskutil info /dev/disk1
   Device Identifier:        disk1
   Device Node:              /dev/disk1
   Whole:                    Yes
   Part of Whole:            disk1
   Device / Media Name:      ST2000LM003 HN-M201RAD

   Volume Name:              Not applicable (no file system)
   Mounted:                  Not applicable (no file system)
   File System:              None

[...]

PR #2045, opened 2018-02-14
feat(GUI): warn the user on large drive selection

Shou:
We warn the user when they select a large drive to confirm they want to
flash in case the device is important.

Fixes: https://github.com/resin-io/etcher/issues/1916
Change-Type: patch
Changelog-Entry: Warn the user on selection of large drives.

### DIFF
--- a/lib/gui/app/components/drive-selector/controllers/drive-selector.js
+++ b/lib/gui/app/components/drive-selector/controllers/drive-selector.js
@@ -75,17 +75,24 @@ module.exports = function (
       return $q.resolve(false)
     }
 
-    if (constraints.isDriveSizeRecommended(drive, selectionState.getImage())) {
-      return $q.resolve(true)
+    if (!constraints.isDriveSizeRecommended(drive, selectionState.getImage())) {
+      return WarningModalService.display({
+        confirmationLabel: 'Yes, continue',
+        description: [
+          messages.warning.unrecommendedDriveSize(selectionState.getImage(), drive),
+          'Are you sure you want to continue?'
+        ].join(' ')
+      })
     }
 
-    return WarningModalService.display({
-      confirmationLabel: 'Yes, continue',
-      description: [
-        messages.warning.unrecommendedDriveSize(selectionState.getImage(), drive),
-        'Are you sure you want to continue?'
-      ].join(' ')
-    })
+    if (constraints.isDriveSizeLarge(drive)) {
+      return WarningModalService.display({
+        confirmationLabel: 'Yes, continue',
+        description: messages.warning.largeDriveSize(drive)
+      })
+    }
+
+    return $q.resolve(true)
   }
 
   /**
--- a/lib/shared/drive-constraints.js
+++ b/lib/shared/drive-constraints.js
@@ -277,6 +277,29 @@ exports.isDriveSizeRecommended = (drive, image) => {
 }
 
 /**
+ * @summary 64GB
+ * @private
+ * @constant
+ */
+exports.LARGE_DRIVE_SIZE = 64e9
+
+/**
+ * @summary Check whether a drive's size is 'large'
+ * @public
+ *
+ * @param {Object} drive - drive
+ * @returns {Boolean} whether drive size is large
+ *
+ * @example
+ * if (constraints.isDriveSizeLarge(drive)) {
+ *   console.log('Impressive')
+ * }
+ */
+exports.isDriveSizeLarge = (drive) => {
+  return _.get(drive, [ 'size' ], UNKNOWN_SIZE) > exports.LARGE_DRIVE_SIZE
+}
+
+/**
  * @summary Drive/image compatibility status messages.
  * @public
  * @type {Object}
@@ -330,7 +353,16 @@ exports.COMPATIBILITY_STATUS_MESSAGES = {
    * @description
    * The drive contains the image and therefore cannot be written to.
    */
-  CONTAINS_IMAGE: 'Drive Contains Image'
+  CONTAINS_IMAGE: 'Drive Contains Image',
+
+  /**
+   * @property {String} LARGE_DRIVE
+   * @memberof COMPATIBILITY_STATUS_MESSAGES
+   *
+   * @description
+   * The drive is large and therefore likely not a medium you want to write to.
+   */
+  LARGE_DRIVE: 'Large Drive'
 }
 
 /**
@@ -413,6 +445,13 @@ exports.getDriveImageCompatibilityStatuses = (drive, image) => {
       statusList.push({
         type: exports.COMPATIBILITY_STATUS_TYPES.WARNING,
         message: exports.COMPATIBILITY_STATUS_MESSAGES.SYSTEM
+      })
+    }
+
+    if (exports.isDriveSizeLarge(drive)) {
+      statusList.push({
+        type: exports.COMPATIBILITY_STATUS_TYPES.WARNING,
+        message: exports.COMPATIBILITY_STATUS_MESSAGES.LARGE_DRIVE
       })
     }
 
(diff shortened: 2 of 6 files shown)


Shou:
This was split off from https://github.com/resin-io/etcher/pull/1917 which I'm changing to become a PR that moves when the drive selector warning dialog appears to the flash step.

tests/shared/drive-constraints.spec.js
```diff
@@ -946,6 +946,48 @@ describe('Shared: DriveConstraints', function () {
...
+        name: 'My Drive',
+        isReadonly: false,
+        isSystem: false,
+        disabled: false,
+        mountpoints: [
+          {
+            path: this.mountpoint
+          }
+        ],
+        size: 64000000001
```
jviotti: Can export the large size constant and use it here?

tests/shared/drive-constraints.spec.js
```diff
@@ -946,6 +946,48 @@ describe('Shared: DriveConstraints', function () {
...
+ 
+    describe('given a drive bigger than the unusually large drive size', function () {
+      it('should return true', function () {
+        m.chai.expect(constraints.isDriveSizeLarge(this.drive)).to.be.true
+      })
+    })
+ 
+    describe('given a drive smaller than the unusually large drive size', function () {
+      it('should return false', function () {
+        this.drive.size = 63999999999
```
jviotti: Same here

tests/shared/drive-constraints.spec.js
```diff
@@ -1080,6 +1122,17 @@ describe('Shared: DriveConstraints', function () {
       })
     })
 
+    describe('given the drive is unusually large', function () {
+      it('should return the large drive size warning', function () {
+        this.drive.size = 64000000001
```
jviotti: And here :P

tests/shared/models/available-drives.spec.js
```diff
@@ -347,6 +347,27 @@ describe('Model: availableDrives', function () {
...
           })
+ 
+          it('should not auto-select a single large size drive', function () {
+            m.chai.expect(selectionState.hasDrive()).to.be.false
+ 
+            availableDrives.setDrives([
+              {
+                device: '/dev/sdb',
+                name: 'Foo',
+                size: 64000000001,
```
jviotti: And here :P

Shou:
Done!

jhermsmeier:
@Shou can you add a screenshot to the PR?

lib/shared/drive-constraints.js
```diff
@@ -276,6 +276,29 @@ exports.isDriveSizeRecommended = (drive, image) => {
   return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'recommendedDriveSize' ], UNKNOWN_SIZE)
 }
 
+/**
+ * @summary 64GB
+ * @private
+ * @constant
+ */
+exports.LARGE_DRIVE_SIZE = 64000000000
```
jhermsmeier: Can we make this a factor, or use the e-expression, to improve clarity & readability – i.e. `64 * 1000 * 1000` or `64e6`?

lib/shared/messages.js
```diff
@@ -78,6 +78,13 @@ module.exports = {
         'The image does not appear to contain a partition table,',
         'and might not be recognized or bootable by your device.'
       ].join(' ')
+    },
+
+    largeDriveSize: (drive) => {
+      return [
+        'Drive <%= drive.description %> (<%= drive.device %>) is unusually large for an SD card or USB stick.',
```
jhermsmeier: This needs to use string interpolation, as these are not lodash templates anymore

jhermsmeier:
@Shou can you use our commit format for all commits in a PR, or squash them? Since we do normal merges now, all commits end up in master, and that might mess with versionist (see [FD thread](https://www.flowdock.com/app/rulemotion/r-etcherprv/threads/EffFc_cek8w1itge4RgdretbTCM)).

Shou:
@jhermsmeier is squash merging being deprecated? I usually do that when I merge

Shou:
Squashed and rebased on master now

Shou:
Here's a screenshot of multiple labels:
image

Will submit a new PR for spacing between the two.

The warning dialog:
image

jhermsmeier:
Agree with @konmouz – also, maybe but an empty line between the reason and the confirmation question, so that it isn't overlooked when skimming the message.

Shou:
Here we go

image

konmouz:
@Shou, looks good to me

lib/shared/drive-constraints.js
```diff
@@ -276,6 +276,29 @@ exports.isDriveSizeRecommended = (drive, image) => {
   return _.get(drive, [ 'size' ], UNKNOWN_SIZE) >= _.get(image, [ 'recommendedDriveSize' ], UNKNOWN_SIZE)
 }
 
+/**
+ * @summary 64GB
+ * @private
+ * @constant
+ */
+exports.LARGE_DRIVE_SIZE = 64e9
```
lurch: Is the `@private` label correct, given that it's included in `exports`? If it's private, does that mean it shouldn't really be used in `tests/shared/drive-constraints.spec.js` ? \*shrug\*

lib/gui/app/components/drive-selector/controllers/drive-selector.js
```diff
@@ -75,17 +75,24 @@ module.exports = function (
       return $q.resolve(false)
     }
 
-    if (constraints.isDriveSizeRecommended(drive, selectionState.getImage())) {
-      return $q.resolve(true)
+    if (!constraints.isDriveSizeRecommended(drive, selectionState.getImage())) {
+      return WarningModalService.display({
+        confirmationLabel: 'Yes, continue',
```
lurch: I wonder if this `'Yes, continue'` string should also be added to the messages class?

lib/gui/app/components/drive-selector/controllers/drive-selector.js
```diff
@@ -75,17 +75,24 @@ module.exports = function (
...
+        description: [
+          messages.warning.unrecommendedDriveSize(selectionState.getImage(), drive),
+          'Are you sure you want to continue?'
```
lurch: Perhaps this `'Are you sure you want to continue?'` should be folded into `messages.warning.unrecommendedDriveSize` ?

lib/shared/store.js
```diff
@@ -144,6 +144,9 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
           constraints.isDriveValid(drive, image),
           constraints.isDriveSizeRecommended(drive, image),
 
+          // We don't want to auto-select large drives
```
lurch: :tada:

lib/shared/drive-constraints.js
lurch: LOL, only just noticed that you made these changes at @jviotti 's request. Ignore my comment then :wink:

lurch:
Is it "obvious enough" that clicking on the "X" is the way for the user to say "No" to this question? What happens if the user clicks on the X, does it then unselect the "large drive" that they'd selected?

smithincanton:
I'm not sure if this is should be part of this issue or it should be entered on it's own.  Using balenaEtcher-Portable-1.5.53.exe on Windows 10.  Selected the ISO image I wanted to write (but at that time was a bad ISO I didn't know about), selected the right drive I wanted to write to (16GB random drive), and clicked Flash.   Etcher failed because of the bad ISO.  All good.  The bad ISO was still in the Image selection so I clicked Change and selected the correct ISO then clicked on Flash.  After the failed Write the selection of the USB drive reverted to the first drive in the list, a 64GB data drive not the one that was previously selected.   My own fault for not double checking what drive was being written to, but I don't think the drive selection should change like that middle of setting up a Write.

lurch:
@smithincanton Sounds similar to #983 or https://github.com/balena-io/etcher/issues/2498#issuecomment-428482551 ?

PR #2049, opened 2018-02-15
test: ensure no leftover template expressions

Shou:
We test for leftover template expressions, including Angluar and
CommonJS template literals. This will ensure we correctly and only
use template literals instead of accidental pure strings or Angular
templates.

Change-Type: patch
Changelog-Entry: Test and ensure there are no leftover template expressions.

### DIFF
--- a/tests/shared/messages.spec.js
+++ b/tests/shared/messages.spec.js
@@ -30,4 +30,14 @@ describe('Shared: Messages', function () {
       m.chai.expect(_.every(_.map(category, _.isFunction))).to.be.true
     })
   })
+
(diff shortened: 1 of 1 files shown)


jhermsmeier:
I don't think this is necessary – for one, we're getting rid of Angular, and secondly lodash templates are also already gone – when adding new messages it would be rather obvious that neither of those are used anymore.

Shou:
I'll close this then – the linter takes care of expressions in pure strings:

```sh
error  Unexpected template string expression  no-template-curly-in-string
```

PR #2050, opened 2018-02-15
refactor: remove stale invalid key check in store

Shou:
We remove a piece of code checking whether `_.keys` returns any non-string
values in its array, but per the Lodash documentation `_.keys` always returns an
array of strings.

Change-Type: patch
Changelog-Entry: Remove stale `invalidKey` check in store.


lurch:
Yeah, looks like the 'logic' from here got moved to https://github.com/resin-io/etcher/blob/master/lib/gui/app/models/settings.js#L159 ?

PR #2056, opened 2018-02-16
fix(app): Fix enabling debug output

jhermsmeier:
This fixes enabling debug output via the DEBUG env var

Change-Type: patch


lurch:
Hmm? Just out of curiosity how come this wasn't picked up in #1789 ? (which is where it was introduced)
Or are there other changes that have taken place since then, which means that #1789 **was** correct at the time?

jhermsmeier:
Not quite sure, might have been Electron version updates?

PR #2060, opened 2018-02-19
fix(GUI): make all class label text bold

Shou:
We make all tags with `.label` have bold text and remove the need for
`<b>` tags.

Change-Type: patch
Changelog-Entry: Make all `.label` tags' text bold and remove need for `<b>` tags.


jhermsmeier:
@Shou unstaged sass changes

Shou:
@jhermsmeier done.

The only visual change is that the `dangerous` label in the settings is bold again.

New
image

Old
image

jhermsmeier:
Why was it de-bolded before?

konmouz:
@Shou, using bold font for this is definitely preferable.

Shou:
@jhermsmeier must've been an accident that got overlooked – I suspect it came with the danger/warning labels below the drives.

PR #2062, opened 2018-02-19
refactor: consolidate store-state nil-checking

Shou:
We make the nil-checking of store state fields generic through a
`verifyNoNilFields` function that throws an error if any fields are
nil.

Change-Type: patch
Changelog-Entry: Consolidate store state nil-checking with helper
function.

### DIFF
--- a/lib/shared/store.js
+++ b/lib/shared/store.js
@@ -29,6 +29,52 @@ const utils = require('./utils')
 const packageJSON = require('../../package.json')
 
 /**
+ * @summary Verify and throw if any state fields are nil
+ * @function
+ * @public
+ *
+ * @param {Object} object - state object
+ * @param {Array<Array<String>> | Array<String>} fields - array of object field paths
+ * @param {String} name - name of the state we're dealing with
+ * @throws
+ *
+ * @example
+ * const fields = [ 'type', 'percentage' ]
+ * verifyNoNilFields(action.data, fields)
+ */
+const verifyNoNilFields = (object, fields, name) => {
+  const nilFields = _.filter(fields, (field) => {
+    return _.isNil(_.get(object, field))
+  })
+  if (nilFields.length) {
+    throw new Error(`Missing ${name} fields: ${nilFields.join(', ')}`)
+  }
+}
+
+/**
+ * @summary FLASH_STATE fields that can't be nil
+ * @constant
+ * @private
+ */
+const flashStateNoNilFields = [
+  'type',
+  'percentage',
+  'eta',
+  'speed'
+]
+
+/**
+ * @summary SELECT_IMAGE fields that can't be nil
+ * @constant
+ * @private
+ */
+const selectImageNoNilFields = [
+  'path',
+  'extension',
+  'size'
+]
+
+/**
  * @summary Application default state
  * @type {Object}
  * @constant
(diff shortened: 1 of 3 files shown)


lib/shared/store.js
```diff
@@ -176,48 +201,32 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
         })
       }
 
-      if (!action.data.type) {
-        throw errors.createError({
-          title: 'Missing state type'
-        })
-      }
+      const fields = [
```
jhermsmeier: This should be a constant way higher up, out of the local scope; we wouldn't want an array to be allocated for this on each call

lib/shared/store.js
```diff
@@ -319,24 +328,19 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
     // place where all the image extension / format handling
     // takes place, to avoid having to check 2+ locations with different logic
     case ACTIONS.SELECT_IMAGE: {
-      if (!action.data.path) {
-        throw errors.createError({
-          title: 'Missing image path'
-        })
-      }
+      const noNilFields = [
```
jhermsmeier: Same here – let's try to avoid allocating things that are constant over and over

lib/shared/store.js
```diff
@@ -28,6 +28,31 @@ const fileExtensions = require('./file-extensions')
...
+ * const fields = [ 'type', 'percentage' ]
+ * verifyNoNilFields(action.data, fields)
+ */
+const verifyNoNilFields = (object, fields, name) => {
+  const nilFields = _.filter(fields, (field) => {
+    return _.isNil(_.get(object, field))
+  })
+  if (nilFields.length) {
+    throw errors.createError({
+      title: `Missing ${name} fields: ${nilFields.join(', ')}`
```
jhermsmeier: We can just use ```throw new Error(`Missing ${name} fields: ${nilFields.join(', ')}`)``` here

Shou:
@jhermsmeier done – moved the constants to the global scope and removed `errors` usage

lib/shared/store.js
```diff
@@ -28,6 +28,52 @@ const fileExtensions = require('./file-extensions')
...
+ * @public
+ *
+ * @param {Object} object - state object
+ * @param {Array<Array<String>> | Array<String>} fields - array of object field paths
+ * @param {String} name - name of the state we're dealing with
+ * @throws
+ *
+ * @example
+ * const fields = [ 'type', 'percentage' ]
+ * verifyNoNilFields(action.data, fields)
```
lurch: Should the example show the `name` parameter being used?

lib/shared/store.js
```diff
@@ -28,6 +28,52 @@ const fileExtensions = require('./file-extensions')
 const utils = require('./utils')
 const packageJSON = require('../../package.json')
 
+/**
+ * @summary Verify and throw if any state fields are nil
+ * @function
+ * @public
+ *
+ * @param {Object} object - state object
+ * @param {Array<Array<String>> | Array<String>} fields - array of object field paths
```
lurch: Is this line correct? Should it just be
```
 * @param {Array<String>} fields - array of object field paths
```
?

lib/shared/store.js
```diff
@@ -28,6 +28,52 @@ const fileExtensions = require('./file-extensions')
...
+ * @param {String} name - name of the state we're dealing with
+ * @throws
+ *
+ * @example
+ * const fields = [ 'type', 'percentage' ]
+ * verifyNoNilFields(action.data, fields)
+ */
+const verifyNoNilFields = (object, fields, name) => {
```
lurch: Is it worth creating a standalone unit-test for this function? Or are the existing unit-tests sufficient?
(I note for example that there's nothing testing that multiple fields are missing at once, which I guess means the `.join(', ')` part isn't being tested?)

lib/shared/store.js
Shou: It's correct, it just passes its argument to `_.get` which can take a string or an array of string for more complex path traversal.

lib/shared/store.js
Shou: Will submit a PR for this

lib/shared/store.js
Shou: That's true, I'll add tests for multiple missing fields.

lib/shared/store.js
lurch: Ahh, then perhaps that behaviour ought to be unit-tested too then?

PR #2063, opened 2018-02-19
feat: resolve jsx files

Shou:
We resolve `.jsx` files such that they get handled by babel and bundled
with webpack.

Change-Type: patch
Changelog-Entry: Resolve JSX files

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -49,7 +49,7 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.js?$/,
+        test: /\.jsx?$/,
         include: [ path.resolve(__dirname, 'lib/gui') ],
         use: {
           loader: 'babel-loader',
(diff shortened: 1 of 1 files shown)


Shou:
You can test this with a JSX file with the `jsx-components-svg-icon` branch

jhermsmeier:
Can you elaborate a bit more on why we need this, what this changes, and how exactly this will work?

Shou:
It's to tell Webpack to handle `.jsx` files and to treat them as modules, AFAICT it doesn't imply something else. If you try to import a JSX module with the config in `master` you just get an error right now saying it can't find the module.

webpack.config.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.js?$/,
+        test: /\.jsx?$/,
```
lurch: LOL, looks like the previous regex would have matched both `foo.j` and `foo.js` - was that intentional?
Whereas now it'll only match `foo.js` or `foo.jsx`

PR #2064, opened 2018-02-19
upgrade(package): Bump electron 1.7.11 -> 1.8.2

jhermsmeier:
This updates Electron to v1.8.2 in order to support arm64 builds

Change-Type: patch
Changelog-Entry: Update Electron to v1.8.2


lurch:
Does this also bump us from Node v6 to Node v8? Does that have wider implications? (e.g. does https://github.com/resin-io/etcher/blob/master/Makefile#L107 need changing? Are all the prebuild packages we're using now going to be using a different ABI version?)

lurch:
Looks like Appveyor is failing because it can't find a suitable prebuild of `winusb-driver-generator`

jhermsmeier:
No, this doesn't bump our Node version we're using for running CI and stuff. Electron's internal Node version is bumped to 8.2.1, but `make electron-develop` builds native modules against whatever Electron we're using, so no problem there.

The NODE_VERSION in the Makefile is for the CLI, and native modules are rebuilt for every package's respective arch & Node / Electron ABI, so that doesn't matter either.

> Are all the prebuild packages we're using now going to be using a different ABI version?

Different from what?

lurch:
Does it "make sense" to be using Node v6 for the CLI and Node v8 inside Electron? Would it be simpler (or at least more consistent) if they're both using Node v8?

> > Are all the prebuild packages we're using now going to be using a different ABI version?

> Different from what?

Sorry, I meant different from what they were using before this PR. And as `winusb-driver-generator` is failing on the AppVeyor build, I guess that means the answer to my question is "yes" ;)

jhermsmeier:
> I meant different from what they were using before this PR

Yes yes, naturally they are, as Node's major version changed (that always comes with a module ABI change)

> And as winusb-driver-generator is failing on the AppVeyor build

Yeah, that's just because the last prebuilds were made before Electron v1.8 was out, so that's easily fixed

> Does it "make sense" to be using Node v6 for the CLI and Node v8 inside Electron? Would it be simpler (or at least more consistent) if they're both using Node v8?

Sure, but it's not necessary as such; we'll get to it at some point

lurch:
I've just created https://github.com/resin-io-modules/winusb-driver-generator/issues/16 :slightly_smiling_face:

jhermsmeier:
Superseded by https://github.com/resin-io/etcher/pull/2165

PR #2065, opened 2018-02-20
fix(package): Add fsevents to platform specific deps

jhermsmeier:
This adds `fsevents` to the platform specific dependencies,
in order to avoid shrinkwrap disagreements between platforms.

Change-Type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
   "platformSpecificDependencies": [
     "7zip-bin-mac",
     "7zip-bin-win",
-    "7zip-bin-linux"
+    "7zip-bin-linux",
+    "fsevents"
   ],
   "dependencies": {
     "angular": "1.6.3",


jviotti:
There is a small typo in the commit message: "pacakge", but looks good otherwise

jhermsmeier:
Oops, fixed @jviotti

lurch:
Apologies for lagging behind, but when/where did `fsevents` come from? Should this PR also include a corresponding change the the shrinkwrap file?

jhermsmeier:
> but when/where did fsevents come from?

A dependency's dependency, more specifically `watchpack`, which is part of `webpack`

> Should this PR also include a corresponding change the the shrinkwrap file?

Nope, we've been keeping it out of the shrinkwrap before, so no mending necessary :)

lurch:
I still don't quite understand how all this shrinkwrapping stuff (and all its quirks) works, but should the other platform-specific modules we're using like `win-drive-clean` and `winusb-driver-generator` also be added to this `platformSpecificDependencies` section?

PR #2076, opened 2018-02-23
chore: only publish production packages to Bintray

jviotti:
The devel channels will be completely deprecated. The deb/rpm snapshot
builds will still be accessible through GitHub Releases as part of Resin
CI builds.

Change-Type: patch


jviotti:
I will delete the Bintray devel apps as soon as this is merged.

Issue #1805, opened 2017-10-24
Implement continuous flashing mode


jhermsmeier:
This would be a mode with no target selection, once "flash" is hit all available targets are written to.

Depends on: #1806

lurch:
Duplicate of #951 ?

jhermsmeier:
Duplicate of #951

Issue #1804, opened 2017-10-24
Support streaming from network sources


jhermsmeier:
Support streaming from network sources such as HTTP streams.

zvin:
This is implemented in etcher-sdk [example](https://github.com/balena-io-modules/etcher-sdk/blob/master/examples/http-zip.ts) but I don't know how this should look in etcher itself.

Issue #1093, opened 2017-02-10
Support decompressing while streaming


jviotti:
This affects the [`image-stream`](https://github.com/resin-io/etcher/tree/master/lib/image-stream) module. Currently, all the decompression handlers in that module rely on the file being present on the disk.

Some decompression methods don't lend themselves to streaming out of the box, therefore we should try to "mimic" this behaviour by probably streaming to the disk (at least partly) until we can decompress.

jviotti:
Maybe @jhermsmeier can help here, since this is very related to the writer, and requires hardcore streaming skills :) ?

lurch:
> therefore we should try to "mimic" this behaviour by probably streaming to the disk (at least partly) until we can decompress.

For compression methods not designed for streaming, I guess in the pathological case this might end up needing to buffer the _entire_ file on disk before decompressing? (which obviously kinda negates the benefit of streaming in the first place)

jviotti:
> I guess in the pathological case this might end up needing to buffer the entire file on disk before decompressing? (which obviously kinda negates the benefit of streaming in the first place)

Its definitely a possibility for some formats, however its something we're willing to accept. The speed advantage of streaming while flashing is lost, but the benefit of downloading the image automatically from the internet is still preserved.

BTW, notice that we will only support streaming from our own catalog (we will not offer a custom URL input, at least for now).

alexandrosm:
streaming-while-writing will only be for catalog entries. There, we can
dictate that only suitable compression methods are used. The whole benefit
of the catalog is to have greater predictability in the inputs so we don't
have to jump through so many hoops.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

lurch:
> There, we can dictate that only suitable compression methods are used.

Yeah, that's what I was thinking / hoping :+1:

jhermsmeier:
> Some decompression methods don't lend themselves to streaming out of the box

Are there any others than `.zip` we'd need to consider?

jviotti:
Apart from zip:

- gzip is easy to stream, however we need to buffer a bit of it in order to be able to inspect the header, estimate the uncompressed size, and return an object with the uncompressed size + the image stream

- xz has the same problem as gzip (we need to read its uncompressed size on the fly before returning the stream to the user)

jviotti:
Some design ideas for this:

[...]

Issue #950, opened 2016-12-06
Select drive as source


alexandrosm:
This is a three-purpose feature request.

Overtly, it's about having the ability to select a drive as a source. So, imagine having two sd drives connected, and selecting the one as a source, and the other as a destination. I think this is a perfectly legitimate Etcher feature. Interestingly, a system drive would be a legitimate thing to select as a source in this case.

Covertly, this is about allowing the selection of things "other than files from the filesystem" as sources. In the future we could add urls, torrents, and other things as sources. If we work out the interaction for having files and drives, we can extend it to other things as they become available.

Even more covertly, this is one of the features we'd need for a hardware version of Etcher. ;)

jviotti:
I like this idea. In order to achieve it, we should first sort out how to determine how much of the drive to copy, which is exactly what we need for the backup feature as well (https://github.com/resin-io/etcher/issues/266).

I'll mark this issue as "pending design", so we can start figuring out how to properly present this paradigm shift to the user while we sort out the technical side in the backup ticket,

WasabiFan:
Theoretically, if you always represent these data sources/destinations as abstract streams, would you just be able to drop-in a disk stream in place of a file one (or a Node.js buffer or whatever internal structure you use)? My thinking is essentially that you could have both the source and destination be completely generic internally and then just specialize the UI to facilitate selection. With something like this, it would presumably be easy for someone to implement a new source or destination, because you wouldn't need to adapt internal switching logic.

lurch:
@WasabiFan Yeah, and that's kinda covered by the `etcher-image-write` and `etcher-image-stream` modules as mentioned in https://github.com/resin-io/etcher/blob/master/docs/ARCHITECTURE.md

I agree with @jviotti that this PR sounds very much like a logical follow-on from #266

alexandrosm:
To me it sounds like a precursor. If you had a drive as a source and a file as a destination, voilà, backup

konmouz:
We can change the UI from 'Image' and 'Drives' to 'Source' and 'Destination'. Probably when you select a drive as source, it should look unavailable on the destination side. Regarding backup, it is true that it will be covered via this feature, however, it might worth having a 'backup' CTA as a more direct approach (mainly for new users).

alexandrosm:
agreed that there may be a need for a backup side-flow even if one could
put it together with these primitives.

having abstract source and destination/target selectors was one of the
earliest things we were talking about with @jviotti and I kinda let it drop
[...]

PR #2080, opened 2018-02-23
feat(sdk): Move CLI writer logic into SDK writer

jhermsmeier:
This moves the preparation logic from the CLI into the SDK
in preparation for further SDK rearchitecturing, and to allow
standalone usage of SDK.

Things this will do / make easier:

- Move towards standalone use capability of the Etcher SDK, which in turn facilitates continuous flashing with a persistent child process (see https://github.com/resin-io/etcher/issues/1805), and also elevation-at-startup to re-enable USBBoot on Linux
- Control multiple write processes with less hassle (i.e. for https://github.com/resin-io/etcher/issues/267)
- Facilitate the move to source & destination for creating backups, network streaming etc. (https://github.com/resin-io/etcher/issues/1804, https://github.com/resin-io/etcher/issues/950, https://github.com/resin-io/etcher/issues/1093)

Change-Type: minor
Changelog-Entry: Move CLI write preparation logic into SDK
Connects To: https://github.com/resin-io/etcher/issues/950, https://github.com/resin-io/etcher/issues/267, https://github.com/resin-io/etcher/issues/1805

### DIFF
--- a/lib/sdk/writer/index.js
+++ b/lib/sdk/writer/index.js
@@ -16,6 +16,11 @@
 
 'use strict'
 
+const os = require('os')
+const fs = require('fs')
+const EventEmitter = require('events').EventEmitter
+const mountutils = require('mountutils')
+const drivelist = require('drivelist')
 const stream = require('readable-stream')
 const Pipage = require('pipage')
 const BlockMap = require('blockmap')
@@ -24,9 +29,53 @@ const BlockWriteStream = require('./block-write-stream')
 const BlockReadStream = require('./block-read-stream')
 const ChecksumStream = require('./checksum-stream')
 const ProgressStream = require('./progress-stream')
+const imageStream = require('../image-stream')
+const diskpart = require('../../cli/diskpart')
+const constraints = require('../../shared/drive-constraints')
+const errors = require('../../shared/errors')
 const debug = require('debug')('etcher:writer')
-const EventEmitter = require('events').EventEmitter
 const _ = require('lodash')
+
+/* eslint-disable prefer-reflect */
+
+/**
+ * @summary Timeout, in milliseconds, to wait before unmounting on success
+ * @constant
+ * @type {Number}
+ */
+const UNMOUNT_ON_SUCCESS_TIMEOUT_MS = 2000
+
+/**
+ * @summary Helper function to run a set of async tasks in sequence
+ * @private
+ * @param {Array<Function>} tasks - set of tasks
+ * @param {Function} callback - callback(error)
+ * @example
+ * runSeries([
+ *   (next) => first(next),
+ *   (next) => second(next),
+ * ], (error) => {
+ *   // ...
+ * })
+ */
+const runSeries = (tasks, callback) => {
+  /**
+   * @summary Task runner
+   * @param {Error} [error] - error
+   * @example
+   * run()
+   */
+  const run = (error) => {
+    const task = tasks.shift()
+    if (error || task == null) {
+      callback(error)
+      return
+    }
+    task(run)
+  }
+
+  run()
+}
 
 /**
  * @summary ImageWriter class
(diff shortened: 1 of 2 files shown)


lib/sdk/writer/index.js
```diff
@@ -24,10 +29,54 @@ const BlockWriteStream = require('./block-write-stream')
...
+const runSeries = (tasks, callback) => {
+  /**
+   * @summary Task runner
+   * @param {Error} [error] - error
+   * @example
+   * run()
+   */
+  const run = (error) => {
+    const task = tasks.shift()
+    if (error || task == null) {
```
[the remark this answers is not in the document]
jhermsmeier: The task is actually `null` when there's no more tasks to work since it shifts tasks off of the array, and there'll be no more items in it – it's the end condition basically. Unless I'm missing the obvious that is :)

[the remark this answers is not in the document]
Shou: Ah fair enough, I thought `callback` was meant to handle errors only

lib/sdk/writer/index.js
```diff
@@ -48,29 +102,200 @@ class ImageWriter extends EventEmitter {
...
+        device: this.options.path
+      })
+
+      if (!selectedDrive) {
+        const selectionError = errors.createUserError({
+          title: 'The selected drive was not found',
+          description: `We can't find ${this.options.path} in your system. Did you unplug the drive?`,
+          code: 'EUNPLUGGED'
+        })
+        debug('state:device-select', this.options.path, 'NOT OK')
```
lurch: Again a small nit, but the 'failure' debug-string here is identical to the one on line 123, so I guess if you see `'state:device-select', this.options.path, 'NOT OK'` (however that happens to get formatted) then I guess it's not immediately obvious whether that's coming from line 123 or line 140?
But I've never used the `debug` module, so maybe this is an invalid concern \*shrug\*

lib/sdk/writer/index.js
```diff
@@ -48,29 +102,200 @@ class ImageWriter extends EventEmitter {
...
+   * @summary Unmount the destination device
+   * @param {Function} callback - callback(error)
+   * @private
+   * @example
+   * writer.unmountDevice((error) => {
+   *   // ...
+   * })
+   */
+  unmountDevice (callback) {
+    if (os.platform() === 'win32') {
```
lurch: Would have been nice if the comment at https://github.com/resin-io/etcher/pull/2080/files#diff-e662af03d4832b4cbfad654fb2c58554L80 could have been preserved ;-)

lib/sdk/writer/index.js
```diff
@@ -48,29 +102,200 @@ class ImageWriter extends EventEmitter {
...
+   *   .on('progress', onProgress)
+   *   .on('finish', resolve)
+   */
+  flash () {
+    const tasks = [
+      (next) => { this.checkSelectedDevice(next) },
+      (next) => { this.unmountDevice(next) },
+      (next) => { this.removePartitionTable(next) },
+      (next) => { this.openSource(next) },
+      (next) => { this.openDestination(next) }
```
lurch: Almost looks like a task-based equivalent of pipage :wink:

lib/sdk/writer/index.js
```diff
@@ -24,10 +29,54 @@ const BlockWriteStream = require('./block-write-stream')
...
+
+/**
+ * @summary Timeout, in milliseconds, to wait before unmounting on success
+ * @constant
+ * @type {Number}
+ */
+const UNMOUNT_ON_SUCCESS_TIMEOUT_MS = 2000
+
+/**
+ * @summary Helper function to run a set of async tasks in sequence
```
lurch: I wonder if such a "generic mechanism" should be moved to a more "shared" location, rather than only being in `lib/sdk/writer/index.js` ?

PR #2082, opened 2018-02-23
minifix: remove stale selection-state clear jsdoc example

Shou:
Change-Type: patch


Shou:
Oh multi-writes fixed this actually

PR #2089, opened 2018-02-26
fix(scripts): Add missing types to architecture-convert.sh usage

lurch:
Change-type: patch

### DIFF
--- a/scripts/build/architecture-convert.sh
+++ b/scripts/build/architecture-convert.sh
@@ -25,7 +25,7 @@ function usage() {
   echo "Options"
   echo ""
   echo "    -r <architecture>"
-  echo "    -t <type (debian|redhat|node|docker)>"
+  echo "    -t <type (debian|redhat|node|docker|electron-builder|appimage)>"
   exit 1
 }
 


scripts/build/architecture-convert.sh
```diff
@@ -25,7 +25,7 @@ function usage() {
   echo "Options"
   echo ""
   echo "    -r <architecture>"
-  echo "    -t <type (debian|redhat|node|docker)>"
+  echo "    -t <type (debian|redhat|node|docker|electron-builder|appimage)>"
```
jhermsmeier: Shouldn't `electron-builder` be just `electron`?

lurch: https://github.com/resin-io/etcher/blob/master/scripts/build/architecture-convert.sh#L59

But if you want to change it, that's obviously another PR entirely... :wink:

lurch: (and I dunno if there's any guarantee that the architectures used by `electron-builder` will always match up with the architectures used by `electron`?)

jhermsmeier: Oh okay, I see – well, that's mildly confusing.

lurch: Not really... the `Makefile` doesn't interact directly with `electron`, only with `electron-builder`. And `architecture-convert.sh` is only called by the `Makefile`, or by [other scripts](https://github.com/resin-io/etcher/tree/master/scripts/build) that in turn get called from the `Makefile`.

jhermsmeier: But electron-builder is just the thing that builds, not the thing being built, so I find the arch referring to that rather than electron mildly confusing, if you see what I mean. Doesn't matter too much though.

lurch: ..but it's `electron-builder` that "consumes" the output of `architecture-convert.sh`, IYSWIM.
This code was originally added by @jviotti in #1510 , if that helps explain how this naming came about.

PR #2090, opened 2018-02-26
refactor: move memoize function to shared utils

Shou:
We move the memoize function to `lib/shared/utils.js` and expose it to
modules across the project.

Change-Type: patch
Changelog-Entry: Move memoize function to shared utils.

### DIFF
--- a/lib/shared/utils.js
+++ b/lib/shared/utils.js
@@ -78,3 +78,62 @@ exports.percentageToFloat = (percentage) => {
 
   return percentage / exports.PERCENTAGE_MAXIMUM
 }
+
+/**
+ * @summary Memoize a function
+ * @function
+ * @private
+ *
+ * @description
+ * This workaround is needed to avoid AngularJS from getting
+ * caught in an infinite digest loop when using `ngRepeat`
+ * over a function that returns a mutable version of an
+ * ImmutableJS object.
+ *
+ * The problem is that every time you call `myImmutableObject.toJS()`
+ * you will get a new object, whose reference is different from
+ * the one you previously got, even if the data is exactly the same.
+ *
+ * @param {Function} func - function that returns an ImmutableJS list
+ * @param {Function} comparer - function to compare old and new args and state
+ * @returns {Function} memoized function
+ *
+ * @example
+ * const getList = () => {
+ *   return Store.getState().toJS().myList;
+ * };
+ *
+ * const memoizedFunction = memoize(getList, angular.equals);
+ */
+exports.memoize = (func, comparer) => {
+  let previousTuples = []
+
+  return (...restArgs) => {
+    let areArgsInTuple = false
+    let state = Reflect.apply(func, this, restArgs)
+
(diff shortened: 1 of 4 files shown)


tests/gui/components/drive-selector.spec.js
```diff
@@ -20,71 +20,17 @@ const _ = require('lodash')
...
-      })
-
-      it('should reflect different arguments', function () {
-        const memoizedParameter = controller.memoizeImmutableListReference(_.identity)
-
-        m.chai.expect(memoizedParameter(false)).to.be.false
-        m.chai.expect(memoizedParameter(true)).to.be.true
-      })
-
+    describe('.memoize()', function () {
```
lurch: I don't fully understand the code in this PR, but why is this test in `drive-selector.spec.js` rather than `utils.spec.js` ? AFAICT there's nothing DriveController-specific about it?

tests/gui/components/drive-selector.spec.js
```diff
@@ -20,71 +20,17 @@ const _ = require('lodash')
...
-      _WarningModalService_
-    ) {
-      $controller = _$controller_
-      $rootScope = _$rootScope_
-      $q = _$q_
-      $uibModalInstance = {}
-      WarningModalService = _WarningModalService_
-    }))
-
-    beforeEach(() => {
```
lurch: Was this code all defunct then?

lib/shared/utils.js
```diff
@@ -78,3 +78,62 @@ exports.percentageToFloat = (percentage) => {
...
+ * @returns {Function} memoized function
+ *
+ * @example
+ * const getList = () => {
+ *   return Store.getState().toJS().myList;
+ * };
+ *
+ * const memoizedFunction = memoize(getList, angular.equals);
+ */
+exports.memoize = (func, comparer) => {
```
lurch: Given how specific this is to AngularJS and lists, I wonder if just `memoize` is a "too generic" function name? \*shrug\*

tests/gui/components/drive-selector.spec.js
Shou: The other isn't a GUI test so Angular isn't available (because `window` isn't), since we use `.memoize()` in the DriveSelectorController in an Angular-specific way (`angular.equals`) I think it makes most sense to keep it here.

tests/gui/components/drive-selector.spec.js
Shou: It is now, but wasn't before. We needed access to the controller to access the memoize method, but now we don't.

tests/gui/components/drive-selector.spec.js
lurch: Ahhh, that makes sense - I hadn't noticed that the test here was using `angular.equals` but the tests in `utils.spec.js` were using `_.isEqual` :)  (as you can probably tell, I know nothing about Angular)

tests/gui/components/drive-selector.spec.js
lurch: Meh :wink:

PR #2100, opened 2018-03-01
fix(gui): Disable throttling timers when in background

jhermsmeier:
This disables Electron throttling timers when not in the foreground.

Change-Type: patch

### DIFF
--- a/lib/gui/etcher.js
+++ b/lib/gui/etcher.js
@@ -46,7 +46,10 @@ electron.app.on('ready', () => {
     fullscreenable: false,
     autoHideMenuBar: true,
     titleBarStyle: 'hidden-inset',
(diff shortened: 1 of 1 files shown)


jhermsmeier:
This is more of a precaution, rather than a reaction to observed behavior, and we'd also want our timers to be reliable I reckon (since this is not some background tab in a browser)

lurch:
OTOH, I guess if Etcher is able to "detect" when it's in the background, maybe it _might_ make sense to reduce the frequency of the drivelist scans? \*shrug\*

jhermsmeier:
@lurch that's an entirely different thing – this is about the VM throttling internal timer ticks to a lower frequency when in background. What you're thinking about we could already do regardless of this setting.

Issue #1113, opened 2017-02-18
CRC/MD5 checksum in filename or extra file


ThomDietrich:
I've recently switched over our image flashing recommendation to Etcher. Great piece of software.
I see a lot of issues regarding the creation of a checksum for the image. Could you please explain how this checksum is validated and how I can provide a checksum for an image file, so the user can be sure the file he/she's flashing is indeed identical to the one I've created and uploaded once. I'm sure there is a solution in place, that I didn't catch up with yet 😄 

Thanks!

jviotti:
Hi @ThomDietrich ,

Thanks for reaching out. The way validation works is the following:

- Etcher will calculate a CRC32 checksum of the select image *while* its flashed to the drive
- Once the image has been flashed, Etcher will re-calculate a CRC32 of the drive
- If the CRC32 checksum calculate from the drive matches the one we initially calculated from the image, then we know the flash was successful

This means that you don't have to pass a checksum to Etcher at all (it does it all under the hood). There is still a chance where the image the user downloaded is not complete (e.g: the download stopped halfway through). In that case, you can provide the right CRC32 checksum of the image on your website, an ask users to compare it with the final checksum they see after flashing the image.

We have some plans to provide an Etcher "catalog", where users will be able to browse and discover images that Etcher knows about, which will contain the real image checksum, and that will be validated as well. Check https://resin.io/blog/the-future-of-etcher/ for details!

Please re-open if you have any other questions!

ThomDietrich:
Hello @jviotti,
thank you for the detailed explanation. Matches my assumptions and that sounds good so far.

> you can provide the right CRC32 checksum of the image on your website

This is the one aspect I was curious about. If this is the current situation, I need to rephrase this issue as a **Feature Request**:

Would you consider comparing the initially calculated checksum with the one provided in the image filename or in a separate checksum file?

The easiest approach would be a short hash CRC32 directly in the filename: `imagefile-c45ad668.img`. This feature is probably a oneliner without any risk. You just need to show an additional success indication in the UI if the file hash matches any part of the filename.

*A more sophisticated way to suply a checksum is via an extra files (`MD5SUMS`, `SHA1SUMS`, `SHA256SUMS` or `SHA512SUMS`, `imagefile.img.sfv`, `imagefile.img.md5`, ...). These are provided by many distributors but probably never actually downloaded by the end user, hence not as important for Etcher to check against.*

Wdyt?

alexandrosm:
[...]

PR #2108, opened 2018-03-08
fix(perf): Remove support for CRC32 checksumming

jhermsmeier:
As crc32-stream was identified as a massive performance bottleneck,
we remove it, and default to Node's crypto API with md5 instead.

Change-Type: minor


lurch:
LOL #643 ;-)

I guess the knock-on effect on #1113 (if / when that gets implemented) is that the hash-embedded-in-the-filename will now be much longer?

PR #2112, opened 2018-03-09
fix: stop autoselecting empty value in store

Shou:
We fix store autoselection, which selects an empty value when one
drive is selected and then ejected, leaving one drive that is
supposed to be autoselectable. Now it instead properly autoselects the
last drive.

Change-Type: patch
Changelog-Entry: Stop store autoselection from selecting empty values on
drive ejection.

### DIFF
--- a/lib/shared/store.js
+++ b/lib/shared/store.js
@@ -175,6 +175,23 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
       }
 
       const newState = state.set('availableDrives', Immutable.fromJS(drives))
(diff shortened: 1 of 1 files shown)


Shou:
How to reproduce:

0. Plug in two autoselectable drives
1. Select one drive
2. Unplug the selected drive
3. 'Selected drive' will now be an empty value
4. (optional) Plug drive back in, it will be autoselected again – not supposed to happen with two drives

lurch:
Ahhh, I'd always wondered why DESELECT_DRIVE was being done at "the end" rather than "the start" of this code-section ;)

Issue #2078, opened 2018-02-23
Embedded image icon not being displayed


lurch:
- **Etcher version:** latest GH master
- **Operating system and architecture:** Linux x64
- **Image flashed:** `ev3dev-jessie-ev3-generic-2017-09-14.zip`
- **Do you see any meaningful error information in the DevTools?** 
```
gui.js:113926 Fri Feb 23 2018 11:58:17 GMT+0000 (GMT) Select image ({"name":"ev3dev-jessie","version":"ev3dev-jessie-ev3-generic-2017-09-14","url":"http://www.ev3dev.org","supportUrl":"http://www.ev3dev.org/support","logo":true,"bmap":true,"path":"/home/qpc/andynas-cache/ev3dev-jessie-ev3-generic-2017-09-14.zip","size":{"original":1887436800,"final":{"estimation":false,"value":1887436800}},"extension":"img","archiveExtension":"zip","hasMBR":true,"hasGPT":false,"partitions":[{"type":11,"id":null,"name":null,"firstLBA":8192,"lastLBA":106496,"extended":false},{"type":131,"id":null,"name":null,"firstLBA":106496,"lastLBA":3686400,"extended":false},{"type":0,"id":null,"name":null,"firstLBA":0,"lastLBA":0,"extended":false},{"type":0,"id":null,"name":null,"firstLBA":0,"lastLBA":0,"extended":false}]})
gui.js:4213 Warning: Failed prop type: Invalid prop `contents` of type `string` supplied to `SVGIcon`, expected `array`.
    in SVGIcon
printWarning @ gui.js:4213
```

When flashing the above image with Etcher 1.3.1 then a nice icon is shown above the "Select image" button. But trying to do the same with the `master` version of Etcher doesn't display the icon, and the above error is displayed in DevTools. I _believe_ this might be due to the comment I left [here](https://github.com/resin-io/etcher/pull/1677#discussion_r169924274).

PR #2113, opened 2018-03-09
minifix: use svg contents list in main template

Shou:
We use a list instead of element with `svg-icon` in `main.tpl.html`, as
required by the `svg-icon` component and will return an error
otherwise.

Fixes: https://github.com/resin-io/etcher/issues/2078
Change-Type: patch
Changelog-Entry: Use SVG contents list in main template.

### DIFF
--- a/lib/gui/app/pages/main/templates/main.tpl.html
+++ b/lib/gui/app/pages/main/templates/main.tpl.html
@@ -3,7 +3,7 @@
     <div class="box text-center relative" os-dropzone="image.selectImageByPath($file)">
 
       <div class="center-block">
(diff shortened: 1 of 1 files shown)


lurch:
Is this the fix for #2078 ? Perhaps the commit-message or PR-comment should mention that? ;)

Shou:
@lurch that's right, thanks :)

jhermsmeier:
@Shou could you add a description to the commit message about what the actual problem is, and how & why this fixes it?

Shou:
@jhermsmeier how's this?

PR #2114, opened 2018-03-09
feat(CSS): use roboto font instead of helvetica

Shou:
We replace Helvetica with Roboto as the main font.

Change-Type: patch
Changelog-Entry: Replace Helvetica as the main font with Roboto.


Shou:
Roboto

image

Helvetica

image

konmouz:
@Shou can you use medium or bold weight for the main CTAs - Select image, Select drive, Flash?

Shou:
@konmouz I'll tackle that in another PR (and probably some other inconsistencies between the designs) but here's what it'll look like

image

PR #2115, opened 2018-03-09
feat(GUI): add colorised multi-writes progress status dots

Shou:
We add colorised progress status dots with quantities of `flashing`,
`validating`, `succeeded`, and `failed` devices.

Change-Type: patch
Changelog-Entry: Add colorised multi-writes progress status dots.

### DIFF
--- a/lib/gui/app/modules/progress-status.js
+++ b/lib/gui/app/modules/progress-status.js
@@ -28,7 +28,10 @@ const utils = require('../../../shared/utils')
  *
  * @example
  * const status = progressStatus.fromFlashState({
- *   type: 'write',
+ *   flashing: 1,
+ *   validating: 0,
+ *   succeeded: 0,
+ *   failed: 0,
  *   percentage: 55,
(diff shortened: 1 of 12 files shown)


lib/gui/app/modules/progress-status.js
```diff
@@ -37,25 +40,26 @@ const utils = require('../../../shared/utils')
  * // '55% Flashing'
  */
 exports.fromFlashState = (state) => {
-  const isChecking = state.type === 'check'
+  const isFlashing = Boolean(state.flashing)
```
lurch: Perhaps something like `const anyFlashing = (state.flashing >= 1)` would be more explicit?

lurch:
Will the status-dots be hidden if only one image & drive is being flashed?

lib/gui/app/modules/progress-status.js
jhermsmeier: Not preferred by the linter though, iirc

PR #2118, opened 2018-03-13
fix(scanner): Enable usbboot on Linux if run as root

jhermsmeier:
This re-enables the usbboot scanner dynamically if Etcher
is run as root on Linux.

Change-Type: patch

### DIFF
--- a/lib/gui/app/modules/drive-scanner.js
+++ b/lib/gui/app/modules/drive-scanner.js
@@ -61,6 +61,8 @@ permissions.isElevated().then((elevated) => {
     scanner.subscribe(adapter)
     scanner.start()
   }
-}).catch(_.noop)
+}).catch((error) => {
+  console.warn('Could not add usbboot adapter:', error)
+})
 
 module.exports = scanner
--- a/lib/gui/app/modules/drive-scanner.js
+++ b/lib/gui/app/modules/drive-scanner.js
@@ -29,7 +29,7 @@ const permissions = require('../../../shared/permissions')
  * @type {String}
  * @constant
  */
-const BLOBS_DIRECTORY = path.join(__dirname, '..', '..', 'blobs')
+const BLOBS_DIRECTORY = path.join(__dirname, '..', '..', '..', 'blobs')
 
 const scanner = SDK.createScanner({
   blockdevice: {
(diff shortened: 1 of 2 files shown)


lurch:
I wonder if it'd make sense (or even be possible?) to move this logic to `lib/sdk/adapters/index.js` so that it remains grouped together? \*shrug\*

lib/gui/app/modules/drive-scanner.js
```diff
@@ -51,4 +52,15 @@ const scanner = SDK.createScanner({
...
+// NOTE: Enable USBBoot on Linux if run as root
+permissions.isElevated().then((elevated) => {
+  if (elevated && process.platform === 'linux') {
+    const UsbbootAdapter = require('../../../sdk/adapters/usbboot')
+    const adapter = new UsbbootAdapter()
+    scanner.stop()
+    scanner.subscribe(adapter)
+    scanner.start()
+  }
+}).catch(_.noop)
```
zvin: Maybe this should console.warn ?

Shou:
After testing it, there are some path issues –  `lib/gui/blobs/usbboot/raspberrypi/bootcode.bin` has been moved to `lib/blobs/usbboot/raspberrypi/bootcode.bin`.

image

And the icon path is an easy fix as well thankfully:

```diff
diff --git a/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html b/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
index 084f5a2..4723e80 100644
--- a/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
+++ b/lib/gui/app/components/drive-selector/templates/drive-selector-modal.tpl.html
@@ -11,7 +11,7 @@
       ng-click="modal.toggleDrive(drive)">
         <img class="list-group-item-section" alt="Drive device type logo"
           ng-if="drive.icon"
-          ng-src="./assets/{{drive.icon}}.svg"
+          ng-src="../assets/{{drive.icon}}.svg"
           width="25"
           height="30">
         <div
```

lib/gui/app/modules/drive-scanner.js
jhermsmeier: @zvin yup, good point – done!

PR #2119, opened 2018-03-13
[WIP]: feat(writer): Use direct I/O in flashing pipeline

jhermsmeier:
This uses [`ronomon/direct-io`](https://github.com/ronomon/direct-io) for aligned buffers and
enables direct I/O through the `O_DIRECT` flag, avoiding
copying memory to kernel-space for each read / write

Change-Type: minor
Connects To: https://github.com/resin-io/etcher/issues/1523

### DIFF
--- a/lib/sdk/writer/block-stream.js
+++ b/lib/sdk/writer/block-stream.js
@@ -16,11 +16,13 @@
 
 'use strict'
 
+const directIo = require('@ronomon/direct-io')
 const stream = require('readable-stream')
 const debug = require('debug')('etcher:writer:block-stream')
 
 const MIN_BLOCK_SIZE = 512
 const CHUNK_SIZE = 64 * 1024
+const DIRECT_IO_ALIGNMENT = 4096
 
 /**
  * @summary BlockStream class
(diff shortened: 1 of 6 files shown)


lib/sdk/writer/block-stream.js
```diff
@@ -16,11 +16,13 @@
 
 'use strict'
 
+const directIo = require('@ronomon/direct-io')
 const stream = require('readable-stream')
 const debug = require('debug')('etcher:writer:block-stream')
 
 const MIN_BLOCK_SIZE = 512
 const CHUNK_SIZE = 64 * 1024
+const DIRECT_IO_ALIGNMENT = 4096
```
zvin: This should be defined in one place only. (It was already defined above in block-read-stream.js).

jhermsmeier:
I've also noticed that this would require a little more work, as some image source streams do use "normal" buffers, which would have to be copied to aligned buffers. Marking this WIP for now.

jhermsmeier:
Closing, as this turns out to be more involved than initially thought, since we'd have to align buffers differently for each destination device with multiwrites. But the performance gains this yielded so far are well worth revisiting this.

PR #2124, opened 2018-03-19
refactor: multi-writes preparatory changes

Shou:
We add some preparatory changes including new utility functions, as well
as changes throughout the codebase that reflect the change from single
drives to a list of drives, given multi-writes is coming.

Change-Type: patch
Changelog-Entry: Various preparatory changes to account for
multi-writes.

### DIFF
--- a/lib/shared/models/flash-state.js
+++ b/lib/shared/models/flash-state.js
@@ -114,34 +114,36 @@ exports.unsetFlashingFlag = (results) => {
  * });
  */
 exports.setProgressState = (state) => {
-  if (!_.isString(state.type)) {
-    throw new Error(`Invalid state type: ${state.type}`)
-  }
+  const {
+    flashing,
+    verifying,
+    succeeded,
+    failed,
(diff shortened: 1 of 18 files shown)


lib/shared/models/flash-state.js
```diff
@@ -120,13 +120,21 @@ exports.setProgressState = (state) => {
...
   // is in proper, thanks to the status quantities.
+  // Also the below code will become a lot simpler.
   const isValidating = state.type === 'check'
   const type = isValidating ? 'validating' : 'flashing'
+  const pickedStates = _.pick(state, [
+    'flashing',
+    'validating',
+    'succeeded',
+    'failed'
+  ])
```
Shou: This is to prepare for when those fields are part of the state sent.

jhermsmeier:
@Shou this still only shows one device as verifying when flashing multiple devices

Shou:
@jhermsmeier seems to be working here

image

Issue #1914, opened 2017-12-09
Categories= key missing in Linux desktop file


probonopd:
The `etcher.desktop` file should include a line

```
Categories=Utility;
```

References:
* https://travis-ci.org/AppImage/appimage.github.io/builds/313897555#L709
* https://github.com/AppImage/appimage.github.io/pull/199#partial-pull-merging

---

The desktop file is lacking a `Categories=` entry in the `*.desktop` file.

According to the [menu spec](https://standards.freedesktop.org/menu-spec/latest/apa.html),

> By including one of the Main Categories in an application's desktop entry file, the application will be ensured that it will show up in a section of the application menu dedicated to this category. If multiple Main Categories are included in a single desktop entry file, the entry may appear more than once in the menu. 

Hence, please add at least one of the following in the `Categories=` key.

Main Category | Description | Notes
-- | -- | --
AudioVideo | Application for presenting, creating, or processing multimedia (audio/video) |  
Audio | An audio application | Desktop entry must include AudioVideo as well
Video | A video application | Desktop entry must include AudioVideo as well
Development | An application for development |  
Education | Educational software |  
Game | A game |  
Graphics | Application for viewing, creating, or processing graphics |  
Network | Network application such as a web browser |  
Office | An office type application |  
Science | Scientific software |  
Settings | Settings applications | Entries may appear in a separate menu or as part of a 		"Control Center"
System | System application, "System Tools" such as say a log viewer or network monitor |  
Utility | Small utility application, "Accessories"

In addition, you could specify one or more from the longer list of [Additional Categories](https://standards.freedesktop.org/menu-spec/latest/apas02.html).

Please test the result with `desktop-file-validate` and make sure it passes.

__electron-builder offers native support for this since [v19.22.1](https://github.com/electron-userland/electron-builder/releases/tag/v19.22.1).__. Please see the `category` key at https://www.electron.build/configuration/linux-other.

jhermsmeier:
This will be fixed by https://github.com/resin-io/etcher/pull/1905 as our electron-builder config already sets the category to "Utility"

jhermsmeier:
Closing, as https://github.com/resin-io/etcher/pull/1905 has been merged, and this should be fixed with the next release.

PR #2125, opened 2018-03-19
 style: add metainformation for linux menus 

mmmm1998:
Now, Etcher, on the basis of the `.desktop` file, is placed in the menus category 'Other'.
With this changes, Etcher will be moved to the menu category for system utilities ('System' category).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,7 @@ COMPANY_NAME = Resinio Ltd
 APPLICATION_NAME = $(shell jq -r '.displayName' package.json)
 APPLICATION_DESCRIPTION = $(shell jq -r '.description' package.json)
 APPLICATION_COPYRIGHT = $(shell cat electron-builder.yml | shyaml get-value copyright)
+APPLICATION_MENU_CATEGORIES = $(shell cat electron-builder.yml | shyaml get-value linux.category)
 
 BINTRAY_ORGANIZATION = resin-io
 BINTRAY_REPOSITORY_DEBIAN = debian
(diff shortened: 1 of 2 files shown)


lurch:
Related to #1914 - which implies that this issue has been fixed already?

mmmm1998:
I am not sure, but `electron-builder` build deb package, using own meta information from electron-builder.yml, and `electron-create-appdir.sh` and `electron-create-appimage-linux.sh` create app image for Linux, and their `.desktop` generate separetly and with using different infromation (`electron-builder` get info for `.desktop` from `electron-builder.yml` and `electron-create-appdir.sh` just using  options from script argument and string template inside script), isn't it?
So #1914 add `Category` for deb building, but not for app image building, because for app image `.desktop` is generated by `electron-create-appdir.sh`. And then you use app image on Linux and install `.desktop` file, `Etcher` appears in category 'Others'.

lurch:
Hmmm, I guess this also partially relates to #1625  (which is why the AppImages aren't also being built by `electron-builder`).

@mmmm1998 It might be better to do something like https://github.com/resin-io/etcher/blob/master/Makefile#L111 inside `electron-create-appdir.sh` to ensure that it always uses the same category as `electron-builder.yml` ? \*shrug\*

mmmm1998:
Well, @ lurch, maybe i have invented better solution.
I have added new argument to `electron-create-appdir.sh`, that takes menu categories.
And have added categories variable, using `shyaml` (like https://github.com/resin-io/etcher/blob/master/Makefile#L111), in `Makefile`. So now, menu categories pass to `electron-create-appdir.sh` in the same way as `APPLICATION_NAME` variable .

thundron:
Closing as we moved on to https://github.com/balena-io/scripts.git usage (related file [here](https://github.com/balena-io/scripts/blob/master/electron/electron-create-appdir.sh))

PR #2135, opened 2018-03-23
fix(GUI): fix status dots to their position

Shou:
We ensure the status dots don't move about and stick to their position
by using absolute positioning relative to its parent element.

Change-Type: patch
Changelog-Entry: Fix status dots to their position.


jhermsmeier:
@Shou can we have screenshots to pepper this PR? :)

lurch:
image

PR #2140, opened 2018-03-24
fix: use correct usbbot blob path in appimages

Shou:
We use the `app.asar` relative path to read the usbboot blob files
instead of relative to the normal project root.

Change-Type: patch
Changelog-Entry: Use correct usbboot blob path in AppImages.

### DIFF
--- a/lib/gui/app/modules/drive-scanner.js
+++ b/lib/gui/app/modules/drive-scanner.js
@@ -23,13 +23,16 @@ const path = require('path')
 const settings = require('../models/settings')
 const SDK = require('../../../sdk')
 const permissions = require('../../../shared/permissions')
+const packageJSON = require('../../../../package.json')
 
 /**
  * @summary The Etcher "blobs" directory path
  * @type {String}
  * @constant
  */
-const BLOBS_DIRECTORY = path.join(__dirname, '..', '..', '..', 'blobs')
+const BLOBS_DIRECTORY = packageJSON.packageType === 'AppImage'
+  ? 'bin/resources/app.asar/lib/blobs'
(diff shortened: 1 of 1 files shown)


Shou:
@lurch that's Windows-specific, isn't it? This change only affects AppImages, outside of that the behaviour stays the same

jhermsmeier:
Yes, this has nothing to do with that – this is usbboot / USB adapter specific

lib/gui/app/modules/drive-scanner.js
```diff
@@ -29,7 +29,9 @@ const permissions = require('../../../shared/permissions')
  * @type {String}
  * @constant
  */
-const BLOBS_DIRECTORY = path.join(__dirname, '..', '..', '..', 'blobs')
+const BLOBS_DIRECTORY = process.env.APPIMAGE && process.env.APPDIR
```
jhermsmeier: Should we use the package.json's `.packageType` property here? i.e. `pkgJson.packageType === 'AppImage'`

Shou: We use the `process.env.APPIMAGE` check in another place, is the `pkgJson.packageType` way better?

jhermsmeier: I'd like to think it's more reliable due to us controlling it; the env tends to be the wild west on other people's computers ;)

lurch: ...and IIRC there is/was code that only copied selected env-vars into the environment of sub-processes?

Shou: Here we go

PR #2146, opened 2018-03-26
feat(GUI): conform to the styleguide buttons

Shou:
We conform to the Resin styleguide buttons, as they are in
[rendition][1], but with Etcher colours. This means bold fonts, slightly
rounder corners, and slightly slimmer buttons.

[1]: https://resin-io-modules.github.io/rendition/?selectedKind=Button&selectedStory=Standard&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

Change-Type: patch
Changelog-Entry: Conform to the Resin styleguide buttons.

### DIFF
--- a/lib/gui/app/scss/components/_button.scss
+++ b/lib/gui/app/scss/components/_button.scss
@@ -17,16 +17,17 @@
 .button {
   @extend .btn;
 
-  padding: 10px;
-  padding-top: 11px;
+  height: 36px;
 
-  border-radius: 2px;
+  border-radius: 3px;
   border: 0;
 
   letter-spacing: .5px;
   outline: none;
 
   position: relative;
+
+  font-weight: bold;
 
   > .glyphicon {
     top: 2px;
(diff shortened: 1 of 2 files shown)


Shou:
After

image

Before

image

Shou:
@jhermsmeier yeah, that's how the rendition buttons are. Maybe we should wait for @konmouz and see what it's supposed to be like.

jhermsmeier:
Hmm :/ I think in this case the old buttons do look better; maybe this is something that needs mending in Rendition? /cc @LucianBuzzo

Shou:
Basically, they are centred to the X-height of the font so it creates the illusion that they aren't centred while it technically is whereas the old button text is cap height centred – which is more visually centred because there are more letters expanding into the capline than there are descenders.

konmouz:
@Shou, I suggest to follow rendition for consistency and make any changes there if possible

PR #2153, opened 2018-03-27
feat: expose all flash state fields to the store

Shou:
We expose all the flash state fields to the store, and mirror
`speed`'s format with `totalSpeed` with MB as the size unit.
We also format the progress state messages to use information
available with multiwrites.

Change-Type: patch
Changelog-Entry: Expose all flash state fields to the store.

### DIFF
--- a/lib/gui/app/app.js
+++ b/lib/gui/app/app.js
@@ -182,15 +182,20 @@ app.run(() => {
     }
 
     const currentFlashState = flashState.getFlashState()
+    const stateType = !currentFlashState.flashing && currentFlashState.verifying
+      ? `Verifying ${currentFlashState.verifying}`
+      : `Flashing ${currentFlashState.flashing}`
 
     // NOTE: There is usually a short time period between the `isFlashing()`
     // property being set, and the flashing actually starting, which
     // might cause some non-sense flashing state logs including
     // `undefined` values.
     analytics.logDebug(
-      `Progress (${currentFlashState.type}): ` +
+      `${stateType} devices, ` +
       `${currentFlashState.percentage}% at ${currentFlashState.speed} MB/s ` +
-      `(eta ${currentFlashState.eta}s)`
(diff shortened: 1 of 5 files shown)


lib/shared/models/flash-state.js
```diff
@@ -114,36 +114,36 @@ exports.unsetFlashingFlag = (results) => {
  * });
  */
 exports.setProgressState = (state) => {
+  // Preserve only two decimal places
+  const PRECISION = 2
```
jhermsmeier: I think 1 decimal is enough; 2 looks a bit odd

lib/gui/app/app.js
```diff
@@ -182,15 +182,25 @@ app.run(() => {
     }
 
     const currentFlashState = flashState.getFlashState()
+    const {
+      flashing,
+      verifying,
+      failed
+    } = currentFlashState
```
jhermsmeier: I think we can just use `currentFlashState.flashing` etc. in the following lines; no need to destructure here

lib/shared/models/flash-state.js
```diff
@@ -114,36 +114,36 @@ exports.unsetFlashingFlag = (results) => {
  * });
  */
 exports.setProgressState = (state) => {
+  // Preserve only two decimal places
```
jhermsmeier: *one* ;)

lib/shared/models/flash-state.js
```diff
@@ -114,36 +114,36 @@ exports.unsetFlashingFlag = (results) => {
...
   const {
-    flashing,
-    verifying,
-    succeeded,
-    failed,
     percentage,
     eta,
-    speed
+    speed,
+    totalSpeed
```
jhermsmeier: Again, no need to destructure here I think

Shou:
I've updated this one to remove the unnecessary destructuring now

PR #2154, opened 2018-03-27
fix(windows): Disable rebuilding bindings in electron-builder

jhermsmeier:
We disable `npm rebuild` and `node-gyp rebuild` being run through
`electron-builder`, as we already rebuild native bindings against
the respective CPU arch through `make electron-develop`, and
electron-builder doesn't appear to respect package's install / rebuild scripts.

Change-Type: patch


lurch:
I guess this will make electron-builder slightly faster too? :)

PR #2155, opened 2018-03-27
chore: don't lint frontend files (ie HTML) on the Node CLI pipeline

jviotti:
Change-Type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "concourse-build-electron": "make webpack",
     "concourse-test": "npm test",
     "concourse-test-electron": "npm test",
-    "concourse-test-node-cli": "make lint test-cli sanity-checks"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "concourse-build-electron": "make webpack",
     "concourse-test": "npm test",
     "concourse-test-electron": "npm test",
-    "concourse-test-node-cli": "make lint test-cli sanity-checks"
+    "concourse-test-node-cli": "make lint-js lint-cpp lint-spell test-cli"
```
lurch: Might make sense to add a phoney target (e.g. `lint-cli`) to the Makefile which depends on `lint-js lint-cpp lint-spell` ?
Why was `sanity-checks` removed?

PR #2156, opened 2018-03-27
fix: ensure flash quantity fields are finite

Shou:
We replace the `_.identity` predicate with `_.isFinite` to ensure
the flash quantity fields are numbers.

Change-Type: patch
Changelog-Entry: Ensure flash quantity fields are finite.

### DIFF
--- a/lib/shared/store.js
+++ b/lib/shared/store.js
@@ -246,14 +246,14 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
 
       verifyNoNilFields(action.data, flashStateNoNilFields, 'flash')
 
-      if (_.every(_.pick(action.data, [
+      if (!_.every(_.pick(action.data, [
         'flashing',
         'verifying',
         'succeeded',
         'failed'
-      ]), _.identity)) {
+      ]), _.isFinite)) {
         throw errors.createError({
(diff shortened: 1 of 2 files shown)


jhermsmeier:
@Shou what's the reason for this extensive type checking? Do we really need this?

Shou:
@jhermsmeier we could consolidate the `isFinite` and `null` checking here, if that's what you mean

Shou:
Rebased this now.

jhermsmeier:
Yeah nah, I mean these values shouldn't ever be anything other than a number, and throwing in store changes means that it'll generate an endless stream of error dialogs should something go wrong (i.e. as with the "Invalid state percentage" errors) making it near impossible to quit the application or debug it. So the question somewhat remains; do we *really* need this here? I think we could ensure this by other means (i.e. tests, sane defaults)

Shou:
Yeah, I guess it's a roundabout way of adding types, when we could just be using TypeScript directly and enforce it at compile-time, especially given we want to converge with the frontend team. This will work for the GUI code, and `d.ts` typings could be used for code outside the GUI so we don't have to rewrite anything (not that we can either, yet).

Should we retrofit this to be only for field existence, or remove it completely then?

PR #2148, opened 2018-03-26
chore(webpack): Enable realtime progress display

jhermsmeier:
This enables realtime progress display for webpack, which can be
governed by setting WEBPACK_PROGRESS to various values.
See https://github.com/dominique-mueller/simple-progress-webpack-plugin

Change-Type: patch

PR #2157, opened 2018-03-28
chore(package): Replace custom missing require check

jhermsmeier:
This replaces the custom bash script with rules from the eslint-node plugin
to check for `require()`d modules that don't exist.
This will also check requires of local modules, and installed packages,
yielding a more thorough check.

Change-Type: patch
Connects to: https://github.com/resin-io/etcher/pull/2148#issuecomment-376463769


lurch:
Neat!

PR #2158, opened 2018-03-29
refactor: inline middle-ellipsis package as util

Shou:
We remove our dependency on the `middle-ellipses` package and include
and expose the function in `lib/shared/utils`, alongside the regular
Angular filter it exposes. This allows use of the middle ellipsis
method in React.

Change-Type: patch
Changelog-Entry: Inline middle-ellipsis package as util.

### DIFF
--- a/lib/gui/app/utils/middle-ellipsis/filter.js
+++ b/lib/gui/app/utils/middle-ellipsis/filter.js
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2018 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


jhermsmeier:
@Shou lgtm, can we move this to the GUI, instead of putting it in lib/shared? As we're moving towards having the SDK standalone etc; lib/shared has to go away, and we'll save us some work by not putting new things in there

Shou:
Moved to `lib/gui/app/utils` now

lib/gui/app/utils/middle-ellipsis/filter.js
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+/**
+ * The purpose of this module is to provide utilities
+ * to work with sizes in bytes.
```
lurch: Typo? ;-)

Shou: Copy-paste from the package – we're just inlining its contents in our codebase  – thanks for noticing!

lurch: > we're just inlining its contents in our codebase

Does the original author need to be credited?

Shou: I didn't really think about that since it's @jviotti's package and he approved, but that's a good call.

Issue #2032, opened 2018-02-11
Image title overflows with Japanese characters


Shou:
![image](https://user-images.githubusercontent.com/819413/36073782-2b262b16-0f2e-11e8-8690-a7f7bde5e854.png)

jhermsmeier:
Is this an image name that contains Japanese characters, or does the ellipsing make them show up out of thin air?

lurch:
Is this fixed by #2160 ?

Shou:
That it is, thanks @lurch!

PR #2159, opened 2018-03-29
fix(GUI): hide the size label given multiple devices

Shou:
We hide the size label when there are multiple selected devices, as this
doesn't make sense because you could only logically show the sum size,
which isn't very helpful on its own.

Change-Type: patch
Changelog-Entry: Hide the size label given multiple devices.

### DIFF
--- a/lib/gui/app/pages/main/templates/main.tpl.html
+++ b/lib/gui/app/pages/main/templates/main.tpl.html
@@ -70,7 +70,11 @@
               <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
               {{ drive.getDrivesTitle() | middleEllipses:20 }}
             </span>
-            <span class="step-drive step-size">{{ main.selection.getCurrentDrive().size | closestUnit }}</span>
+            <span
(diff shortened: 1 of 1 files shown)


Shou:
After

image

Before

image

lib/gui/app/pages/main/templates/main.tpl.html
```diff
@@ -70,7 +70,11 @@
               <!-- middleEllipses errors on undefined, therefore fallback to empty string -->
               {{ drive.getDrivesTitle() | middleEllipses:20 }}
             </span>
-            <span class="step-drive step-size">{{ main.selection.getCurrentDrive().size | closestUnit }}</span>
+            <span
+              ng-if="main.selection.getSelectedDevices().length < 2"
```
jhermsmeier: Should this be `.length === 1`, as we also don't need to render the size when no drives are selected?

Shou: Good point, it's also probably easier to understand the intent with `=== 1`. Done that now.

lurch:
I guess the screenshots here are also still showing #2032 ?

PR #2160, opened 2018-03-29
fix(GUI): line wrap selector size subtitles wholly

Shou:
We line wrap the main-page image and drive size labels as a whole
instead of partially, if the drive or image title is long enough for a
line wrap.

Change-Type: patch
Changelog-Entry: Line wrap selector size subtitles wholly


Shou:
After

image

Before

image

PR #2166, opened 2018-04-02
fix(lint): Fix html-lint error about "lang" attr

jhermsmeier:
This avoids a new error from `html-lint` about the "lang" attr
on `<html>` tags, which isn't applicable to templates.

Change-Type: patch
Depends On: #2167


jhermsmeier:
As the error message contains UTF-8 chars, the ASCII-only sanity check fails – #2167 fixes that

PR #2167, opened 2018-04-02
test(lint): Replace ASCII-only check with ESLint rule

jhermsmeier:
This replaces the `ensure-all-text-files-only-ascii` sanity-check
with an ESLint rule to only check identifiers in code against being
printable ASCII, as there are places where UTF-8 in content is required,
which would make the previous check fail.

Change-Type: patch
Connects To: #2166


jhermsmeier:
Tests will still require https://github.com/resin-io/etcher/pull/2166 to pass

PR #2170, opened 2018-04-03
minifix: use multi-drive methods with drive-list warning button

Shou:
Change-Type: patch
Changelog-Entry: Use multi-drive methods with drive-list warning button.


Shou:
After
image

Before
image

PR #2173, opened 2018-04-04
fix(gui): Ensure image/drive size displayed on new line

jhermsmeier:
This forces the image and device size to always be displayed on the
second line for visual consistency.

### Before:

step-name-before

### After:

step-name-name

Change-Type: patch


lurch:
@jhermsmeier Could you also show a screenshot of when an image has been selected but a drive hasn't been selected yet?

jhermsmeier:
@lurch here ya go :)

step-name-partial

lurch:
Nice. I'd forgotten that the "Change" link / button only appears _after_ selecting a disk ;-)

PR #2183, opened 2018-04-06
fix(writer): Fix erronous event handling in write pipeline

jhermsmeier:
This fixes the use and handling of events in the write pipeline,
such that the pipeline would not be prematurely stalled or terminated.
Also, a new `fail` event is introduced, to signal non-fatal errors.

Change-Type: patch

### DIFF
--- a/lib/sdk/writer/index.js
+++ b/lib/sdk/writer/index.js
@@ -217,6 +217,7 @@ class ImageWriter extends EventEmitter {
 
     mountutils.unmountDisk(destination.device.device, (error) => {
       debug('state:unmount', destination.device.device, error ? 'NOT OK' : 'OK')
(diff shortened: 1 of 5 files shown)


lib/sdk/writer/index.js
```diff
@@ -286,12 +288,13 @@ class ImageWriter extends EventEmitter {
...
       debug('state:destination-open', destination.device.raw, error ? 'NOT OK' : 'OK')
       destination.fd = fd
+      destination.error = error
       callback.call(this, error)
     })
   }
 
   /**
-   * @summary Check a destinstation against the drive constraints
+   * @summary Check a destination against the drive constraints
```
lurch: Something to add to the codespell dictionary? ;)

PR #2189, opened 2018-04-09
Enable nodeGypRebuild for Etcher builds

brownjohnf:
This will ensure we have all bindings built, even when using cached
modules.

Change-Type: patch


lurch:
Does this conflict with #2154 ?

brownjohnf:
It turns out that the concourse pipeline doesn't honor the change to the electron builder config. You need to override it in the .resinci file. I've updated the PR to do that.

@jviotti do you think we want to make the node gyp prebuild default to true in the electron pipeline? this seems like it will be an issue for all electron bindings, even in other projects.

PR #2196, opened 2018-04-12
fix(GUI): pass empty string fallback to middle-ellipsis filter

Shou:
We ensure that the middle-ellipsis filter doesn't explode when it's
passed an undefined value by passing a default empty string in those
template functions.

Change-Type: patch


lurch:
Would modifying the middle-ellipsis filter _itself_ be a more robust solution? \*shrug\*

jhermsmeier:
This can be closed now, right @Shou?

PR #2198, opened 2018-04-12
fix(GUI): display untitled device when device lacks description

Shou:
We fallback to `'Untitled Device'` when the device lacks a
`.description` field.

Change-Type: patch
Changelog-Entry: Display Untitled Device when the device lacks a
description field.

### DIFF
--- a/tests/gui/pages/main.spec.js
+++ b/tests/gui/pages/main.spec.js
@@ -260,8 +260,8 @@ describe('Browser: MainPage', function () {
     let DriveSelectionController
 
     const drivePaths = process.platform === 'win32'
-      ? [ 'E:\\', 'F:\\' ]
-      : [ '/dev/disk1', '/dev/disk2' ]
+      ? [ '\\\\.\\PhysicalDrive1', '\\\\.\\PhysicalDrive2', '\\\\.\\PhysicalDrive3' ]
+      : [ '/dev/disk1', '/dev/disk2', '/dev/disk3' ]
     const drives = [
       {
         device: drivePaths[0],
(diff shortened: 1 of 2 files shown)


jhermsmeier:
While this should not be needed anymore after https://github.com/resin-io/etcher/pull/2207 – it might make sense to default to "Untitled" at the appropriate place (which is perhaps drivelist)? Wdyt @Shou?

Shou:
@jhermsmeier we should probably leave `drivelist` as-is, I think, in case someone might want to rely on there not being a label for the device, should that ever happen. We can add a null-check at the GUI layer and return `'Untitled'` there.

jhermsmeier:
Good point, let's do that then

Shou:
@jhermsmeier done

tests/gui/pages/main.spec.js
```diff
@@ -260,8 +260,8 @@ describe('Browser: MainPage', function () {
     let DriveSelectionController
 
     const drivePaths = process.platform === 'win32'
-      ? [ 'E:\\', 'F:\\' ]
-      : [ '/dev/disk1', '/dev/disk2' ]
+      ? [ 'E:\\', 'F:\\', 'G:\\' ]
```
jhermsmeier: These aren't device paths, but volume paths – should be `\\.\PhysicalDrive1` etc.

tests/gui/pages/main.spec.js
```diff
@@ -280,6 +280,14 @@ describe('Browser: MainPage', function () {
         mountpoints: [ drivePaths[1] ],
```
jhermsmeier: Mildly unrelated to this particular PR, but mountpoints have the structure
```js
[{ path: '/private/var/vm', label: 'VM' }, ... ]
```
now

Shou: Yeah I noticed but I left it so we can mend this in another PR, they're not essential to those tests atm either.

tests/gui/pages/main.spec.js
Shou: Done

PR #2202, opened 2018-04-16
feat(GUI): remove unnecessary status dots

Shou:
We remove usage of the status dots except when failed devices occur, in
which case we still display the red failed dot and quantity. We also use
singular and plural depending on the quantity of failed devices.

Change-Type: patch

### DIFF
--- a/lib/gui/app/pages/main/templates/main.tpl.html
+++ b/lib/gui/app/pages/main/templates/main.tpl.html
@@ -111,12 +111,13 @@
           <span>ETA: {{ main.state.getFlashState().eta | secondsToDate | amDateFormat:'m[m]ss[s]' }}</span>
         </p>
 
-        <div class="target-status-wrap" ng-if="main.state.isFlashing()">
-          <div class="target-status-line target-status-{{ type }}"
-            ng-repeat="(type, quantity) in main.state.getFlashQuantities()">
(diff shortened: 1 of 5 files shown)


Shou:
image

lib/gui/app/pages/main/templates/main.tpl.html
```diff
@@ -111,12 +111,13 @@
...
-              <span class="target-status-dot"></span>
-              <span class="target-status-quantity">{{ quantity }}</span>
-              <span class="target-status-message">{{ main.progressMessage[type]() }}</span>
+        <div class="target-status-wrap" ng-if="main.state.getFlashState().failed">
+          <div class="target-status-line target-status-failed">
+            <span class="target-status-dot"></span>
+            <span class="target-status-quantity">{{ main.state.getFlashState().failed }}</span>
+            <span class="target-status-message">{{
+              main.progressMessage.failed(main.state.getFlashState().failed)
+            }}</span>
```
jhermsmeier: Succeeded devices aren't displayed at all anymore, even though there's code for it in `lib/shared/messages.js`?

Shou: Yeah, it'll return with the success page PR so I left it in purely out of convenience.

PR #2203, opened 2018-04-16
feat(GUI): make the progress button blue on verification

Shou:
We make the progress button blue on verification.

Change-Type: patch
Changelog-Entry: Make the progress button blue on verification.


Shou:
image

PR #2204, opened 2018-04-16
minifeat(GUI): prefix multiple devices label with quantity

Shou:
Change the `Multiple Devices (n)` label on selected devices to a
quantity-prefixed form `n Devices`.

Change-Type: patch


Shou:
image

PR #2206, opened 2018-04-16
feat(GUI): display succeeded and failed devices on finish screen

Shou:
We display the quantity of succeeded and failed devices using status
dots on the finish screen.

Change-Type: patch
Changelog-Entry: Display succeeded and failed device quantities on the
finish screen.

### DIFF
--- a/lib/gui/app/modules/image-writer.js
+++ b/lib/gui/app/modules/image-writer.js
@@ -126,16 +126,25 @@ exports.performWrite = (image, drives, onProgress) => {
       console.log(message)
     })
 
+    const flashResults = {
+      devices: {
+        succeeded: 0,
+        failed: 0
+      }
+    }
     ipc.server.on('fail', (error) => {
       console.log('Fail:', error)
     })
 
-    const flashResults = {}
     ipc.server.on('done', (results) => {
       _.merge(flashResults, results)
     })
 
-    ipc.server.on('state', onProgress)
+    ipc.server.on('state', (progress) => {
+      flashResults.devices.succeeded = progress.succeeded
+      flashResults.devices.failed = progress.failed
+      onProgress(progress)
+    })
 
     ipc.server.on('ready', (data, socket) => {
       ipc.server.emit(socket, 'write', {
--- a/lib/gui/app/pages/finish/controllers/finish.js
+++ b/lib/gui/app/pages/finish/controllers/finish.js
@@ -16,10 +16,12 @@
 
 'use strict'
 
+const _ = require('lodash')
 const settings = require('../../../models/settings')
 const flashState = require('../../../../../shared/models/flash-state')
 const selectionState = require('../../../../../shared/models/selection-state')
 const analytics = require('../../../modules/analytics')
+const messages = require('../../../../../shared/messages')
 
 module.exports = function ($state) {
   /**
@@ -30,11 +32,13 @@ module.exports = function ($state) {
   this.settings = settings
 
   /**
-   * @summary Source checksum
-   * @type {String}
+   * @summary Flash state
+   * @type {Object}
    * @public
    */
-  this.checksum = flashState.getLastFlashSourceChecksum()
+  this.flash = flashState
+
+  this.progressMessage = messages.progress
 
   /**
    * @summary Restart the flashing process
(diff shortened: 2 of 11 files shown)


Shou:
image

lib/gui/app/modules/image-writer.js
```diff
@@ -126,16 +126,23 @@ exports.performWrite = (image, drives, onProgress) => {
...
     ipc.server.on('done', (results) => {
       _.merge(flashResults, results)
     })
 
-    ipc.server.on('state', onProgress)
+    ipc.server.on('state', (progress) => {
+      flashResults.devices = {
+        succeeded: progress.succeeded,
+        failed: progress.failed
+      }
```
jhermsmeier: Can we create this object once (and also init values to zero), and only update the values in this handler, to avoid thrashing the GC with all the created objects as well as potential ReferenceErrors (i.e. when accessing this before a state event happened)?

lib/gui/app/modules/image-writer.js
```diff
@@ -126,16 +126,23 @@ exports.performWrite = (image, drives, onProgress) => {
       console.log(message)
     })
 
+    const flashResults = { errors: [] }
     ipc.server.on('fail', (error) => {
+      flashResults.errors.push(error)
```
jhermsmeier: Isn't this information available through the flash results from the writer once it finished? Why do we need to track it again? The success page doesn't show until the flash results are in anyways, correct?

jhermsmeier:
The spacing between the "Flash Complete", the dots, and the checkmark icon seems a bit off to me – is that intentional / the only way they fit?

lib/gui/app/pages/finish/controllers/finish.js
```diff
@@ -55,4 +59,22 @@ module.exports = function ($state) {
...
+   *
+   * @returns {String} formatted errors
+   *
+   * @example
+   * const errors = FinishController.formattedErrors()
+   * console.log(errors)
+   */
+  this.formattedErrors = () => {
+    const errors = _.map(_.get(flashState.getFlashResults(), [ 'results', 'errors' ]), (error) => {
+      return `${error.device}: ${error.code}`
```
jhermsmeier: Should this include the error message as well?

Shou: The EUNPLUGGED error doesn't seem to have one, but I've instead opted for `error.message || error.code`.

Shou:
@jhermsmeier the intention is to fit numbers up to 999 with the spacing, although ideally it'd stretch to fit

jhermsmeier:
@Shou regarding the spacing, I mean vertically – the vertical alignment / rhythm seems off. Also, should these also get the proper pluralization treatment?

Shou:
My bad, forgot to pass the quantities!

image

lib/gui/app/pages/finish/controllers/finish.js
```diff
@@ -55,4 +59,23 @@ module.exports = function ($state) {
...
+   *
+   * @returns {String} formatted errors
+   *
+   * @example
+   * const errors = FinishController.formattedErrors()
+   * console.log(errors)
+   */
+  this.formattedErrors = () => {
+    const errors = _.map(_.get(flashState.getFlashResults(), [ 'results', 'errors' ]), (error) => {
+      console.log(error)
```
jhermsmeier: Is it intended for this to stick around?

lib/sdk/writer/index.js
```diff
@@ -636,7 +636,13 @@ class ImageWriter extends EventEmitter {
...
         bytesRead: this.bytesRead,
         bytesWritten: this.bytesWritten,
-        checksum: this.checksum
+        checksum: this.checksum,
+        errors: Array.from(this.destinations).map(([ device, dest ]) => {
+          if (dest.error) {
+            dest.error.device = device
+          }
+          return dest.error
+        }).filter(Boolean)
```
jhermsmeier: Why not filter first? 

```js
Array.from(this.destinations).filter((dest) => {
  return dest.error
}).map(([ device, dest ]) => {
  dest.error.device = device
  return dest.error
})
```

jhermsmeier: Also, the `device` property will need adding to `errors.toJSON()` in `lib/shared/errors.js`

Shou:
@konmouz how does this look?

image

Issue #2071, opened 2018-02-21
Unsafe Drives Displayed/Selectable


turbouser:
- Etcher version: Etcher-Portable-1.3.1-x64.exe
- Operating system and architecture: Windows 8.1 x64
- Image flashed: 2017-11-29-raspbian-stretch-lite.img
- Do you see any meaningful error information in the DevTools? No

Unsafe mode is NOT checked, but the following drives are being offered in addition to the expected flash drive:

WD My Book 1230 USB Device - 4 TB
WDC WD1600ADFS75SLR2 - 160 GB

These drives have never been shown in any version of Etcher prior to 1.3.1 (unless Unsafe mode was checked).

jhermsmeier:
This is because the drive detection has undergone a significant change in 1.3.1, which now eliminates devices based on how they're connected. I'm guessing those two drives are connected via USB?

turbouser:
WD My Book is an external 4 TB USB drive used for daily system backups

WDC WD1600ADFS75SLR2 is an internal SATA system drive.

Allowing Etcher to write to either of these drives would be a disaster here.  I would prefer they both remain hidden as they have been in all previous versions of Etcher.

lurch:
Just FYI the drive-detection component that's recently had a big change is https://github.com/resin-io-modules/drivelist
So if you'd like to install that and let us know what it outputs for your drives, that'd be really helpful.

Also, in the next version of Etcher there'll be a feature ( #2045 ) that'll warn you about any drives bigger than 64GB.

turbouser:
I don't see an executable version of drivelist for Windows 8.1 x64.

lurch:
Pinging @jhermsmeier for drivelist-on-Windows help :)

turbouser:
Any chance you could add a user option to hide unsafe drives that are showing up so there's absolutely no chance of accidentally selecting one of them?

lurch:
You mean like a blacklist? We previously discussed that in #1728 which details the complexities of such an approach.

jhermsmeier:
Etcher can only do so much; and this generally is a thing where you need to pay attention to what you're doing – if it's displaying devices that should be deemed "unsafe", then that's a bug – but I don't think putting in safety-cages at every turn is a good idea. It hurts usability considerably, and in the end you still need to be aware of what's going on.

lurch:
@jhermsmeier the "internal SATA system drive" shouldn't be being displayed if unsafe mode is disabled, right?

[...]

PR #2209, opened 2018-04-17
fix(adapters): Always ignore RAID attached devices

jhermsmeier:
This force-ignores RAID attached devices, to ensure they aren't flashed, not even intentionally,
as that can have yet unknown consequences.

Change-Type: patch
Changelog-Entry: Exclude RAID devices from drive selection list
Connects To: https://github.com/resin-io/etcher/issues/2071

### DIFF
--- a/lib/sdk/adapters/blockdevice/index.js
+++ b/lib/sdk/adapters/blockdevice/index.js
@@ -79,6 +79,11 @@ class BlockDeviceAdapter extends EventEmitter {
       this.emit('error', error)
       callback && callback(error)
     }).filter((drive) => {
+      // Always ignore RAID attached devices, as they are in danger-country;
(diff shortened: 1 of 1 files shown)


lurch:
Just out of curiosity: For which OSes is drivelist currently able to detect RAID devices?
And does this change just ignore the RAID device itself, or does it also ignore the sub-devices that actually compose the RAID?

jhermsmeier:
On Windows it enumerates the devices belonging to the RAID, i.e. the individual block devices themselves; not sure how it is on Mac OS, but so far we can detect RAID on Win & Mac

jhermsmeier:
Thinking about it, we should prob. add an `.isRAID` property in drivelist instead of relying on the `.busType`

PR #2211, opened 2018-04-18
fix(image-writer): Remove use of _.isError

jhermsmeier:
`_.isError()` returns `true` for anything that has a `name` and `message`
property, causing the check here to always keep the plain object as error.

Change-Type: patch


jhermsmeier:
@zvin these here, are objects are from the child process via IPC as JSON, whereas the error events the docs talk about are errors of the IPC lib itself (socket errors, etc)

PR #2215, opened 2018-04-18
Revert "feat(lib): Use win-drive-clean instead of diskpart"

jhermsmeier:
This reverts commit 47fc1b7357bdb9e9aa8e2d7476690435087d984e
in order to prevent a possible regression, until properly investigated
and fixed.

Change-Type: patch
Changelog-Entry: Revert using native binding to clean disks on Windows

### DIFF
--- a/lib/cli/diskpart.js
+++ b/lib/cli/diskpart.js
@@ -16,13 +16,70 @@
 
 'use strict'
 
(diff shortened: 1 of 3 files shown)


lurch:
Effectively reverts #2023

Issue #2197, opened 2018-04-12
Raspberry Pi Compute Module (and rpi0) enumeration fails


curcuz:
- **Etcher version:**  1.4.1
- **Operating system and architecture:**  OSX 10.13.4
- **Image flashed:**  resinOS v2.12.5-rev2 prod
- **Do you see any meaningful error information in the DevTools?**  yes, see below:
![screen shot 2018-04-12 at 20 35 37](https://user-images.githubusercontent.com/2338223/38697292-d8922baa-3e91-11e8-817f-95545f169dcd.png)

```
gui.js:105102 {"name":"Error","message":"ENOENT: no such file or directory, open 'lib/blobs/usbboot/raspberrypi/bootcode.bin'","cause":{"errno":-2,"code":"ENOENT","syscall":"open","path":"lib/blobs/usbboot/raspberrypi/bootcode.bin"},"isOperational":true,"stack":"Error: ENOENT: no such file or directory, open 'lib/blobs/usbboot/raspberrypi/bootcode.bin'","errno":-2,"code":"ENOENT","syscall":"open","path":"lib/blobs/usbboot/raspberrypi/bootcode.bin"}
console.(anonymous function) @ gui.js:105102
gui.js:105102 etcher:sdk:usbboot Serial number index: 0 +1m
gui.js:105102 etcher:sdk:usbboot [20:18] -> 10% +0ms
gui.js:105102 etcher:sdk:usbboot Opening device: 20:18 +0ms
gui.js:105102 etcher:sdk:usbboot Claiming interface: 0 +0ms
gui.js:105102 Unhandled rejection Error: ENOENT: no such file or directory, open 'lib/blobs/usbboot/raspberrypi/bootcode.bin'
```

lurch:
Just FYI: even without this error, you'd still be unable to enumerate the CM1 or RPi Zero on OSX, due to a bug in the bootROM in the BCM2835 https://www.raspberrypi.org/documentation/hardware/computemodule/cm-emmc-flashing.md

curcuz:
@lurch thx for the info - FYI this also happens on CM3

Shou:
This doesn't seem to affect Linux btw, usbboot runs as usual here.

PR #2216, opened 2018-04-19
fix(usbboot): Move blob handling to SDK

jhermsmeier:
This moves the usbboot blob handling into the SDK to avoid
root dirname conflicts through shimmed __dirname in bundled UI
and different contexts of execution.

Change-Type: patch
Connects To: #2197


mariob022:
LGTM, and confirmed works on Linux with AppImages and npm start

PR #2218, opened 2018-04-19
fix(gui): De-serialize errors from flashResults

jhermsmeier:
Change-Type: patch


lurch:
A bit like a return of the `Robot` protocol? :wink:

PR #2220, opened 2018-04-20
test: ensure webpack paths are handled correctly

Shou:
We ensure the paths are rewritten correctly to exclude the `sdk` and
`shared` folders, and `package.json`, from the Webpack bundle.

Change-Type: patch


Shou:
We can expand this in the future, e.g. test for https://github.com/resin-io/etcher/pull/2195 when that's in.

PR #2229, opened 2018-04-23
refactor: use word target instead of drive/device

Shou:
Change-Type: patch

### DIFF
--- a/lib/gui/app/pages/main/controllers/drive-selection.js
+++ b/lib/gui/app/pages/main/controllers/drive-selection.js
@@ -39,10 +39,10 @@ module.exports = function (DriveSelectorService) {
 
     // eslint-disable-next-line no-magic-numbers
     if (drives.length === 1) {
-      return _.head(drives).description || 'Untitled Device'
+      return _.head(drives).description || 'Untitled Target'
     }
 
-    return `${drives.length} Devices`
+    return `${drives.length} Targets`
   }
 
   /**
--- a/lib/shared/messages.js
+++ b/lib/shared/messages.js
@@ -35,13 +35,13 @@ module.exports = {
     succeeded: (quantity) => {
       // eslint-disable-next-line no-magic-numbers
       const plural = quantity === 1 ? '' : 's'
-      return `Succeeded device${plural}`
+      return `Succeeded target${plural}`
     },
 
     failed: (quantity) => {
       // eslint-disable-next-line no-magic-numbers
       const plural = quantity === 1 ? '' : 's'
-      return `Failed device${plural}`
+      return `Failed target${plural}`
     }
   },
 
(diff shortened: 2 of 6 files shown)


lib/gui/app/pages/main/controllers/drive-selection.js
```diff
@@ -39,10 +39,10 @@ module.exports = function (DriveSelectorService) {
 
     // eslint-disable-next-line no-magic-numbers
     if (drives.length === 1) {
-      return _.head(drives).description || 'Untitled Device'
+      return _.head(drives).description || 'Untitled Target'
```
jhermsmeier: This would still be a device, imo

lib/shared/messages.js
```diff
@@ -84,16 +84,16 @@ module.exports = {
...
 
     containsImage () {
-      return 'Drive Contains Image'
+      return 'Target Contains Image'
     },
 
     // The drive is large and therefore likely not a medium you want to write to.
     largeDrive () {
-      return 'Large Drive'
+      return 'Large Target'
```
jhermsmeier: I'm thinking maybe this could be left as "Large Device", or maybe "Large Storage"?

lib/shared/messages.js
```diff
@@ -156,8 +156,8 @@ module.exports = {
 
     notEnoughSpaceInDrive: () => {
       return [
-        'Not enough space on the drive.',
-        'Please insert larger one and try again.'
+        'Not enough space on the target.',
+        'Please insert a larger target and try again.'
```
jhermsmeier: "insert a larger target" sounds weird to me – maybe we should use "insert a larger device" here, or rather "connect a larger device"?

lib/shared/messages.js
```diff
@@ -200,17 +200,17 @@ module.exports = {
 
     driveUnplugged: () => {
       return [
-        'Looks like Etcher lost access to the drive.',
+        'Looks like Etcher lost access to the target.',
         'Did it get unplugged accidentally?',
-        '\n\nSometimes this error is caused by faulty readers that don\'t provide stable access to the drive.'
+        '\n\nSometimes this error is caused by faulty readers that don\'t provide stable access to the target.'
```
jhermsmeier: Here, in the second mention, "target" seems awkwardly unspecific, maybe "target" -> "target device"?

lib/shared/messages.js
```diff
@@ -200,17 +200,17 @@ module.exports = {
...
       ].join(' ')
     },
 
     inputOutput: () => {
       return [
-        'Looks like Etcher is not able to write to this location of the drive.',
-        'This error is usually caused by a faulty drive, reader, or port.',
-        '\n\nPlease try again with another drive, reader, or port.'
+        'Looks like Etcher is not able to write to this location of the target.',
+        'This error is usually caused by a faulty target, reader, or port.',
```
jhermsmeier: "target" -> "device" here? Not sure a "faulty target" makes sense

lib/shared/messages.js
```diff
@@ -200,17 +200,17 @@ module.exports = {
...
+        '\n\nPlease try again with another target, reader, or port.'
```
jhermsmeier: Same here; target -> device?

lib/shared/messages.js
lurch: Yeah, "Large Target" makes me think of 
image

PR #2268, opened 2018-04-24
refactor(GUI): make the finish notification message concise

Shou:
We make the finish notification message print the device name as usual
when there's one target, and instead list quantity of successful and
failed devices when there are multiple. Previously it would list all
device names, and wouldn't specify how many were successful or failures.

Change-Type: patch


Shou:
image
image

Shou:
Updated, with new screenshots above.

Issue #2267, opened 2018-04-24
[Bug Report] Etcher 1.4.3 - Drive Select GUI Error


Richyread:
- **Etcher version:** 1.4.3
- **Operating system and architecture:** Win 10(64)
- **Image flashed:** Raspbian Stretch Lite (13-03-2018)
- **Do you see any meaningful error information in the DevTools?** See screenshot attached.

I have upgraded my Etcher installation this morning and encountered an issue with writing an image to an SD Card - specifically that the device/drive selection step appears to accept and then immediately cancel the choice within the GUI ~ see shot below for better description 😄 

![etcher_143_error](https://user-images.githubusercontent.com/7470850/39170481-a54fff5a-4793-11e8-8876-8b28cae020e2.gif)

**Workflow prior to upgrade:**

- Insert SD Card into reader

- Close Win10's 'Please Format Drive' prompts

- Launcher Etcher

- Select image file (including .ZIP archived images)

- Select USB Card Reader

- Click 'Flash'

Etcher would then in chain of actions, Format the Drive(s), Extract the image, Write the image, Tidy up, Eject SD Reader.

Since the upgrade I'm unable to proceed past the 'Select Drive' stage.

Screenshot of Etcher_Console, below.

![etcher_143_console](https://user-images.githubusercontent.com/7470850/39170494-adfd0634-4793-11e8-8342-41b1126cb348.png)

Hope this helps - let me know if you need anything further etc.

Many thanks,

Richy.

PecoraInPannaCotta:
Same on ubuntu 17.10

nex86:
same issue. going back to 1.3.1!

Shou:
What happens when you double-click?

nex86:
it keeps deselecting even when you double click.

PecoraInPannaCotta:
If you go faster than the GUI, you can actually flash the ISO, so is confirmed is just a UI problem

fivestones:
I wish I were as fast as you :-)

Shou:
I've identified the issue and fixed it in the above PR. We'll have 1.4.4 out shortly.

PR #2272, opened 2018-04-24
fix(store): keep single warning-drives selected

Shou:
We ensure that drive-image pairs with warnings don't get deselected when
there is only one drive available. This happenened because there was no
check for any previous selected devices. Comes with a test case.

Fixes: https://github.com/resin-io/etcher/issues/2267
Change-Type: patch
Changelog-Entry: Keep single warning-drive-image pairs selected.


lurch:
Perhaps https://github.com/resin-io/etcher/blob/master/docs/MANUAL-TESTING.md should be updated too?

jhermsmeier:
"Comes with a test case." :D

PR #2276, opened 2018-04-25
feat(GUI): add convenience localstorage class

Shou:
We add a class `Storage` and accompanying helper methods that makes
localStorage usage easier.

Change-Type: patch
Changelog-Entry: Add a convenience Storage class on top of localStorage.

### DIFF
--- a/lib/gui/app/models/local-settings.js
+++ b/lib/gui/app/models/local-settings.js
@@ -21,6 +21,7 @@ const _ = require('lodash')
 const fs = require('fs')
 const path = require('path')
 const os = require('os')
+const Storage = require('./storage')
 
 /**
  * @summary Local storage settings key
@@ -28,6 +29,7 @@ const os = require('os')
  * @type {String}
  */
 const LOCAL_STORAGE_SETTINGS_KEY = 'etcher-settings'
+const settingsStorage = new Storage(LOCAL_STORAGE_SETTINGS_KEY)
 
 /**
  * @summary Local settings filename
@@ -86,7 +88,7 @@ exports.readAll = () => {
   const workdirConfigPath = path.join(process.cwd(), RCFILE)
   const settings = {}
   return Bluebird.try(() => {
-    _.merge(settings, JSON.parse(window.localStorage.getItem(LOCAL_STORAGE_SETTINGS_KEY)))
+    _.merge(settings, settingsStorage.getAll())
   }).return(readConfigFile(homeConfigPath))
     .then((homeConfig) => {
       _.merge(settings, homeConfig)
@@ -114,9 +116,8 @@ exports.readAll = () => {
  * });
  */
 exports.writeAll = (settings) => {
-  const INDENTATION_SPACES = 2
   return Bluebird.try(() => {
-    window.localStorage.setItem(LOCAL_STORAGE_SETTINGS_KEY, JSON.stringify(settings, null, INDENTATION_SPACES))
+    settingsStorage.setAll(settings)
   })
 }
 
@@ -137,6 +138,6 @@ exports.writeAll = (settings) => {
  */
 exports.clear = () => {
   return Bluebird.try(() => {
-    window.localStorage.removeItem(LOCAL_STORAGE_SETTINGS_KEY)
+    settingsStorage.clearAll()
   })
 }
--- a/lib/gui/app/models/storage.js
+++ b/lib/gui/app/models/storage.js
@@ -54,7 +54,8 @@ class Storage {
       // JSON.parse(null) === null, so we fallback to {}
       return JSON.parse(window.localStorage.getItem(this.superkey)) || {}
     } catch (err) {
-      return {}
+      this.setAll({})
+      throw err
     }
   }
 
@@ -138,8 +139,7 @@ class Storage {
     }
 
     // eslint-disable-next-line lodash/prefer-lodash-method
-    const json = JSON.stringify(Object.assign(obj, { [key]: result }), null, INDENTATION_SPACES)
-    window.localStorage.setItem(this.superkey, json)
+    this.setAll(Object.assign(obj, { [key]: result }))
     return result
   }
 
--- a/lib/gui/app/models/storage.js
+++ b/lib/gui/app/models/storage.js
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2018 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


Shou:
We'll need `localStorage` usage in the file picker and this PR helps with that.

lib/gui/app/models/storage.js
```diff
@@ -0,0 +1,162 @@
...
+   * })
+   */
+  modify (key, func, defaultValue) {
+    const obj = this.getAll()
+
+    let result = func(defaultValue)
+    // eslint-disable-next-line no-undefined
+    if (obj[key] !== undefined) {
+      result = func(obj[key])
+    }
```
lurch: Should there be an `else` here to prevent `func` being called twice?
Also, it's not clear from the docstring that `func` will actually get called on `defaultValue`, rather than `key` being set to `defaultValue` directly.

lib/gui/app/models/local-settings.js
```diff
@@ -17,13 +17,15 @@
...
 const Bluebird = require('bluebird')
+const Storage = require('./storage')
 
 /**
  * @summary Local storage settings key
  * @constant
  * @type {String}
  */
 const LOCAL_STORAGE_SETTINGS_KEY = 'etcher-settings'
+const settingStorage = new Storage(LOCAL_STORAGE_SETTINGS_KEY)
```
lurch: `settingsStorage` instead of `settingStorage` ?

Shou: Yeah it's more correct, it just felt awkward to pronounce and type. I'll change it back.

lib/gui/app/models/storage.js
Shou: Good point, initially it was that way until the linter complained, but I'll find a different way around it.

tests/gui/models/storage.spec.js
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2017 resin.io
```
lurch: Um, 2018? :stuck_out_tongue_winking_eye:

tests/gui/models/storage.spec.js
```diff
@@ -0,0 +1,97 @@
...
+    it('should return a value', function () {
+      const value = this.testStorage.modify('fieldA', (fieldA) => {
+        return fieldA + 1
+      })
+      m.chai.expect(value).to.equal(2)
+    })
+
+    it('should use the fallback default value if field doesn\'t exist', function () {
+      const FALLBACK = 0
+      m.chai.expect(this.testStorage.modify('fieldC', _.identity, FALLBACK)).to.equal(FALLBACK)
```
lurch: Perhaps this test should use a different modify-function, to make it clearer that `fieldC` is actually being set to `_.identity(FALLBACK)` and not simply `FALLBACK`? (IYSWIM)

lib/gui/app/models/storage.js
```diff
@@ -0,0 +1,164 @@
...
+   * for (const key in potatoStorage.getAll()) {
+   *   console.log(key)
+   * }
+   */
+  getAll () {
+    try {
+      // JSON.parse(null) === null, so we fallback to {}
+      return JSON.parse(window.localStorage.getItem(this.superkey)) || {}
+    } catch (err) {
+      return {}
```
jhermsmeier: Not sure if this would be a good way to handle errors. In which situation would we expect this to fail? And if it fails, this would effectively clear all settings to default values once .setAll() is called – which in turns seems rather unsafe.

lib/gui/app/models/storage.js
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @example
+   * potatoStorage.setAll({
+   *   location: 'somewhere',
+   *   freshness: 100,
+   *   edible: true
+   * })
+   */
+  setAll (value) {
+    window.localStorage.setItem(this.superkey, JSON.stringify(value, null, INDENTATION_SPACES))
```
jhermsmeier: Analog to .getAll(), why does this not have error handling? There could certainly be the case of a value accidentally containing something that's not stringify-able, i.e. circular refs.

lib/gui/app/models/storage.js
Shou: Yeah, what @lurch said above is the error case I had in mind. I assume the case where localStorage fails in another way is extremely rare (but probably not impossible?), but how do you suggest we handle any potential errors more gracefully here? I guess we could check if it's the most common mistake, a parse error, and then return an empty object if so, and in any other error case we unleash the error upon whatever code uses it?

lib/gui/app/models/storage.js
Shou: Good call, but if we do try to stringify a circular reference, shouldn't we propagate the error and make it visible? We should probably log it with analytics too.

tests/gui/models/storage.spec.js
Shou: Thanks!

tests/gui/models/storage.spec.js
Shou: Ah, my bad, I'm too used to higher-order functions, but it's clearer without one here. I wish the linter allowed use of one-liner functions, i.e. `x => x` is just identity... I'll have a look into that.

lib/gui/app/models/storage.js
```diff
@@ -0,0 +1,164 @@
...
+    let result = null
+    // eslint-disable-next-line no-undefined
+    if (obj[key] === undefined) {
+      result = func(defaultValue)
+    } else {
+      result = func(obj[key])
+    }
+
+    // eslint-disable-next-line lodash/prefer-lodash-method
+    const json = JSON.stringify(Object.assign(obj, { [key]: result }), null, INDENTATION_SPACES)
```
zvin: You could use setAll here

lib/gui/app/models/storage.js
jhermsmeier: @Shou @lurch nothing apart from the code we run would have access to localStorage (unless the user willingly does something in the devtools), so imo this should explode if it fails, as .setAll() does

lib/gui/app/models/storage.js
Shou: Thanks

lib/gui/app/models/storage.js
Shou: How about this commit? We reset the value to an empty object and propagate the error subsequently so we don't allow the user to get stuck with a broken Etcher

tests/gui/models/storage.spec.js
```diff
@@ -0,0 +1,97 @@
...
+      })
+      m.chai.expect(value).to.equal(2)
+    })
+
+    it('should use the fallback default value if field doesn\'t exist', function () {
+      const FALLBACK = 1.5
+      m.chai.expect(this.testStorage.modify('fieldC', _.ceil, FALLBACK)).to.equal(2)
+    })
+
+    it('should be undefined if no fallback default value is given', function () {
```
lurch: Will this also set `fieldC` in `testStorage` to undefined, even if `fieldC` didn't exist before?

Shou: Good call, it seems it will, and that's a bug – luckily just a small one given we control which keys are used and don't really even fall into the bug right now.

PR #2277, opened 2018-04-25
feat(GUI): add react middle-ellipsis component

Shou:
We make a React middle-ellipsis component, and sneak in attribution to
jviotti.

Change-Type: patch
Changelog-Entry: Add a React middle-ellipsis component.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 # Javascript files must retain LF line-endings (to keep eslint happy)
 *.js text eol=lf
+*.jsx text eol=lf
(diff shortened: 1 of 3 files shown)


Shou:
This is also part of preparatory PRs for the file picker.

lurch:
> should be guard against null values being passed to middleEllipsis? Preferably in middleEllipsis itself

I suggested that previously? https://github.com/resin-io/etcher/pull/2196#issuecomment-380880012

Shou:
I'm gonna close this as I don't need it in the new file selector anymore, but I'll address the above feedback in a new PR.

PR #2295, opened 2018-04-30
fix(GUI): fix multi-writes analytics

Shou:
We make the analytics block into a function `handleErrorLogging` and
use it in the fail event that happens during multi-writes. Previously
error events would only be handled when single drives were flashed.

Change-Type: patch
Changelog-Entry: Fix multi-writes analytics by reusing existing logic in
multi-write events.

### DIFF
--- a/lib/gui/app/modules/image-writer.js
+++ b/lib/gui/app/modules/image-writer.js
@@ -61,6 +61,35 @@ const getApplicationEntryPoint = () => {
 }
 
 /**
+ * @summary Handle a flash  error and log it to analytics
+ * @function
+ * @private
+ *
+ * @param {Error} error - error object
+ * @param {Object} analyticsData - analytics object
+ *
+ * @example
+ * handleErrorLogging({ code: 'EUNPLUGGED' }, { image: 'resin.img' })
+ */
+const handleErrorLogging = (error, analyticsData) => {
+  if (error.code === 'EVALIDATION') {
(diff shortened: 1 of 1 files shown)


lib/gui/app/modules/image-writer.js
```diff
@@ -275,21 +311,7 @@ exports.flash = (image, drives) => {
...
-    } else if (error.code === 'ENOSPC') {
-      analytics.logEvent('Out of space', analyticsData)
-    } else if (error.code === 'ECHILDDIED') {
-      analytics.logEvent('Child died unexpectedly', analyticsData)
-    } else {
-      analytics.logEvent('Flash error', _.merge({
-        error: errors.toJSON(error)
-      }, analyticsData))
-    }
+    handleErrorLogging(error, analyticsData)
```
jhermsmeier: We might get some double-reporting here, if all devices fail over the course of the flash, I think (?)

lib/gui/app/modules/image-writer.js
```diff
@@ -127,9 +156,16 @@ exports.performWrite = (image, drives, onProgress) => {
     })
 
     const flashResults = {}
+    const analyticsData = {
```
lurch: Rather than creating the `analyticsData` object twice, would it make sense for `exports.flash` to pass its `analyticsData` object into  `exports.performWrite` ?

PR #2296, opened 2018-05-01
fix(sdk): load usbboot adapter on start

zvin:
For some reason loading it later didn't work

Change-Type: patch
Changelog-Entry: Load usbboot adapter on start on GNU/Linux if running
as root.

### DIFF
--- a/lib/sdk/adapters/index.js
+++ b/lib/sdk/adapters/index.js
@@ -19,6 +19,8 @@
 const _ = require('lodash')
 const os = require('os')
 
+const permissions = require('../../shared/permissions')
+
 /**
  * @summary The list of loaded adapters
  * @type {Object[]}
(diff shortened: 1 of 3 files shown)


lib/sdk/adapters/index.js
```diff
@@ -28,10 +31,8 @@ const ADAPTERS = [
   require('./blockdevice')
 ]
 
-// We don't support usbboot on GNU/Linux yet, given
-// that some distributions require root permissions
-// to open USB devices.
-if (os.platform() !== 'linux') {
+// On GNU/Linux, we only support usbboot when running as root.
+if ((os.platform() !== 'linux') || (process.geteuid() === permissions.UNIX_SUPERUSER_USER_ID)) {
```
lurch: Isn't this duplicating the functionality already contained inside `permissions.isElevated()` ? Or are you doing it this way simply to avoid using Promises?

zvin: We can't use isElevated here as it returns a promise and we want to set `exports` __now__.

PR #2298, opened 2018-05-02
feat(GUI): add drive quantity to flash analytics

Shou:
We add a field `driveAmount` to the flash analytics events in the image
writer.

Change-Type: patch

### DIFF
--- a/lib/gui/app/modules/image-writer.js
+++ b/lib/gui/app/modules/image-writer.js
@@ -160,6 +160,7 @@ exports.performWrite = (image, drives, onProgress) => {
     const analyticsData = {
       image,
       drives,
+      driveCount: drives.length,
       uuid: flashState.getFlashUuid(),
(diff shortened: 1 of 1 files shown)


lib/gui/app/modules/image-writer.js
```diff
@@ -255,6 +255,7 @@ exports.flash = (image, drives) => {
   const analyticsData = {
     image,
     drives,
+    driveAmount: drives.length,
```
jhermsmeier: How about `driveCount` here? `driveAmount` sounds a bit weird to me :D

Shou: Done

Shou:
Gonna rebase this on top of https://github.com/resin-io/etcher/pull/2301 and https://github.com/resin-io/etcher/pull/2295

Shou:
Here we go

Issue #2241, opened 2018-04-23
Update UI to match 7" screen size


jviotti:
We will use the OEM version of the ODroid HDMI touch screen 7”, which is 800x480.

![screen 02-01](https://user-images.githubusercontent.com/2192773/39320644-ba1e1104-4952-11e8-9818-145b6ec0a892.png)

Issue #2263, opened 2018-04-23
Display an scrollable area with drive selection information


jviotti:
![screen 03-02](https://user-images.githubusercontent.com/2192773/39321217-6432d566-4954-11e8-8dd1-533800e27608.png)

![screen 02-05](https://user-images.githubusercontent.com/2192773/39321221-67631250-4954-11e8-84de-1064bcff2603.png)

PR #2299, opened 2018-05-02
feat(GUI): center content independent to window resolution

Shou:
We pave way for different sizes of Etcher windows by dynamically
centering the content with flexbox.

Connects-To: #2263
Closes: #2241
Change-Type: patch
Changelog-Entry: Increase the vertical window size to 480px.

### DIFF
--- a/lib/gui/app/components/safe-webview.js
+++ b/lib/gui/app/components/safe-webview.js
@@ -65,8 +65,10 @@ const API_VERSION_PARAM = 'api-version'
  * should only be changed when truly necessary as it introduces breaking changes.
  * This version number is exposed to the banner such that it can determine what
  * features are safe to utilize.
+ *
+ * See `git blame -L n` where n is the line below for the history of version changes.
  */
-const API_VERSION = 1
+const API_VERSION = 2
 
 /**
  * @summary Webviews that hide/show depending on the HTTP status returned
--- a/lib/gui/app/pages/finish/styles/_finish.scss
+++ b/lib/gui/app/pages/finish/styles/_finish.scss
@@ -16,6 +16,7 @@
 
 .page-finish {
   margin-top: -25px;
+  flex: 1;
 }
 
 .page-finish .button-label {
--- a/lib/gui/app/pages/main/styles/_main.scss
+++ b/lib/gui/app/pages/main/styles/_main.scss
@@ -14,12 +14,17 @@
  * limitations under the License.
  */
 
-.page-main {
-margin-top: 50px;
-}
-
 svg-icon > img[disabled] {
   opacity: $disabled-opacity;
+}
+
+.page-main {
+  flex: 1;
+  align-self: center;
+}
+
+.page-main > .col-xs {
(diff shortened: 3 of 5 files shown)


Shou:
image
vs
image

jhermsmeier:
@konmouz @Shou @jviotti do we want to hide this behind a Etcher Pro feature flag? I think it looks a bit empty on desktop

lurch:
Should this also change the version-number sent to the online success-banner URL, so that a (different) success banner can be displayed that takes full advantage if the increased space? (while the old smaller-sized success banner still gets displayed by older versions of Etcher)

Shou:
@jhermsmeier the extra space should be used by https://github.com/resin-io/etcher/issues/2263 once that's implemented so maybe this shouldn't be merged until there's progress on that so we can still do releases if we have to.

Shou:
@lurch I wonder if we should increase the API variable given that the interface hasn't changed? Instead we could check the version string in the banner? Or maybe that's a slippery slope into a tangled mess.

lurch:
Probably easier to just check a single version-variable than it is to do `if ((programVersion == xxx) || (programVersion == yyy) || (programVersion == zzz) || ...` ?

jhermsmeier:
> so maybe this shouldn't be merged until there's progress on that so we can still do releases if we have to

@Shou can you add a "Depends on" to the PR description, so we're immediately aware of it (without reading the whole thread) when looking at this?

Shou:
I'll go with the API version increase. Also here's what it looks like right now with the current banner.

image

Gonna push these changes soon, and more screenshots to follow.

jviotti:
@jhermsmeier @Shou We should not be hiding this behind any Etcher Pro specific environment variable. If the new window feels empty, then we can revise the design.

jhermsmeier:
@jviotti yeah, but no need to revise the design with what @Shou mentioned, as https://github.com/resin-io/etcher/issues/2263 will fill the space – I think we should just wait until that's done before we merge this, as @Shou suggested, to avoid awkward looks

Shou:
Yup, I probably have to make some small adjustments after rebasing as well, but it's mostly done.

lib/gui/app/components/safe-webview.js
```diff
@@ -66,7 +66,7 @@ const API_VERSION_PARAM = 'api-version'
  * This version number is exposed to the banner such that it can determine what
  * features are safe to utilize.
  */
-const API_VERSION = 1
+const API_VERSION = 2
```
lurch: Perhaps it'd be worth adding some comments, clarifying what changes each of the different API_VERSIONs signify?

Shou: That's a good idea, but it'd grow with every version change, instead maybe we could just have a comment like below.

```
// For history use: git blame -L n
```

lurch: > but it'd grow with every version change

I guess it depends on how often you expect `API_VERSION` to change...

PR #2300, opened 2018-05-02
minifix(GUI): negate predicate to show help icon

Shou:
Change-Type: patch


Shou:
Accidentally removed the `!` while refactoring

Issue #2245, opened 2018-04-23
Add a cancel button the UI


jviotti:
This will allow the user to cancel all flashes (for now), and potentially particular ones in the future.

jviotti:
This is blocked by https://github.com/resin-io/etcher/issues/2234

jviotti:
We also need some designs from @konmouz

lurch:
Duplicate of #1791 ?

jviotti:
![screen 03-02](https://user-images.githubusercontent.com/2192773/39320474-3c398d04-4952-11e8-91cb-b1fc81eb3dff.png)

jviotti:
We should also add a modal that says "Are you sure you want to cancel?" or something like that

konmouz:
<img width="1210" alt="Screenshot 2019-06-26 at 12 49 22" src="https://user-images.githubusercontent.com/15323961/60177452-d7447f80-9810-11e9-8735-d59d34a6f649.png">

Updated design - https://zpl.io/aRzJezg

Issue #2234, opened 2018-04-23
Allow cancellation of on-going flashes


jviotti:
We can start by supporting cancellation of all drives at the same thing. We might want to consider the possibility of cancelling a single one later on.

Issue #1791, opened 2017-10-18
STOP Button would be very useful!


rsmatias:
It would be great to have a STOP button in the WRITE operation!

Thanks.
RSM

- **Etcher version:**
- **Operating system and architecture:**
- **Image flashed:**
- **Do you see any meaningful error information in the DevTools?**

jhermsmeier:
That's already planned, and will be possible once #1671 has landed

rsmatias:
Nice! Thanks a lot for your quick reply.

Regards,
RSM

lurch:
This request was originally to be able to stop the "write" operation, but we've also had a request from @Gallore to be able to cancel the "verify" operation.

PR #2301, opened 2018-05-02
feat(GUI): add button to cancel flash process

Shou:
We add a cancel button next to the flash progress bar that gracefully
aborts the flash process.

Closes: https://github.com/resin-io/etcher/issues/1791
Closes: https://github.com/resin-io/etcher/issues/2234
Closes: https://github.com/resin-io/etcher/issues/2245
Change-Type: patch
Changelog-Entry: Add a button to cancel the flash process.

### DIFF
--- a/lib/gui/app/modules/image-writer.js
+++ b/lib/gui/app/modules/image-writer.js
@@ -30,6 +30,7 @@ const permissions = require('../../../shared/permissions')
 const windowProgress = require('../os/window-progress')
 const analytics = require('../modules/analytics')
 const packageJSON = require('../../../../package.json')
+const selectionState = require('../../../shared/models/selection-state')
 
 /**
  * @summary Number of threads per CPU to allocate to the UV_THREADPOOL
@@ -137,6 +138,13 @@ exports.performWrite = (image, drives, onProgress) => {
         return errors.fromJSON(data)
       })
       _.merge(flashResults, event)
+    })
+
+    ipc.server.on('abort', () => {
+      terminateServer()
+      resolve({
+        cancelled: true
+      })
     })
 
     ipc.server.on('state', onProgress)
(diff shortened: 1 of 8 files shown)


jhermsmeier:
@Shou can we have 💻 📷 🖼 to go along with this?

Shou:
Here we go!

image

kriekprinsloo:
Wow.  That is a nice button.

lib/gui/app/modules/image-writer.js
```diff
@@ -296,3 +304,29 @@ exports.flash = (image, drives) => {
...
+    validateWriteOnSuccess: settings.get('validateWriteOnSuccess')
+  })
+
+  if (ipc.server) {
+    const [ socket ] = ipc.server.sockets
+    ipc.server.emit(socket, 'cancel')
+  } else {
+    const error = new Error('Can\'t abort, not connected to IPC server')
+    analytics.logException(error)
+  }
```
jhermsmeier: I think this will be better off in a try / catch block, to also catch situations where `socket` would be null, etc... i.e.

lurch:
Does this work for both cancelling the write-phase and also cancelling the verify-phase?

lib/gui/app/modules/image-writer.js
```diff
@@ -296,3 +304,28 @@ exports.flash = (image, drives) => {
...
+/**
+ * @summary Cancel write operation
+ * @function
+ * @public
+ *
+ * @example
+ * imageWriter.cancel()
+ */
+exports.cancel = () => {
+  analytics.logEvent('Cancel', {
```
lurch: If a user is flashing multiple drives, I guess they might cancel after some drives have completed but before all drives have completed - I wonder if that'd be useful analytics-data to log?

lurch:
A thought I've just had: On the "success dots" on the final page, would it be worth displaying a separate counter for "cancelled devices", or will they always just show up as "failed devices"?
Or will cancelling mean that the final page won't ever be reached, even if the user cancels after some of their devices have been fully written? (e.g. if a user is flashing 10 drives, and 9 of them finish after 5 minutes, but the 10th card is much slower and still has an ETA of 20 minutes, the user might cancel the flashing process running on the last drive)

lib/gui/app/pages/main/templates/main.tpl.html
```diff
@@ -106,6 +106,12 @@
             <span ng-bind="flash.getProgressButtonLabel()"></span>
         </progress-button>
 
+        <button class="button button-link button-abort-write"
```
lurch: I wonder if it'd be worth adding a tooltip to this button, to make its purpose more obvious?

jhermsmeier:
> Does this work for both cancelling the write-phase and also cancelling the verify-phase?

Yup :)

> On the "success dots" on the final page, would it be worth displaying a separate counter for "cancelled devices"

Aborting the process will not send you to the success page ;)

lib/gui/app/modules/image-writer.js
Shou: Because of how multi-writes works right now, drives basically finish at the same time. In the future that might change so we'll have to worry about that then though.

lib/gui/app/pages/main/templates/main.tpl.html
jhermsmeier: Yeah, @Shou @konmouz wdyt?

Issue #2262, opened 2018-04-23
Implement autoselection and hide the drive selection step


jviotti:
There will still be a drive step, but instead of allowing a drive selection modal, we will directly show the summary of selected available drives (https://github.com/resin-io/etcher/issues/2263). So basically, if the user puts a drive inside one of the ports, then the drive is selected by default.

jviotti:
This "mode" should be enabled by an environment variable specific for this feature

PR #2306, opened 2018-05-04
feat(GUI): env var toggle autoselecting all valid drives

Shou:
We introduce an environment variable
`ETCHER_DISABLE_EXPLICIT_DRIVE_SELECTION` that both enables
autoselection of drives and disables explicit drive selection by hiding
the buttons allowing this. All valid drives are autoselected, i.e. any
drive-image pair that does not result in an error, however warnings are
accepted.

Closes: https://github.com/resin-io/etcher/issues/2262
Change-Type: patch
Changelog-Entry: Introduce env var to toggle autoselection of all
drives.

### DIFF
--- a/lib/gui/app/pages/main/controllers/drive-selection.js
+++ b/lib/gui/app/pages/main/controllers/drive-selection.js
@@ -18,6 +18,7 @@
 
 const _ = require('lodash')
 const angular = require('angular')
+const prettyBytes = require('pretty-bytes')
 const settings = require('../../../models/settings')
 const selectionState = require('../../../../../shared/models/selection-state')
 const analytics = require('../../../modules/analytics')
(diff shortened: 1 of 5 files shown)


konmouz:
@Shou I have some issues with the vertical spacing. Please make sure to follow the spacing from Zeplin. For instance ‘no targets’ should be vertically aligned with the button box and the grey text below should be placed much lower.

Also, ‘n Devices’ should be white in order to be a bit more clear. Are we listing the drive names?

Shou:
Yeah, we will list the drives shortly. I'm working on that in a separate PR right now. Gonna follow up here with your requested changes in a bit.

lib/gui/app/pages/main/controllers/drive-selection.js
```diff
@@ -42,9 +43,35 @@ module.exports = function (DriveSelectorService) {
...
+   * > '32 GB'
+   */
+  this.getDrivesSubtitle = () => {
+    const drive = selectionState.getCurrentDrive()
+
+    if (drive) {
+      return prettyBytes(drive.size)
+    }
+
+    return 'Please plug at least one target device'
```
jhermsmeier: Missing a word here, I believe: "plug" -> "plug in"

Shou: Yeah, I wonder if 'insert' would be better?

jhermsmeier: Yeah, I think we should go with "insert"

PR #2311, opened 2018-05-08
feat(gui): Add ability to read settings from a config file

jhermsmeier:
This adds the capability to configure settings via a `.etcherrc` file,
either in the user's home directory, or the current working directory.
The defined settings are merged with localStorage settings, and preceding
configuration files.
If both are present, the current working directory takes precedence.

Change-Type: minor
Changelog-Entry: Add support for configuration files
Connects To: #1356

### DIFF
--- a/lib/gui/app/models/local-settings.js
+++ b/lib/gui/app/models/local-settings.js
@@ -17,6 +17,10 @@
 'use strict'
 
 const Bluebird = require('bluebird')
+const _ = require('lodash')
+const fs = require('fs')
+const path = require('path')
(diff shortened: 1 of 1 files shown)


zvin:
It would be nice not to put files directly in the users home directory.
We should use `~/.config/etcherrc` on Linux (XDG) and equivalents on other platforms.

zvin:
Also if it's a json file, we could keep the `.json` extension.

zvin:
There are some linting problems left.

PR #2313, opened 2018-05-08
feat(GUI): add env var to toggle fullscreen

Shou:
We add an environment variable to toggle Etcher in fullscreen.

Closes: https://github.com/resin-io/etcher/issues/2307
Change-Type: patch
Changelog-Entry: Add environment variable to toggle fullscreen.


lib/gui/etcher.js
```diff
@@ -40,10 +40,10 @@ electron.app.on('ready', () => {
     height: 380,
     useContentSize: true,
     show: false,
-    resizable: false,
+    resizable: Boolean(process.env.ETCHER_FULLSCREEN),
```
jhermsmeier: It still shouldn't be resizable if fullscreen I think

Shou: Seems to be required for some reason along with fullscreening. Although weirdly enough you can remove `fullscreenable` and it's still fullscreen.

curcuz:
@Shou why aren't we using the `kiosk` property of `browserWindow` ?https://electronjs.org/docs/api/browser-window -

Shou:
Ah, I wasn't aware what it did, they don't explain it very well in the docs. I wonder if Kiosk mode will prevent the error dialogs from appearing on top? Otherwise it seems perfect for our use-case.

PR #2314, opened 2018-05-08
feat(GUI): use gtk3 dark theme mode

Shou:
We enable the `darkTheme` mode for GTK-3 applications (mainly Linux)
that suits Etcher's dark theme better, making the window title bar dark.

Change-Type: patch
Changelog-Entry: Use GTK-3 darkTheme mode.


Shou:
image

vs

image

PR #2318, opened 2018-05-09
minifix(gui): Only enable Kiosk Mode when FULLSCREEN is set

jhermsmeier:
This fixes the `kiosk` setting always being true, and causing
the operating system's desktop to disappear.

Change-Type: patch


jhermsmeier:
This is what was happening; note the missing bar at the very top

screen shot 2018-05-09 at 17 03 39

Issue #2258, opened 2018-04-23
Make use of Resin's update locks to not update the Etcher Pro when the user is using it


jviotti:
We can only allow updates when the user is not flashing, an in fact didn't even select an image yet.

PR #2326, opened 2018-05-10
feat(gui): Add resin update lock based on inactivity

jhermsmeier:
This adds functionality to acquire & release the update lock
when running under resinOS, re-using the `ELECTRON_RESIN_UPDATE_LOCK`
environment variable from `resin-electronjs`.

Further this adds the `ETCHER_INTERACTION_TIMEOUT_MS` env var, to
facilitate adjusting the inactivity period required to release the lock.

Change-Type: minor
Connects To: https://github.com/resin-io/etcher/issues/2258

### DIFF
--- a/lib/gui/app/app.js
+++ b/lib/gui/app/app.js
@@ -46,6 +46,7 @@ const selectionState = require('../../shared/models/selection-state')
 const driveScanner = require('./modules/drive-scanner')
 const osDialog = require('./os/dialog')
 const exceptionReporter = require('./modules/exception-reporter')
+const updateLock = require('./modules/update-lock')
 
 /* eslint-disable lodash/prefer-lodash-method */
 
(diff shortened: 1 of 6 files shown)


lib/gui/app/modules/update-lock.js
```diff
@@ -0,0 +1,184 @@
...
+    this.lockTimer.signal()
+
+    // When extending, check that we have the lock,
+    // and acquire it, if not
+    if (process.env.ELECTRON_RESIN_UPDATE_LOCK) {
+      UpdateLock.check((checkError, isLocked) => {
+        if (checkError) {
+          analytics.logException(checkError)
+        }
+        if (!isLocked) {
```
zvin: What will be the error of `isLocked` if `checkError` is not `null`?

jhermsmeier: `isLocked` will be undefined in that case. Reasoning behind the mere `!isLocked` is that we'll still want to try to get the lock in that case

lib/gui/app/app.js
```diff
@@ -284,6 +285,23 @@ app.run(($window) => {
...
+   * @private
+   * @example
+   * window.addEventListener('click', extendLock)
+   */
+  const extendLock = () => {
+    updateLock.extend()
+  }
+
+  $window.addEventListener('click', extendLock)
+  $window.addEventListener('touchstart', extendLock)
```
Shou: It probably makes more sense to use `touchend` here given that a user could _in theory_ pretend Etcher Pro is a pet.

PR #2327, opened 2018-05-10
feat(GUI): use new design blue and spacing

Shou:
We use the new design blue, spacing, and order of step buttons.

Change-Type: patch


jhermsmeier:
@Shou what's with the missing "Change" in the first screenshot?

Shou:
@jhermsmeier that's with the autoselection enabled

PR #2331, opened 2018-05-11
feat(gui): Add ability to set analytics tokens via env

jhermsmeier:
This adds the ability to set the Sentry & Mixpanel API tokens
via environment variables.

Change-Type: patch
Connects To: #2329

### DIFF
--- a/lib/gui/app/modules/analytics.js
+++ b/lib/gui/app/modules/analytics.js
@@ -23,8 +23,10 @@ const settings = require('../models/settings')
 
 resinCorvus.install({
   services: {
-    sentry: _.get(packageJSON, [ 'analytics', 'sentry', 'token' ]),
-    mixpanel: _.get(packageJSON, [ 'analytics', 'mixpanel', 'token' ])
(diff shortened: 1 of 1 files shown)


lib/gui/app/modules/analytics.js
```diff
@@ -23,8 +23,10 @@ const settings = require('../models/settings')
 
 resinCorvus.install({
   services: {
-    sentry: _.get(packageJSON, [ 'analytics', 'sentry', 'token' ]),
-    mixpanel: _.get(packageJSON, [ 'analytics', 'mixpanel', 'token' ])
+    sentry: process.env.ANALYTICS_SENTRY_TOKEN ||
+      _.get(packageJSON, [ 'analytics', 'sentry', 'token' ]),
+    mixpanel: process.env.ANALYTICS_MIXPANEL_TOKEN ||
```
jhermsmeier: I think we should prefix these with `ETCHER_`, but that'd deviate from the build process, wdyt @resin-io/etcher?

Issue #2310, opened 2018-05-08
Use new colours


Shou:
As per the latest Zeplin designs below.

![screen 02-04](https://user-images.githubusercontent.com/819413/39770788-eec0c174-52e7-11e8-8ad7-e0bad88ad774.png)

PR #2341, opened 2018-05-15
feat(GUI): use design background and drive size ordering

Shou:
We use the new design background color, and order the drive step size in
accordance with the new design as well.

Related: https://github.com/resin-io/etcher/issues/2310
Change-Type: patch
Changelog-Entry: Use new design background color and drive step size
ordering.


Shou:
New
image

Old
image

Issue #2254, opened 2018-04-23
Disable desktop notifications inside the container


jviotti:
Might need to tweak some GTK themes, and integrate a minimal window manager such as Flexbox.

jhermsmeier:
This changed, as @curcuz said we want to completely disable the desktop notifications instead

PR #2343, opened 2018-05-15
feat(gui): Add desktop notification setting

jhermsmeier:
This adds a setting to disable desktop notifications, to be controlled
via configuration file

Change-Type: feat
Connects To: #2254


lurch:
Seems a bit strange that this behaviour is controlled from a settings option, whereas all the other Etcher-pro related features are controlled from env-vars?

Issue #2264, opened 2018-04-23
Add support for an environment variable to blacklist drives through their unique port IDs


jviotti:
Blocked by https://github.com/resin-io/etcher/issues/2235.

The UI should check if a drive matches what's on this comma-separated environment variable and not show such drive on the drive selection dialog, nor allow auto selection to it.

jviotti:
Right. I meant port IDs. I'll clarify :D

Shou:
https://github.com/resin-io-modules/drivelist/pull/278

for reference

PR #2345, opened 2018-05-16
feat: blacklist devices by device path

Shou:
We use `devicePath` instead of `device` to blacklist drives using the
`ETCHER_BLACKLISTED_DRIVES` environment variable.

Closes: https://github.com/resin-io/etcher/issues/2264
Change-Type: patch

### DIFF
--- a/lib/gui/app/app.js
+++ b/lib/gui/app/app.js
@@ -224,7 +224,7 @@ app.run(($timeout) => {
     $timeout(() => {
       if (BLACKLISTED_DRIVES.length) {
         const allowedDrives = drives.filter((drive) => {
-          return !BLACKLISTED_DRIVES.includes(drive.device)
+          return !BLACKLISTED_DRIVES.includes(drive.devicePath)
         })
         availableDrives.setDrives(allowedDrives)
       } else {


lib/gui/app/app.js
```diff
@@ -224,7 +224,7 @@ app.run(($timeout) => {
     $timeout(() => {
       if (BLACKLISTED_DRIVES.length) {
         const allowedDrives = drives.filter((drive) => {
-          return !BLACKLISTED_DRIVES.includes(drive.device)
+          return !BLACKLISTED_DRIVES.includes(drive.devicePath)
```
jhermsmeier: I think it'd be worth allowing to blacklist based on both, the `.device` and the `.devicePath`, i.e.
```js
return !(
  BLACKLISTED_DRIVES.includes(drive.devicePath) ||
  BLACKLISTED_DRIVES.includes(drive.device)
)
```

lurch: ...or perhaps there could be separate `BLACKLISTED_DRIVES` and `BLACKLISTED_DEVICES` env-vars?

PR #2347, opened 2018-05-16
feat(GUI): hide unsafe mode option with env var

Shou:
We hide the unsafe mode option toggle with an env var
`ETCHER_HIDE_UNSAFE_MODE` that also enables unsafe mode.

Closes: https://github.com/resin-io/etcher/issues/2243
Change-Type: patch
Changelog-Entry: Hide unsafe mode option toggle with an env var.

### DIFF
--- a/lib/gui/app/modules/drive-scanner.js
+++ b/lib/gui/app/modules/drive-scanner.js
@@ -23,7 +23,7 @@ const permissions = require('../../../shared/permissions')
 const scanner = SDK.createScanner({
   blockdevice: {
     get includeSystemDrives () {
-      return settings.get('unsafeMode')
+      return settings.get('unsafeMode') && !process.env.ETCHER_HIDE_UNSAFE_MODE
     }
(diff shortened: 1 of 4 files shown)


lib/gui/app/modules/drive-scanner.js
```diff
@@ -23,7 +23,7 @@ const permissions = require('../../../shared/permissions')
 const scanner = SDK.createScanner({
   blockdevice: {
     get includeSystemDrives () {
-      return settings.get('unsafeMode')
+      return settings.get('unsafeMode') || process.env.ETCHER_HIDE_UNSAFE_MODE
```
jhermsmeier: This would turn unsafeMode **on** when this env var is set to any value – iirc we want it to be off? @curcuz?

lurch:
Not that it really matters, but I guess the commit message should have been updated to `... that also disables unsafe mode.` :wink:

PR #2333, opened 2018-05-11
feat(GUI): add electron-native file-picker component

Shou:
We add a file-picker written with Rendition/React. It is activated with
the `ETCHER_EXPERIMENTAL_FILE_PICKER` environment variable. Further
customisation can be done with the `ETCHER_FILE_BROWSER_CONSTRAIN_FOLDER`
variable that takes a path and allows one to constrain the file-picker to
a folder.

Related: https://github.com/resin-io/etcher/issues/2238
Related: https://github.com/resin-io/etcher/issues/2285
Change-Type: patch
Changelog-Entry: Add electron-native file-picker component.

PR #2349, opened 2018-05-16
fix: add missing files module

Shou:
We add a convenience module for file and path operations.
Tests included.

Change-Type: patch

### DIFF
--- a/lib/gui/app/components/file-selector/file-selector/file-selector.jsx
+++ b/lib/gui/app/components/file-selector/file-selector/file-selector.jsx
@@ -18,6 +18,7 @@
 
 const _ = require('lodash')
 const os = require('os')
+const fs = require('fs')
 const path = require('path')
 const React = require('react')
 const propTypes = require('prop-types')
@@ -224,7 +225,7 @@ class RecentFilesUnstyled extends React.PureComponent {
   render () {
     const existing = (fileObjs) => {
       return _.filter(fileObjs, (fileObj) => {
-        return files.exists(fileObj.fullpath)
+        return fs.existsSync(fileObj.fullpath)
       })
     }
 
@@ -460,7 +461,7 @@ class FileSelector extends React.PureComponent {
   }
 
   setFilesProgressively (dirname) {
-    return files.getDirectory(dirname).then((basenames) => {
+    return fs.readdirAsync(dirname).then((basenames) => {
       const fileObjs = basenames.map((basename) => {
         return {
           dirname: this.state.path,
@@ -496,6 +497,8 @@ class FileSelector extends React.PureComponent {
 
     this.setFilesProgressively(file.fullpath).then(() => {
       this.setState({ path: file.fullpath })
+    }).catch((error) => {
+      this.setState({ error: error.message })
     })
   }
 
--- a/lib/shared/files.js
+++ b/lib/shared/files.js
@@ -1,0 +1,188 @@
+/*
+ * Copyright 2018 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const _ = require('lodash')
+const path = require('path')
+const Bluebird = require('bluebird')
+const fs = Bluebird.promisifyAll(require('fs'))
+
+/* eslint-disable lodash/prefer-lodash-method */
+
+/**
+ * @summary Get file metadata
+ * @function
+ * @private
+ *
+ * @param {String} dirname - directory name
+ * @param {String} [basename] - custom basename to append
+ * @returns {Object} file metadata
+ *
+ * @example
+ * try {
+ *   const file = files.getFileMetadataSync('/home/user')
+ *   console.log(`Is ${file.basename} a directory? ${file.isDirectory}`)
+ * } catch (error) {
+ *   console.error(error)
+ * }
+ */
+exports.getFileMetadataSync = (dirname, basename = '') => {
+  // TODO(Shou): use path.parse object information here
+  const fullpath = path.join(dirname, basename)
+  const pathObj = path.parse(fullpath)
+
+  // TODO(Shou): this is not true for Windows, figure out Windows hidden files
+  const isHidden = pathObj.base.startsWith('.')
+  const stats = fs.statSync(fullpath)
+
+  return {
+    basename: pathObj.base,
+    dirname: pathObj.dir,
+    fullpath,
+    extension: pathObj.ext.replace('.', ''),
+    name: pathObj.name,
+    isDirectory: stats.isDirectory(),
+    isHidden,
+    size: stats.size
+  }
+}
+
+/**
+ * @summary Get file metadata asynchronously
+ * @function
+ * @private
+ *
+ * @param {String} fullpath - full path
+ * @returns {Promise<Object>} promise of file metadata
+ *
+ * @example
+ * files.getFileMetadataAsync('/home/user').then((file) => {
+ *   console.log(`Is ${file.basename} a directory? ${file.isDirectory}`)
+ * })
+ */
+exports.getFileMetadataAsync = (fullpath) => {
+  const pathObj = path.parse(fullpath)
+
+  // NOTE(Shou): this is not true for Windows
+  const isHidden = pathObj.base.startsWith('.')
+
+  return fs.statAsync(fullpath).then((stats) => {
+    return {
+      basename: pathObj.base,
+      dirname: pathObj.dir,
+      fullpath,
+      extension: pathObj.ext.replace('.', ''),
+      name: pathObj.name,
+      isDirectory: stats.isDirectory(),
+      isHidden,
+      size: stats.size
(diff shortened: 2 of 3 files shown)


jhermsmeier:
Ref #2333

lib/shared/files.js
```diff
@@ -0,0 +1,167 @@
...
+}
+
+exports.getFileMetadataAsync = (dirname, basename = '') => {
+  // TODO(Shou): use path.parse object information here
+  const fullpath = path.join(dirname, basename)
+  const pathObj = path.parse(fullpath)
+
+  // NOTE(Shou): this is not true for Windows
+  const isHidden = pathObj.base.startsWith('.')
+  return fs.lstatAsync(fullpath).then((stats) => {
```
jhermsmeier: Same here

lib/shared/files.js
```diff
@@ -0,0 +1,167 @@
...
+ * console.log(`Is ${file.basename} a directory? ${file.isDirectory}`)
+ */
+exports.getFileMetadataSync = (dirname, basename = '') => {
+  // TODO(Shou): use path.parse object information here
+  const fullpath = path.join(dirname, basename)
+  const pathObj = path.parse(fullpath)
+
+  // TODO(Shou): this is not true for Windows, figure out Windows hidden files
+  const isHidden = pathObj.base.startsWith('.')
+  const stats = fs.lstatSync(fullpath)
```
jhermsmeier: Why is this stat()-ing the symlink rather than the file it links to?

lib/shared/files.js
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * @param {String} fullpath - full path
+ * @returns {Boolean}
+ *
+ * @example
+ * console.log('I have pictures: ' + files.exists('~/Pictures'))
+ */
+exports.exists = (fullpath) => {
+  try {
+    return Boolean(fs.statSync(fullpath))
```
lurch: Why not use https://nodejs.org/api/fs.html#fs_fs_existssync_path ?

lib/shared/files.js
```diff
@@ -0,0 +1,167 @@
...
+ * @function
+ * @private
+ *
+ * @param {String} fullpath - path string
+ * @returns {Array<Object>} - all subpaths as file objects
+ *
+ * @example
+ * const subpaths = files.subpaths('/home/user/Downloads')
+ * console.log(subpaths.map(file => file.fullpath))
+ * > [ '/', '/home', '/home/user', '/home/user/Downloads' ]
```
lurch: Perhaps it'd also be useful to give an example of how it splits a Windows-style path into "subpaths" ?

lib/shared/files.js
jhermsmeier: Because "Note that fs.exists() is deprecated" ;)

lib/shared/files.js
```diff
@@ -0,0 +1,167 @@
...
+ */
+exports.subpaths = (fullpath) => {
+  if (!_.isString(fullpath)) {
+    return null
+  }
+
+  const dirs = _.compact(_.update(fullpath.split(path.sep), [ '0' ], (root) => {
```
jhermsmeier: This method should be checking with `path.isAbsolute()`, and split with `path.basename()`and `path.dirname()` – as splitting on `path.sep` can break in horrible ways with escaped chars, and on Windows, both `/` and `\` are legit path separators

lib/shared/files.js
lurch: ..." but fs.existsSync() is not" ;)

jhermsmeier: and I can't read, apparently :D

lib/shared/files.js
Shou: Absolutely, for now this is an experimental Linux-only file-picker but we'll definitely give it a proper Windows treatment as it develops.

lib/shared/files.js
Shou: Couldn't we just `path.resolve` instead of checking with `path.isAbsolute`?

jhermsmeier: Sure, if you always want an absolute path

Shou:
Thanks guys, I've updated it now.

tests/shared/files.spec.js
```diff
@@ -0,0 +1,56 @@
...
+    it('should handle a root directory', function () {
+      const { root } = path.parse(__dirname)
+      const dirs = files.splitPath(root)
+      m.chai.expect(dirs).to.deep.equal([ root ])
+    })
+
+    it('should handle relative paths', function () {
+      let dir
+      if (process.platform === 'win32') {
+        dir = 'C:relative\\dir\\test'
```
jhermsmeier: What about the more common case of the unprefixed relative path on Windows?

lib/shared/files.js
```diff
@@ -0,0 +1,248 @@
...
+ * @returns {Boolean}
+ *
+ * @example
+ * console.log('I have pictures: ' + files.exists('~/Pictures'))
+ */
+exports.exists = (fullpath) => {
+  try {
+    return fs.existsSync(fullpath)
+  } catch (err) {
+    return err.code !== 'ENOENT'
```
lurch: Now that you've switched to using `existsSync`, presumably this error-handling is redundant?

lib/shared/files.js
```diff
@@ -0,0 +1,248 @@
...
+ *
+ * @param {String} dirname - directory path
+ * @returns {Array<String>}
+ *
+ * @example
+ * files.getDirectoryFilenames(os.homedir())
+ * > [ 'Downloads', 'Pictures', 'Videos', 'Documents' ]
+ */
+exports.getDirectoryFilenames = (dirname) => {
+  return fs.readdirAsync(dirname)
```
jhermsmeier: Why is this being wrapped?

lib/shared/files.js
```diff
@@ -0,0 +1,248 @@
...
+ * @summary Get file metadata
+ * @function
+ * @private
+ *
+ * @param {String} dirname - directory name
+ * @param {String} [basename] - custom basename to append
+ * @returns {Object} file metadata
+ *
+ * @example
+ * const file = getFileMetadata('/home/user')
```
lurch: Should this example be updated to use `getFileMetadataSync` ?

lib/shared/files.js
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * @summary Get file metadata asynchronously
+ * @function
+ * @private
+ *
+ * @param {String} fullpath - full path
+ * @returns {Object} file metadata
+ *
+ * @example
+ * const file = getFileMetadata('/home/user')
```
lurch: Should this example be updated to use `getFileMetadataAsync` ?

lib/shared/files.js
```diff
@@ -0,0 +1,248 @@
...
+  }
+}
+
+/**
+ * @summary Get file metadata asynchronously
+ * @function
+ * @private
+ *
+ * @param {String} fullpath - full path
+ * @returns {Object} file metadata
```
jhermsmeier: This returns a Promise, which fulfils with an Object

lib/shared/files.js
jhermsmeier: What's the error handling case here? try / catch? Should that be illustrated in the example?

lib/shared/files.js
```diff
@@ -0,0 +1,248 @@
...
+ *     return `Directory: ${file.fullpath}`
+ *   }
+ *
+ *   return `File: ${file.basename} in ${dirname}`
+ * })
+ */
+exports.getDirectoryContents = (dirname) => {
+  return fs.readdirAsync(dirname).then((contents) => {
+    return Bluebird.map(contents, (basename) => {
+      return exports.getFileMetadata(dirname, basename)
```
lurch: Should this be either `getFileMetadataSync` or `getFileMetadataAsync` ?

tests/shared/files.spec.js
lurch: Is the `C:` supposed to be there at the start?

Shou: Yeah, apparently it's an esoteric way of specifying relative paths in Windows.

lib/gui/app/components/file-selector/file-selector/file-selector.jsx
```diff
@@ -224,7 +225,7 @@ class RecentFilesUnstyled extends React.PureComponent {
   render () {
     const existing = (fileObjs) => {
       return _.filter(fileObjs, (fileObj) => {
-        return files.exists(fileObj.fullpath)
+        return fs.existsSync(fileObj.fullpath)
```
jhermsmeier: What happens if this throws?

lib/shared/files.js
```diff
@@ -0,0 +1,188 @@
...
+  const isHidden = pathObj.base.startsWith('.')
+  const stats = fs.statSync(fullpath)
+
+  return {
+    basename: pathObj.base,
+    dirname: pathObj.dir,
+    fullpath,
+    extension: pathObj.ext.replace('.', ''),
+    name: pathObj.name,
+    isDirectory: stats.isDirectory(),
```
jhermsmeier: Maybe worth having the `stats` object on the metadata in here? Would allow for easier future extensions; i.e. `metadata.stats.isDirectory()` (same for the async method obv.)

Shou: I would do that but at least `isDirectory` needs to be a value because of filters so either way we need to do some custom handling of it.

lib/gui/app/components/file-selector/file-selector/file-selector.jsx
jhermsmeier: Omg, right 🤦‍♂️

lurch:
Just out of curiosity, how come the `files` module being missing wasn't picked up by #2157 ?

jhermsmeier:
Good question. Reason is that it's required from JSX, and that wasn't being linted – see https://github.com/resin-io/etcher/pull/2358

PR #2351, opened 2018-05-17
fix(diskpart): Fix diskpart argv when tmpdir contains spaces

jhermsmeier:
This escapes the diskpart script filename when shelling out,
to avoid failure when the username and thus the `os.tmpdir()` path
contains spaces.

Change-Type: patch
Connects To: https://github.com/resin-io/etcher/issues/1991

### DIFF
--- a/lib/cli/diskpart.js
+++ b/lib/cli/diskpart.js
@@ -69,7 +69,7 @@ const runDiskpart = (commands, callback) => {
   }, (writeError) => {
     debug('write %s:', filename, writeError || 'OK')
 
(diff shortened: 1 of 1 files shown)


lurch:
Wasn't there some kind of issue with `execFile` and elevated commands? I can't remember, but perhaps @jviotti will remember more of the details... (or perhaps I'm just getting confused with something totally different!)

jhermsmeier:
@lurch not that I remember, and it works just fine as far as I can see

Issue #2312, opened 2018-05-08
Loading screen when Etcher Pro is updating


Shou:
Has to be a static page that fits the Etcher Pro screen (1024x600px).

Design is a pre-requisite.

lurch:
1024x600 ?! Wasn't Etcher only recently tweaked to be 800x480? Does that mean #2299 is out of date already? :wink:

konmouz:
cc @Shou 

![screen 04-01 copy](https://user-images.githubusercontent.com/15323961/40113112-3815fa88-5900-11e8-9cc8-b4b9cdfb6194.jpg)

jhermsmeier:
As this has to be a static *image*, rather than a page, I don't think we'll be able to display which version we're updating for now, as we'd have to render the image prior to updating for that.

Shou:
There's an updated version on Zeplin now. Looks good :)

curcuz:
@konmouz we need to get rid of the spinner - this is a static image

konmouz:
@curcuz, we removed the spinner in the new version.

<img width="809" alt="screen shot 2018-05-23 at 15 05 49" src="https://user-images.githubusercontent.com/15323961/40429517-cdeb05ea-5e9a-11e8-9158-eb16bbf6bb25.png">

konmouz:
Update screen design - @konmouz 
Implement screen - @thundron

konmouz:
<img width="1213" alt="Screenshot 2019-06-24 at 16 20 46" src="https://user-images.githubusercontent.com/15323961/60031153-18675300-969c-11e9-915b-78a092b9f2a3.png">

https://zpl.io/VDKA3vw -- @thundron please update screen

PR #2352, opened 2018-05-17
feat(GUI): display a loading page when updating

Shou:
We add an updating overlay that should only be shown during updates for
the Etcher Pro. This can be set using the `updatingState` field in the
Redux store, and setting it to `true`, and `false` when finished. We may
extend this to become an object containing more metadata in the future.

Closes: #2312
Change-Type: patch


Shou:
image

curcuz:
@Shou this needs to be a PNG image file that is pushed directly to the framebuffer via `fbi`

Shou:
In that case this should do, right? I just exported the mock-up and added square edges.

untitled

jhermsmeier:
Closing, as incorrect solution

PR #2356, opened 2018-05-18
chore(webpack): Reduce build log verbosity

jhermsmeier:
Travis CI is complaining about 10k+ lines of log output again.
This was initially set to `verbose` to facilitate debugging on concourse, which has since concluded.
But as it's generating 2000+ lines of output, it's masking CI build results

Change-Type: patch


lurch:
Relates to #2148

Shou:
@jhermsmeier I think this made the logs longer

PR #2362, opened 2018-05-22
fix(GUI): file-picker performance and design improvements

Shou:
- Replace onClick arrow functions in all components that use them for
  efficiency reasons: 300-500% speed-up
- Sort by folders and ignore case for better UX
- Remove use of `rendition.Button` in files, leading to a 10-20%
  performance increase when browsing files
- Proper sidebar width and spacing
- Recents and favorites are now filtered by existence async for a tiny
  performance improvement
- Make Breadcrumbs and Icon pure components to stop frequent re-rendering
- Initial support for array constraints
- Use first constraint as initial path instead of homedir if a
  constraint is set
- Use correct design height on modal, `calc(100vh - 20px)`
- Reset scroll position when browsing a new folder
- Fuse Bluebird `.map()` and `.reduce()` in
`files.getAllFilesMetadataAsync`.
- Use `localeCompare`'s own case-insensitive option instead of calling
`.toLowerCase()` twice on `n-2` files compared.
- Use 16px font sizes in sidebar and files to match design.
- Disable `$locationProvider.html5Mode.rewriteLinks`, which seemed to
  take 50ms of the directory changing time.
- Leave file extension as-is in `files.getFileMetadataSync` and the
  async counterpart for a very minor performance improvement.

Change-Type: patch

### DIFF
--- a/lib/gui/app/components/file-selector/controllers/file-selector.js
+++ b/lib/gui/app/components/file-selector/controllers/file-selector.js
@@ -16,7 +16,10 @@
 
 'use strict'
 
+const os = require('os')
 const settings = require('../../../models/settings')
+
+/* eslint-disable lodash/prefer-lodash-method */
 
 module.exports = function (
   $uibModalInstance
@@ -43,9 +46,25 @@ module.exports = function (
    * @example
    * FileSelectorController.getFolderConstraint()
    */
-  this.getFolderConstraint = () => {
+  this.getFolderConstraints = () => {
     // TODO(Shou): get this dynamically from the mountpoint of a specific port in Etcher Pro
-    // TODO: Make this handle multiple constraints
-    return settings.get('fileBrowserConstraintPath')
+    return settings.has('fileBrowserConstraintPath')
+      ? settings.get('fileBrowserConstraintPath').split(',')
+      : []
+  }
+
+  /**
+   * @summary Get initial path
+   * @function
+   * @public
+   *
+   * @returns {String} - path
+   *
+   * @example
+   * <file-selector path="FileSelectorController.getPath()"></file-selector>
+   */
+  this.getPath = () => {
+    const [ constraint ] = this.getFolderConstraints()
+    return constraint || os.homedir()
   }
 }
--- a/lib/gui/app/app.js
+++ b/lib/gui/app/app.js
@@ -330,6 +330,13 @@ app.config(($provide) => {
   })
 })
 
+app.config(($locationProvider) => {
+  // NOTE(Shou): this seems to invoke a minor perf decrease when set to true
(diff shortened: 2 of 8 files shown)


lib/gui/app/components/file-selector/controllers/file-selector.js
```diff
@@ -41,8 +45,27 @@ module.exports = function (
    * @example
    * FileSelectorController.getFolderConstraint()
    */
-  this.getFolderConstraint = () => {
+  this.getFolderConstraints = utils.memoize(() => {
     // TODO(Shou): get this dynamically from the mountpoint of a specific port in Etcher Pro
+    // FIXME(Shou): splitting on ',' is unsafe
```
lurch: On Unix, the `$PATH` variable is separated by `:` characters, on Windows the `%PATH%` variable is separated by `;` characters. Might make sense to re-use those semantics in Etcher?

Shou: We're moving to a JSON config file so this is only a temporary 'hack' given that all `,:;` can appear in a filename leading to some jagged character escaping

Shou:
image

lurch:
Should dot-folders be hidden by default?

Shou:
@lurch we intend to use rendition filters to implement folder hiding (and searching) but that will be in a separate PR.

jhermsmeier:
@Shou looks like there are some unintentional design changes?

Before, on **master**:

screen shot 2018-05-22 at 15 50 50

This branch:

screen shot 2018-05-22 at 18 49 49

jhermsmeier:
@Shou okay, why is the character spacing off though?

lurch:
There's a scrollbar on @Shou 's screenshot, but no scrollbar on @jhermsmeier 's screenshot?

Shou:
@jhermsmeier the font size wasn't 16px, I've changed that now, and in the files – screenshot below.

@lurch good point, I should make the scrollbar custom so that we don't have to see the OS-native scrollbars which could potentially be ugly.

image

lurch:
What does the file-size shown underneath the folders mean? I assume that for performance-reasons it probably _isn't_ the size of the folder's contents? Or does Jonas _really_ only have 192 bytes in his `Movies` folder?

Shou:
@lurch I assume it depends on what filesystem you use, because I'm used to seeing 4kB with EXT4.

jhermsmeier:
@lurch the scrollbars on Mac OS only permanently show if you have a non-touchy mouse attached, otherwise they fade out if unused :) And the folder sizes are fs dir entry sizes – they shouldn't be displayed for folders

Shou:
Done

image

PR #2367, opened 2018-05-24
minifix(GUI): move success banner back down

Shou:
Change-Type: patch

### DIFF
--- a/lib/gui/app/scss/main.scss
+++ b/lib/gui/app/scss/main.scss
@@ -151,7 +151,7 @@ body,
     bottom: 0;
     left: 0;
     width: 100%;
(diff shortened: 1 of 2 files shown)


lurch:
> We need to take care of the colour change in etcher-homepage to match.

Perhaps this is a good place to use the webview-api-version thing, as I guess there'll be lots of people still running the 'old' v1.4.4 as well as people running the 'new' version (whenever that gets released); and I guess you'd want the success-banner to display with the 'correct' background colours in both versions of Etcher?

Shou:
@lurch absolutely, it's good the version was bumped!

Issue #2342, opened 2018-05-15
Storage.modify shouldn't save undefined values


Shou:
Neither fallback nor values given that are undefined should be saved to the JSON string.

Issue #2265, opened 2018-04-23
Make the Resin container capable of loading an .env file with settings


jviotti:
We will use these variables to configure what LEDs correspond to what drives, etc.

jhermsmeier:
We should be able to re-use https://github.com/resin-io/etcher/pull/2311 for this

lurch:
> We should be able to re-use #2311 for this

I guess that presumes that everything that's configurable through an env-var, is also configurable through the localSettings? Currently IIRC there are several places in the code which **only** look at `process.env` .

lurch:
Has this been closed by #2368 and #2369 ?

jhermsmeier:
Not quite yet, some changes on the Etcher Pro side still tbd

PR #2368, opened 2018-05-25
feat(gui): Persist local settings to config file

jhermsmeier:
This switches the settings to use a local configuration file, instead of `localStorage` to facilitate loading and manipulating the settings outside of a `window` context. Further, this decouples the settings from the applications state tracking redux store, and relocates models from `lib/shared/models` to `lib/gui/app/models`, as they weren't used in SDK code.

Previously added functionality to also respect a configuration file in the current working directory has been removed to simplify the mechanism and deconflict precedence.

Change-Type: minor
Connects To: https://github.com/resin-io/etcher/issues/2265


lib/gui/app/models/local-settings.js
```diff
@@ -38,10 +48,19 @@ const settingsStorage = new Storage(LOCAL_STORAGE_SETTINGS_KEY)
  */
 const RCFILE = '.etcher.json'
 
+/**
+ * @summary Configuration file path
+ * @type {String}
+ * @constant
+ */
+const HOME_CONFIG_PATH = process.platform === 'win32'
+  ? path.join(USER_DATA_DIR, RCFILE)
```
lurch: Is this special-casing of the filename still relevant on win32 now, given that `USER_DATA_DIR` will always end with `etcher` ?

lurch:
Does this also fix #2342 ?

> and relocates models from lib/shared/models to lib/gui/app/models

Oooh, interesting - were they not used by any of the Etcher CLI code?

lib/gui/app/models/local-settings.js
```diff
@@ -68,6 +87,35 @@ const readConfigFile = (filename) => {
   })
 }
 
+/**
+ * @summary Read a local .etcherrc file
```
lurch: **Write** a local .etcherrc file? :wink:

lib/gui/app/modules/image-writer.js
```diff
@@ -313,6 +313,7 @@ exports.flash = (image, drives) => {
   analytics.logEvent('Flash', analyticsData)
 
   return exports.performWrite(image, drives, (state) => {
+    console.log('performWriate.onProgress', state)
```
zvin: performWrite

lib/gui/app/models/settings.js
```diff
@@ -156,9 +157,16 @@ exports.set = (key, value) => {
     }))
   }
 
+  const previousValue = settings[key]
+
   settings[key] = value
 
   return localSettings.writeAll(settings)
+    .catch((error) => {
```
zvin: NIT: `writeAll` returns a Bluebird promise so you can use `.tapCatch`

jhermsmeier: Aye, added!

lib/gui/app/models/local-settings.js
jhermsmeier: Yup, mended!

lib/gui/app/models/local-settings.js
jhermsmeier: Ah, good catch – this is obsolete now, as is `RCFILE`

lib/gui/app/models/local-settings.js
```diff
@@ -16,32 +16,42 @@
...
+
+/**
+ * @summary Configuration file path
+ * @type {String}
+ * @constant
+ */
+const CONFIG_PATH = path.join(USER_DATA_DIR, 'config.json')
 
 /**
  * @summary Read a local .etcherrc file
```
lurch: Perhaps this should be changed to `Read the local configuration file` to match the jsdoc on line 82?

lib/gui/app/models/settings.js
jhermsmeier: Had to revert this, as this method isn't a thing with the version of Bluebird we're using atm yet

lib/gui/app/models/local-settings.js
jhermsmeier: Done

lib/gui/app/models/store.js
```diff
@@ -550,3 +508,34 @@ module.exports = _.merge(redux.createStore(storeReducer, DEFAULT_STATE), {
...
+    const nextState = module.exports.getState()
+    if (!_.isEqual(nextState, currentState)) {
+      currentState = nextState
+      onChange(currentState)
+    }
+  }
+
+  const unsubscribe = module.exports.subscribe(changeHandler)
+  changeHandler()
+  return unsubscribe
```
Shou: Do we need the `changeHandler()` between `const unsubscribe` and `return unsubscribe` or could we consolidate the const-setting with the return statement? Either works for me I'm just curious if I've missed something subtle.

PR #2369, opened 2018-05-25
feat: Use settings for feature control

jhermsmeier:
This changes the feature switches to be queried through the settings module

**NOTE:** This PR targets #2368 as base, as it depends on it

Change-Type: minor
Connects To: #2368, https://github.com/resin-io/etcher/issues/2265

### DIFF
--- a/lib/gui/app/models/local-settings.js
+++ b/lib/gui/app/models/local-settings.js
@@ -88,8 +88,8 @@ const readConfigFile = (filename) => {
       } else {
         try {
           data = JSON.parse(contents)
-        } catch(e) {
-          // ignore (?)
+        } catch (parseError) {
+          // Ignore (?)
         }
         resolve(data)
       }
--- a/lib/gui/app/models/local-settings.js
+++ b/lib/gui/app/models/local-settings.js
@@ -36,10 +36,14 @@ const JSON_INDENT = 2
  * - `$XDG_CONFIG_HOME/etcher` or `~/.config/etcher` on Linux
  * - `~/Library/Application Support/etcher` on macOS
  * See https://electronjs.org/docs/api/app#appgetpathname
+ * NOTE: The ternary is due to this module being loaded both,
+ * in Electron's main and renderer processes
  * @constant
  * @type {String}
  */
-const USER_DATA_DIR = electron.remote.app.getPath('userData')
+const USER_DATA_DIR = electron.app
+  ? electron.app.getPath('userData')
+  : electron.remote.app.getPath('userData')
 
 /**
  * @summary Local settings filename
@@ -73,15 +77,21 @@ const HOME_CONFIG_PATH = process.platform === 'win32'
  */
 const readConfigFile = (filename) => {
   return new Bluebird((resolve, reject) => {
-    fs.readFile(filename, (error, buffer) => {
+    fs.readFile(filename, { encoding: 'utf8' }, (error, contents) => {
+      let data = {}
       if (error) {
         if (error.code === 'ENOENT') {
-          resolve({})
+          resolve(data)
         } else {
           reject(error)
         }
       } else {
-        resolve(JSON.parse(buffer.toString()))
+        try {
+          data = JSON.parse(contents)
+        } catch(e) {
+          // ignore (?)
+        }
+        resolve(data)
       }
     })
   })
--- a/lib/gui/app/modules/drive-scanner.js
+++ b/lib/gui/app/modules/drive-scanner.js
@@ -23,7 +23,7 @@ const permissions = require('../../../shared/permissions')
 const scanner = SDK.createScanner({
   blockdevice: {
     get includeSystemDrives () {
-      return settings.get('unsafeMode') && !process.env.ETCHER_HIDE_UNSAFE_MODE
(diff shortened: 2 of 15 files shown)


lib/gui/etcher.js
```diff
@@ -20,31 +20,29 @@ const electron = require('electron')
...
+ * electron.app.on('ready', createMainWindow)
+ */
+const createMainWindow = () => {
   mainWindow = new electron.BrowserWindow({
     width: 800,
     height: 380,
     useContentSize: true,
     show: false,
-    resizable: Boolean(process.env.ETCHER_FULLSCREEN),
+    resizable: Boolean(config.fullscreen),
```
lurch: Is this right? AFAICT the fact that you've got `const config = settings.getDefaults()` earlier on means that `config` will always be set to the values at https://github.com/resin-io/etcher/blob/d37ea80eff76fe2fdda767bdd4ed5c41f291e4e7/lib/gui/app/models/settings.js#L36 and thus `config.fullscreen` will _always_ be undefined, rather than the value that you've set in the actual config file?

lurch: Oh, I've just seen the `settings.load()` part down at the bottom of the file :laughing:

lib/gui/app/pages/main/controllers/image-selection.js
```diff
@@ -156,7 +157,7 @@ module.exports = function (
   this.openImageSelector = () => {
     analytics.logEvent('Open image selector')
 
-    if (process.env.ETCHER_EXPERIMENTAL_FILE_PICKER) {
+    if (settings.get('experimentalFilePicker')) {
```
lurch: I wonder if this would be a good time to rename this setting something other than `experimentalFilePicker`? i.e. before it starts getting written into actual config files? (after which time it might become harder to rename it?)
IYSWIM.

lib/gui/app/models/local-settings.js
```diff
@@ -73,15 +77,21 @@ const HOME_CONFIG_PATH = process.platform === 'win32'
...
+          resolve(data)
         } else {
           reject(error)
         }
       } else {
-        resolve(JSON.parse(buffer.toString()))
+        try {
+          data = JSON.parse(contents)
+        } catch (parseError) {
+          // Ignore (?)
```
Shou: Maybe we should `logDebug` or even `logException` this?

lib/gui/app/modules/drive-scanner.js
```diff
@@ -23,7 +23,7 @@ const permissions = require('../../../shared/permissions')
 const scanner = SDK.createScanner({
   blockdevice: {
     get includeSystemDrives () {
-      return settings.get('unsafeMode') && !process.env.ETCHER_HIDE_UNSAFE_MODE
+      return settings.get('unsafeMode') && !settings.get('disableUnsafeMode')
```
Shou: I think `hideUnsafeMode` is a better description given that it's disabled by default, and we're actually hiding the ability to toggle unsafe mode while still keeping it disabled

lib/gui/etcher.js
```diff
@@ -81,4 +82,32 @@ electron.app.on('ready', () => {
...
+electron.app.on('before-quit', () => {
+  process.exit(EXIT_CODES.SUCCESS)
 })
+
+settings.load().then((localSettings) => {
+  Object.assign(config, localSettings)
+}).catch((error) => {
+  // TODO: What do if loading the config fails?
+  console.error('Error loading settings:')
+  console.error(error)
```
Shou: We should probably `logDebug` and/or `logException` this as well, although the `console.error`s take care of the former.

lib/gui/app/modules/drive-scanner.js
lurch: IMHO if this was indeed a `hideUnsafeMode` setting, then that'd imply that this `includeSystemDrives` function would simply be `return settings.get('unsafeMode')` (i.e. unsafeMode could remain enabled, even if the option to toggle it was hidden).
However as that's not the case, I think `disableUnsafeMode` is perhaps the best name? :man_shrugging:

Shou: That's true, but our intent with this option (and these options generally) is to create an environment rather than just an option you toggle every now and then as a user, as such it's safe to assume it will remain disabled.

lib/gui/etcher.js
jhermsmeier: @Shou I don't think we can use analytics here, as this is the main entry point (and the analytics not set up yet), as well as there being no point to opt out of them before this runs

lib/gui/app/models/local-settings.js
jhermsmeier: @Shou actually same thing as I mentioned below here, as this is also loaded from the main entry point

lib/gui/etcher.js
Shou: Fair, let's just log them then.

lib/gui/app/modules/drive-scanner.js
jhermsmeier: Yeah, I think sticking with "disable" is more appropriate still

PR #2383, opened 2018-06-08
fix(webpack): Exclude all node externals / node_modules

jhermsmeier:
This adds `webpack-node-externals` to exclude node_modules,
immensely reducing bundle size and avoiding complex exclusion rules
for the etcher-sdk

Change-Type: patch


lurch:
Does that mean that any of the current "complex exclusion rules" can be removed? Or are they orthogonal to all this?

PR #2165, opened 2018-04-02
upgrade(package): Update Electron to v1.8.4

jhermsmeier:
This updates Electron to v1.8.4 in order to support arm64 builds,
and mitigate 2 vulnerabilities: [CVE-2018-1000136] and [CVE-2018-1000118]

[CVE-2018-1000136]: https://nvd.nist.gov/vuln/detail/CVE-2018-1000136
[CVE-2018-1000118]: https://nvd.nist.gov/vuln/detail/CVE-2018-1000118

Depends On:
- [ ] resin-io-modules/winusb-driver-generator#26

Change-Type: patch
Changelog-Entry: Update Electron to v1.8.4

PR #2384, opened 2018-06-08
Update Electron to v2

jhermsmeier:
**TODO**

- [x] resolve Webview related error
- [x] resolve "electron: error while loading shared libraries: libgtk-3.so.0: cannot open shared object file: No such file or directory" on linux / up2
- [x] https://github.com/resin-io/etcher/pull/2412
- [x] resolve "/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.17' not found (required by /etcher/node_modules/electron/dist/electron)"
- [ ] Update [resin-ci/pipelines/electron/docker](https://github.com/resin-io/resin-concourse/tree/master/pipelines/electron/docker) according to Dockerfile changes

Change-Type: patch
Connects To: #2165

### DIFF
--- a/scripts/build/docker/compile-template.js
+++ b/scripts/build/docker/compile-template.js
@@ -31,11 +31,11 @@ const template = fs.readFileSync(path.join(currentDirectory, 'Dockerfile.templat
 _.each([
   {
     architecture: 'i686',
-    image: 'erwinchang/ubuntu-12.04-32bit-build'
+    image: 'erwinchang/ubuntu-14.04-32bit-build'
   },
   {
     architecture: 'x86_64',
-    image: 'ubuntu:12.04'
+    image: 'ubuntu:14.04'
   },
   {
     architecture: 'armv7hf',
--- a/scripts/build/docker/compile-template.js
+++ b/scripts/build/docker/compile-template.js
@@ -31,7 +31,7 @@ const template = fs.readFileSync(path.join(currentDirectory, 'Dockerfile.templat
 _.each([
   {
     architecture: 'i686',
-    image: 'erwinchang/ubuntu-14.04-32bit'
+    image: 'i386/ubuntu:14.04'
   },
   {
     architecture: 'x86_64',
--- a/scripts/build/docker/compile-template.js
+++ b/scripts/build/docker/compile-template.js
@@ -31,7 +31,7 @@ const template = fs.readFileSync(path.join(currentDirectory, 'Dockerfile.templat
 _.each([
   {
     architecture: 'i686',
-    image: 'erwinchang/ubuntu-14.04-32bit-build'
+    image: 'erwinchang/ubuntu-14.04-32bit'
   },
   {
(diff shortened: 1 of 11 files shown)


lurch:
Wow, big version-jump :arrow_up:  :slightly_smiling_face:  Good luck with all the testing! :crossed_fingers:

scripts/build/docker/compile-template.js
```diff
@@ -31,11 +31,11 @@ const template = fs.readFileSync(path.join(currentDirectory, 'Dockerfile.templat
 _.each([
   {
     architecture: 'i686',
-    image: 'erwinchang/ubuntu-12.04-32bit-build'
+    image: 'erwinchang/ubuntu-14.04-32bit-build'
   },
   {
     architecture: 'x86_64',
-    image: 'ubuntu:12.04'
+    image: 'ubuntu:14.04'
```
lurch: As you're probably aware, the reason that Etcher (for Linux) was being built on Ubuntu 12.04 was because that's the version that https://electronjs.org/docs/tutorial/support claims Electron (for x86 and amd64 Linux versions) was built on, and (at the time) Etcher was aiming for compatibilty with as many different Linux distros as possible (IIRC this effectively came down to the [glibc](https://github.com/resin-io/etcher/search?q=glibc&type=Issues) version used by that old Ubuntu 12.04 version).
But given that Ubutunu 12.04 is now EOL with no security updates, it probably makes sense to drop support for it (just like you did for WindowsXP). Ubuntu 14.04 LTS is supported until April 2019 https://www.ubuntu.com/info/release-end-of-life

IIRC some of the custom repos and packages that we had to add to these Docker scripts were purely because we were using such an old version of Ubuntu, so it might be worth revisiting those and seeing which ones can be removed now you're buidling on an Ubuntu 14.04 base?

lurch:
`upgrade(package): Update Electron to 2.0.4`

Might be worth keeping the PR's title in sync too? :wink:

scripts/build/docker/compile-template.js
```diff
@@ -31,7 +31,7 @@ const template = fs.readFileSync(path.join(currentDirectory, 'Dockerfile.templat
 _.each([
   {
     architecture: 'i686',
-    image: 'erwinchang/ubuntu-14.04-32bit-build'
+    image: 'erwinchang/ubuntu-14.04-32bit'
```
lurch: Looks like you could use the official `i386/ubuntu:14.04` image which is probaly kept more up-to-date than the erwinchang image?
https://hub.docker.com/_/ubuntu/
https://hub.docker.com/r/i386/ubuntu/

Jacalz:
I would recommend you to update to version 3 instead of version 2.

Jacalz:
Discussion should probably continue [here](https://github.com/balena-io/etcher/pull/2578) 😉

lurch:
I guess this PR should probably be closed now? :wink:

Issue #2348, opened 2018-05-16
Handle incorrect files in new file-picker gracefully


Shou:
Right now the UX is not ideal when selecting incorrect files in the new file-picker, this should at least match the old UX.

Issue #1823, opened 2017-10-31
Implement a new file / source picker


jhermsmeier:
Instead of a file picker, have a modal with 3 options:
  - type url
  - duplicate whole source
  - select file from source

If "select file" is preferred, show a very simple HTML file picker.
Also investigate the need for an on-screen keyboard for typing URLs (add to resin-electronjs?)

Shou:
Instead of a modal should we make it a page and move towards the V2 design? This also comes with the added benefit of allowing me to write it in React so we don't get more Angular into the codebase.

@konmouz could we get a mockup of the image selection screen with URL inputs and drive selection? We could potentially reuse the drive modal here if needed and just have a "select drive" button, if you feel like that's better. Also based on the V2 design I assume we want the recent selections in there too.

konmouz:
@Shou I'll take care of that within this or next week.

Shou:
Awesome.

lurch:
Sounds like this'd be a great place for closer interaction with https://github.com/resin-io-modules/resin-components ?

Shou:
@konmouz what's the status on this now?

konmouz:
@Shou I can push this and deliver today

jhermsmeier:
Duplicate of #950

curcuz:
ping @konmouz - any news?

konmouz:
@curcuz I need to redesign this and add some features but should be ready soon

PR #2388, opened 2018-06-11
refactor(gui): Refactor file picker fs I/O

jhermsmeier:
This refactors the experimental file picker to avoid fs i/o in as many places as possible to improve performance. Further, rendering performance is improved by avoiding unnecessary element state changes invalidating components. Also, recent files & favorites have been temporarily disabled due to lack of need for Etcher Pro.

screen shot 2018-06-11 at 19 49 23

screen shot 2018-06-11 at 19 50 07

Errors / warnings from within the file picker are native OS dialogs, as this is a React component and doesn't have access to the Angular modal controllers / components;

screen shot 2018-06-11 at 19 50 15

screen shot 2018-06-11 at 19 59 06

Change-Type: patch
Connects To: https://github.com/resin-io/etcher/issues/2348, https://github.com/resin-io/etcher/issues/1823

### DIFF
--- a/lib/gui/app/components/file-selector/file-selector/file-list.jsx
+++ b/lib/gui/app/components/file-selector/file-selector/file-list.jsx
@@ -1,0 +1,303 @@
+/*
+ * Copyright 2018 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const React = require('react')
+const propTypes = require('prop-types')
+const styled = require('styled-components').default
+const rendition = require('rendition')
+const colors = require('./colors')
+
+const prettyBytes = require('pretty-bytes')
+const files = require('../../../models/files')
+const middleEllipsis = require('../../../utils/middle-ellipsis')
+const supportedFormats = require('../../../../../shared/supported-formats')
+
+const debug = require('debug')('etcher:gui:file-selector')
+
+/**
+ * @summary Character limit of a filename before a middle-ellipsis is added
+ * @constant
+ * @private
+ */
+const FILENAME_CHAR_LIMIT = 20
+
+/**
+ * @summary Pattern to match all supported formats for highlighting
+ * @constant
+ * @private
+ */
+const SUPPORTED_FORMATS_PATTERN = new RegExp(`^\\.(${supportedFormats.getAllExtensions().join('|')})$`, 'i')
+
+/**
+ * @summary Flex styled component
+ * @function
+ * @type {ReactElement}
+ */
+const Flex = styled.div`
+  display: flex;
+  flex: ${ props => props.flex };
+  flex-direction: ${ props => props.direction };
+  justify-content: ${ props => props.justifyContent };
+  align-items: ${ props => props.alignItems };
+  flex-wrap: ${ props => props.wrap };
+  flex-grow: ${ props => props.grow };
+`
+
+/**
+ * @summary Anchor flex styled component
+ * @function
+ * @type {ReactElement}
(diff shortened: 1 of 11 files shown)


lib/gui/app/components/file-selector/file-selector/file-list.jsx
```diff
@@ -0,0 +1,318 @@
...
+const supportedFormats = require('../../../../../shared/supported-formats')
+
+const debug = require('debug')('etcher:gui:file-selector')
+
+/**
+ * @summary Character limit of a filename before a middle-ellipsis is added
+ * @constant
+ * @private
+ */
+const FILENAME_CHAR_LIMIT = 20
```
zvin: This is not used anywhere.

lib/gui/app/components/file-selector/file-selector/file-list.jsx
```diff
@@ -0,0 +1,318 @@
...
+
+  componentDidMount () {
+    process.nextTick(() => {
+      this.readdir(this.state.path)
+    })
+  }
+
+  onHighlight (file) {
+    debug('FileList:onHighlight', file)
+    // this.setState({ highlighted: file.path })
```
zvin: Shouold this be there?

lib/gui/app/components/file-selector/file-selector/file-list.jsx
```diff
@@ -0,0 +1,318 @@
...
+    this.props.onHighlight(file)
+  }
+
+  onSelect (file) {
+    debug('FileList:onSelect', file.path, file.isDirectory)
+    this.props.onSelect(file)
+  }
+
+  componentDidUpdate () {
+    console.timeEnd('FileList:componentDidUpdate')
```
zvin: Shouldn't we use `debug` for that?

lib/gui/app/models/files.js
```diff
@@ -16,133 +16,92 @@
...
+    this.dirname = components.dir
+    this.basename = components.base
+    this.name = components.name
+    this.ext = components.ext
+    this.isHidden = components.name.startsWith('.')
+    this.isFile = stats.isFile()
+    this.isDirectory = stats.isDirectory()
+    this.isBlockDevice = stats.isBlockDevice()
+    this.isCharacterDevice = stats.isCharacterDevice()
+    this.stats = stats
```
zvin: the only thing left in `stats` that we use is `size`. Why not copy it like the rest and remove `FileEntry.stats`?

lurch:
Looking at those screenshots - how does this file-picker work if you have two different images which get shortened to the same ellipses-filename? (which I guess is quite likely if you e.g. have different versions of Raspbian in the same folder?)

jhermsmeier:
@lurch yeah, it doesn't work very well in that situation, which is something we've already noticed; will need some more design / ux work – but then, this is experimental stuff, so it'll be rough round the edges atm

package.json
```diff
@@ -40,8 +40,7 @@
     "fsevents"
   ],
   "dependencies": {
-    "@fortawesome/fontawesome": "1.1.5",
-    "@fortawesome/fontawesome-free-solid": "5.0.9",
+    "@fortawesome/fontawesome-free-webfonts": "^1.0.9",
```
curcuz: @jhermsmeier does this mean we depend on internet connection now?

jhermsmeier: Nope, the font itself is in the npm package

lurch:
> it doesn't work very well in that situation, which is something we've already noticed

I guess one option would be to display the full filename on mouse-hover? Although I guess that doesn't help if you're using a touch screen ;)

lib/gui/app/components/file-selector/file-selector/file-list.jsx
jhermsmeier: `debug` doesn't support timing like this that mark timings in the devtools' perf timeline

lib/gui/app/models/files.js
jhermsmeier: Yeah true, done

lib/gui/app/components/file-selector/file-selector/file-list.jsx
jhermsmeier: Nope, removed

lib/gui/app/components/file-selector/file-selector/file-list.jsx
jhermsmeier: Good catch, fixed!

jhermsmeier:
@konmouz wdyt about this updated fade color:

screen shot 2018-06-12 at 18 51 41

PR #2400, opened 2018-06-20
feat(gui): Add full filename to file selector

jhermsmeier:
This adds the display of the full filename to the file selector.

screen shot 2018-06-20 at 18 11 23

screen shot 2018-06-20 at 18 29 23

Change-Type: minor


lurch:
In the bottom example, where the filename is still "too long", does mouse-hovering over the filename at the bottom show the full filename as a tooltip?

Issue #2385, opened 2018-06-11
Fix settings module not being available when packaged


jhermsmeier:

jhermsmeier:
Fixed by https://github.com/resin-io/etcher/pull/2402

PR #2402, opened 2018-06-20
fix(app): Fix settings being unavailable when packaged

jhermsmeier:
This fixes an issue where the settings model would be missing
from Etcher when packaged, as it's used in two different contexts;
namely the webpack bundle and the main process.

Change-Type: patch
Connects To: #2385

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -6,7 +6,6 @@ nodeGypRebuild: false
 publish: null
 files:
   - lib
-  - "!lib/gui/app"
   - lib/gui/app/index.html
   - generated
   - build/**/*.node


electron-builder.yml
```diff
@@ -6,7 +6,6 @@ nodeGypRebuild: false
 publish: null
 files:
   - lib
-  - "!lib/gui/app"
```
lurch: Sorry if this is a stupid question, but if it was only the settings that were affected, should this be modified to _just_ pull in the settings-related files, rather than the whole of `lib/gui/app`? :man_shrugging:

jhermsmeier:
No, not at all - and yes, that might have been the better (or more correct) way to go; but the settings module depends on quite a few other things which depend on others, etc... so this is the safest way atm. Definitely some improvement required here.

Issue #2377, opened 2018-05-30
Constrain file selection to physical slot


jhermsmeier:

jhermsmeier:
Closed by https://github.com/resin-io/etcher/pull/2405

PR #2405, opened 2018-06-22
feat(gui): Enable device specific constraints for file selection

jhermsmeier:
This adds the ability to restrict the file selection to a given device,
only making its mountpoints accessible.

screen shot 2018-06-22 at 11 21 57

screen shot 2018-06-22 at 11 22 01

Change-Type: patch
Connects To: https://github.com/resin-io/etcher/issues/2377

### DIFF
--- a/lib/gui/app/components/file-selector/controllers/file-selector.js
+++ b/lib/gui/app/components/file-selector/controllers/file-selector.js
@@ -48,11 +48,10 @@ module.exports = function (
    * @example
    * FileSelectorController.getFolderConstraint()
    */
-  this.getFolderConstraints = utils.memoize(() => {
-    // TODO(Shou): get this dynamically from the mountpoint of a specific port in Etcher Pro
+  this.getFolderConstraint = utils.memoize(() => {
     return settings.has('fileBrowserConstraintPath')
-      ? settings.get('fileBrowserConstraintPath').split(',')
-      : []
+      ? settings.get('fileBrowserConstraintPath')
+      : ''
   }, angular.equals)
 
   /**
@@ -66,7 +65,6 @@ module.exports = function (
    * <file-selector path="FileSelectorController.getPath()"></file-selector>
    */
   this.getPath = () => {
-    const [ constraint ] = this.getFolderConstraints()
-    return constraint || os.homedir()
+    return this.getFolderConstraint() ? '/' : os.homedir()
   }
 }
--- a/lib/gui/app/models/files.js
+++ b/lib/gui/app/models/files.js
@@ -18,6 +18,7 @@
 
 const fs = require('fs')
 const path = require('path')
+const drivelist = require('drivelist')
 
 /* eslint-disable lodash/prefer-lodash-method */
 /* eslint-disable no-undefined */
@@ -70,7 +71,6 @@ class FileEntry {
     this.isFile = stats.isFile()
     this.isDirectory = stats.isDirectory()
     this.size = stats.size
-    this.stats = stats
   }
 }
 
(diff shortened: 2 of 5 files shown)


lib/gui/app/components/file-selector/controllers/file-selector.js
```diff
@@ -50,9 +50,10 @@ module.exports = function (
    */
   this.getFolderConstraints = utils.memoize(() => {
```
lurch: This used to return an array, but now returns a string, so perhaps it should be renamed `getFolderConstraint` ?
And should the `TODO` line now be removed?

lurch:
How does it look if the "constrained device" has multiple mountpoints? In the example shown, is the user allowed to navigate back to "Root" (in case they want to switch to a different mountpoint) but prevented from navigating to "Root / Volumes" ?

I wonder if it might be worth retaining the existing "folder constraint" setting (perhaps some people might find it useful?) and add a new "device constraint" setting? :man_shrugging:

lib/gui/app/models/files.js
```diff
@@ -138,3 +138,30 @@ exports.splitPath = (fullpath, subpaths = []) => {
...
+/**
+ * @summary Get contraint path device
+ * @param {String} pathname - device path
+ * @param {Function} callback - callback(error, constraintDevice)
+ * @example
+ * files.getConstraintDevice('/dev/disk2', (error, device) => {
+ *   // ...
+ * })
+ */
+exports.getConstraintDevice = (pathname, callback) => {
```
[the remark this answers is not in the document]
jhermsmeier: We're using the physical device path, which does not change

jhermsmeier:
> How does it look if the "constrained device" has multiple mountpoints? In the example shown, is the user allowed to navigate back to "Root" (in case they want to switch to a different mountpoint) but prevented from navigating to "Root / Volumes" ?

Yes, multiple mountpoints are handled, and navigating to subpaths in between will kick you back to Root, where all mountpoints are enumerated.

> I wonder if it might be worth retaining the existing "folder constraint" setting (perhaps some people might find it useful?) and add a new "device constraint" setting?

This feature is nothing that's going to have use in desktop Etcher really, so I don't see why – did you have something specific in mind?

lib/gui/app/components/file-selector/controllers/file-selector.js
jhermsmeier: Yup, done!

lurch:
> did you have something specific in mind?

The only thing I can think of off the top of my head is if a user has all their disk images in a particular folder, and they might like Etcher to always default to looking in this specifc folder, rather than always starting in their home folder? But perhaps I'm getting confused, and that could be implemented in some other way and doesn't actually have anything to do with this PR... :wink: :man_shrugging: 

Or perhaps in a "locked-down environment" (I'm clutching at straws here :laughing: ) the IT admins might only want users to be able to choose image files from a particular read-only directory, rather than allowing users to flash any arbitrary image? :man_shrugging: (probably a pretty unlikely scernario, given that Etcher requires admin-rights anyway)

lib/gui/app/models/files.js
```diff
@@ -138,3 +138,30 @@ exports.splitPath = (fullpath, subpaths = []) => {
 
   return exports.splitPath(dir, [ base ].concat(subpaths))
 }
+
+/**
+ * @summary Get contraint path device
```
lurch: contraint -> constraint

jhermsmeier: Fixed

PR #2412, opened 2018-07-04
upgrade(package): Update winusb-driver-generator

jhermsmeier:
This updates `winusb-driver-generator` to the latest version,
which supports building under VS 2015 and running under Electron 2.0+

Change-Type: patch


lurch:
Did you manage to test if it also works out-of-the-box on e.g. Windows 7 with no "extra" CRT libs installed?

jhermsmeier:
@lurch not yet – but doing that today

PR #2420, opened 2018-07-20
fix(GUI): Additional data on events

thundron:
Application session UUID for global app events
Flashing workflow UUID for every flashing session
Refine events for when flashing ends with success/failure count on each flash instance

Change-type: patch

### DIFF
--- a/lib/gui/app/models/store.js
+++ b/lib/gui/app/models/store.js
@@ -80,6 +80,8 @@ const selectImageNoNilFields = [
  * @private
  */
 const DEFAULT_STATE = Immutable.fromJS({
+  applicationSessionUuid: '',
+  flashingWorkflowUuid: '',
   availableDrives: [],
   selection: {
     devices: new Immutable.OrderedSet()
@@ -111,7 +113,9 @@ const ACTIONS = _.fromPairs(_.map([
   'SELECT_DRIVE',
   'SELECT_IMAGE',
   'DESELECT_DRIVE',
-  'DESELECT_IMAGE'
+  'DESELECT_IMAGE',
+  'SET_APPLICATION_SESSION_UUID',
+  'SET_FLASHING_WORKFLOW_UUID'
 ], (message) => {
   return [ message, message ]
 }))
@@ -497,6 +501,14 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
 
     case ACTIONS.DESELECT_IMAGE: {
       return state.deleteIn([ 'selection', 'image' ])
+    }
+
+    case ACTIONS.SET_APPLICATION_SESSION_UUID: {
+      return state.set('applicationSessionUuid', action.data)
+    }
+
+    case ACTIONS.SET_FLASHING_WORKFLOW_UUID: {
+      return state.set('flashingWorkflowUuid', action.data)
     }
 
     default: {
--- a/lib/gui/app/modules/image-writer.js
+++ b/lib/gui/app/modules/image-writer.js
@@ -23,6 +23,7 @@ const os = require('os')
 const ipc = require('node-ipc')
 const isRunningInAsar = require('electron-is-running-in-asar')
 const electron = require('electron')
+const store = require('../models/store')
 const settings = require('../models/settings')
 const flashState = require('../models/flash-state')
 const errors = require('../../../shared/errors')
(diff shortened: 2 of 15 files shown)


jhermsmeier:
Commit messages usually start lower-case, and `add` is not a scope we have defined; see the [commit guidelines](https://github.com/resin-io/etcher/blob/master/docs/COMMIT-GUIDELINES.md) (these types / scopes are used for automatic changelog generation)

jhermsmeier:
Also, I think the change-type should be "minor", if not "patch", as this doesn't introduce breaking changes in UI / UX for the users, or much else

thundron:
Actually, I received some errors when starting them with lower-case... Will reword in a moment, anyways I was following this guide: https://github.com/resin-io/process/blob/master/process/Commit_and_PR_Guidelines.md

lib/gui/app/pages/finish/controllers/finish.js
```diff
@@ -57,10 +59,20 @@ module.exports = function ($state) {
       selectionState.deselectImage()
     }
     selectionState.deselectAllDrives()
-    analytics.logEvent('Restart', options)
+    analytics.logEvent('Restart', _.assign({
```
jhermsmeier: Can we use `Object.assign()` here instead?

lib/gui/app/pages/main/controllers/image-selection.js
```diff
@@ -76,18 +77,30 @@ module.exports = function (
...
       return
     }
 
     Bluebird.try(() => {
       let message = null
 
       if (supportedFormats.looksLikeWindowsImage(image.path)) {
-        analytics.logEvent('Possibly Windows image', image)
+        analytics.logEvent('Possibly Windows image', {
+          image,
```
jhermsmeier: This will need to be merged; i.e. `Object.assign({ applicationSessionUuid, flashingWorkflowUuid }, image)` as this would otherwise change the data structure received by mixpanel

lib/gui/app/modules/image-writer.js
```diff
@@ -327,7 +346,14 @@ exports.flash = (image, drives) => {
...
     // eslint-disable-next-line no-magic-numbers
     if (drives.length > 1) {
-      analytics.logEvent('Write failed', analyticsData)
+      const { results } = flashState.getFlashResults()
+      const event = _.assign({
+        errors: results.errors,
+        devices: results.devices
+      },
+      analyticsData,
+      { status: 'failed' })
```
jhermsmeier: Why explicitly add `{ status: 'failed' }`? Is this something that's left over from development?

lib/gui/app/modules/image-writer.js
```diff
@@ -306,19 +314,30 @@ exports.flash = (image, drives) => {
...
   return exports.performWrite(image, drives, (state) => {
     flashState.setProgressState(state)
   }).then(flashState.unsetFlashingFlag).then(() => {
+    const { results } = flashState.getFlashResults()
+    const event = _.assign({
+      errors: results.errors,
+      devices: results.devices
+    },
+    analyticsData,
+    { status: 'finished' })
```
jhermsmeier: Why explicitly add `{ status: 'finished' }`? Is this something that's left over from development?

lib/gui/app/modules/image-writer.js
```diff
@@ -74,20 +75,26 @@ const getApplicationEntryPoint = () => {
  * handleErrorLogging({ code: 'EUNPLUGGED' }, { image: 'resin.img' })
  */
 const handleErrorLogging = (error, analyticsData) => {
+  const evt = _.assign({
```
jhermsmeier: Might `eventData` be a more descriptive name for this?

lib/gui/app/models/store.js
```diff
@@ -499,6 +503,16 @@ const storeReducer = (state = DEFAULT_STATE, action) => {
       return state.deleteIn([ 'selection', 'image' ])
     }
 
+    case ACTIONS.SET_APPLICATION_SESSION_UUID: {
+      return state
+        .set('applicationSessionUuid', action.data)
```
jhermsmeier: Why is there a new line here?

thundron: There are newlines everywhere before each new `case`, so I figured I'd keep them in the added cases too

lib/gui/app/modules/image-writer.js
thundron: It's to make filtering easier on Mixpanel

lib/gui/app/modules/image-writer.js
thundron: It's to make filtering easier on Mixpanel

lib/gui/app/pages/finish/controllers/finish.js
thundron: The linter prefers `_.assign` over `Object.assign`... :/

lib/gui/app/pages/main/controllers/image-selection.js
thundron: Yeah true, I didn't think about that. The image has a very different structure, but I'll just merge it with new data for now, we'll be able to change it later.

lib/gui/app/models/store.js
jhermsmeier: I mean

```js
return state
        .set('applicationSessionUuid', action.data)
```

There's no chaining going on here, so no point in sticking an EOL in front of a method call ;)

lib/gui/app/modules/image-writer.js
jhermsmeier: Can we put that on the first argument, which is the object literal, to avoid creating more objects than needed?

lib/gui/app/pages/finish/controllers/finish.js
jhermsmeier: Ugh. I thought we had already changed that setting...

jhermsmeier: I guess we'll keep that for now then; and have another 👀 at the linter settings.

lib/gui/app/pages/main/controllers/image-selection.js
jhermsmeier: How's this `image`'s structure different from the one before?

thundron: @jhermsmeier Sorry, I meant it's different from a classic event, so there's gonna be properties of the image on the same level as the various tracking UUIDs. It's not a problem, just something I thought could be improved.

PR #2430, opened 2018-08-01
fix(writer): Set EVALIDATION on blockmap checksum errors

jhermsmeier:
This patches validation errors coming form the `blockmap` module to have
the `EVALIDATION` error code, in order to trigger the appropriate code
paths and analytics.

Change-type: patch

### DIFF
--- a/lib/sdk/writer/index.js
+++ b/lib/sdk/writer/index.js
@@ -515,6 +515,12 @@ class ImageWriter extends EventEmitter {
       const pipeline = this._createVerifyPipeline(destination)
 
       pipeline.on('error', (error) => {
+        // NOTE: As the `blockmap` module doesn't set our custom error codes,
+        // we need to patch `EVALIDATION` into a range checksum error here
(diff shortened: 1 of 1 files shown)


lib/sdk/writer/index.js
```diff
@@ -515,6 +515,12 @@ class ImageWriter extends EventEmitter {
       const pipeline = this._createVerifyPipeline(destination)
 
       pipeline.on('error', (error) => {
+        // NOTE: As the `blockmap` module doesn't set our custom error codes,
+        // we need to patch `EVALIDATION` into a range checksum error here
+        if (error.message && error.message.startsWith('Invalid checksum for range')) {
+          error.code = 'EVALIDATION'
+          this.emit('fail', { device: destination.device.device, error })
```
zvin: Is it intended to emit both `error` and `fail` events ?

jhermsmeier: That's something we've discussed last evening briefly; it shouldn't cause any trouble, but we need to test this to be sure.

PR #2433, opened 2018-08-02
fix(cli): Don't use electron to get USER_DATA_DIR in CLI

jhermsmeier:
A previous change erroneously relied solely on `electron` to determine
the user's data directory, which of course, isn't available when running
in the CLI. So this adds a fallback to the same mechanic for CLI should
this code not be loaded in Electron.

Change-type: patch


thundron:
retest

mikesimos:
@resin-ci retest

PR #2439, opened 2018-08-06
fix(test): Move available-drives, flash-state and selection-state tests to gui tests

zvin:
These files only test gui code and make `make test-cli` fail.

Change-type: patch
Signed-off-by: Alexis Svinartchouk <alexis@resin.io>


thundron:
@zvin `Error: The commit title should not exceed 72 characters`
Maybe change to something like `Move tests to correct scope` and then specify them in the message body

PR #2448, opened 2018-08-14
fix(GUI): Remove results data from cancel event

thundron:
Change-type: patch
Signed-off-by: Lorenzo Alberto Maria Ambrosi <lorenzoa@resin.io>


thundron:
@resin-ci retest

PR #2484, opened 2018-09-20
Drive selector button

amdomanska:
Convert angular component to React using Rendition library

### DIFF
--- a/lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
+++ b/lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
@@ -16,46 +16,51 @@
 
 'use strict'
 
+const _ = require('lodash')
 const React = require('react')
-const ReactDOM = require('react-dom')
 const propTypes = require('prop-types')
-const Color = require('color')
+
+/* eslint-disable no-magic-numbers */
 
 const middleEllipsis = require('./../../utils/middle-ellipsis')
 
 const { Provider, Txt } = require('rendition')
-const { StepButton, StepNameButton, StepSelection,
-  DetailsText, ChangeButton } = require('./../../styled-components')
+const {
+  StepButton,
+  StepNameButton,
+  StepSelection,
+  DetailsText,
+  ChangeButton
+} = require('./../../styled-components')
 
 class DriveSelectorButton extends React.PureComponent {
-
-  allDevicesFooter() {
-    return this.props.selectedDevices.map((device) =>
-      <Txt key={device.device} tooltip={device.description + '(' + device.displayName + ')'}>
+  allDevicesFooter () {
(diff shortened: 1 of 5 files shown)


amdomanska:
@resinci retest

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -0,0 +1,118 @@
...
+  reselectDrive: propTypes.func,
+  driveSize: propTypes.string,
+  hasCompatibilityStatus: propTypes.func,
+  getCompatibilityStatuses: propTypes.func,
+  drives: propTypes.func,
+  image: propTypes.func,
+  selectedDevices: propTypes.array,
+}
+
+module.exports = DriveSelectorButton
```
LucianBuzzo: Can you use a named export here?

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -0,0 +1,118 @@
...
+      return this.props.selectedDevices.map((device) =>
+        <Txt key={device.device} tooltip={device.description + '(' + device.displayName + ')'}>
+          { middleEllipsis(device.description, 14) }
+        </Txt>
+       )
+    }
+  }
+
+  render() {
+    if (this.props.hasDrive || !this.props.shouldShowDrivesButton) {
```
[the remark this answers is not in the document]
amdomanska: It's the property set in settings, it is decided outside component.

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
amdomanska: I cannot as long as we need to use React2Angular component - I haven't found the way for it to accept the named export. Once the whole main page will be converted to React this is going to be changed.

LucianBuzzo: @amdomanska How are you importing this module? modules and `require` statements should have nothing to do with `react2angular`.

thundron:
@resin-ci retest

thundron:
@resin-ci retest

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
thundron: Tests fail for some reason when we use named exports, I still need to take a closer look at this

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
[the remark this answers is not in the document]
thundron: Confirmed, it's in the angular side for now. When we transition all of this logic to React, we'll be able to change names too

Issue #2483, opened 2018-09-20
Create Devices Details Modal


amdomanska:
https://www.flowdock.com/app/rulemotion/resin-frontend/threads/wEBKD3m_S0YR2gZDxXNBntrJlkg

jviotti:
@amdomanska Let's not add Flowdock links in issues, as the open source community will not be able to see them

PR #2486, opened 2018-09-21
feat(GUI): Add Drives Details modal

amdomanska:
Add Drivers Details modal window

Type-change: minor
Fixes: #2483
Signed-off-by: amdomanska <aga@resin.io>

drivers details modal


lib/gui/app/components/details-modal/details-modal.jsx
```diff
@@ -0,0 +1,121 @@
...
+  cursor: pointer;
+  &:hover {
+    color: ${colors.dark.background};
+  }
+`
+
+class DetailsModal extends React.Component {
+
+  renderDetails() {
+    let lines = []
```
LucianBuzzo: NIT: This would be better as `return this.props.details.map(....)`

lib/gui/app/components/details-modal/details-modal.jsx
```diff
@@ -0,0 +1,121 @@
...
+                  plaintext
+                  onClick={this.props.callback}
+                  align='left'
+                >
+                &times;
+                </CloseButton>
+              </ModalHeader>
+            </React.Fragment>
+          }
+          primaryButtonProps={{
```
LucianBuzzo: Is this necessary since you have a `<CloseButton>` element in the modal header?

lib/gui/app/components/details-modal/details-modal.jsx
```diff
@@ -0,0 +1,121 @@
...
+const ModalHeader = styled(Flex) `
+  text-align: left;
+  align-items: baseline;
+  font-size: 12px;
+  color: ${colors.light.soft.foreground};
+  padding: 11px 20px;
+  border-bottom: 1.5px solid ${colors.light.soft.background};
+`
+
+const ModalTitle = styled(Txt) `
```
LucianBuzzo: You could use a `Box` element with the flex attribute applied here instead of creating a new styled-component

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -27,9 +27,19 @@ const { Provider, Txt } = require('rendition')
 const { StepButton, StepNameButton, StepSelection,
   DetailsText, ChangeButton } = require('./../../styled-components')
 
+const DetailsModal = require('./../details-modal/details-modal')
+
 class DriveSelectorButton extends React.PureComponent {
 
-  allDevices() {
+  constructor(props) {
+  super(props)
```
LucianBuzzo: NIT: I think this block needs to be indented one level

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -27,9 +27,19 @@ const { Provider, Txt } = require('rendition')
...
+
+  this.state = {
+    show: false
+  }
+}
+
+  allDevicesFooter() {
     let devices = []
```
LucianBuzzo: This would be better written as a `.map`

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -44,6 +54,14 @@ class DriveSelectorButton extends React.PureComponent {
     return devices
   }
 
+  selectedDevicesDetails() {
+    let details = []
```
LucianBuzzo: NIT: this would be better written as a `.map`

lib/gui/app/components/details-modal/details-modal.jsx
amdomanska: But I removed all unnecessary params

LucianBuzzo: If it needs to be hidden can you just set it to `display: hidden` or something similar? On a side note, we should either add an "empty" modal to rendition, or use the default rendition button position here, this code feels a bit messy

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -64,19 +87,30 @@ class DriveSelectorButton extends React.PureComponent {
...
+              )
             }
             <DetailsText>
               {this.allDevicesFooter()}
             </DetailsText>
           </StepSelection>
+          {this.state.show ?
+            (
+              <DetailsModal
+                title={'DRIVERS DETAILS'}
```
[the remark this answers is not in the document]
amdomanska: @LucianBuzzo Actually it should be "DRIVER" if there is one and "DRIVERS" when there is more. Not a bad idea :)

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -64,19 +87,30 @@ class DriveSelectorButton extends React.PureComponent {
...
+                >
+                  Change
+                </ChangeButton>
+              )
             }
             <DetailsText>
               {this.allDevicesFooter()}
             </DetailsText>
           </StepSelection>
+          {this.state.show ?
```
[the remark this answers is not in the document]
amdomanska: nice! I haven't thought I can do that with components.

lib/gui/app/components/drive-selector-button/drive-selector-button.jsx
```diff
@@ -22,13 +22,25 @@ const propTypes = require('prop-types')
...
 
+const DetailsModal = require('./../details-modal/details-modal')
+
 class DriveSelectorButton extends React.PureComponent {
 
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      show: false
```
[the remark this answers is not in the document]
amdomanska: it will be changed. I'm writing the second modal for Driver Selector right now and it has that changed. But I will change it in this PR.

lib/gui/app/components/details-modal/details-modal.jsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+DetailsModal.propTypes = {
+  title: propTypes.string,
+  details: propTypes.array,
+  callback: propTypes.func
+}
+
+module.exports = DetailsModal
```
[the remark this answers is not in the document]
amdomanska: @LucianBuzzo what's the difference?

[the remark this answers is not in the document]
amdomanska: @LucianBuzzo actually I cannot make react2angular module to work with named exports (in `.js` file) I constantly have `Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.` Seems straightforward but doesn't work. @thundron is there a chance you know how to use it? In the worst case it could stay like that for now and will be changed for named exports once the whole page will be converted. What do you think?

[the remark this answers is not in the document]
LucianBuzzo: How are you requiring the module?

PR #2488, opened 2018-09-24
Use Resin CI scripts to build Etcher

jviotti:
Change-type: patch


jviotti:
Depends on https://github.com/resin-io/scripts/pull/6

PR #2491, opened 2018-10-01
fix(GUI): Warn when requiring usb fails

zvin:
Change-type: patch
Signed-off-by: Alexis Svinartchouk <alexis@resin.io>


lib/sdk/adapters/usbboot/usb.js
```diff
@@ -28,6 +29,7 @@ const usb = (() => {
   try {
     return require('usb')
   } catch (error) {
+    analytics.logException(error);
```
thundron: Extra semicolon at EOL
Other than this, lgtm

PR #2496, opened 2018-10-04
Use fork of node-usb to disable libudev

thundron:
Change-type: patch
Signed-off-by: Lorenzo Alberto Maria Ambrosi <lorenzoa@resin.io>


zvin:
@thundron maybe we should fork node-usb to resin-io-modules ?

thundron:
> @thundron maybe we should fork node-usb to resin-io-modules ?

Without doubt, I'll do that once it is fully working in order to keep things clean

PR #2509, opened 2018-10-17
docs(support): Move support.md file to docs folder

amdomanska:
For new webpage support.md support.md file needs to be moved. As long as the old homepage is accessible both copies of the file need to be kept but once the new homepage will be published the old copy should be removed 

Change-type: patch
Signed-off-by: amdomanska <aga@resin.io>


thundron:
We removed the support link in the new website, closing

PR #2533, opened 2018-11-01
Lossless compression + updated dependencies

mrmajik45:
Lossless compression: Smaller!

Updated dependencies for fewer bugs and better performance!


thundron:
I'm having a look at the compressed images asap, meanwhile you could consider reducing the commits number to be more functional oriented as opposed to having a commit for each single change, especially the merge commits.
One note about dependencies: we have a very strict policy for updating dependencies in order to successfully build the final package like using `npm 3.10.10` (that comes with `node 6 if you fancy installing/using that), so for that you can split the dependencies updates in one/several separate and incremental PRs so we can check that each one doesn't break :) 

Thank you nonetheless for the effort!

PR #2539, opened 2018-11-08
Use https for fetching sub modules

Jacalz:
This switches over the sub module to using https instead of ssh. It simplifies a lot for people packaging the application and you won't need to have ssh configured correctly. 👍

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "scripts/resin"]
 	path = scripts/resin
-	url = git@github.com:balena-io/scripts.git
+	url = https://github.com/balena-io/scripts.git
+	branch = master


Jacalz:
Could anybody please have a look at this? Without this, Etcher can't be built for the Solus operati9ng system, which is a real bummer!

Jacalz:
Pinging @thundron on this :+1:

thundron:
@Jacalz We're checking if it creates any issues with people that use two-factor auth

Jacalz:
I am pretty certain that it doesn’t, you don’t need to be logged in for it to fetch sub modules, I think...

thundron:
Trying it now, I'll let you know in some minutes 👍 (for cloning you shouldn't need auth, but better be safe)

Jacalz:
This is rebased against master now :+1:
You are free to merge when you have time :slightly_smiling_face:

thundron:
@resin-ci retest

thundron:
@resin-ci retest

PR #2552, opened 2018-11-22
fix(GUI): Reject drives with null size

thundron:
Change-type: patch
Changelog-entry: Reject drives with null size (fixes pretty-bytes error)


zvin:
This filtering could also be here https://github.com/balena-io/etcher/blob/1c5bab63a530ec08dbe851fe0075ed60a1a1a67e/lib/gui/app/app.js#L233 
I don't mind, it has the same effect

PR #2556, opened 2018-11-26
fix(GUI): Fix update notification error

thundron:
Remove "this" and use "exports" instead

Change-type: patch
Changelog-entry: Fix update notifier error popping up on v1.4.1->1.4.8

### DIFF
--- a/lib/gui/app/components/update-notifier.js
+++ b/lib/gui/app/components/update-notifier.js
@@ -79,7 +79,7 @@ exports.shouldCheckForUpdates = (options) => {
     return true
   }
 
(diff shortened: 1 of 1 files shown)


thundron:
@zvin Short answer: no.

Long answer: worked fine until 1.4.1 (afaik from running each version separately), but users didn't receive the notification until 1.4.3 went out (first-to-last version notified) and so I think most of the users were still on 1.3.1 that had no issues

zvin:
@thundron this is the kind of stuff `tsc` would notice.

PR #2558, opened 2018-11-26
Updated Chocolatey section of Readme

Eate:
Previously, the section header was a level 3 header, and I changed it to a lever 4 header like all the other install variants so it is the same level header. I also added uninstall instructions like the other sections had.


thundron:
@resin-ci retest

PR #2562, opened 2018-11-27
Update URLs in GUI to balena

chrisys:
Change-type: patch
Signed-off-by: Chris Crocker-White <chriscw@balena.io>


chrisys:
@resin-ci recheck

PR #2571, opened 2018-11-30
Restyle success screen and enlarge UI elements

thundron:
Change-type: patch
Changelog-entry: Added React component for the Flash Another button


lib/gui/app/components/flash-another/flash-another.jsx
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2016 resin.io
```
lurch: 2018? :wink:

thundron:
Changing things quite a bit, will need re-approval

Issue #2281, opened 2018-04-26
Multiple drives labeled as containing source image


jhermsmeier:
- **Etcher version:** at least 1.3.1+
- **Operating system and architecture:** Linux
- **Image flashed:** N/A

If the selected image is on `/dev/sdc`:

![image](https://user-images.githubusercontent.com/244907/39318568-c58586e0-497e-11e8-8fc6-14ee5428d1db.png)

If the selected image is on `/dev/sda`:

![image](https://user-images.githubusercontent.com/244907/39318675-07935f08-497f-11e8-89dc-dbc6879a6441.png)

```js
[ { enumerator: 'lsblk',
    busType: 'UNKNOWN',
    busVersion: '0.0',
    device: '/dev/sda',
    raw: '/dev/sda',
    description: 'Samsung SSD 850',
    error: null,
    size: 120034123776,
    blockSize: null,
    logicalBlockSize: null,
    mountpoints: [ { path: '/' }, { path: '/boot/efi' } ],
    isReadOnly: false,
    isSystem: true,
    isVirtual: null,
    isRemovable: null,
    isCard: null,
    isSCSI: null,
    isUSB: null,
    isUAS: null },
  { enumerator: 'lsblk',
    busType: 'UNKNOWN',
    busVersion: '0.0',
    device: '/dev/sdb',
    raw: '/dev/sdb',
    description: 'SD Transcend',
    error: null,
    size: 31914983424,
    blockSize: null,
    logicalBlockSize: null,
    mountpoints: [],
    isReadOnly: false,
    isSystem: false,
    isVirtual: null,
    isRemovable: null,
    isCard: null,
    isSCSI: null,
    isUSB: null,
    isUAS: null },
  { enumerator: 'lsblk',
    busType: 'UNKNOWN',
    busVersion: '0.0',
    device: '/dev/sdc',
    raw: '/dev/sdc',
    description: 'AS2115',
    error: null,
    size: 1000204886016,
    blockSize: null,
    logicalBlockSize: null,
    mountpoints: [ { path: '/media/jonas/R5-D4' } ],
    isReadOnly: false,
    isSystem: true,
    isVirtual: null,
    isRemovable: null,
    isCard: null,
    isSCSI: null,
    isUSB: null,
    isUAS: null } ]
```

lurch:
[...]

PR #2582, opened 2018-12-11
Change "Drive Contains Image" label.

zvin:
Use "Drive Mountpoint Contains Image" instead as the image may not be on
this drive but on a drive mounted in one of the mountpoins of this
drive.
We still don't want to allow flashing this drive in that situation.

Change-type: patch

Connects-To: #2281

### DIFF
--- a/lib/shared/messages.js
+++ b/lib/shared/messages.js
@@ -97,7 +97,7 @@ module.exports = {
     },
 
     containsImage () {
-      return 'Drive Contains Image'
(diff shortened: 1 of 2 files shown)


lurch:
Would "Drive Mountpoint" make sense to a Windows user?

zvin:
@lurch I didn't knew but it does https://docs.microsoft.com/en-us/windows-server/storage/disk-management/assign-a-mount-point-folder-path-to-a-drive

PR #2584, opened 2018-12-18
Remove forwarding of SafeWebview console messages

thundron:
Change-type: patch


lib/gui/app/components/safe-webview/safe-webview.jsx
```diff
@@ -250,12 +250,6 @@ class SafeWebview extends react.PureComponent {
 
     if (message.command === 'error') {
       analytics.logException(message.data)
```
zvin: Do we still care about `console.error` messages?

thundron: I personally don't as I think errors on the webviews should be separated from etcher, but I think we wanted to send those on Sentry anyway?
If not we can get rid of this whole event tuple [here](https://github.com/balena-io/etcher/blob/ce5d2ec657294042f2dea72bed8b9646fff5cf18/lib/gui/app/components/safe-webview/safe-webview.jsx#L105)

PR #2586, opened 2018-12-19
v1.4.9

thundron:
### Fixes

- Fix update notifier error popping up on v1.4.1->1.4.8

### Misc

- Added React component for the Flash Results button
- Added React component for the Flash Another button
- Restyle success screen and enlarge UI elements
- Use https for fetching sub modules
- Add `.wic` image extension as supported format


thundron:
@zvin Yes, otherwise we'd actually have the opposite behaviour as originally planned (e.g. we want to send events only below a certain threshold, not above)

PR #2587, opened 2018-12-21
New dmg package background assets

thundron:
Change-type: patch


thundron:
screenshot 2018-12-22 at 10 51 37
@apexider This is how it looks
I had to turn down the size of the images you gave me (the `.tiff` particularly) because they were too big for the window

apexider:
Like that it does not look good. Can you use @2x images so it looks good on retina displays? Let me know if you want me to provide the image again differently.

Attached the file as png:

etcher_install_window 2x

thundron:
@apexider It's not possible unfortunately, the image stretches beyond the window's size making it look horrible

Issue #2128, opened 2018-03-21
Documentation on how to use bmap files with etcher


symgryph:
Just a documentation question on bmaptool support, where do I put the bmap file and what do I name it to?

Thomas

jhermsmeier:
The documentation around that has been removed, but never really replaced with something.
All you need to do though, is to put both the image, bmap and a manifest into a zip archive. The manifest is just a JSON file in a `.meta` directory within the archive. The basic structure is the following:

```
archive.zip
├── imagefile.img
└── .meta/
    ├── manifest.json
    ├── image.bmap
    └── logo.svg
```

I'd recommend checking out https://github.com/ev3dev/ev3dev/releases for an example.

symgryph:
Thank you much for this. It really helps. I will be happy to submit a document with the entire process if you would like in markdown format. Should I do a pull request with the file?

Also, what method do you use to generate the bmap files? I use the bmaptool method, but its quite cumbersome, and only works on linux. Have you developed a more streamlined method of making bmap files?

Perhaps a docker appliance?

Thomas

jhermsmeier:
Sure, that'd be awesome if you'd PR it! We have a somewhat resinOS specific process to blockmap those images, which doesn't work with any image, and otherwise we rarely use bmaps to be honest :) The process they recommend is quite cumbersome, and also platform specific unfortunately, and we've been planning to create alternative methods, but it's not high on the list.

jcormier:
A 2019 example of a zipped image with bmap. https://github.com/balena-io-modules/etcher-sdk/commit/574978b8444eab0867d5ec4302d2c40ecac42707

jcormier:
bmap file needs to be named image.bmap and placed in the .meta folder

Then blockMap should be true in the console log after you select the image and it will print all the different blocks.

![image](https://user-images.githubusercontent.com/5613721/112865316-a731ac80-9086-11eb-9118-eaf78bb1d385.png)

jcormier:
Example bash which renames wic image files from Yocto builds and zips them up

```bash
					# Rename wic to .img and zip up sd card images
					for f in *.wic; do
						# Replace .wic with .img
						img="$(basename -s .wic "$f").img"

						# Cleanup leftover .meta folders
						rm -rf .meta

						cp "$f" "$img"
						cp "$f".bmap "$img".bmap

						# Create .meta folder for balenaEtcher bmap support
						mkdir .meta
						cp "$f".bmap .meta/image.bmap

[...]

PR #2630, opened 2019-02-14
Remove etcher-cli

zvin:
Change-type: major


lurch:
Oh! Will https://www.balena.io/etcher/cli be disappearing too? (and BTW, the "report issues" link on that page is broken)

RobertCNelson:
Why was this removed? It made things so much more useful when jumping between beaglebone images and was better than bmap-tools which needed a *.bmap file (bmap-tools is only cli based)..  I guess i better re-enable *.bmap files for beagleboard.org

YUCK!  Please revert this..

Regards,

lurch:
In theory, writing a .bmap file (if it was properly created) can/should be faster than writing a .img file. And Etcher also supports writing bmaps. See #2128

LongLiveCHIEF:
In fact, the UI never works at all, on Windows _or_ Linux (as admin or regular user), and the only way I was able to create images with etcher was with the cli.

👎

lurch:
"In fact, the UI never works at all..."

I'm sure thousands of people would disagree with you :wink: 
Have you created an issue describing your problem in more detail?

LongLiveCHIEF:
Trying to narrow it down. I've ruled out some of the existing known issues
(like docker running). I wound up having to re-install both Windows and
Ubuntu shortly after leaving this comment, and have since used etcher via
the UI just fine on Windows.

I haven't re-installed all my normal suite of tools yet, so I'll keep
checking it as I go to see if something causes a conflict.

Etcher logs were non-deterministic at the time.  I even used a different
USB drive, so I literally wiped the slate clean.

My gut tells me it could actually be the drive that was causing the
problem, but I haven't tested it since.

rikvermeer:
I loved this product for the cli... back to dd :'(

goodbye my lover <3
hope to see you back

NicolasDorier:
Is there any other alternative to flash an image in cli?

thundron:
@NicolasDorier you can use examples from [etcher-sdk](https://github.com/balena-io-modules/etcher-sdk), they're not fully documented yet but you can give them a try (cc @rikvermeer in case you want to give it a try too)

NicolasDorier:
@thundron I replaced by using `dd bs=8M "if=$IMAGE" "of=/dev/sdx"` it does not check integrity though.

PR #2631, opened 2019-02-14
Dont use drivelist directly

zvin:
Enable the experimental file picker if you want to test this
```json
{
	"fileBrowserConstraintPath": "/dev/sda",
	"experimentalFilePicker": true
}
```
in `~/.config/balena-etcher/config.json`

### DIFF
--- a/lib/gui/app/models/files.js
+++ b/lib/gui/app/models/files.js
@@ -19,7 +19,8 @@
 const Bluebird = require('bluebird')
 const fs = Bluebird.promisifyAll(require('fs'))
 const path = require('path')
-const drivelist = require('drivelist')
+
+const driveScanner = require('../modules/drive-scanner')
 
 /* eslint-disable lodash/prefer-lodash-method */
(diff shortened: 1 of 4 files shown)


lib/gui/app/models/files.js
```diff
@@ -143,26 +144,13 @@ exports.splitPath = (fullpath, subpaths = []) => {
...
-
-    const constraintDevice = devices.find((device) => {
-      return device.device === pathname ||
-        device.devicePath === pathname
-    })
-
-    callback(null, constraintDevice)
-  })
+exports.getConstraintDevice = (pathname) => {
+  // This supposes the drive scanner is ready
```
thundron: Why "supposes"? I'd rather it be sure about it, what happens if not?

zvin: If it's not ready, it'll return `undefined`.
This is called when you open the experimental file picker so the scanner is already started and ready at this point.

PR #2651, opened 2019-02-22
Update electron links

zrzka:
Original [link works](https://electronjs.org/docs/tutorial/supported-platforms), but the page says the it was moved to the support.md. This PR updates this link to the correct/working one:

* https://electronjs.org/docs/tutorial/support#supported-platforms

It also updates electron link to https://electronjs.org/

Change-type: patch
Signed-off-by: Robert Vojta <robert@balena.io>

Issue #2652, opened 2019-02-22
AppImages link leads to a file/folder, which doesn't exist (user documentation)


zrzka:
* Open user documentation, chapter [Deactivate desktop shortcut prompt on GNU/Linux](https://github.com/balena-io/etcher/blob/master/docs/USER-DOCUMENTATION.md#deactivate-desktop-shortcut-prompt-on-gnulinux)
* There's AppImages link, which leads to https://github.com/balena-io/etcher/blob/master/docs/appimage

But this file/folder doesn't exist.

lurch:
Looks like changing `[AppImages](appimage)` to `[AppImages][appimage]` would fix that?

zrzka:
Thanks, did add to #2653.

lurch:
I guess this can be closed :wink:

PR #2653, opened 2019-02-22
Fix user documentation

zrzka:
* Same electron link updates as in #2651
* According to [this page](https://electronjs.org/docs/tutorial/support#macos), macOS 10.10 (Yosemite) is required, not macOS 10.9


zrzka:
One more commit to fix the #2652 as well.

zrzka:
Closing. Cherry picked to 1.5.7 PR #2674.

PR #2655, opened 2019-02-22
Fix blurred background image

### LINKED ISSUE #2643: Blurred background in the DMG (macOS installer)
See attached image. Repo contains `background.png` and `background@2x.png`, but the DMG contains TIFF (`background.tiff`), which contains just the standard resolution version (not both). And it leads to a blurred background image. Compare _balenaEtcher.app_ & _Applications_ with _An open source project by balena_.

screenshot 2019-02-22 at 13 35 37

zrzka:
This PR fixes #2643. See my [comment here](https://github.com/balena-io/etcher/issues/2643#issuecomment-466585838). In other words, `background.tiff` in the repo contains just the standard resolution version. But because the date & time of all these files is same, `make assets/dmg/background.tiff` does nothing. It's explained in the linked comment.

Change-type: patch
Signed-off-by: Robert Vojta <robert@balena.io>


zrzka:
Another approach would to be remove `background.tiff` from the repo, add it to `.gitignore` and rebuild it every single time. Which is probably safer and always works way.

lurch:
@zrzka I suggested something similar to that once, since there's a lot of "generated files" (e.g. the CSS files) checked into git; but I was told that it was a deliberate decision to keep those in github. I _think_ the reasoning was something like it makes Etcher easier to build for people who don't have **all** the required build-tools installed :man_shrugging:

zrzka:
@lurch In this case, I'd rather see it removed & ignored. `tiffutil` is part of macOS for quite a long time. Even books from 2003 mention it. In other words, everyone who has macOS has `tiffutil`. And with this change we can avoid same problems in the future. But I don't know rationale behind this decission, so, just my $.02 :)

zrzka:
@zvin okay, will rebase all PRs.

zrzka:
@zvin is it possible to enable [Allow rebase merging](https://help.github.com/en/articles/configuring-commit-rebasing-for-pull-requests) for the repo? So, one can rebase it here (GH UI) as well?

zvin:
@zrzka Sorry I don't have access to settings on this repo

zrzka:
Closing. Cherry picked to 1.5.7 PR #2674.

PR #2686, opened 2019-03-10
Bump dependencies

szb512:


szb512:
It will bump dependencies to the latest version.

PR #2688, opened 2019-03-11
Use APPDIR from env in the child writer

zvin:
Change-type: patch


zvin:
@resin-ci retest

nazrhom:
@resin-ci retest

PR #2690, opened 2019-03-12
Bump dependencies (round 2)

szb512:
This will bump dependencies.


zvin:
You're bumping major versions without testing, please stop.

PR #2691, opened 2019-03-13
Update electron-builder scripts

thundron:
Change-type: patch

### DIFF
--- a/.resinci.json
+++ b/.resinci.json
@@ -16,7 +16,7 @@
       "appId": "io.balena.etcher",
       "copyright": "Copyright 2016-2019 Balena Ltd",
       "productName": "balenaEtcher",
-      "nodeGypRebuild": true,
+      "nodeGypRebuild": false,
       "files": [
         "!lib/gui/app",
         "lib/gui/app/index.html",


package.json
```diff
@@ -21,9 +21,9 @@
   "scripts": {
     "test": "make lint test sanity-checks",
     "start": "electron lib/start.js",
-    "configure": "node-gyp configure",
-    "build": "node-gyp build",
-    "install": "node-gyp rebuild",
+    "configure": "npx node-gyp configure",
+    "build": "npx node-gyp build",
+    "install": "npx node-gyp rebuild",
```
zvin: We don't need this.

npm-shrinkwrap.json
```diff
@@ -14910,6 +15447,17 @@
         }
       }
     },
+    "winusb-driver-generator": {
```
zvin: fsevents and winusb-driver-generator shouldn't be in teh shrinkwrap file.

thundron: I don't even know how they got there, I'm on OSX

PR #2698, opened 2019-03-20
New parameter in webview for opt-out analytics

thundron:
Change-type: patch


thundron:
@dimitrisnl I don't understand, could you please comment the exact line? The condition is supposed to be in the website in order to stop tracking based on the parameter we pass here

dimitrisnl:
I meant that passing along undefined & null values make it harder to understand. We should strip those beforehand.

I was thinking of something like

```
if (settings.get('errorReporting’) === false) {
    url.searchParams.set('optOutAnalytics', true))
}
```

lurch:
Presumably this works for both the "during flashing" webview **and** the "flash finished" webview?

dimitrisnl:
Yes @lurch 	It's for both of them

PR #2702, opened 2019-03-21
Trim

zvin:
This needs to be enabled on the settings screen.
Already works (with raw images) but may be a bit slow when writing small or unaligned chunks.
This can probably be optimized in etcher-sdk and / or blockmap.
Still faster than writing the whole image.

### DIFF
--- a/lib/gui/app/pages/settings/templates/settings.tpl.html
+++ b/lib/gui/app/pages/settings/templates/settings.tpl.html
@@ -46,6 +46,17 @@
   <div class="checkbox">
     <label>
       <input type="checkbox"
+        tabindex="8"
+        ng-model="settings.currentData.trim"
+        ng-change="settings.toggle('trim')">
+
+      <span>Trim ext{2,3,4} partitions before writing (raw images only)</span>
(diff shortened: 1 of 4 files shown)


zvin:
This needs to be enabled on the settings screen.
Already works but may be a bit slow when writing small or unaligned chunks.
This can probably be optimized in etcher-sdk and / or blockmap.
Still faster than writing the whole image.

lib/gui/app/pages/settings/templates/settings.tpl.html
```diff
@@ -43,6 +43,17 @@ <h1 class="title space-bottom-large">Settings</h1>
...
   </div>
 
+  <div class="checkbox">
+    <label>
+      <input type="checkbox"
+        tabindex="8"
+        ng-model="settings.currentData.trim"
+        ng-change="settings.toggle('trim')">
+
+      <span>Trim ext{2,3,4} partitions before writing (raw images only)</span>
```
lurch: Would it be possible to add a tooltip or something explaining what this actually means / does?

PR #2721, opened 2019-04-01
Fix error when config.analytics was undefined

zvin:
Changelog-entry: Fix error when config.analytics was undefined
Change-type: patch


zvin:
@resin-ci retest

zvin:
@resin-ci retest

PR #2724, opened 2019-04-02
Use full path to wmic as some systems don't have it in their PATH

zvin:
Changelog-entry: (Windows): Use full path to wmic as some systems don't have it in their PATH
Change-type: patch


lurch:
Seems kinda ironic that a lot of work as put into drivelist to get rid of WMI, and now it's being added into Etcher :wink:
https://github.com/balena-io-modules/drivelist/search?q=WMI&type=Commits

PR #2726, opened 2019-04-03
Add sample property to Mixpanel events

thundron:
Change-type: patch

### DIFF
--- a/lib/gui/app/modules/analytics.js
+++ b/lib/gui/app/modules/analytics.js
@@ -50,16 +50,28 @@ resinCorvus.install({
 
 let mixpanelSample = DEFAULT_PROBABILITY
 
-getConfig(configUrl)
-  .then((config) => {
+/**
+ * @summary Init analytics configurations
+ * @example initConfig()
+ */
+const initConfig = async () => {
+  let validatedConfig = null
+  try {
+    const config = await getConfig(configUrl)
     const mixpanel = _.get(config, [ 'analytics', 'mixpanel' ], {})
-    const probability = mixpanel.probability || DEFAULT_PROBABILITY
-    mixpanelSample = probability
-    const validatedConfig = isClientEligible(probability) ? validateMixpanelConfig(mixpanel) : null
-    resinCorvus.setConfigs({
-      mixpanel: validatedConfig
-    })
+    mixpanelSample = mixpanel.probability || DEFAULT_PROBABILITY
+    if (isClientEligible(mixpanelSample)) {
+      validatedConfig = validateMixpanelConfig(mixpanel)
+    }
+  } catch (err) {
+    resinCorvus.logException(err)
+  }
+  resinCorvus.setConfigs({
+    mixpanel: validatedConfig
   })
+}
+
+initConfig()
 
 /**
  * @summary Get etcher configs stored online
@@ -129,10 +141,7 @@ exports.logDebug = resinCorvus.logDebug
  * });
  */
 exports.logEvent = (message, data) => {
-  const sessionData = {
-    mixpanelSample
-  }
-  resinCorvus.logEvent(message, { ...data, sessionData })
+  resinCorvus.logEvent(message, { ...data, sample: mixpanelSample })
 }
 
 /**
--- a/lib/gui/app/modules/analytics.js
+++ b/lib/gui/app/modules/analytics.js
@@ -31,7 +31,7 @@ const mixpanelToken = settings.get('analyticsMixpanelToken') ||
 
 const configUrl = settings.get('configUrl') || 'http://balena.io/etcher/static/config.json'
 
-const DEFAULT_PROBABILITY = 1
+const DEFAULT_PROBABILITY = 0.1
 
 const services = {
   sentry: sentryToken,
@@ -48,10 +48,16 @@ resinCorvus.install({
   }
 })
 
+let mixpanelSample = DEFAULT_PROBABILITY
+
 getConfig(configUrl)
   .then((config) => {
+    const mixpanel = _.get(config, [ 'analytics', 'mixpanel' ], {})
+    const probability = mixpanel.probability || DEFAULT_PROBABILITY
+    mixpanelSample = probability
+    const validatedConfig = isClientEligible(probability) ? validateMixpanelConfig(mixpanel) : null
     resinCorvus.setConfigs({
-      mixpanel: getMixpanelConfig(config)
+      mixpanel: validatedConfig
     })
   })
 
@@ -70,14 +76,8 @@ function getConfig(url) {
  * @param {Object} - config
  */
 // eslint-disable-next-line
-function getMixpanelConfig(config) {
-  const analytics = config.analytics || {}
-  const mixpanel = analytics.mixpanel || {}
-  const probability = mixpanel.probability || DEFAULT_PROBABILITY
-  if (Math.random() > probability) {
-    return null
-  }
-  return validateMixpanelConfig(mixpanel)
+function isClientEligible(probability) {
+  return Math.random() > probability
 }
 
 /**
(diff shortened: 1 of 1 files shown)


lib/gui/app/modules/analytics.js
```diff
@@ -70,10 +73,7 @@ function getConfig(url) {
  * @param {Object} - config
  */
 // eslint-disable-next-line
-function getMixpanelConfig(config) {
-  const analytics = config.analytics || {}
-  const mixpanel = analytics.mixpanel || {}
-  const probability = mixpanel.probability || DEFAULT_PROBABILITY
+function isClientEligible(mixpanel, probability) {
```
zvin: The function name seems wrong, also it should only take a mixpanel parameter as probability is included in it.

lib/gui/app/modules/analytics.js
```diff
@@ -50,9 +50,12 @@ resinCorvus.install({
 
 getConfig(configUrl)
   .then((config) => {
+    const probability = _.get(config, [ 'analytics', 'mixpanel', 'probability' ], DEFAULT_PROBABILITY)
+    const mixpanel = _.get(config, [ 'analytics', 'mixpanel' ])
     resinCorvus.setConfigs({
-      mixpanel: getMixpanelConfig(config)
+      mixpanel: isClientEligible(mixpanel, probability)
     })
+    settings.set('sample', probability)
```
zvin: Can we rename `sample` to something like `mixpanelSample` ?
Also why do we put it in settings?
It could just be local to this file.

lib/gui/app/modules/analytics.js
```diff
@@ -128,7 +128,13 @@ exports.logDebug = resinCorvus.logDebug
  *   image: '/dev/disk2'
  * });
  */
-exports.logEvent = resinCorvus.logEvent
+exports.logEvent = (message, data) => {
+  const sessionData = {
+    sample: settings.get('sample') || DEFAULT_PROBABILITY
+  }
+  const extendedData = _.merge(data, sessionData)
```
zvin: This can be written like
```javascript
const sample = settings.get('sample') || DEFAULT_PROBABILITY;
const extendedData = { ...data, sample };
```
now that we have electron 3.

lib/gui/app/modules/analytics.js
```diff
@@ -48,11 +48,17 @@ resinCorvus.install({
   }
 })
 
+let mixpanelSample = DEFAULT_PROBABILITY
+
 getConfig(configUrl)
   .then((config) => {
+    const mixpanel = _.get(config, [ 'analytics', 'mixpanel' ]) || {}
```
zvin:
```js
const mixpanel = _.get(config, [ 'analytics', 'mixpanel' ], {})
```

lib/gui/app/modules/analytics.js
```diff
@@ -48,11 +48,17 @@ resinCorvus.install({
...
+    const probability = mixpanel.probability || DEFAULT_PROBABILITY
+    const validatedConfig = isClientEligible(probability) ? validateMixpanelConfig(mixpanel) : null
     resinCorvus.setConfigs({
```
zvin: Do we need to call setConfigs if the config is null?

lib/gui/app/modules/analytics.js
```diff
@@ -48,11 +48,17 @@ resinCorvus.install({
...
-      mixpanel: getMixpanelConfig(config)
+      mixpanel: validatedConfig
     })
+    mixpanelSample = mixpanel.probability
```
zvin: This may set `mixpanelSample` to undefined, use `probability` instead.

lib/gui/app/modules/analytics.js
```diff
@@ -128,7 +128,13 @@ exports.logDebug = resinCorvus.logDebug
  *   image: '/dev/disk2'
  * });
  */
-exports.logEvent = resinCorvus.logEvent
+exports.logEvent = (message, data) => {
+  const sessionData = {
+    mixpanelSample
+  }
+  const extendedData = { ...data, sessionData }
```
zvin: You want 
```js
resinCorvus.logEvent(message, { ...data, mixpanelSample })
```

lib/gui/app/modules/analytics.js
```diff
@@ -70,14 +76,8 @@ function getConfig(url) {
...
 // eslint-disable-next-line
-function getMixpanelConfig(config) {
-  const analytics = config.analytics || {}
-  const mixpanel = analytics.mixpanel || {}
-  const probability = mixpanel.probability || DEFAULT_PROBABILITY
-  if (Math.random() > probability) {
-    return null
-  }
-  return validateMixpanelConfig(mixpanel)
+function isClientEligible(probability) {
```
zvin: You can use `mixpanelSample` and remove the parameter.

lib/gui/app/modules/analytics.js
zvin: or 
```
resinCorvus.logEvent(message, { ...data, sample: mixpanelSample })
```
according to the PR name.

nazrhom:
@thundron its failing to clone on CI because the scripts repo was moved https://ci.balena-dev.com/teams/main/pipelines/github-events-resin/jobs/is-pr/builds/17850 We should update the reference to master and never leave it pointing to a branch

lib/gui/app/modules/analytics.js
thundron: I think so, let me check again on `resin-corvus`

lib/gui/app/modules/analytics.js
thundron: I don't like the idea of the function depending on a parameter outside of its context tbh

lib/gui/app/modules/analytics.js
thundron: We need to do so in order to clear the queue of events that stacks, since we're loading Mixpanel deferred in order to retrieve the online config first

lib/gui/app/modules/analytics.js
```diff
@@ -128,7 +128,12 @@ exports.logDebug = resinCorvus.logDebug
  *   image: '/dev/disk2'
  * });
  */
-exports.logEvent = resinCorvus.logEvent
+exports.logEvent = (message, data) => {
+  const sessionData = {
+    sample: mixpanelSample
+  }
+  resinCorvus.logEvent(message, { ...data, sessionData })
+}
```
zvin: Do we need this extra `sessionData` layer? It will only contain `sample`.

lib/gui/app/modules/analytics.js
zvin: This means that we need a `.catch` for this `getConfig()` or the mixpanel events will accumulate forever.

lib/gui/app/modules/analytics.js
thundron: We don't... *for now*, so I guess it's fine to remove it.
*For now* 👀

thundron: It seems like we do in order to use the spread operator

thundron: We don't for now, added it to make it more clear that we can add w/e to that context.
We can put it back when/if needed

lib/gui/app/modules/analytics.js
thundron: Added

lib/gui/app/modules/analytics.js
```diff
@@ -48,10 +48,22 @@ resinCorvus.install({
...
+      mixpanel: validatedConfig
+    })
+  })
+  .catch((err) => {
+    resinCorvus.logException(err)
     resinCorvus.setConfigs({
-      mixpanel: getMixpanelConfig(config)
+      mixpanel: null
     })
   })
```
zvin:
```
async function initConfig() {
    let validatedConfig = null;
    try {
        const config = await getConfig(configUrl)
        const mixpanel = _.get(config, [ 'analytics', 'mixpanel' ], {})
        mixpanelSample = _.get(mixpanel, 'probability', DEFAULT_PROBABILITY)
        if (isClientEligible(mixpanelSample)) {
            validatedConfig = validateMixpanelConfig(mixpanel)
        }
    } catch (err) {
        resinCorvus.logException(err)
    } finally {
        resinCorvus.setConfigs({ mixpanel: validatedConfig })
    }
}

initConfig()
```

PR #2734, opened 2019-04-16
Fix reading images from network drives on windows when the tmp dir has spaces

zvin:
Changelog-entry: (Windows): Fix reading images from network drives when the tmp dir has spaces
Change-type: patch


thundron:
@resin-ci retest

PR #2767, opened 2019-05-06
[WIP] Add WiFi management

thundron:


lib/gui/app/components/wifi/react/actions.js
```diff
@@ -0,0 +1,195 @@
...
+        })
+      })
+      .then((res) => {
+        return res.value
+      })
+      .catch((err) => {
+        const message = (!_.isEmpty(err.data) && _.head(err.data)) || err.message
+
+        // In case there's no registered connection yet
+        if (message === 'The device \'wlan0\' has no connections available for activation.') {
```
lurch: I have no context for where / how this is going to be used, but just out of curiosity I'm wondering if it'd be safer to use a regex like `The device \'.*\' has no connections available for activation\.` rather than relying on a hardcoded `wlan0` string?

thundron: @lurch Development leftover 👀

PR #2789, opened 2019-05-15
Build packages that support both ia32 and x64 on windows

zvin:
Change-type: patch

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -1,8 +1,8 @@
 appId: io.balena.etcher
 copyright: Copyright 2016-2019 Balena Ltd
 productName: balenaEtcher
-npmRebuild: false
-nodeGypRebuild: false
+npmRebuild: true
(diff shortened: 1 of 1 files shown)


lurch:
Just out of curiosity, can Windows-ia32 and Windows-x64 both build Windows-ia32 and Windows-x64 packages, or can only Windows-x64 build Windows-ia32 and Windows-x64 packages? (IYSWIM!)

zvin:
replaced by #2791

thundron:
> Just out of curiosity, can Windows-ia32 and Windows-x64 both build Windows-ia32 and Windows-x64 packages, or can only Windows-x64 build Windows-ia32 and Windows-x64 packages? (IYSWIM!)

@lurch I don't think it's a two-ways compilation but haven't tried to build a x64 app on ia32 yet tbh. Sounds too strange though, but we've seen worse so might give it a shot :P

PR #2791, opened 2019-05-16
Build packages that support both ia32 and x64 on windows

zvin:
Changelog-entry: windows installer and portable version support both ia32 and x64
Change-type: patch


zvin:
@balena-ci retest

zvin:
replaced by #2799

PR #2795, opened 2019-05-17
Upgrade rendition to v8

thundron:


npm-shrinkwrap.json
```diff
@@ -86,7 +86,6 @@
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/@babel/helper-annotate-as-pure/-/helper-annotate-as-pure-7.0.0.tgz",
       "integrity": "sha512-3UYcJUj9kvSLbLbUIfQTqzcy5VX7GRZ/CCDrnOaZorFFM01aXp1+GJwuFGV4NDDoAS+mOUyHcO6UD/RfqOks3Q==",
-      "dev": true,
```
zvin: The `dev: true` removals seem strange, did you use npm 6.7.0 ?

thundron: Yes I did, I'm now running the installation step again (also `@types/react-dom` was missing)

thundron: Same results

lib/gui/app/components/progress-button/progress-button.jsx
```diff
@@ -49,6 +53,10 @@ const ProgressButtonStripes = keyframes `
   }
 `
 
+const ProgressButtonStripesRule = css `
```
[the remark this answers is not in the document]
thundron: eslint rule

thundron:
Screenshot 2019-06-03 at 16 08 47

lib/gui/app/styled-components.js
```diff
@@ -16,34 +16,70 @@
...
-  Button, Txt, Flex
+  Button,
+  Txt,
+  Flex,
+  Provider
 } = require('rendition')
+const {
+  height,
+  space
+} = require('styled-system')
```
sradevski: I didn't check, so this is just to confirm, do you have `styled-system` as a dependency or you rely on rendition for it to be there?

thundron: @sradevski Confirmed, `"styled-system": "^4.1.0"`

lib/gui/app/components/drive-selector/target-selector.jsx
```diff
@@ -0,0 +1,137 @@
...
+  ThemedProvider
+} = require('./../../styled-components')
+const { Txt } = require('rendition')
+const middleEllipsis = require('./../../utils/middle-ellipsis')
+const { bytesToClosestUnit } = require('./../../../../shared/units')
+
+const TargetSelector = (props) => {
+  const targets = props.selection.getSelectedDrives()
+
+  let buttonTemplate = props.show && (
```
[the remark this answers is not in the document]
thundron: Sounds good, gonna change it then. I thought about doing this just to avoid repeating the `<ThemedProvider>` component but it's w/e, better use the same style everywhere

package.json
```diff
@@ -39,6 +39,7 @@
   ],
   "dependencies": {
     "@fortawesome/fontawesome-free-webfonts": "^1.0.9",
+    "@types/react-dom": "^16.8.4",
```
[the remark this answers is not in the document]
thundron: Gotcha. Saw that it was missing and added it as a normal dependency to be safer inside the package

thundron:
> Overall looks good. What I didn't see is you importing `CircularStd` and `Nunito` anywhere, so maybe that is missing?

Yes they were indeed missing. Added them in the last push

thundron:
@sradevski This still needs some improvements, I'm unsure if it's better to put them in a separate PR though (need to convert the menu icons to rendition in order to fix them properly and permanently)

sradevski:
@thundron I think it would be better if that is a separate PR, so as not to overcrowd this one with changes.

thundron:
@sradevski Ok then, sounds good. Gonna test the packages as soon as they're ready 👍

lurch:
Dunno if it's intentional or not, but running the AppImage on Ubuntu 16.04, and the fonts look like they're always displayed in **bold** style? (And they're maybe a bit bigger too, so hopefully this won't cause any alignment or word-wrapping issues anywhere in the app)

image

And I just spotted that whilst the old "Select drive" button is now "Select target", the title on the window that pops up after clicking that button is still "Select a Drive".

thundron:
@lurch Yes we have many ongoing changes, this was a middle step that we needed in order to advance with other stuff (reducing font weight as you mentioned, converting "Drive" to generic "Target" to match the fact that you can flash sdcard/usb/physical devices, etc)

thundron:
The bigger font size is needed considering that Etcher will run on a small screen soon™ :)

PR #2800, opened 2019-05-27
waffle.io removal and adding a link to the license

mhajder:


zvin:
@mhajder You need to add a `Change-type: patch` and optionally a `Changelog-entry: some message` in your commit message (check previous commits for reference).

PR #2803, opened 2019-05-27
Fix elevation on windows when the path contains "&" or "'"

### LINKED ISSUE #2809: Elevation issues on windows when the path contains "&" or "'"
- **Etcher version:** 1.5.43
- **Operating system and architecture:** Windows

zvin:
fixes #2809
Change-type: patch

### DIFF
--- a/lib/shared/permissions.js
+++ b/lib/shared/permissions.js
@@ -14,16 +14,23 @@
  * limitations under the License.
  */
 
+/* eslint-disable lodash/prefer-lodash-method,quotes,no-magic-numbers,require-jsdoc */
+
 'use strict'
 
-const os = require('os')
-const nativeModule = require('./native-module')
 const Bluebird = require('bluebird')
 const childProcess = Bluebird.promisifyAll(require('child_process'))
-const sudoPrompt = Bluebird.promisifyAll(require('sudo-prompt'))
-const commandJoin = require('command-join')
+const fs = require('fs')
 const _ = require('lodash')
+const os = require('os')
+const sudoPrompt = Bluebird.promisifyAll(require('sudo-prompt'))
+const { promisify } = require('util')
+
 const errors = require('./errors')
+
+const { tmpFileDisposer } = require('./utils')
+
+const writeFileAsync = promisify(fs.writeFile)
 
 /**
  * @summary The user id of the UNIX "superuser"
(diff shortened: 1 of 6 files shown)


zvin:
@balena-ci retest

tests/shared/permissions.spec.js
```diff
@@ -32,70 +34,32 @@ describe('Shared: permissions', function () {
...
+            "C:\\Users\\Alice & Bob's Laptop\\\"what\"\\balenaEtcher",
+            [
+              '"a Laser"',
+              'arg1',
+              "'&/ ^ \\",
+              '" $ % *'
+            ],
+            {
+              key: 'value',
+              key2: ' " ^ & = + $ % / \\',
```
lurch: Might be also worth adding a `\'` (i.e. an actual single-quote character) to the test string for key2? (both here and on the unix test)

lurch:
Looks like this is changing things to create a temp-script, and then execute that; rather than executing a single in-line command? IIRC there were problems a long time ago with the temp-script approach, which is why it was changed to the in-line command approach?
Perhaps @jviotti remembers more, or perhaps I'm just misremembering things or getting confused with some other issue? :man_shrugging:

tests/shared/permissions.spec.js
zvin: Added

tests/shared/permissions.spec.js
```diff
@@ -32,70 +34,32 @@ describe('Shared: permissions', function () {
...
+          `set "key=value"${os.EOL}` +
+          `set "key2= " ' ^^ ^& = + $ % / \\"${os.EOL}` +
+          `set "key3=8"${os.EOL}` +
+          `"C:\\Users\\Alice & Bob's Laptop\\\\"what\\"\\balenaEtcher" "\\"a Laser\\"" "arg1" "'&/ ^ \\" "\\" $ % *"`
+        )
       })
     })
 
-    describe('given linux', function () {
+    describe('given unix', function () {
```
lurch: You've changed "linux" -> "unix" here, but a couple of lines below it still says `this.osPlatformStub.returns('linux')` ?
I've no idea whether that's important or not though :wink:

zvin:
@balena-ci retest

lib/shared/permissions.js
```diff
@@ -90,68 +97,61 @@ exports.isElevatedUnixSync = () => {
...
+  return `set "${name}=${String(value)}"`
+}
 
-  if (isWindows) {
-    // This is a trick to make the binary afterwards catch
-    // the environment variables set just previously.
-    return _.concat(argv, [ 'call' ])
+// Exported for tests
+exports.createLaunchScript = (command, argv, environment) => {
+  const isWin = os.platform() === 'win32'
```
lurch: Might be nice to change `isWin` to `isWindows`, to match code elsewhere?

tests/shared/permissions.spec.js
zvin: osx and linux should behave the same, updated the tests to test both 'linux' and 'darwin'

lib/shared/permissions.js
zvin: updated

Roboblunder:
It was not fixed.

PR #2805, opened 2019-05-28
[WIP] Add etcher-pro logic

thundron:
Change-type: patch
Changelog-entry: Add env configs & balena lockfile management


lib/gui/etcher.js
```diff
@@ -124,29 +134,67 @@ const createMainWindow = () => {
...
+  }
+
+  // Listen for a 'resin-update-lock' to either enable, disable or check
+  // the update lock from the renderer process (i.e. the app)
+  if (config.resinUpdateLock) {
+    const lockFile = require('lockfile')
+    ipcMain.on('resin-update-lock', (event, command) => {
+      switch (command) {
+        case 'lock':
+          lockFile.lock('/tmp/resin/resin-updates.lock', (error) => {
```
lurch: Hmm, not `/tmp/balena/updates.lock` ? https://www.balena.io/docs/learn/deploy/release-strategy/update-locking/
Perhaps it'd be worth adding a `lockFilePath` const (instead of repeating the filename in multiple places), to make it easier to switch in future?

thundron: @lurch Good catch! We'll also try to find a proper place for the default settings in `write-env-config.js`, this is just a stub to speed up development of etcher-pro stuff so many things will change and many will need to be adjusted

PR #2806, opened 2019-05-28
Rollback update webpack

thundron:


lurch:
Oh dear, did something slip through testing? :disappointed:

thundron:
@lurch Looks like it references absolute paths for node_modules loaded through `sass-loader`, thus referencing the ones of the builder instead of local package :/

lurch:
Is that something that's worth fixing with a patch to webpack? :man_shrugging:

PR #2830, opened 2019-06-13
Make window size configurable

zvin:
Change-type: patch
Changelog-entry: Make window size configurable


thundron:
Linting rules 🚓

PR #2853, opened 2019-07-12
Add docker files for balenaCloud

thundron:
Change-type: patch
Changelog-entry: Add docker files for balenaCloud


lurch:
pigpio probably isn't needed in the buildstep? :man_shrugging:

thundron:
@lurch It is required to install the `pigpio` node module

Issue #2858, opened 2019-07-21
Unable to open an image from the mapped/network drive


zrzka:
- **Etcher version:** 1.5.51
- **Operating system and architecture:** Windows 10 Home x86-64
- **Image flashed:** Any
- **Environment:** Windows 10 Home is executed in the VMware Fusion on a Mac
  * This is probably irrelevant, because the problem is with mapped network drives

VMware Fusion can share folders from the host OS. It's implemented via mapped network drives.

<img width="1699" alt="Screenshot 2019-07-21 at 10 14 28" src="https://user-images.githubusercontent.com/1084172/61588896-16ad8400-aba3-11e9-8618-18ec4da45ffa.png">

When I open an image from `Z:\SharedWithVM\...`, etcher thinks that the location is `C:\SharedWithVM\...` (`Z` -> `C`) and fails to open the image. Nothing interesting in logs, just `ENOENT` with wrong path: 

<img width="979" alt="Screenshot 2019-07-21 at 10 30 26" src="https://user-images.githubusercontent.com/1084172/61588962-0518ac00-aba4-11e9-8fb2-618d88bbbced.png">

I've found an old [issue](https://github.com/balena-io/etcher/issues/2278) with [suggested workaround](https://github.com/balena-io/etcher/issues/2278#issuecomment-474992829) (comes from the [Microsoft Support](https://support.microsoft.com/en-us/help/3035277/mapped-drives-are-not-available-from-an-elevated-prompt-when-uac-is-co)). I have this value already set ...

<img width="902" alt="Screenshot 2019-07-21 at 10 21 11" src="https://user-images.githubusercontent.com/1084172/61588985-54f77300-aba4-11e9-8ae5-37ed20c41e55.png">

... and it still doesn't work.

It works if the image is copied to a local drive.

lurch:
IIRC this is because the "mapped drive letter" is only visible to user-level processes, and isn't visible to admin-level processes.
The Etcher file-selector GUI runs at a user-level (and so the mapped drive is visible), but Etcher needs to switch to an admin-level process to have enough permissions to write raw disk images (and so the mapped drive "disappears"). The drive letter switching from Z:\ to C:\ looks very strange though! :confused: 

Slightly-related discussion in #1436

zrzka:
[...]

PR #2862, opened 2019-07-22
Don't use wmic's ProviderName if it's empty

zvin:
Change-type: patch


lurch:
Is it worth adding another unit-test for this?

lib/gui/app/os/windows-network-drives.js
```diff
@@ -107,6 +107,8 @@ const getWindowsNetworkDrives = async () => {
       // eslint-disable-next-line no-magic-numbers
       return [ str.slice(0, colonPosition + 1), _.trim(str.slice(colonPosition + 1)) ]
     })
+    // eslint-disable-next-line no-magic-numbers
+    .filter((couple) => couple[1].length > 0)
```
zrzka: I'd like to see a note somewhere why we're doing this. I mean, we know now, but in a couple of days ... This VMware Tools case seems like a special one worth documenting.

thundron: imho in a comment block over that line

lurch:
In the `wmic-output.txt` it probably makes sense to use two different drive letters, rather than `Z:` both times? :wink:

lib/gui/app/os/windows-network-drives.js
zvin: This is not vmwere specific.
The second part is supposed to be `\\<ip address>\folder`.
If it's an empty string we don't use it.

zvin:
@lurch we use `Z:` and the test checks that the non empty value for `Z:`is being used.

lurch:
But surely in any real wmic output, you'll never get the same drive-letter appearing twice? Feel free to ignore me if I'm just misunderstanding what the test is doing.

lurch:
For future reference, this PR relates to #2858

PR #2864, opened 2019-07-25
Allow typescript files

thundron:
Change-type: patch
Changelog-entry: Allow typescript files


.gitattributes
```diff
@@ -1,4 +1,5 @@
 # Javascript files must retain LF line-endings (to keep eslint happy)
+*.ts text eol=lf
```
zvin: We can add `.tsx` files too

thundron:
@balena-ci retest

PR #2875, opened 2019-08-02
Fix auto updater

thundron:


lib/gui/etcher.js
```diff
@@ -21,16 +21,28 @@ const path = require('path')
...
+const { version, packageType } = require('../../package.json')
 /* eslint-disable lodash/prefer-lodash-method */
+/* eslint-disable no-magic-numbers */
 
 const config = settings.getDefaults()
-
 const configUrl = settings.get('configUrl') || 'https://balena.io/etcher/static/config.json'
+const packageTypes = [
+  'rpm',
+  'deb',
```
lurch: Is this correct? I thought that when you install a deb (or rpm) package, it was then up to the system package-manager to keep it up to date, rather than having the package itself directly "auto update"?

thundron: @lurch it is: those are the packages that don't support auto-update

thundron: I should probably choose a better name, like `ignoredPackages` or something along those lines

lurch: Does that mean that the `const packageUpdatable = _.includes(packageTypes, packageType)` logic is wrong?

thundron: @lurch *big face palm*
It was previously a `indexOf ... < 0`, then I mixed things up.
Fixing
*end big face palm*

lib/gui/etcher.js
```diff
@@ -21,16 +21,25 @@ const path = require('path')
...
 const analytics = require('./app/modules/analytics')
 const { getConfig } = require('../shared/utils')
+const { version, packageType } = require('../../package.json')
 /* eslint-disable lodash/prefer-lodash-method */
+/* eslint-disable no-magic-numbers */
 
 const config = settings.getDefaults()
-
 const configUrl = settings.get('configUrl') || 'https://balena.io/etcher/static/config.json'
+const packageTypes = [
```
zvin: `UPDATABLE_PACKAGE_TYPES` would be a better name

package.json
```diff
@@ -50,6 +50,7 @@
     "bluebird": "^3.5.3",
     "bootstrap-sass": "^3.3.6",
     "color": "^2.0.1",
+    "compare-versions": "3.5.0",
```
zvin: Use `semver`, it's already in the `package.json`.

lib/gui/etcher.js
```diff
@@ -40,14 +49,20 @@ const configUrl = settings.get('configUrl') || 'https://balena.io/etcher/static/
 const checkForUpdates = async (interval) => {
   // We use a while loop instead of a setInterval to preserve
   // async execution time between each function call
-  while (true) {
-    try {
-      const release = await autoUpdater.checkForUpdates()
-      if (release.updateInfo.stagingPercentage) {
-        await autoUpdater.downloadUpdate()
+  // eslint-disable-next-line no-unmodified-loop-condition
+  while (packageUpdatable && !packageUpdated) {
```
zvin: `checkForUpdates` should not be called if `packageUpdatable` is `false`

lurch:
Is the info at https://github.com/balena-io/etcher/blob/master/docs/USER-DOCUMENTATION.md#simulate-an-update-alert now out of date?

lib/gui/etcher.js
```diff
@@ -40,14 +49,20 @@ const configUrl = settings.get('configUrl') || 'https://balena.io/etcher/static/
 const checkForUpdates = async (interval) => {
   // We use a while loop instead of a setInterval to preserve
   // async execution time between each function call
-  while (true) {
-    try {
-      const release = await autoUpdater.checkForUpdates()
-      if (release.updateInfo.stagingPercentage) {
-        await autoUpdater.downloadUpdate()
+  // eslint-disable-next-line no-unmodified-loop-condition
```
zvin: This shouldn't be needed now

lib/gui/etcher.js
```diff
@@ -110,7 +125,7 @@ const createMainWindow = () => {
     autoUpdater.on('error', (err) => {
       analytics.logException(err)
     })
-    if (settings.get('updatesEnabled')) {
+    if (packageUpdatable && settings.get('updatesEnabled')) {
```
zvin: We should remove the test for `settings.get('updatesEnabled')` because it might be changed later and is already checked in `checkForUpdates`

lib/gui/etcher.js
```diff
@@ -21,16 +21,25 @@ const path = require('path')
...
 const analytics = require('./app/modules/analytics')
 const { getConfig } = require('../shared/utils')
+const { version, packageType } = require('../../package.json')
 /* eslint-disable lodash/prefer-lodash-method */
+/* eslint-disable no-magic-numbers */
 
 const config = settings.getDefaults()
-
 const configUrl = settings.get('configUrl') || 'https://balena.io/etcher/static/config.json'
+const UPDATABLE_PACKAGE_TYPES = [
```
lurch: Just out of curiosity, what's the rationale for some `const` variables being camelCase and others being UPPER_CASE ?

lib/gui/etcher.js
```diff
@@ -40,14 +49,19 @@ const configUrl = settings.get('configUrl') || 'https://balena.io/etcher/static/
...
-      const release = await autoUpdater.checkForUpdates()
-      if (release.updateInfo.stagingPercentage) {
-        await autoUpdater.downloadUpdate()
+  while (!packageUpdated) {
+    if (settings.get('updatesEnabled')) {
+      try {
+        const release = await autoUpdater.checkForUpdates()
+        const isOutdated = semver.compare(release.updateInfo.version, version) > 0
+        const shouldUpdate = parseInt(release.updateInfo.stagingPercentage, 10) > 0
+        if (shouldUpdate && isOutdated && !packageUpdated) {
```
zvin: No need to check for `!packageUpdated` here, it's already `false` if we're in this `while` loop.

lib/gui/etcher.js
```diff
@@ -40,14 +49,19 @@ const configUrl = settings.get('configUrl') || 'https://balena.io/etcher/static/
...
-    try {
-      const release = await autoUpdater.checkForUpdates()
-      if (release.updateInfo.stagingPercentage) {
-        await autoUpdater.downloadUpdate()
+  while (!packageUpdated) {
+    if (settings.get('updatesEnabled')) {
+      try {
+        const release = await autoUpdater.checkForUpdates()
+        const isOutdated = semver.compare(release.updateInfo.version, version) > 0
+        const shouldUpdate = parseInt(release.updateInfo.stagingPercentage, 10) > 0
```
lurch: Are you sure this is necessary? Nothing I can find at https://www.electron.build/auto-update#staged-rollouts suggests you manually need to check the stagingPercentage value?

thundron: @lurch It's because we manually check and then download, so it's an extra safety step before triggering the download for those connected with a data plan. One thing I didn't check is if `release.updateInfo.version` is unchanged in case no release is available (i.e. a newer release's `stagingPercentage` is 0), but I will

lib/gui/etcher.js
thundron: Just a quick change without realising it's not ane env variable, going back to camel case in the next push

PR #2880, opened 2019-08-12
Move lib/shared to lib/gui/app/modules

thundron:
Change-type: patch
Changelog-entry: Move lib/shared to lib/gui/app/modules


lurch:
Looking at https://github.com/balena-io/etcher/tree/9ef90008ae21588c1fac9954e334fa2f31a1b516/lib would it be worth moving `lib/gui/*` into `lib/` :grey_question:

thundron:
> Looking at https://github.com/balena-io/etcher/tree/9ef90008ae21588c1fac9954e334fa2f31a1b516/lib would it be worth moving `lib/gui/*` into `lib/` ❔

Eh, maybe. I'm not sure about having `start.js`, `etcher.js` and `menu.js` in the same folder

PR #2896, opened 2019-09-05
Update style of progress bar

thundron:
Change-type: patch
Changelog-entry: Updated style of progress bar

### DIFF
--- a/lib/gui/app/modules/progress-status.js
+++ b/lib/gui/app/modules/progress-status.js
@@ -24,6 +24,7 @@ const units = require('../../../shared/units')
  * @summary Make the progress status subtitle string
  *
  * @param {Object} state - flashing metadata
+ * @param {Boolean} plain - with or without html additions
  *
  * @returns {String}
  *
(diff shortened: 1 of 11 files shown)


lib/gui/app/modules/progress-status.js
```diff
@@ -61,11 +62,13 @@ exports.fromFlashState = (state) => {
...
     // eslint-disable-next-line no-eq-null
     if (state.percentage != null) {
-      return `${state.percentage}% Flashing`
+      const txt = (plain) ? `${state.percentage}%` : `<strong>${state.percentage}%</strong>`
+      return `Flashing... ${txt}`
     }
     return `${units.bytesToClosestUnit(state.position)} flashed`
   } else if (isValidating) {
-    return `${state.percentage}% Validating`
+    const txt = (plain) ? `${state.percentage}%` : `<strong>${state.percentage}%</strong>`
```
zvin: We should avoid html tags in strings.
Looks like we need an `FlashProgressLabel` component.

tests/gui/modules/progress-status.spec.js
```diff
@@ -22,7 +22,7 @@ describe('Browser: progressStatus', function () {
     })
 
     it('should report 0% if percentage == 0 but speed != 0', function () {
-      m.chai.expect(progressStatus.fromFlashState(this.state)).to.equal('0% Flashing')
+      m.chai.expect(progressStatus.fromFlashState(this.state, true)).to.equal('Flashing... 0%')
```
lurch: Is it worth having a test for `.fromFlashState(this.state, false)` ? Or does `tests/gui/pages/main.spec.js` provide sufficient coverage for that?

PR #2906, opened 2019-09-16
Fix entrypoint when options are passed to electron

zvin:
Change-type: patch
Changelog-entry: Fix entrypoint when options are passed to electron


thundron:
This might also fix https://github.com/balena-io/etcher/pull/2880 , be sure to test the Windows package since that's where I had most issues

Issue #2707, opened 2019-03-26
Small error on website


lurch:
I just noticed that at the bottom of https://www.balena.io/etcher/ it says "Etcher GUI is based on the Electron framework, which only supports macOS 10.9 and newer versions." but https://github.com/balena-io/etcher/blob/master/docs/USER-DOCUMENTATION.md#running-in-older-macos-versions says "Etcher GUI is based on the Electron framework, which only supports macOS 10.10 (Yosemite) and newer versions.".

Related to https://github.com/balena-io/etcher/commit/b6ad6e0a85a84b081327d5adb49d4b2e3164bd8d

lurch:
According to https://github.com/balena-io/etcher/pull/2964#discussion_r343234253 I've got it the wrong way around; the website is correct but the `USER-DOCUMENTATION.md` is wrong.

shawaj:
Closing this as it appears to be fixed on both the website and the user documentation now.

lurch:
Since I opened this issue, Etcher has migrated to a newer version of Electron, but good to see everything has been ironed out now :slightly_smiling_face:

shawaj:
Just noticed it is actually still wrong in the [FAQ file](https://github.com/balena-io/etcher/blob/master/FAQ.md) - will sort a PR

Issue #2519, opened 2018-10-25
diskpart command failed


martijintro:
- **Etcher version: Current** 
- **Operating system and architecture: Win10** 
- **Image flashed: 32GB img** 
- **Do you see any meaningful error information in the DevTools?** 

![image](https://user-images.githubusercontent.com/42765391/47509904-5430eb00-d877-11e8-8185-b14268575df4.png)

cyphermaster:
+1
i have the same issue after the windows 10 october update

martijintro:
any beta version?

martijintro:
no answer, so not interested for an adequate solution so it seems?!?

thundron:
@martijintro Sorry for the late reply, this is actually a known issue that goes all the way back to node itself, which has been updated and is included in the latest Electron versions that we're bumping as seen in this PR https://github.com/balena-io/etcher/pull/2578 which is in the works

thundron:
@martijintro In the meantime, formatting the disk beforehand *may* help

docholiday:
@thundron Just for information - I'm getting the same error in etcher 1.4.8 with my internal Ricoh SD Card Reader. Luckily I could solve it by using an external reader by Kingston...

martijintro:
"@martijintro In the meantime, formatting the disk beforehand may help as seen here"

I don't format the disk before using your tool; i just use "HDD Low Level Format Tool" to quick wipe the file system from the drive. (SD card with sandisk USB 3.0 reader)

cyoung:
This is still an issue in 1.4.9.

When looking at the disk in "Computer Management", Windows states:

"Offline (The disk is offline because it has a signature collision with another disk that is online)"

After using "Computer Management" to remove the partitions, then disconnecting and re-connecting the USB-to-microUSB adapter, Etcher was able to flash the image on the microSD card.

martijintro:
ok thx, will try it as soon as possible...

jimmyeao:
Same issue here, deleting the second partition in disk management then connecting to a different USB port solves the issue. Back to win32diskimager until this is fixed..

lurch:
> "Offline (The disk is offline because it has a signature collision with another disk that is online)"

Signature collision was also mentioned (with a workaround?) in #2967

thundron:
@martijintro Hey there, wondering if this issue is still happening and if you tried with the latest version, since you closed it. We since upgraded Electron, so it might be worth a shot if you'd like to help us figure it out

[...]

PR #2932, opened 2019-10-14
Electron6

zvin:


package.json
```diff
@@ -9,7 +9,7 @@
     "sleepDays": 7,
     "semverRange": "<2.0.0"
   },
-  "main": "lib/start.js",
+  "main": "generated/etcher.js",
```
roman-mazur: would you elaborate on how the entry point works atm?

I'm a bit confused by the fact, we remove usage of lib/start here, but keep using it in webpack config and update it to use `gui/etcher` instead of `generated/etcher`.

zvin: Etcher has 2 "main" functions:
 * one launches the GUI: `lib/gui/etcher.js`;
 * the other launches the child writer process (when the env var `ELECTRON_RUN_AS_NODE` is set): `lib/gui/modules/child-writer.js`.

`lib/start.js` is the entrypoint that decides which one of the above needs to run (based on `ELECTRON_RUN_AS_NODE` being set or not)

Before this PR, the entrypoint in `package.json` was `lib/start.js` and it started either `generated/etcher.js` or `lib/gui/modules/child-writer.js` which is weird because one is webpacked and the other not). This means the same code may be ran from different files.

After this PR, the only entrypoint is `generated/etcher.js`, `lib/start.js` and all the things it requires (except node_modules) are webpacked.


In the long term, we shouldn't use webpack at all for the main process (we should only use it for the js running in the renderer process). But for now it is useful as we have mixed js & ts code (and also some [non standard stuff](https://github.com/balena-io/etcher/blob/master/webpack.config.js#L43)).

zvin:
@balena-ci retest

package.json
roman-mazur: Got it, thanks for the explanation.

roman-mazur:
@zvin to confirm, should this help with #2519 ?

It looks good to me in general, but I believe @thundron or @jviotti would give more adequate feedback.

zvin:
@roman-mazur I don't think this will help with #2519.
It is still not merged because it has issues packaging lzma_native on windows.

lurch:
Now that this has been merged, does it affect the minimum-version supported operating systems?
(see also #2707 )

PR #2935, opened 2019-10-16
correct two nomenclature errors, redux

RichMorin:
correct two nomenclature errors

PC keyboards have "Alt" keys; Mac keyboards have "Opt" keys. Although it's possible to use a PC keyboard on a Mac, it's unusual. In any case, all of the macOS (not "Mac OS" for some years now) documentation refers to the "Opt" key.

Change-type: minor
Changelog-entry: correct two nomenclature errors


zvin:
fixed in #3140 , thanks @RichMorin

PR #2939, opened 2019-10-18
Make Etcher work on macOS Catalina

thundron:
* "notarize" macOS builds;
 * use `sudo` instead of `sudo-prompt` to elevate writer process;
 * update electron-builder


afterSignHook.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @example
+ * await notarize({
+ *   // electron-builer context
+ * })
+ */
+const notarizing = async (context) => {
+  const { electronPlatformName, appOutDir } = context
+  if (electronPlatformName !== 'darwin') {
+    return Promise.resolve()
```
zvin: You can simply `return` here, no need for `Promise.resolve()`, this is already an async function.

afterSignHook.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  const appName = context.packager.appInfo.productFilename
+  const appleId = 'accounts+apple@balena.io'
+
+  return notarize({
```
zvin: `await` would make more sense than `return` here.

afterSignHook.js
thundron: The linter complains about the function returning void then a Promise otherwise

afterSignHook.js
thundron: IIRC returning a Promise in an async function is the same as `return await`ing a Promise

zvin: `await` expresses the intent, which is to wait for `notarize` to finish. It doesn't return anything anyway.

afterSignHook.js
zvin: The linter is old and wrong, disable it for this file.

afterSignHook.js
zvin: Also it makes clear why this function is `async`. It shouldn't be async if there's no `await` in it.

lib/shared/catalina-sudo/sudo-askpass.osascript.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env osascript -l JavaScript
+
+/* eslint-disable */
+
+ObjC.import('stdlib')
+
+const app = Application.currentApplication()
+app.includeStandardAdditions = true
+
+const result = app.displayDialog('Please enter your sudo password.', {
```
lurch: I don't have a Mac myself, but will the average Mac user know what "your sudo password" means?
I wonder if it might make sense to copy the text used in the screenshots on https://github.com/jorangreef/sudo-prompt ?
:man_shrugging:

package.json
```diff
@@ -52,7 +52,6 @@
     "color": "^2.0.1",
     "d3": "^4.13.0",
     "debug": "^3.1.0",
-    "electron-is-running-in-asar": "^1.0.0",
```
lurch: Should the shrinkwrap file have updated too? :man_shrugging:

zvin: no

zvin: It is also required by resin-corvus.

PR #2966, opened 2019-11-11
Revamp finish

thundron:
- [ ] Need to create new online success-banner to match new layout


lib/gui/app/components/finish/finish.jsx
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright 2019 resin.io
```
lurch: resin.io or balena.io ?

lib/gui/app/components/finish/finish.jsx
```diff
@@ -0,0 +1,124 @@
...
+                <span onClick={() => osOpenExternal('https://resin.io?ref=etcher_success')}>
+                  <SvgIcon paths={[ '../../assets/balena.svg' ]}
+                    width='auto'
+                    height='20px'>
+                  </SvgIcon>
+                </span>
+                <div className="section-footer">
+                  <span className="caption caption-small footer-right"
+                    manifest-bind="version"
+                    onClick={() => osOpenExternal('https://github.com/resin-io/etcher/blob/master/CHANGELOG.md')}></span>
```
lurch: I guess the URLs here should also be balena rather than resin?

lib/gui/app/components/finish/finish.tsx
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright 2019 resin.io
```
lurch: resin.io -> balena.io ?

thundron: Err, yes, again. Damn copyright!

PR #2982, opened 2019-12-02
Refactor image selection to remove angular

LucianBuzzo:
Change-type: patch
Signed-off-by: Lucian <lucian.buzzo@gmail.com>


lib/gui/app/components/image-selector/index.js
```diff
@@ -24,11 +26,10 @@ const angular = require('angular')
 const { react2angular } = require('react2angular')
 
 const MODULE_NAME = 'Etcher.Components.ImageSelector'
-const SelectImageButton = angular.module(MODULE_NAME, [])
+const ImageSelector = angular.module(MODULE_NAME, [])
 
-SelectImageButton.component(
+ImageSelector.component(
   'imageSelector',
   react2angular(require('./image-selector.jsx'))
```
thundron: NIT: We can get rid of `load-ng-injector` in the utils by adding a third argument to the module definition like so:
```javascript
react2angular(require('/path/to/module/'), [], [ 'WarningModalService', 'FileSelectorService' ])
```

PR #2983, opened 2019-12-03
chore: move flash step to React

sradevski:
Changelog-entry: chore: move flash step to React
Change-type: patch
Signed-off-by: Stevche Radevski <stevche@balena.io>


zvin:
@sradevski would you mind converting this to typescript?

sradevski:
@zvin yesterday on a call we talked about moving it to React + javascript first, and they convert it to typescript, so I would rather do it in a separate PR.

PR #2985, opened 2019-12-03
Refactor image-selection

LucianBuzzo:
Change-type: patch
Changelog-entry: Use React instead of Angular for image selection
Signed-off-by: Lucian <lucian.buzzo@gmail.com>


lib/gui/app/components/image-selector/image-selector.jsx
```diff
@@ -32,67 +42,316 @@ const {
...
+ * @summary Get the basename of the selected image
+ * @function
+ * @public
+ *
+ * @returns {String} basename of the selected image
+ *
+ * @example
+ * const imageBasename = getImageBasename();
+ */
+const getImageBasename = () => {
```
LucianBuzzo: remove this function

PR #2988, opened 2019-12-03
Convert the drive selection step to React

thgreasi:
Change-type: patch
Changelog-entry: Convert the drive selection step to React
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>


lib/gui/app/pages/main/DriveSelector.jsx
```diff
@@ -0,0 +1,243 @@
...
+ * @summary Open drive selector
+ * @function
+ * @public
+ * @param {Object} DriveSelectorService - drive selector service
+ *
+ * @example
+ * openDriveSelector(DriveSelectorService);
+ */
+const openDriveSelector = (DriveSelectorService) => {
+  DriveSelectorService.open().then((drive) => {
```
zvin: It would have been nice to convert all `.then`s to `async` functions but we can leave it for later.

thgreasi: I actually just de-indented all the methods while converting the service to a component and didn't check exactly what each method was doing.

thgreasi: Updated @zvin

PR #2993, opened 2019-12-04
Remove experimental file picker

zvin:
Change-type: patch


zvin:
@sradevski it was a react file picker that replaced the operating system file picker when enabled.
It was moved here https://github.com/balena-io-modules/balena-electronjs/tree/master/src/ui/file-selector

PR #3001, opened 2019-12-06
Move a couple of files to typescript and remove unnecessary $timeout

sradevski:


lib/gui/app/pages/main/DriveSelector.tsx
```diff
@@ -0,0 +1,163 @@
...
+import * as SvgIcon from '../../components/svg-icon/svg-icon.jsx';
+import * as selectionState from '../../models/selection-state';
+import * as settings from '../../models/settings';
+import * as store from '../../models/store';
+import * as analytics from '../../modules/analytics';
+import * as exceptionReporter from '../../modules/exception-reporter';
+import * as driveConstraints from '../../../../shared/drive-constraints';
+import * as utils from '../../../../shared/utils';
+
+const StepBorder = styled.div<{
```
sradevski: This is the only thing that's new in order to fix the typing issues (we were passing `disabled` to a div, which doesn't exist).

lib/gui/app/pages/main/DriveSelector.tsx
```diff
@@ -14,183 +14,150 @@
...
- * @function
- * @public
- *
- * @returns {String} - 'list' of drives separated by newlines
- *
- * @example
- * console.log(getDriveListLabel())
- * > 'My Drive (/dev/disk1)\nMy Other Drive (/dev/disk2)'
- */
+'use strict';
```
zvin: not needed in typescript

PR #3003, opened 2019-12-06
Revamp finish for remove angular

LucianBuzzo:


lib/gui/app/styled-components.tsx
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright 2018 resin.io
```
lurch: Was this supposed to be "2019 balena.io" ?

PR #3009, opened 2019-12-10
Refactor drive selector and confirm modal to React

thundron:
Change-type: patch
Changelog-entry: Refactor drive selector and confirm modal to React


lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,325 @@
...
+   * @param {Object} drive - drive
+   * @returns {Promise}
+   *
+   * @example
+   * shouldChangeDriveSelectionState(drive)
+   *    .then((shouldChangeDriveSelectionState) => {
+   *        if (shouldChangeDriveSelectionState) doSomething();
+   *    });
+   */
+  const shouldChangeDriveSelectionState = async (drive) => {
```
zvin: No need for `async` as there's no `await` in this function.

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,325 @@
...
+              setConfirmModal({ open: false })
+            }
+          } catch (error) {
+            analytics.logException(error)
+          }
+        }
+      })
+    }
+
+    return Bluebird.resolve()
```
zvin: No need to return a Bluebird here

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,325 @@
...
+   *
+   * @example
+   * selectDriveAndClose({
+   *   device: '/dev/disk2',
+   *   size: 999999999,
+   *   name: 'Cruzer USB drive'
+   * });
+   */
+  const selectDriveAndClose = (drive, closeFn) => {
+    return shouldChangeDriveSelectionState(drive).then((canChangeDriveSelectionState) => {
```
zvin: Please use `await` instead of `.then`

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,325 @@
...
+   * @param {Object} drive - drive
+   * @param {Object} evt - event
+   *
+   * @example
+   * <div tabindex="1" onKeyPress="keyboardToggleDrive(drive, evt)">
+   *   Tab-select me and press enter or space!
+   * </div>
+   */
+  const keyboardToggleDrive = (drive, evt) => {
+    console.log(evt.keyCode)
```
zvin: leftover `console.log`

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,325 @@
...
+    >
+      <div>
+        <ul style={{
+          height: '250px',
+          overflowX: 'hidden',
+          overflowY: 'auto',
+          padding: '0'
+        }}>
+          {_.map(getDrives(), (drive, index) => {
+            drive.link = `https://blabla${index}.com`
```
zvin: blabla?

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,320 @@
...
+   *   device: '/dev/disk2',
+   *   size: 999999999,
+   *   name: 'Cruzer USB drive'
+   * });
+   */
+  const selectDriveAndClose = async (drive, closeFn) => {
+    const canChangeDriveSelectionState = await shouldChangeDriveSelectionState(drive)
+
+    if (canChangeDriveSelectionState) {
+      selectionState.selectDrive(availableDrives.device)
```
zvin: This probably should be `selectionState.selectDrive(drive.device)`.

Do we really want this behaviour ?
Double adds the drive to the selection even if it was already selected.

thundron: > This probably should be

Correct, fixing

lib/gui/app/components/confirm-modal/ConfirmModal.jsx
```diff
@@ -16,19 +16,37 @@
...
+    rejectionLabel,
+    title,
+    message,
+    cancel,
+    done
+  } = options
 
-const MODULE_NAME = 'Etcher.Components.TargetSelector'
-const SelectTargetButton = angular.module(MODULE_NAME, [])
+  const modalProps = {
```
sradevski: Let's apply this directly to the modal, instead of having it as a separate object.

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,320 @@
...
+const store = require('../../models/store')
+const analytics = require('../../modules/analytics')
+const availableDrives = require('../../models/available-drives')
+const selectionState = require('../../models/selection-state')
+const { bytesToClosestUnit } = require('../../../../shared/units')
+const utils = require('../../../../shared/utils')
+const { open: openExternal } = require('../../os/open-external/services/open-external')
+const ConfirmModal = require('../confirm-modal/ConfirmModal.jsx')
+
+const DriveSelectorModal = (props) => {
```
sradevski:
```suggestion
const DriveSelectorModal = ({close}) => {
```
And remove the line below.

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,320 @@
...
+    if (_.includes([ ENTER, SPACE ], evt.keyCode)) {
+      toggleDrive(drive)
+    }
+  }
+
+  const hasStatus = hasListDriveImageCompatibilityStatus(selectionState.getSelectedDrives(), selectionState.getImage())
+
+  return (
+    <Modal
+      className='modal-drive-selector-modal'
```
sradevski: Why is the class still necessary?

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,320 @@
...
+      title='Select a Drive'
+      done={close}
+      action='Continue'
+      style={{
```
sradevski: Let's use the default padding if possible

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
```diff
@@ -0,0 +1,320 @@
...
+   */
+  const installMissingDrivers = (drive) => {
+    if (drive.link) {
+      analytics.logEvent('Open driver link modal', {
+        url: drive.link,
+        applicationSessionUuid: store.getState().toJS().applicationSessionUuid,
+        flashingWorkflowUuid: store.getState().toJS().flashingWorkflowUuid
+      })
+
+      setConfirmModal({
```
sradevski: Since these are the only possible settings for the Confirm modal in this component, how about specifying all of the props directly on the modal in JSX, and only having `[showConfirmModal, setShowConfirmModal] =React.useState(false)` in state instead?

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
thundron: To preserve style of the drives list

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
thundron: Definitely, after refactoring the confirm modal a bit it makes more sense, it's still useful to have a `confirmModal` object though so we just spread it with (e.g.) `{...confirmModal.options}` and avoid having multiple variables around

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
thundron: It adds a padding to the left which makes the list uncentered

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
sradevski: If there were multiple places where this was used, it does make sense, but it doesn't really make sense as it is now. Even then, I would rather have 2 modals with different props, rather than using an object, as it is easier to follow what is logic and what is presentation

thundron: The thing is that the values are still dynamic so we would either need to set/unset each of them every time, as opposed to just having the simplified version of the confirm-modal (which is gone with these changes) like:
```javascript
setConfirmModal({
        open: true,
        options: {
          width: 400,
          title: drive.linkTitle,
          cancel: () => setConfirmModal({ open: false }),
          done: async (shouldContinue) => {
            try {
              if (shouldContinue) {
                openExternal(drive.link)
              } else {
                setConfirmModal({ open: false })
              }
            } catch (error) {
              analytics.logException(error)
            }
          },
          action: 'Yes, continue',
          cancelButtonProps: {
            children: 'Cancel'
          },
          children: drive.linkMessage || `Etcher will open ${drive.link} in your browser`
        }
      })
```

lib/gui/app/components/confirm-modal/ConfirmModal.jsx
thundron: Deleted with the latest changes

lib/gui/app/components/drive-selector/DriveSelectorModal.jsx
thundron: They're not 2+ modals, it's the same modal with different parameters. While not all of them are dynamic (e.g. action, width, cancel and cancelButtonProps are not), it still leaves us with 3 variables just for a modal... It's not the end of the world of course, but I'd rather have just an object that we can possibly reuse in the future as it can easily be extracted

PR #3026, opened 2020-01-02
Remove remaining angular and convert everything to typescript

zvin:


lurch:
I guess the PR needs to be re-titled "Remove remaining angular and convert lots of modules to typescript" :wink:

lib/shared/permissions.ts
```diff
@@ -0,0 +1,206 @@
...
+	lines.push(..._.map(environment, setEnvVarFn));
+	lines.push([command, ...argv].map(escapeFn).join(' '));
+	return lines.join(os.EOL);
+}
+
+async function elevateScriptWindows(
+	path: string,
+): Promise<{ cancelled: boolean }> {
+	// 'elevator' imported here as it only exists on windows
+	// TODO: replace this with sudo-prompt once https://github.com/jorangreef/sudo-prompt/issues/96 is fixed
```
lurch: https://github.com/jorangreef/sudo-prompt/issues/96 claims to be fixed now?

typings/resin-corvus/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'sudo-prompt';
```
lurch: Huh, `typings/resin-corvus/index.d.ts` and `typings/sudo-prompt/index.d.ts` **both** say
```typescript
declare module 'sudo-prompt';
```
??? :confused:

webpack.config.js
```diff
@@ -23,7 +23,6 @@ const nodeExternals = require('webpack-node-externals')
 const commonConfig = {
   mode: 'production',
   optimization: {
-    // Minification breaks angular.
```
lurch: Does the removal of this comment imply that minification could be turned on? Or is that only relevant to web-apps and not applicable to desktop-apps?

zvin: It was disabled in the first place because it was breaking angular.
Now that we no longer have angular it could be enabled.
But we'll keep it disabled for easier debugging.

lib/shared/permissions.ts
zvin: @lurch, thanks, noted, that will be for another PR.

PR #3046, opened 2020-01-31
Etcher pro leds

zvin:


lib/gui/app/components/image-selector/image-selector.tsx
```diff
@@ -185,7 +185,6 @@ export class ImageSelector extends React.Component<
 						title,
 					},
 				});
-				return;
```
thundron: This is needed to disable selecting unsupported images (the check above)

zvin: @thundron , this is a warning, the images are still supported.
This was introduces recently & is a bug, see https://github.com/balena-io/etcher/issues/3053

Issue #3056, opened 2020-02-04
macOS: version. 1.5.74 app not initializing 


RandyMcMillan:
I have a fresh install of Mojave 10.14.6 (18G103) running.
[-1580857850329.log](https://github.com/balena-io/etcher/files/4156610/-1580857850329.log)

/Applications/balenaEtcher.app/Contents/Resources/app/node_modules/raven-js/dist/raven.js:58 {"stack":"TypeError: Cannot convert undefined or null to object\n    at Function.entries (<anonymous>)\n    at Object.init (file:///Applications/balenaEtcher.app/Contents/Resources/app/generated/gui.js:3951:49)\n    at main (file:///Applications/balenaEtcher.app/Contents/Resources/app/generated/gui.js:3633:12)","message":"Cannot convert undefined or null to object"}
console.<computed> @ /Applications/balenaEtcher.app/Contents/Resources/app/node_modules/raven-js/dist/raven.js:58

![Screen Shot 2020-02-04 at 3 06 35 PM](https://user-images.githubusercontent.com/152159/73795678-100de500-4760-11ea-81e2-d30d74f620ae.png)
![Screen Shot 2020-02-04 at 3 06 14 PM](https://user-images.githubusercontent.com/152159/73795708-2156f180-4760-11ea-892b-0f1c037383ff.png)

lurch:
Probably a duplicate of #3057 ?

Issue #3060, opened 2020-02-05
Etcher can not running well(version  1.5.74)


a18214264850:
- **Etcher version:** 
- **Operating system and architecture:** 
- **Image flashed:** 
- **Do you see any meaningful error information in the DevTools?** 

in windows version,i tried in system version 1909/1809,but when i open it(no matter whether i turn off uac)it doesn't display bottom at all,but it works well in 1.5.73.

lurch:
See #3062

Issue #3058, opened 2020-02-04
Etcher not running


kovkus:
- **Etcher version:** 1.5.74
- **Operating system and architecture:** MacOS 10.14.6
- **Do you see any meaningful error information in the DevTools?** 
|  ___| |     | |
| |__ | |_ ___| |__   ___ _ __
|  __|| __/ __| '_ \ / _ \ '__|
| |___| || (__| | | |  __/ |
\____/ \__\___|_| |_|\___|_|

Version = 1.5.74, Type = dmg
/Applications/balenaEtcher.app/Contents/Resources/app/node_modules/raven-js/dist/raven.js:58 Wed Feb 05 2020 00:54:58 GMT+0100 (Central European Standard Time) Application start ({"packageType":"dmg","version":"1.5.74","applicationSessionUuid":"b03426ef-29a8-4641-b9b3-b787387bfca3","sample":0.1})
/Applications/balenaEtcher.app/Contents/Resources/app/node_modules/raven-js/dist/raven.js:58 {"stack":"TypeError: Cannot convert undefined or null to object\n    at Function.entries (<anonymous>)\n    at Object.init (file:///Applications/balenaEtcher.app/Contents/Resources/app/generated/gui.js:3951:49)\n    at main (file:///Applications/balenaEtcher.app/Contents/Resources/app/generated/gui.js:3633:12)","message":"Cannot convert undefined or null to object"}

lurch:
Probably a duplicate of #3057 ?

Issue #3057, opened 2020-02-04
Black screen when opening Etcher


flightless-katie:
- **Etcher version:** 1.5.74
- **Operating system and architecture:** Mac OS Mojave 10.14.6
- **Image flashed:** N/A
- **Do you see any meaningful error information in the DevTools?** Yes, see below

After tracing the error as the program is running, I found that in leds.ts this code seems to be throwing: (this is on startup, first time trying Etcher)
```
const ledsMapping = settings.get('ledsMapping');
for (const [drivePath, ledsPaths] of Object.entries(ledsMapping)) { // <-- this line throws
  leds.set('/dev/disk/by-path/' + drivePath, new RGBLed(ledsPaths));
}
```
settings.get(...) is returning undefined, so when the for loop tries to use Object.entries(ledsMapping), it is trying to get the entries of undefined, which causes an error.

If in devtools I set ledsMapping to an empty object `{}` the program continues to the opening menu to select image, etc. It's doing a flash operation now with that manual hack. :)

flightless-katie:
PS, I am not familiar with your codebase or any conventions you may have, but this could be a potential fix, maybe?
```
const ledsMapping = settings.get('ledsMapping') || {};
```

barrypress:
This did work, but not being familiar with the codebase or the dev tools, I had to fumble around a lot before I finally set enough breakpoints (discovering cmd-R was a biggie, too) and pasted in the patch in enough files. I believe I had to do at least gui.js and leds.ts, but not altogether sure which one worked.

However, the fix is not persistent - closing and reopening the pap results in the same gray useless window.

Nightwing531:
Same problem here. I would really like this to get fixed

DickSwart:
Same here

PR #3062, opened 2020-02-05
Initialize leds object map

thundron:
Fixes: #3056 #3057 #3058 #3060
Change-type: patch
Changelog-entry: Initialize leds object map


thundron:
Thanks to @zomars for the contribution, replacing his PR only because it's an important fix to roll out asap

PR #3069, opened 2020-02-17
Update GPG key acquisition to work on Ubuntu 18.04

Angeall:
The former command failed systematically, the keyserver did not respond.
This modified command did the trick


djvar94:
Same problem here on ALL of my machines. Please fix. It really is so annoying

zvin:
The command works fine.
We can not recommend users to use hkp instead of hkps for getting keys.

PR #3071, opened 2020-02-17
Flash from url

thundron:


nazrhom:
@balena-ci retest

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +77,158 @@ function getState() {
...
+const URLSelector = ({ done }: { done: (imageURL: string) => void }) => {
+	const [imageURL, setImageURL] = React.useState('');
+	const [recentImages, setRecentImages]: [
+		string[],
+		(value: React.SetStateAction<string[]>) => void,
+	] = React.useState([]);
+	React.useEffect(() => {
+		const fetchRecentUrlImages = async () => {
+			try {
+				const recentUrlImages: any = await jsonStorage.get('recentUrlImages');
```
zvin, line 89: Please avoid `any`.

zvin: Also provide a default empty list or the code below will break.

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +77,158 @@ function getState() {
...
+				console.error(err);
+			}
+		};
+		fetchRecentUrlImages();
+	}, []);
+	return (
+		<Modal
+			done={async () => {
+				const sanitizedRecentUrls = _.uniq(
+					_.reject([...recentImages, imageURL], _.isEmpty),
```
zvin, line 101: This will not work if `recentImages` is `undefined`.

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +77,158 @@ function getState() {
...
+	const [recentImages, setRecentImages]: [
+		string[],
+		(value: React.SetStateAction<string[]>) => void,
+	] = React.useState([]);
+	React.useEffect(() => {
+		const fetchRecentUrlImages = async () => {
+			try {
+				const recentUrlImages: any = await jsonStorage.get('recentUrlImages');
+				setRecentImages(recentUrlImages);
+			} catch (err) {
```
zvin, line 91: Don't use a try catch here, this has to work.

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +77,158 @@ function getState() {
...
+};
+
+interface Flow {
+	icon: any;
+	onClick: (evt: MouseEvent) => void;
+	label: string;
+}
+
+const FlowSelector = styled(
+	({ flow, ...props }: { flow: Flow; props?: any }) => {
```
zvin, line 154: Please avoid `any`

lib/gui/modules/child-writer.ts
```diff
@@ -131,12 +131,23 @@ async function writeAndValidate(
 	return result;
 }
 
+type Source =
```
zvin, line 134: `Source` and `SourceOptions` are defined several times.

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +77,158 @@ function getState() {
...
+		}
+	}
+`;
+
+type Source = typeof sourceDestination.File | typeof sourceDestination.Http;
+
+interface SourceOptions {
+	imagePath: string;
+	SourceType: Source;
+	sourceParams?: any[];
```
zvin, line 179: Please avoid `any`

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +77,158 @@ function getState() {
...
-export class ImageSelector extends React.Component<
-	ImageSelectorProps,
-	ImageSelectorState
+export class SourceSelector extends React.Component<
+	SourceSelectorProps,
+	SourceSelectorState
 > {
 	private unsubscribe: () => void;
+	private afterSelected: SourceSelectorProps['afterSelected'];
+	public flows: Flow[];
```
zvin, line 202: No need for this

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -363,15 +518,9 @@ export class ImageSelector extends React.Component<
...
-								<StepButton onClick={this.openImageSelector}>
-									Select image
-								</StepButton>
-								<Footer>
-									{mainSupportedExtensions.join(', ')}, and{' '}
-									<Underline tooltip={extraSupportedExtensions.join(', ')}>
-										many more
-									</Underline>
-								</Footer>
+								{_.map(this.flows, flow => {
```
zvin, line 521: No need to define `this.flows` and use `_.map`, just use `<FlowSelector />` two times.

lib/gui/app/components/svg-icon/svg-icon.tsx
```diff
@@ -95,31 +95,36 @@ export class SVGIcon extends React.Component<SVGIconProps> {
...
-				// This means the path to the icon should be
-				// relative to *this directory*.
-				// TODO: There might be a way to compute the path
-				// relatively to the `index.html`.
-				const imagePath = path.join(baseDirectory, 'assets', relativePath);
-
-				const contents = _.attempt(() => {
-					return fs.readFileSync(imagePath, {
-						encoding: 'utf8',
+				if (_.startsWith(relativePath, 'http')) {
```
zvin, line 98: Why this changes? SVGIcon is not used in this pull request.

lib/gui/app/components/svg-icon/svg-icon.tsx
```diff
@@ -95,31 +95,36 @@ export class SVGIcon extends React.Component<SVGIconProps> {
...
+					svgData = relativePath;
+					return true;
+				} else {
```
zvin, line 101: No need for this `else`, the `if` has a `return`.

lib/gui/app/pages/main/Flash.tsx
```diff
@@ -71,7 +71,10 @@ const getErrorMessageFromCode = (errorCode: string) => {
 	return '';
 };
 
-async function flashImageToDrive(goToSuccess: () => void): Promise<string> {
+async function flashImageToDrive(
+	goToSuccess: () => void,
+	source: any,
```
zvin, line 76: Please avoid `any`

lib/gui/app/pages/main/Flash.tsx
```diff
@@ -160,7 +163,11 @@ const formatSeconds = (totalSeconds: number) => {
 	return `${minutes}m${seconds}s`;
 };
 
-export const Flash = ({ shouldFlashStepBeDisabled, goToSuccess }: any) => {
+export const Flash = ({
+	shouldFlashStepBeDisabled,
+	goToSuccess,
+	source,
+}: any) => {
```
zvin, line 170: Please avoid `any`

lib/gui/app/pages/main/Flash.tsx
```diff
@@ -299,7 +306,11 @@ export const Flash = ({ shouldFlashStepBeDisabled, goToSuccess }: any) => {
 					done={() => handleFlashErrorResponse(true)}
 					action={'Retry'}
 				>
-					<Txt>{errorMessage}</Txt>
+					<Txt>
+						{_.map(errorMessage.split('\n'), (message, key) => (
+							<p key={`${message}-${key}`}>{message}</p>
```
zvin, line 311: `key={key}` should be enough.

lib/gui/app/styled-components.tsx
```diff
@@ -55,30 +57,45 @@ export const ThemedProvider = (props: any) => (
 	<Provider theme={theme} {...props}></Provider>
 );
 
-export const BaseButton = styled(Button)`
-	height: 48px;
+export const Icon = styled(props => <Button plain {...props} />)`
```
zvin, line 60: `IconButton` would be a better name.

lib/gui/app/styled-components.tsx
```diff
@@ -55,30 +57,45 @@ export const ThemedProvider = (props: any) => (
...
+		> svg {
+			font-size: 1em;
+		}
+	}
 `;
 
-export const StepButton = (props: any) => (
-	<BaseButton primary {...props}></BaseButton>
-);
+export const StepButton = styled((props: any) => <Button {...props}></Button>)`
```
zvin, line 73: Please avoid `any`.

lib/gui/app/components/source-selector/source-selector.tsx
thundron: I couldn't find a way around the jsonStorage.get method (promisified) returning `unknown` as a type
As for the default list - it won't break, the default for that method is an empty object

lib/gui/app/pages/main/Flash.tsx
thundron: It gave me warnings using just one or the other, but admit it was in the early stages and things have changed. I'll try that

lib/gui/app/components/svg-icon/svg-icon.tsx
thundron: Leftover, will remove it

lib/gui/app/components/source-selector/source-selector.tsx
thundron: There's no need to but it will look better with all the other flows implemented which we'll have relatively soon, and it costs nothing while reducing the clutter

lib/gui/app/components/source-selector/source-selector.tsx
zvin: It most definitely breaks.
```
/home/alexis/dev/resin.io/etcher/node_modules/resin-corvus/src/resin-corvus.js:233 {"stack":"TypeError: recentImages is not iterable\n    at done (file:///home/alexis/dev/resin.io/etcher/generated/gui.js:4610:61)\n    at HTMLUnknownElement.callCallback (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:336:14)\n    at Object.invokeGuardedCallbackDev (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:385:16)\n    at invokeGuardedCallback (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:440:31)\n    at invokeGuardedCallbackAndCatchFirstError (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:454:25)\n    at executeDispatch (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:584:3)\n    at executeDispatchesInOrder (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:606:7)\n    at executeDispatchesAndRelease (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:713:5)\n    at executeDispatchesAndReleaseTopLevel (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:722:10)\n    at forEachAccumulated (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:694:8)\n    at runEventsInBatch (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:739:3)\n    at runExtractedPluginEventsInBatch (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:880:3)\n    at handleTopLevel (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:5807:5)\n    at batchedEventUpdates$1 (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:24368:12)\n    at batchedEventUpdates (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:1419:12)\n    at dispatchEventForPluginEventSystem (/home/alexis/dev/resin.io/etcher/node_modules/react-dom/cjs/react-dom.development.js:5898:5)","message":"recentImages is not iterable"}
```

zvin: Capture d’écran de 2020-04-02 16-49-01

zvin: For the return type, you'll need to add a typings file for electron-json-storage in the `typings` folder.

lib/gui/app/components/source-selector/source-selector.tsx
thundron: It relies on a method that has an error parameter in the callback and rightly so because it handles opening a file, and things might go wrong.
You reminded me though that we need a warning to pop-up instead of just logging the error, as the error won't probably trigger the dialogue by itself, given that it comes from a callback (so it has to be manually handled anyway)

lib/gui/app/components/source-selector/source-selector.tsx
thundron: It won't be, the default returned value is an empty object

thundron: In case it fails, it still is `[]` with React's `useState`

zvin: See the screenshot above

lib/gui/app/components/source-selector/source-selector.tsx
zvin: For now, when there's an error, nothing happens, the url prompt just stays there as if no button was clicked.

lib/gui/app/components/source-selector/source-selector.tsx
zvin: It won't look better.
You're defining a public attribute that is only used in one place in this class.

lib/gui/app/components/source-selector/source-selector.tsx
thundron: I was able to reproduce, it's because of the default object returned. Adding a check there

thundron: For the return type: there are already the official typings, that's what `promisify` uses. I'll just ditch that with a custom promise, I'd rather have correct typings

zvin: right, I haven't seen @types/electron-json-storag

lib/gui/app/pages/main/Flash.tsx
thundron: this wasn't changed, but sure, pushing the fix now

thundron:
@balena-ci retest

lib/gui/modules/child-writer.ts
```diff
@@ -95,6 +97,7 @@ async function writeAndValidate(
 	onProgress: sdk.multiWrite.OnProgressFunction,
 	onFail: sdk.multiWrite.OnFailFunction,
 ): Promise<WriteResult> {
+	console.log('source', source);
```
zvin, line 100: leftover `console.log`

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -35,15 +38,50 @@ import { replaceWindowsNetworkDriveLetter } from '../../os/windows-network-drive
...
+			jsonStorageCb.get(key, (err, value) => {
+				if (err) {
+					reject(err);
+					throw err;
+				}
+				resolve(value);
+				return value;
+			});
+		});
+	},
```
zvin, line 61: Throwing or returning in a `new Promise((resolve, reject) => { ... })` makes no sense.
Use `util.promisify`:
```typescript
const jsonStorage = {
    get: promisify(jsonStorageCb.get),
    set: promisify(jsonStorageCb.set),
}
```

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +101,145 @@ function getState() {
...
+	const [imageURL, setImageURL] = React.useState('');
+	const [recentImages, setRecentImages]: [
+		string[],
+		(value: React.SetStateAction<string[]>) => void,
+	] = React.useState([]);
+	React.useEffect(() => {
+		const fetchRecentUrlImages = async () => {
+			const recentUrlImages: string[] = await getRecentUrlImages();
+			if (!Array.isArray(recentUrlImages)) {
+				setRecentImages([]);
```
zvin, line 114: The array check and default empty list value should be done in `getRecentUrlImages`

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -35,15 +38,50 @@ import { replaceWindowsNetworkDriveLetter } from '../../os/windows-network-drive
...
+				}
+				resolve(value);
+				return value;
+			});
+		});
+	},
+};
+
+const getRecentUrlImages = () =>
+	jsonStorage.get('recentUrlImages') as Promise<string[]>;
```
zvin, line 77: Please make `'recentUrlImages'` a constant and add an equivalent `setRecentUrlImages` function using it.
Check that `recentUrlImages` is an array and return a default empty array in `getRecentUrlImages`

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -220,12 +357,18 @@ export class ImageSelector extends React.Component<
 			return;
 		}
 
-		const source = new sdk.sourceDestination.File({
-			path: imagePath,
-		});
+		let source;
+		if (SourceType.name === sourceDestination.File.name) {
```
zvin, line 361: `(SourceType === sourceDestination.File)` should work

lib/gui/app/components/source-selector/source-selector.tsx
zvin: Actually just use `localStorage.setItem('recentUrlImages', JSON.stringify(yourData))` and `JSON.load(localStorage.getItem('recentUrlImages'))` and remove electron-json-storage completely

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -35,15 +37,39 @@ import { replaceWindowsNetworkDriveLetter } from '../../os/windows-network-drive
...
 	StepSelection,
-	Underline,
 } from '../../styled-components';
+import { colors } from '../../theme';
 import { middleEllipsis } from '../../utils/middle-ellipsis';
 import { SVGIcon } from '../svg-icon/svg-icon';
 
+const recentUrlImagesKey = 'recentUrlImages';
+
+function normalizeRecentUrlImages(urls: string[]) {
```
zvin, line 50: Please add the array check, as we don't know what we might get from localStorage.
```typescript
function normalizeRecentUrlImages(urls: any): string[] {
	if (!Array.isArray(urls)) {
		urls = [];
	}
	return _.takeRight(urls.filter(_.isString), 5);
}
```

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -35,15 +37,39 @@ import { replaceWindowsNetworkDriveLetter } from '../../os/windows-network-drive
...
+	return _.takeRight(urls.filter(_.isString), 5);
+}
+
+function getRecentUrlImages(): string[] {
```
zvin, line 54: Actually to be 100% safe:
```typescript
function getRecentUrlImages(): string[] {
	let urls = [];
	try {
		urls = JSON.parse(localStorage.getItem(recentUrlImagesKey) || '[]');
	} catch {
		// noop
	}
	return normalizeRecentUrlImages(urls);
}
```

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -73,36 +89,138 @@ function getState() {
...
+			const recentUrlImages: string[] = await getRecentUrlImages();
+			setRecentImages(recentUrlImages);
+		};
+		fetchRecentUrlImages();
+	}, []);
+	return (
+		<Modal
+			done={async () => {
+				const sanitizedRecentUrls = _.uniq(
+					_.reject([...recentImages, imageURL], _.isEmpty),
```
zvin, line 109: Move this logic to `normalizeRecentUrlImages`

```typescript
const sanitizedRecentUrls = normalizeRecentUrlImages([...recentImages, imageURL])
```


```typescript
function normalizeRecentUrlImages(urls: any): string[] {
	if (!Array.isArray(urls)) {
		urls = [];
	}
	return _.chain(urls)
		.filter(_.isString)
		.reject(_.isEmpty)
		.uniq()
		.takeRight(5)
		.value();
}

```

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -220,12 +338,18 @@ export class ImageSelector extends React.Component<
...
+				path: imagePath,
+			});
+		} else {
+			source = new sourceDestination.Http(imagePath);
+		}
+
 		try {
 			const innerSource = await source.getInnerSource();
-			const metadata = (await innerSource.getMetadata()) as sdk.sourceDestination.Metadata & {
+			const metadata = (await innerSource.getMetadata()) as sourceDestination.Metadata & {
```
zvin, line 352: We need the UI to show that we are loading the metadata, maybe don't close the URL selector until we have the metadata.

thundron: Made it so that the button is disabled and it closes when the data's done loading

lib/gui/app/pages/main/MainPage.tsx
```diff
@@ -188,12 +197,17 @@ export class MainPage extends React.Component<
...
-					<div
+					<Flex
 						className="page-main row around-xs"
 						style={{ margin: '110px 50px' }}
 					>
 						<div className="col-xs">
-							<ImageSelector flashing={this.state.isFlashing} />
+							<SourceSelector
+								flashing={this.state.isFlashing}
+								afterSelected={(source: SourceOptions) =>
```
zvin, line 207: Please throw an error if we're using `Http` and `source.imagePath` does not start with `http://` or `https://`
Something like "Only http:// and https:// protocols are supported"

thundron: Added after checking for supported image formats too

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -220,12 +356,28 @@ export class ImageSelector extends React.Component<
...
-		const source = new sdk.sourceDestination.File({
-			path: imagePath,
-		});
+		let source;
+		if (SourceType === sourceDestination.File) {
+			source = new sourceDestination.File({
+				path: imagePath,
+			});
+		} else {
+			if (!_.startsWith(imagePath, 'http')) {
```
zvin, line 365: Please explicitly check for 'http://' or 'https://'.
This code won't throw for `httpwhatever://invalid-url`

lib/shared/messages.ts
```diff
@@ -191,4 +191,8 @@ export const error = {
 			'Please try again, and contact the Etcher team if the problem persists.',
 		].join(' ');
 	},
+
+	unsupportedProtocol: () => {
+		return 'Please insert only URLs that use HTTP or HTTPS.';
```
zvin, line 196: "Only http:// and https:// URLs are supported"

lib/gui/app/components/source-selector/source-selector.tsx
thundron: It will throw as an invalid image so this code is actually redundant for the purpose of a validity check.
I'm just adding it as a way to instruct users, because it throws anyway

PR #3099, opened 2020-03-09
README: adapt gpg key acquisition to work on Ubuntu18.04

Angeall:
The former command failed systematically, the keyserver did not respond.
This modified command did the trick, using hkps.
(Follows https://github.com/balena-io/etcher/pull/3069 that was closed 
for not using hkps)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ installers for all supported operating systems.
 2. Trust Bintray.com's GPG key:
 
     ```sh
-    sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 379CE192D401AB61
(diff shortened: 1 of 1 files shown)


Angeall:
Here is a screenshot to add some context to the problem.

Using Ubuntu 18.04 (Official LTS distribution), I get an error running the command written in the README file.
My commit seems to solve the problem

etcher-gpg

zvin:
Port `11371` is most probably blocked on your network.
Check with `telnet keyserver.ubuntu.com 11371`.
We'll merge that, as other users may be in that case.

Angeall:
You are right, this is the problem, I got a connection refused testing your command.
Indeed, port 443 should be OK in all situations

zvin:
Merged in #3101

Issue #3080, opened 2020-02-24
Can not find icon.png


NelDav:
- **Etcher version:** 1.5.79
- **Operating system and architecture:** Windows 10
- **Image flashed:** 2020-02-13-raspbian-buster-lite
- **Do you see any meaningful error information in the DevTools?** 
> GET file:///C:/Users/DN/AppData/Local/Programs/balena-etcher/resources/app.asar/lib/assets/icon.png net::ERR_FILE_NOT_FOUND

I think etcher should not crash because of a missing image

lurch:
Possibly a duplicate of #3039 ? I wonder why more people aren't getting this error? :man_shrugging:

thundron:
When does this happen?

NelDav:
It happens, directly bevore flashing.

I solved the problem by adding Balena as exception in "Allow an app through Controlled folder acces". (Settings->Update & Security->Windows Security->Virus & threat protection->Manage Settings->Controlled folder access->Allow an app through Controlled folder acces)

lurch:
Ahhhh. Seems like tweaking that setting somehow fixes many Etcher problems? https://github.com/balena-io/etcher/search?q=controlled+folder+access&type=Issues

wahyudwikrisnanto:
Im still had the problem even my setting on controlled folder access is turned off

thundron:
@wahyudwikrisnanto have you tried opening as admin?

zvin:
fixed in 1.5.84

PR #3101, opened 2020-03-09
Updates

zvin:
scale contents in fullscreen mode, electron 7.1.14, readme, lint-sass, ~don't asar pack on windows~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,7 +152,7 @@ lint-ts:
 	resin-lint --fix --typescript typings lib tests scripts/clean-shrinkwrap.ts webpack.config.ts
 
 lint-sass:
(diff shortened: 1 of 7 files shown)


lurch:
> There are issues on sentry reporting missing .node files.
These files are present in the asar archive and we couldn't reproduce
the issue.

This is probably the same thing as reported in #3080 ?  Which appears to be related to some "Windows  Security" feature, but I don't use Windows myself so I dunno why this affects some users but not others.
(i.e. it looks like this PR _might_ fix #3080 too)

nazrhom:
@balena-ci retest

thundron:
@balena-ci retest

PR #3117, opened 2020-04-08
Fixes the Command for macOS drive recovery

wilsonsf:
Changes the documentation to update the disktutil command which didn't fix my case, cause the boot partition was broken.
This way it rewrites the drive into a FAT32 partition editable in Unix/Windows.


zvin:
added in #3115

PR #3131, opened 2020-04-28
Decompress first

zvin:


lurch:
Huh? Won't decompress-before-flashing slow down the overall flashing operation? What about users that don't have enough free disk-space to store the uncompressed image?

lib/gui/app/app.ts
```diff
@@ -89,30 +89,43 @@ analytics.logEvent('Application start', {
 	applicationSessionUuid,
 });
 
+const debouncedLog = _.debounce(console.log, 1000, { maxWait: 1000 });
```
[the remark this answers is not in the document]
zvin: We don't need the time being added.
We don't have any path in the message so we don't need `hideAbsolutePathsInObject` either.
Also this function has nothing to do with analytics.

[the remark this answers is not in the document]
zvin: The `debounce` is here to limit these log messages to 1 per second

zvin:
> Huh? Won't decompress-before-flashing slow down the overall flashing operation? What about users that don't have enough free disk-space to store the uncompressed image?

@lurch it allows us to trim ext partitions and avoid writing unallocated space so it may save some time in some cases.
Compressed `iso`s won't be decompressed.
If there is not enough disk space, images won't be decompressed and everything will work as before.

lurch:
How does Etcher choose where to write the uncompressed files to? Do the uncompressed files get deleted again afterwards, or will users who flash multiple different images suddenly find all their disk space disappearing?
I hope you don't mind me asking questions :slightly_smiling_face:

lib/gui/app/components/finish/finish.tsx
```diff
@@ -27,12 +27,9 @@ import { FlashAnother } from '../flash-another/flash-another';
...
 import { SVGIcon } from '../svg-icon/svg-icon';
 
-const restart = (options: any, goToMain: () => void) => {
-	if (!options.preserveImage) {
-		selectionState.deselectImage();
-	}
+function restart(goToMain: () => void) {
 	selectionState.deselectAllDrives();
-	analytics.logEvent('Restart', options);
+	analytics.logEvent('Restart');
```
thundron, line 32: Why remove the options from the log? It was useful to also have the flash information (e.g. the image flashed/reflashed)

zvin: It was always the same options: `{ preserveImage: true }`

thundron: I must've confused it with another event

PR #3132, opened 2020-04-29
update copyright years

branttaylor:
Noticed in the "About" box that it shows last year, figured I'd get that updated! Also found a few more places to update in comments and code here and there.


zvin:
updated in #3140 , thanks @branttaylor

PR #3142, opened 2020-05-05
Prefer balena-etcher to etcher-bin on Arch Linux

zvin:
Changelog-entry: Prefer balena-etcher to etcher-bin on Arch Linux
Change-type: patch


zvin:
@balena-ci retest

PR #3166, opened 2020-05-21
Init param in correct place

thundron:
Change-type: patch
Changelog-entry: Minor fix - Init isSourceDrive param in correct place

### DIFF
--- a/lib/shared/drive-constraints.ts
+++ b/lib/shared/drive-constraints.ts
@@ -58,8 +58,10 @@ export interface Image {
  * In the context of Etcher, a source drive is a drive
  * containing the image.
  */
-export function isSourceDrive(drive: DrivelistDrive, image: Image): boolean {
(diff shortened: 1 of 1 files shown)


lib/shared/drive-constraints.ts
```diff
@@ -58,8 +58,10 @@ export interface Image {
  * In the context of Etcher, a source drive is a drive
  * containing the image.
  */
-export function isSourceDrive(drive: DrivelistDrive, image: Image): boolean {
-	image = image || {};
```
lurch, line 62: LOL, I _did_ wonder why it was being done that way in the previous PR; but I didn't think it was worth mentioning as I assumed there was a reason for it being done that way :laughing:

thundron: Gotta go fast :P

PR #3176, opened 2020-05-25
fix(gui): Fix userdatadir selection under electron-mocha

thundron:
Change-Type: patch
Signed-off-by: Jonas Hermsmeier <jonas@resin.io>


thundron:
Testing

PR #3180, opened 2020-05-27
Fix flash from url

zvin:


lurch:
"Stop checking file extensions" - is that because of the sometimes-strange file extensions you get when downloading files from URLs? If that's the case, I wonder if it's worth keeping the file-extension-checking code just for images opened from a local drive?

zvin:
@lurch it is not.
We should not care about the file extensions.
If a file has no partition table, the user will be warned about it. This is enough.

lurch:
Oooh, is bulldozer-balena the new name for versionbot? :laughing:

PR #3189, opened 2020-05-29
spectron: Make tests pass on Windows Docker containers

jviotti:
The Spectron test that we have that checks that the browser window is
visible fails when ran inside a Windows Docker container.

In particular, the `isVisible()` function returns `false` when running
in a headless Windows machine.

However, the `isMinimized()` function returns `false`, the `isFocused()`
function returns `true`, and we can fetch the expected browser window
bounds, so we can use all those values in conjunction to reformulate the
test case and avoid `isVisible()`.

The results should be pretty much the same, and the assertions will pass
inside Docker Windows containers.

Change-type: patch

### DIFF
--- a/tests/spectron/runner.spec.ts
+++ b/tests/spectron/runner.spec.ts
@@ -51,7 +51,14 @@ describe('Spectron', function () {
 
 	describe('Browser Window', function () {
 		it('should open a browser window', async function () {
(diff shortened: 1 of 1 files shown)


zvin:
@jviotti Can you please add your explanation above as a comment in the test ?

thundron:
@balena-ci retest

zvin:
@balena-ci retest

PR #3199, opened 2020-06-03
Fix ia32 builds for windows

zvin:
Changelog-entry: Fix ia32 builds for windows
Change-type: patch

### DIFF
--- a/beforeBuild.js
+++ b/beforeBuild.js
@@ -1,0 +1,17 @@
+'use strict'
+
+const cp = require('child_process')
+
+// Rebuild native modules for ia32 and run webpack again for the ia32 part of windows packages
+exports.default = function(context) {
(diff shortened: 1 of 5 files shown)


beforeBuild.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const cp = require('child_process')
+
+// Rebuild native modules for ia32 and run webpack again for the ia32 part of windows packages
```
lurch, line 5: Does this assume that the 64-bit build always runs before the 32-bit build? (and if so, is that guaranteed to happen?)

zvin: Yes and yes unless you specify otherwise.
This was to save some build time on the ci.
I'll change it it rebuild in all cases.

zvin:
cherry-picked in #3195

PR #3202, opened 2020-06-04
Add custom protocol 2

zvin:


thundron:
@balena-ci retest

zvin:
@balena-ci retest

lib/gui/etcher.ts
```diff
@@ -54,6 +57,46 @@ async function checkForUpdates(interval: number) {
 	}
 }
 
+function getCommandLineURL(argv: string[]): string | undefined {
+	if (platform() === 'darwin') {
+		// macOS uses open-url events instead
+		return;
```
thundron, line 63: I looked a bit into it and I was partially wrong so I apologise, but the `open-url` is the event used when opening an URL. When doing something like `./balenaEtcher\ 1.5.97/balenaEtcher.app/Contents/MacOS/balenaEtcher miao.img` (so calling the executable directly) it actually uses argv

thundron: Another strange thing is that the error about the `-psn_...` parameter didn't pop-up with a package I built locally to test the above (removing the check)

PR #3203, opened 2020-06-05
New target selector

thundron:


thundron:
@balena-ci retest

lib/gui/app/scss/modules/_bootstrap.scss
```diff
@@ -25,3 +25,20 @@ html {
...
+  margin: 0;
+}
+
+[uib-tooltip] {
+  cursor: default;
+}
+
+.tooltip {
+  word-wrap: break-word;
+}
```
zvin, line 44: These are no longer needed (except maybe `label`).

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,409 @@
...
+		);
+	}
+	return;
+}
+
+interface TableData extends DrivelistTarget {
+	disabled: boolean;
+}
+
+export const TargetSelectorModal = ({
```
zvin, line 174: This could be a `React.Component`

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,409 @@
...
+ * @summary Get a drive's compatibility status object(s)
+ *
+ * @description
+ * Given a drive, return its compatibility status with the selected image,
+ * containing the status type (ERROR, WARNING), and accompanying
+ * status message.
+ */
+function getDriveStatuses(
+	drive: DrivelistTarget,
+): Array<{ type: number; message: string }> {
```
zvin, line 63: This is redeclaring `TargetStatus`

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,409 @@
...
+	}
+`;
+
+interface DriverlessDrive {
+	link: string;
+	linkTitle: string;
+	linkMessage: string;
+}
+
+interface TargetStatus {
```
zvin, line 103: This should be in `drive-constraints.ts`

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,409 @@
...
+function renderProgress(progress: number) {
+	if (Boolean(progress)) {
+		return (
+			<Flex flexDirection="column">
+				<Txt fontSize={12}>Initializing device</Txt>
+				<InitProgress value={progress} />
+			</Flex>
+		);
+	}
+	return;
```
zvin, line 167: unneeded return

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	::-webkit-progress-value {
+		background-color: #1496e1;
+		border-radius: 14px;
+	}
+`;
+
+function renderProgress(progress: number) {
+	if (Boolean(progress)) {
```
zvin, line 159: unneeded check

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,409 @@
...
+		...props
+	}: {
+		value: number;
+		props?: React.ProgressHTMLAttributes<Element>;
+	}) => {
+		return <progress max="100" value={value} {...props}></progress>;
+	},
+)`
+	/* Reset the default appearance */
+	-webkit-appearance: none;
```
zvin, line 142: Do we still need the `-webkit` prefixes ?

zvin:
You can remove some scss rules that are no longer used like 
https://github.com/balena-io/etcher/blob/master/lib/gui/app/pages/main/styles/_main.scss#L17-L19
https://github.com/balena-io/etcher/blob/master/lib/gui/app/scss/components/_tick.scss#L26-L30

lib/gui/app/components/target-selector/target-selector-modal.tsx
thundron: Not on this one, but we do on the rules below

thundron:
> You can remove some scss rules that are no longer used like
> https://github.com/balena-io/etcher/blob/master/lib/gui/app/pages/main/styles/_main.scss#L17-L19
> https://github.com/balena-io/etcher/blob/master/lib/gui/app/scss/components/_tick.scss#L26-L30

We need the first, for the second one I replaced the tick icon in the flash results so we can remove the whole file

lib/gui/app/components/target-selector/target-selector-modal.tsx
thundron: Why would it though? With a plain function + useState + useEffect it's just simpler and more readable, I think the issue is the way we handle the current application state *everywhere* and the application state itself needs some refactoring, there are times where I see components called multiple times even though the application state is the same (or should be, but it triggers multiple times)

zvin: I don't mind. But a 230 lines function is not exactly simple and readable.

thundron: Touche, I'm probably just being stubborn because with the current application state it's better to have a class. I'll convert in the next push

lib/gui/app/pages/main/DriveSelector.tsx
```diff
@@ -114,22 +122,37 @@ export const DriveSelector = ({
...
-			{showDriveSelectorModal && (
-				<DriveSelectorModal
-					close={() => setShowDriveSelectorModal(false)}
-				></DriveSelectorModal>
+			{showTargetSelectorModal && (
+				<TargetSelectorModal
+					cancel={() => setShowTargetSelectorModal(false)}
+					done={(selectedTargets: DrivelistTarget[]) => {
+						const selectedDrives = getSelectedDrives();
+						if (_.isEmpty(selectedTargets)) {
```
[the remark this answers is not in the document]
zvin: We don't need this `if`, the `else` block works in all cases.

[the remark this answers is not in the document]
thundron: Refactored & removed lodash

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,499 @@
...
+				systemDrives: enrichedDrivesData.filter((drive) => drive.isSystem),
+				disabledRows: drives
+					.filter(
+						(drive) =>
+							!isDriveValid(drive, this.image) || drive.progress !== undefined,
+					)
+					.map((drive) => drive.displayName),
+				selectedList: getSelectedDrives(),
+			});
+		});
```
zvin, line 340: This is duplicated code, please factorize this and the constructor lines.

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,499 @@
...
+						<Txt>{description}</Txt>
+					);
+				},
+			},
+			{
+				field: 'size',
+				label: 'Size',
+				render: (size: number) => {
+					return bytesToClosestUnit(size);
+				},
```
zvin, line 241: `render: bytesToClosestUnit`

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -96,22 +101,15 @@ const TargetsTable = styled(({ refFn, ...props }) => {
...
+		top: 0;
+		background-color: rgb(248, 249, 253);
 		padding: 6px 8px;
 		color: #2a506f;
 	}
 
-	[data-display='table-body']
-		> [data-display='table-row']
-		> [data-display='table-cell']:first-child {
+	[data-display='table-cell']:first-child {
```
thundron: This whole style can be refactored to:
```css
[data-display='table-head'] [data-display='table-cell'] {
		position: sticky;
		top: 0;
	}

	[data-display='table-cell']:first-child {
		padding-left: 15px;
	}

	&& [data-display='table-row']
		> [data-display='table-cell'] {
		padding: 6px 8px;
		color: #2a506f;
	}
```
so the last rule is valid for both table head and body and is more specific than rendition's rules (which in the `head` override the padding)

lib/gui/app/components/target-selector/target-selector-modal.tsx
```diff
@@ -0,0 +1,480 @@
...
+		padding-left: 15px;
+	}
+
+	[data-display='table-body']
+		> [data-display='table-row']
+		> [data-display='table-cell'] {
+		padding: 6px 8px;
+		color: #2a506f;
+	}
```
thundron, line 121: This whole style can be refactored to:
```css
[data-display='table-head'] [data-display='table-cell'] {
  position: sticky;
  top: 0;
}

[data-display='table-cell']:first-child {
  padding-left: 15px;
}

&& [data-display='table-row']
  > [data-display='table-cell'] {
  padding: 6px 8px;
  color: #2a506f;
}
```
so the last rule is valid for both table head and body and is more specific than rendition's rules (which override the padding in the `head`)

zvin: :+1: missing `background-color: rgb(248, 249, 253);` for the header

PR #3222, opened 2020-06-22
Efp restyle

thundron:


lib/gui/app/components/flash-results/flash-results.tsx
```diff
@@ -50,15 +39,19 @@ export function FlashResults({
...
-		bytesToMegabytes(
-			results.sourceMetadata.size /
-				(results.bytesWritten / results.averageFlashingSpeed),
-		),
-		1,
-	);
+	const effectiveSpeed = bytesToMegabytes(
+		results.sourceMetadata.size /
+			(results.bytesWritten / results.averageFlashingSpeed),
+	).toFixed(1);
```
zvin, line 45: `_.round` is prefereable (for 1.65 for example).

lib/gui/app/components/flash-results/flash-results.tsx
```diff
@@ -73,24 +66,22 @@ export function FlashResults({
 					Flash Complete!
 				</Txt>
 			</Flex>
-			<Div className="results" mr="0" mb="0" ml="40px">
-				{_.map(results.devices, (quantity, type) => {
+			<Flex flexDirection="column" mr="0" mb="0" ml="40px">
+				{Object.keys(results.devices).map((type: 'failed' | 'successful') => {
+					const quantity = results.devices[type];
```
zvin, line 71: `Object.entries` would fit better.

lib/gui/app/components/reduced-flashing-infos/reduced-flashing-infos.tsx
```diff
@@ -71,24 +40,36 @@ export class ReducedFlashingInfos extends React.Component<
...
+				<Flex mb={16}>
 					<SVGIcon
 						disabled
-						width="20px"
+						width="21px"
+						height="21px"
 						contents={this.props.imageLogo}
-						fallback={<ImageSvg className="disabled" width="20px" />}
+						fallback={ImageSvg}
+						style={{ marginRight: 9 }}
```
zvin, line 55: `9px` would be clearer

lib/gui/app/components/reduced-flashing-infos/reduced-flashing-infos.tsx
```diff
@@ -71,24 +40,36 @@ export class ReducedFlashingInfos extends React.Component<
...
 					/>
-					<Span>{this.props.imageName}</Span>
-					<Span color="#7e8085">{this.props.imageSize}</Span>
-				</Span>
+					<Txt
+						style={{ marginRight: 9 }}
```
zvin, line 58: `9px` would be clearer

lib/gui/app/components/featured-project/featured-project.tsx
```diff
@@ -34,23 +37,43 @@ export class FeaturedProject extends React.Component<
...
 				'https://assets.balena.io/etcher-featured/index.html';
+			const efpParams = {
+				borderRight: false,
+				darkBackground: true,
+			};
+			let params = '?';
+			for (const [param, value] of Object.entries(efpParams)) {
+				params += `${param}=${value}&`;
+			}
+			endpoint += params;
```
zvin, line 59:
```typescript
const url = new URL((await settings.get('featuredProjectEndpoint')) || 'https://assets.balena.io/etcher-featured/index.html');
url.searchParams.append('borderRight', false);
url.searchParams.append('darkBackground', true);
this.setState({ endpoint: url.toString() });
```

thundron:
@balena-ci retest

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -140,6 +139,9 @@ const URLSelector = ({
 			cancel={cancel}
 			primaryButtonProps={{
 				disabled: loading || !imageURL,
+				style: {
+					display: imageURL ? 'block' : 'none',
+				},
```
zvin, line 144: Don't add the button to the document instead of hiding it with `display: none`

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -61,16 +60,16 @@ import ImageSvg from '../../../assets/image.svg';
...
-		.reject(_.isEmpty)
-		.uniq()
-		.takeRight(5)
-		.value();
+	return urlImages
+		.filter((url: any) => typeof url === 'string')
+		.filter((url: string) => url !== '')
+		.filter((url: string, i: number, urls: string[]) => urls.indexOf(url) === i) // Deduplicate
+		.reverse()
+		.slice(0, 5);
```
zvin, line 72: This will reverse the url list each time you click ok

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -61,16 +60,16 @@ import ImageSvg from '../../../assets/image.svg';
 
 const recentUrlImagesKey = 'recentUrlImages';
 
-function normalizeRecentUrlImages(urls: any): string[] {
-	if (!Array.isArray(urls)) {
-		urls = [];
+function normalizeRecentUrlImages(urlImages: any): string[] {
```
zvin, line 63: `urls` was a better name than `urlUmages`

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -164,20 +166,21 @@ const URLSelector = ({
...
 							<Txt
 								key={recent}
 								onClick={() => {
 									setImageURL(recent);
 								}}
 							>
 								<span>
-									{_.last(_.split(recent, '/'))} - {recent}
+									{middleEllipsis(recent.split('/').pop() || '', 45)} -{' '}
+									{middleEllipsis(recent, 45)}
```
zvin, line 183: If the goal was to avoid a scroll bar, I still get a scroll bar

Capture d’écran de 2020-07-03 12-12-26

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -164,20 +166,21 @@ const URLSelector = ({
...
+						rows={recentImages.map((recent) => (
 							<Txt
 								key={recent}
 								onClick={() => {
 									setImageURL(recent);
 								}}
 							>
 								<span>
-									{_.last(_.split(recent, '/'))} - {recent}
+									{middleEllipsis(recent.split('/').pop() || '', 45)} -{' '}
```
zvin, line 182: `recent.split('/').pop()` is not good enough if the url has query parameters,
`new URL(recent).pathname.split('/').pop()` would be better

lib/gui/app/components/source-selector/source-selector.tsx
zvin: I'm not sure we should limit the url length, just let it show on multiple lines.
You don't know if `middleEllipsis` will make it short enough:

Capture d’écran de 2020-07-03 12-20-48

lib/gui/app/components/source-selector/source-selector.tsx
thundron: This can't be done with current rendition

zvin: Can we just keep the button but disable it ?

thundron: Yes, but we haven't got any primary button disabled outside of the three on the main screen. Isn't it better to not show it unless there is a selection happening? It's pointless otherwise imho

zvin: I don't like the cancel button moving

PR #3247, opened 2020-07-16
Lighter

zvin:
Saves around 4MiB of generated js, splits the child-writer into a separate file.


zvin:
@balena-ci retest

thundron:
@balena-ci retest

PR #3283, opened 2020-08-25
[DON'T MERGE] Save url image stub

thundron:


lib/gui/app/theme.ts
```diff
@@ -67,8 +70,7 @@ export const colors = {
 
 const font = 'SourceSansPro';
 
-export const theme = {
-	colors,
+export const theme = _.merge({}, Theme, {
```
zvin, line 73: You don't need lodash here: `const theme = { ...Theme, otherStuff }`

lib/gui/app/styled-components.tsx
```diff
@@ -134,24 +124,23 @@ const modalFooterShadowCss = css`
 	background-attachment: local, local, scroll, scroll;
 `;
 
-export const Modal = styled(({ style, ...props }) => {
+export const Modal = styled(({ style, children, ...props }) => {
 	return (
 		<Provider
-			theme={{
-				...defaultTheme,
+			theme={_.merge({}, theme, {
```
zvin, line 130: You don't need lodash here: `theme={{ ...theme, otherStuff }}`

lib/gui/app/theme.ts
thundron: properties that are objects are overwritten with the spread operator instead of being merged

lib/gui/app/styled-components.tsx
thundron: same as above

zvin (changes requested):
see the comments

lib/gui/app/components/drive-selector/drive-selector.tsx
```diff
@@ -75,27 +72,12 @@ function isDrivelistDrive(drive: Drive): drive is DrivelistDrive {
...
-		<Table<Drive> ref={refFn} {...props} />
-	</div>
+const DrivesTable = styled((props: GenericTableProps<Drive>) => (
+	<Table<Drive> {...props} />
 ))`
-	[data-display='table-head']
+	[data-display='table-body']
 		> [data-display='table-row']
 		> [data-display='table-cell'] {
-		position: sticky;
```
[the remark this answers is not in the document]
thundron: It's still there, it should be in the generic table, I'll check what's wrong with that

zvin (changes requested):
The sticky header in the target list is gone, please briing it back

Capture d’écran de 2020-09-24 18-18-00

lib/gui/app/components/drive-selector/drive-selector.tsx
```diff
@@ -453,95 +405,92 @@ export class DriveSelector extends React.Component<
...
+								if (this.props.multipleSelection) {
+									const newList = [...selectedList];
+									const selectedIndex = selectedList.findIndex(
+										(drive) => drive.device === row.device,
+									);
+									if (selectedIndex === -1) {
+										newList.push(row);
+									} else {
+										// Deselect if selected
+										newList.splice(selectedIndex, 1);
```
zvin, line 462: The behaviour is different between clicking the radio button or clicking the row for non multipleSelection tables.
This should fix it:

```typescript
								if (
									!isDrivelistDrive(row) ||
									this.driveShouldBeDisabled(row, image)
								) {
											return;
								}
								const index = selectedList.findIndex((d) => d.device === row.device);
								const newList = this.props.multipleSelection ? [...selectedList] : [];
								if (index === -1) {
									newList.push(row);
								} else {
									// Deselect if selected
									newList.splice(index, 1);
								}
								this.setState({
									selectedList: newList,
								});
```

zvin (changes requested):

lib/gui/app/components/drive-selector/drive-selector.tsx
thundron: Fixed, I'm handling some other issues that came up with the latest rendition before pushing again though

lib/gui/app/styled-components.tsx
```diff
@@ -242,6 +247,9 @@ export const Alert = styled((props) => (
 
 export interface GenericTableProps<T> extends BaseTableProps<T> {
 	refFn: (t: BaseTable<T>) => void;
+	data: T[];
+	notAllSelected?: boolean;
```
zvin, line 251: `notAllSelected` is calculated from other props so it should not be in props.

zvin (changes requested):

lib/gui/app/styled-components.tsx
thundron: It's used in the styled component's css which can only access the props. I can make another interface which extends it and adds only the `notAllSelected` property, but it seems a waste to me

zvin: Move the logic calculating `notAllSelected` into the function that takes the props as parameters (`StyledTable`).

zvin:
```typescript
				${(props) => {
					const notAllSelected =
						props.multipleSelection &&
						props.checkedRowsNumber !== 0 &&
						props.checkedRowsNumber !== props.data.length;
					return notAllSelected
						? `
					font-weight: 600;
					color: ${colors.primary.foreground};
					background: ${colors.primary.background};

					::after {
						content: '–';
					}
					`
						: '';
				}}
```

thundron: Right, good call, no need to have the `const ...` either, it's just a short if with the 3 props

lib/gui/app/models/selection-state.ts
```diff
@@ -72,24 +72,24 @@ export function getImage(): SourceMetadata | undefined {
...
-export function getImageLogo() {
-	return getImage()?.logo;
+export function getImageLogo(): string | undefined {
+	return store.getState().toJS().selection.image?.logo;
 }
 
-export function getImageSupportUrl() {
-	return getImage()?.supportUrl;
+export function getImageSupportUrl(): string | undefined {
+	return store.getState().toJS().selection.image?.supportUrl;
```
zvin, line 92: Why are we adding copy pasted code ?
I'd rather not change this or remove it completely, using `getImage().whateverYouNeed` in the caller.

tests/shared/drive-constraints.spec.ts
```diff
@@ -700,11 +700,6 @@ describe('Shared: DriveConstraints', function () {
 				});
 
 				it('should return false if the drive is not large enough and is a source drive', function () {
-					console.log('YAYYY', {
-						...image,
-						path: path.join(this.mountpoint, 'rpi.img'),
-						size: 5000000000,
-					});
```
zvin, line 707: Please refactor the PR to remove this commit and avoid adding this console.log in the first place.

lib/gui/app/components/progress-button/progress-button.tsx
```diff
@@ -44,7 +44,7 @@ const FlashProgressBar = styled(ProgressBar)`
 `;
 
 interface ProgressButtonProps {
-	type: 'decompressing' | 'flashing' | 'verifying';
+	type: FlashState['type'];
```
zvin, line 47: This is not equivalent as type is nullable in FlashState

lib/gui/app/models/settings.ts
```diff
@@ -41,12 +41,15 @@ export const DEFAULT_HEIGHT = 480;
  * NOTE: The ternary is due to this module being loaded both,
```
zvin, line 41: This comment needs to be updated

lib/gui/app/modules/progress-status.ts
```diff
@@ -62,6 +62,12 @@ export function fromFlashState({
 		} else {
 			return { status: 'Finishing...' };
 		}
+	} else if (type === 'downloading') {
+		if (percentage == null) {
+			return { status: 'Downloading...' };
+		} else if (percentage < 100) {
```
zvin, line 68: Can it be > 100 ?

lib/gui/app/os/dialog.ts
```diff
@@ -40,6 +40,12 @@ async function mountSourceDrive() {
  * Notice that by image, we mean *.img/*.iso/*.zip/etc files.
  */
 export async function selectImage(): Promise<string | undefined> {
+	return await openDialog();
+}
+
+export async function openDialog(
+	type: 'openFile' | 'openDirectory' = 'openFile',
+) {
 	await mountSourceDrive();
```
zvin, line 49: `await mountSourceDrive();` should only be in `selectImage`

lib/gui/modules/child-writer.ts
```diff
@@ -302,3 +320,43 @@ ipc.connectTo(IPC_SERVER_ID, () => {
...
+		error: Error,
+	) => void,
+) {
+	const urlImage = new Http({ url: imagePath, avoidRandomAccess: true });
+	const source = await urlImage.getInnerSource();
+	const metadata = await source.getMetadata();
+	const fileName = `${saveUrlImageTo}/${metadata.name}`;
+	let alreadyDownloaded = false;
+	try {
+		alreadyDownloaded = (await fs.stat(fileName)).isFile();
```
zvin, line 339: We can't rely on a file with the same name being present in the downloads folder.
It could be any other file.
It could be a partially downloaded file.

lib/gui/modules/child-writer.ts
```diff
@@ -302,3 +320,43 @@ ipc.connectTo(IPC_SERVER_ID, () => {
...
+	imagePath: string,
+	saveUrlImageTo: string,
+	onProgress: (state: ProgressState) => void,
+	onFail: (
+		destination: sdk.sourceDestination.SourceDestination,
+		error: Error,
+	) => void,
+) {
+	const urlImage = new Http({ url: imagePath, avoidRandomAccess: true });
+	const source = await urlImage.getInnerSource();
```
zvin, line 334: Are we sure we want to save the decompressed image ?

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -405,6 +410,10 @@ export class SourceSelector extends React.Component<
 		});
 	}
 
+	private setDefaultFlowActive(defaultFlowActive: boolean) {
+		this.setState({ defaultFlowActive });
+	}
```
zvin, line 415: We don't need a method for that

zvin (changes requested):

lib/gui/app/modules/progress-status.ts
thundron: I don't think so, I remember we had an issue with the `flashing` type continuously going up but it shouldn't be the case with the `downloading` type. I'll check how it behaves with the condition removed

lib/gui/modules/child-writer.ts
thundron: > It could be any other file.
> It could be a partially downloaded file.

Would checking the size be accurate enough then? Given that I'll change it so that we save the compressed image instead of the uncompressed one

lib/gui/modules/child-writer.ts
thundron: Good call, how do we deal with decompressing the image though? Should we download with a normal axios call then call the `decompressThenFlash` method on that file or flash without decompressing first?

lib/gui/app/components/progress-button/progress-button.tsx
thundron: @zvin what's the issue with it? We should avoid re-defining types we already have - FlashState's `type` is the one we should use for the UI and it extends the one from etcher-sdk's flash progress state since it can also be `downloading`, which isn't a concept of the sdk

thundron: @zvin what's the issue with it though? This just streamlines the type so it's not defined in multiple places

lib/gui/modules/child-writer.ts
thundron: Good call, I'll change that to pipeToDesinations then decompress afterwards, so we can also check the compressed size in case it's a compressed image

tests/shared/drive-constraints.spec.ts
thundron: It was merged into master, we would have to rebase that which I'm really skeptical about

PR #3284, opened 2020-08-25
105

zvin:


lib/gui/app/pages/main/MainPage.tsx
```diff
@@ -163,36 +163,26 @@ export class MainPage extends React.Component<
 		return (
 			<>
 				<Flex
-					id="app-header"
```
thundron, line 166: this is needed to grab the header and move the window

lib/gui/app/pages/main/MainPage.tsx
```diff
@@ -163,36 +163,26 @@ export class MainPage extends React.Component<
...
-					justifyContent="center"
-					style={{
-						width: '100%',
-						height: '50px',
-						padding: '13px 14px',
-						textAlign: 'center',
-						position: 'relative',
-						zIndex: 1,
```
thundron, line 174: this is needed as well for the header to show while showing the webviews (efp and the incoming new success screen)

thundron (changes requested):

thundron (approved):

PR #3288, opened 2020-08-28
Remove all ext2fs built files before rebuilding on windows

zvin:
Changelog-entry: Remove all ext2fs built files before rebuilding on windows
Change-type: patch


zvin:
@balena-ci retest

PR #3295, opened 2020-09-09
Fix content not loading when the app path contains special characters

zvin:
Changelog-entry: Fix content not loading when the app path contains special characters
Change-type: patch


zvin:
@balena-ci retest

zvin:
@balena-ci I self-certify!

PR #3345, opened 2020-11-06
111

zvin:


lurch:
222 :wink:

thundron:
definitely not the best name for a branch :P but it stands for v1.5.111 and the commits make sense

lib/gui/app/components/finish/finish.tsx
```diff
@@ -39,8 +40,19 @@ function restart(goToMain: () => void) {
 	goToMain();
 }
 
+async function getSuccessBannerURL() {
+	return (
+		(await settings.get('successBannerURL')) ??
+		'https://www.balena.io/etcher/success-banner?borderTop=false&darkBackground=true'
```
lurch, line 46: I dunno if it's deliberate, or just an accidental inclusion, but if I load that URL in a web-browser and look at the "Network" tab in the WebInspector, I see that there's a request to `https://www.google-analytics.com/analytics.js` ?

zvin: The URL didn't change.
It is not loaded as it is, see https://github.com/balena-io/etcher/blob/12cd8a39c1ec1883474f2027f335100fe3400bf9/lib/gui/app/components/safe-webview/safe-webview.tsx#L91-L94

lurch: Ah cool. Just a quick follow-up reply to confirm that when I try the same thing with
`https://www.balena.io/etcher/success-banner/?borderTop=false&darkBackground=true&optOutAnalytics=true`
then the Network tab in the web-inspector shows that all requests are to `www.balena.io` :+1:

nazrhom:
@balena-ci retest

zvin:
@balena-ci retest

thundron (approved):

PR #3384, opened 2020-12-14
docs: update macOS version

shawaj:
Update macOS version as latest version of Electron is 10.10 compatible only (Yosemite)

Change-Type: patch
Signed-off-by: Aaron Shaw <aaron@balena.io>


zvin:
cherry-picked in #3377

PR #3389, opened 2020-12-20
docs: fix quote marks

shawaj:
Fix quote mark styling

Change-type: patch
Signed-off-by: Aaron Shaw <aaron@balena.io>


zvin:
cherry-picked in #3394 3394

PR #3427, opened 2021-02-01
Correct the errors in README.md

hmsjy2017:
hkps -> http


zvin:
This is not a typo, see https://en.wikipedia.org/wiki/Key_server_(cryptographic)

PR #3447, opened 2021-03-04
Better separate the Yum and DNF instructions.

duganchen:


zvin:
cherry-picked in #3432

PR #3495, opened 2021-04-23
Etcher dev server

zwhitchcox:
Add webpack dev server and hot module reloading to get live changes and reloads without reloading the whole electron app.

This patch also runs the development environment in development mode, which is much, much faster on builds and rebuilds.


zwhitchcox:
@balena-ci retest

thundron (approved):

PR #3503, opened 2021-05-03
chore: replace make webpack with npm run webpack

SethFalco:
I was running through CONTRIBUTING.md and got tripped up by one of the instructions.

It seems to me this should be updated to `npm run`, instead of `make`?

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -91,7 +91,7 @@ make electron-develop
 
 ```sh
 # Build the GUI
-make webpack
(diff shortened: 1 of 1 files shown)


lurch:
An alternative approach might be to add:
```
webpack:
	npm run webpack
```
to the `Makefile`? But I don't work on this project so I obviously dunno which approach the Etcher devs would prefer!

thundron:
cherry-picked in https://github.com/balena-io/etcher/pull/3562 , waiting for some internal testing to merge 👍

PR #3545, opened 2021-07-02
Add possibile authorization as a query param

JSReds:
Add possibile authorization as a query param

Change-type: minor
Signed-off-by: Andrea Rosci <andrear@balena.io>

### DIFF
--- a/lib/gui/app/components/source-selector/source-selector.tsx
+++ b/lib/gui/app/components/source-selector/source-selector.tsx
@@ -61,6 +61,7 @@ import ImageSvg from '../../../assets/image.svg';
 import SrcSvg from '../../../assets/src.svg';
 import { DriveSelector } from '../drive-selector/drive-selector';
 import { DrivelistDrive } from '../../../../shared/drive-constraints';
+import axios, { AxiosRequestConfig } from 'axios';
 
 const recentUrlImagesKey = 'recentUrlImages';
 
@@ -335,7 +336,20 @@ export class SourceSelector extends React.Component<
 				path: selected,
 			});
 		}
-		return new sourceDestination.Http({ url: selected });
+		const splitUrl = selected.split('?');
+		const url = splitUrl.slice(0, splitUrl.length - 1).join('');
+		const queryParams = splitUrl.length > 1 && splitUrl[splitUrl.length - 1];
+		const urlSearchParams = !!queryParams
+			? new URLSearchParams(queryParams)
+			: undefined;
+		const params: AxiosRequestConfig | undefined = !!urlSearchParams
+			? Object.fromEntries(urlSearchParams.entries())
+			: undefined;
+		const config = params ?? {};
+		return new sourceDestination.Http({
+			url,
+			axiosInstance: axios.create(config),
+		});
 	}
 
 	private reselectSource() {


lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -335,7 +336,17 @@ export class SourceSelector extends React.Component<
 				path: selected,
 			});
 		}
-		return new sourceDestination.Http({ url: selected });
+		let authorization: string | undefined;
+		if (selected.includes('authorization')) {
```
lurch, line 340: Perhaps this should be `'authorization='` to match the line below? I guess you wouldn't want this to trigger if the image URL was something like `http://www.no-authorization-needed.com/download/os.img` :wink: 

(I guess for completeness, you could do full query-string parsing, in case there are other query-params after the authorization param, but perhaps that's overkill?)

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -335,7 +336,19 @@ export class SourceSelector extends React.Component<
 				path: selected,
 			});
 		}
-		return new sourceDestination.Http({ url: selected });
+		const splitUrl = selected.split('?');
+		const queryParams = splitUrl[splitUrl.length - 1];
```
lurch, line 340: I think you need to check that `splitUrl` has a length greater than 1 first? :wink: 
```
foo = 'abc'.split('?')
console.log(foo[foo.length - 1])
abc
foo = 'abc?def'.split('?')
console.log(foo[foo.length - 1])
def
```

JSReds: :pray: thanks for noticing, unfortunately I changed the logic a bit and forgot to add that check.

lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -335,7 +336,19 @@ export class SourceSelector extends React.Component<
...
-		return new sourceDestination.Http({ url: selected });
+		const splitUrl = selected.split('?');
+		const queryParams = splitUrl.length > 1 && splitUrl[splitUrl.length - 1];
+		const urlSearchParams = !!queryParams
+			? new URLSearchParams(queryParams)
+			: undefined;
+		const params: AxiosRequestConfig | undefined = !!urlSearchParams
+			? Object.fromEntries(urlSearchParams.entries())
+			: undefined;
+		const config = params ?? {};
```
JSReds, line 347: This could also be: 

```suggestion
const splitUrl = selected.split('?');
const queryParams = splitUrl.length > 1 && splitUrl[splitUrl.length - 1];
const urlSearchParams = !!queryParams && new URLSearchParams(queryParams);
const params: AxiosRequestConfig | false = !!urlSearchParams && Object.fromEntries(urlSearchParams.entries());
const config = params || {};
```

which is probably shorter, but TBH I prefer in that case to have value or undefined instead of having value or false.
Of course if you prefer the other way around, I'll change it.

thundron: I prefer to have either the value or undefined too, as the `false` value is still a value and could be misinterpreted in the future

thundron:
cherry-picked in https://github.com/balena-io/etcher/pull/3562 , waiting for some internal testing to merge 👍

Issue #3538, opened 2021-06-22
Typo in README?


lurch:
I just noticed that in the https://github.com/balena-io/etcher#dnf part of the README, the
```
rm /etc/yum.repos.d/balena-etcher.repo
rm /etc/yum.repos.d/balena-etcher-source.repo
```
lines appear in **both** the "Update and install:" and also the "Uninstall" sections, which probably isn't right :wink:

PR #3570, opened 2021-07-31
Fix Update and install Etcher with DNF instructions - documentation

gitexel:
Remove wrong commands `rm` from DNF update and install section

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,8 +91,6 @@ zypper rr balena-etcher-source
 
    ```sh
    sudo dnf install -y balena-etcher-electron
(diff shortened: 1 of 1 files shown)


lurch:
I guess this fixes #3538 ?

gitexel:
@lurch yes

lurch:
Did you close this accidentally?

gitexel:
I thought there is another PR open for the same issue.

thundron:
@gitexel I included your commit in another PR where I sanitized the commits, so that we also have one version bump instead of multiple (due to other PRs too)

PR #3581, opened 2021-08-20
add support for basic auth when getting images from URL

flec:
This adds basic auth support for getting images via URL which is required by our customer using balena.
The inputs for username and password are cleared when collapsed.
The changes have been tested on Ubuntu 20.04.

⚠️This pull request depends on https://github.com/balena-io-modules/etcher-sdk/pull/212 and should not be merged without an updated version of etcher-sdk

image
image


lurch:
Is this similar to #3545 or are they separate things?

flec:
I guess they are similar but this seems more user friendly to me. The solution proposed in https://github.com/balena-io/etcher/pull/3545 would probably not be accepted by our customer. An other issue in the proposed solution is the visible password.

thundron:
cherry-picked in https://github.com/balena-io/etcher/pull/3599 , thank you very much for the contribution <3

PR #3599, opened 2021-09-14
Led color settings

thundron:


flec:
Thank you @thundron for considering my PR. Looking at the commits, I think the cherry pick from https://github.com/balena-io/etcher/pull/3581 is missing.

thundron:
@flec I had it in my local version as I'm also trying to update electron version, just pushed it so the remote branch reflects the local commits 👍

thundron:
@balena-ci I self-certify!

PR #3610, opened 2021-10-12
Fix write step for Http file process

JSReds:
Change-type: patch
Signed-off-by: Andrea Rosci <andrear@balena.io>


thundron:
cherry-picked in https://github.com/balena-io/etcher/pull/3605

PR #3628, opened 2021-11-09
Device info draft

thundron:


lib/gui/app/app.ts
```diff
@@ -217,8 +217,7 @@ function prepareDrive(drive: Drive) {
 			disabled: true,
 			icon: 'warning',
 			size: null,
-			link:
-				'https://www.raspberrypi.org/documentation/hardware/computemodule/cm-emmc-flashing.md',
+			link: 'https://www.raspberrypi.org/documentation/hardware/computemodule/cm-emmc-flashing.md',
```
lurch, line 220: This documentation is now at https://www.raspberrypi.com/documentation/computers/compute-module.html#flashing-the-compute-module-emmc

thundron: thanks for the heads up! I'll make sure to include it next CI run, as even a simple doc change for now triggers the whole process all over again :/

thundron:
@balena-ci I self-certify!

lib/gui/app/app.ts
thundron: including in the latest PR!

PR #3667, opened 2021-12-29
patch: fix mesage of null

### LINKED ISSUE #3648: Cannot read property ‘message’ of null
- **Etcher version:** 1.7.1
- **Operating system and architecture:** Arch Linux x86_64
- **Image flashed:** [latest deb](https://github.com/balena-io/etcher/releases/download/v1.7.1/balena-etcher-electron_1.7.1_amd64.deb) via AUR package
- **Do you see any meaningful error information in the DevTools?** 

As reported by users on the [forum](https://forums.balena.io/t/flash-failed-cannot-read-property/350420), verification after flashing ISO files always fails with message: “Cannot read property ‘message’ of null.”

mcraa:
resolves #3648

Change-type: patch
Signed-off-by: Peter Makra peter@balena.io

### DIFF
--- a/lib/shared/permissions.ts
+++ b/lib/shared/permissions.ts
@@ -40,7 +40,7 @@ function sudoExecAsync(
 			cmd,
 			options,
 			(error: Error | undefined, stdout: Std, stderr: Std) => {
-				if (error !== undefined) {
+				if (error) {
 					reject(error);
 				} else {
 					resolve({ stdout, stderr });


mcraa:
@balena-ci I self-certify!

lib/shared/permissions.ts
```diff
@@ -40,7 +40,7 @@ function sudoExecAsync(
 			cmd,
 			options,
 			(error: Error | undefined, stdout: Std, stderr: Std) => {
-				if (error !== undefined) {
+				if (error) {
```
mvduin, line 43: While this is a functional workaround for the issue, this code wasn't actually wrong. The real bug is here:
https://github.com/jorangreef/sudo-prompt/blob/v9.2.1/index.js#L159-L191

It passes the error argument of the `child_process.exec()` callback, which is [documented](https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback) to be `null` on success, directly to the callback of `sudoPrompt.exec()`, which violates [its typescript declaration](https://github.com/jorangreef/sudo-prompt/blob/v9.2.1/index.d.ts#L4).

PR #3676, opened 2022-01-09
Change Linux instructions in README.md

Robertschv:
1. `apt` is better for installations and is more use than `apt-get`.
It is faster than `apt-get`.
It shows more output in the terminal.

2. `rm` requires the use of `sudo` to work.

3. `sudo apt clean` and `sudo rm -rf /var/lib/apt/lists/*` are not required to uninstall Etcher.

4. Use `apt autoremove` to uninstall not required packages and have a cleaner uninstallation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ confidence.
    sudo apt update
    sudo apt install balena-etcher
    ```
-   >Note: after v1.7.9 the package name changed to `balena-etcher` (no electron at the end)
+   >Note: for version 1.7.9 and below the package name was `balena-etcher-electron`
 
 ##### Uninstall
 
@@ -87,7 +87,7 @@ sudo apt autoremove
    ```sh
    sudo dnf install -y balena-etcher
    ```
-   >Note: after v1.7.9 the package name changed to `balena-etcher` (no electron at the end)
+   >Note: for version 1.7.9 and below the package name was `balena-etcher-electron`
 
 ###### Uninstall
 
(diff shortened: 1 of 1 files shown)


hosnifraj0 (approved):
Good 👍

khancyr (approved):
LGTM

Robertschv:
It's been a year since I made this pull request, I have no respond from any maintainer, neither have they approved my pull request.

What is wrong?

mcraa:
Thanks for your concerns.
Nothing is wrong, as you see issues and PRs in the repo things are worked on that actually need a fix.
The update of the documentation is also planned in #4009

Robertschv (commented):
There is no install or update with the old name, but there can be uninstalled.

Robertschv:
I changed the commands to install and remove the package, it doesn't make sense to install a package that is not actually there, the old name is not available anymore and the current way to remove it only works for the old name and not with the new name.

cloudmaker97:
So: I've got troubles today while trying to install it via. the wrong commands, then i needed to unecceasarily removed the obsolote packages just to read a few seconds later, that the -electron package is outdate.. :(

It needs to be merged.

aethernet:
Closing as we cannot merge it. Atm apt is not a recommended way to install etcher anymore.
We might host a new repository in the upcoming weeks and we'll revisit this.

Robertschv:
Which way is the recommend @aethernet ?

The website still shows that I can use the .deb package to install in Debian.

PR #3711, opened 2022-02-22
patch: Fix auto update

mcraa:
When the `stagingPercentage` is not set in the `latest*.yml` it should mean 100%
The previous version expected that the property is always present with a value of `0` or `100` was not prepared for `undefined`

### DIFF
--- a/lib/gui/etcher.ts
+++ b/lib/gui/etcher.ts
@@ -43,7 +43,7 @@ async function checkForUpdates(interval: number) {
 				const release = await autoUpdater.checkForUpdates();
 				const isOutdated =
 					semver.compare(release.updateInfo.version, version) > 0;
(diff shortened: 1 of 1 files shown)


mcraa:
@balena-ci I self-certify!

lib/gui/etcher.ts
```diff
@@ -43,7 +43,7 @@ async function checkForUpdates(interval: number) {
 				const release = await autoUpdater.checkForUpdates();
 				const isOutdated =
 					semver.compare(release.updateInfo.version, version) > 0;
-				const shouldUpdate = release.updateInfo.stagingPercentage || 0 > 0;
```
lurch, line 46: `...  || 0 > 0`

Yeah, you can see why that didn't work, if it was expecting zero to be bigger than zero :rofl:

PR #3712, opened 2022-02-23
patch: add arm64 output

mcraa:


Zarf-Marf (approved):
ok

samjijina:
@mcraa Why was this closed?

mcraa:
in #3714 you have the current arm builds, and our pipeline until it does not have the required hardware won't produce the right output even if the updated configuration files are there

PR #3727, opened 2022-03-14
Make PLATFORM, HOST_ARCH information to be override-able

awesometic:
This can be useful when doing a PLATFORM, HOST_ARCH related experiment.
It is totally not harmful to the normal usage of this Makefile.

Before this patch, the `make info` will report like the below on my 2021 MBP 14.

```
$ make info
Platform            : darwin
Host arch           : aarch64
Target arch         : aarch64
```

After this patch, I can override the HOST_ARCH information.

```
$ HOST_ARCH=x64 make info
Platform            : darwin
Host arch           : x64
Target arch         : x64
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,50 +24,54 @@ SHELL := /bin/bash
 # ---------------------------------------------------------------------
 
 # http://stackoverflow.com/a/12099167
+# You can override the PLATFORM, HOST_ARCH info by doing:
+#
+#   make <target> [PLATFORM=<platform>] [HOST_ARCH=<arch>]
+#
 ifeq ($(OS),Windows_NT)
-	PLATFORM = win32
+	PLATFORM ?= win32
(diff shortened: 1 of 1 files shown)


lurch:
There's already support for overriding `TARGET_ARCH` https://github.com/balena-io/etcher/blob/master/Makefile#L82-L86
Out of curiosity, what's the need for overriding `HOST_ARCH`?

awesometic:
I looked into the scripts again and I realized that I misunderstood how this works.
Right, not really, it isn't needed except setting `TARGET_ARCH` and this is used for letting Resin know which architecture we are going to build.

I think that I was getting dizzy while I was trying to build a macOS universal application.
Shortly I thought that this change is needed to do more experiments with my virtual environments. Sorry. 😅 

Close this PR.

PR #3747, opened 2022-04-22
patch: deb afterinstall and readme updates

mcraa:


mcraa:
@balena-ci I self-certify!

PR #3756, opened 2022-05-02
Fix formatting of ETA

getsnoopy:
This commit fixes the formatting of the estimated time of arrival to use
the proper symbols as per the SI.


Robertschv (approved):
Seems good to me.

“m” is used for meter not minutes, “min” is the correct abbreviation of minutes, and the spaces are needed too.

getsnoopy:
@Robertschv Done!

PR #3780, opened 2022-07-05
patch: introducing github actions (WinGet)

mcraa:
patch: introducing github actions
winget release, EP auto update, release build

Change-Type: patch

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -6,7 +6,7 @@ jobs:
   publish:
     runs-on: windows-latest # action can only be run on windows
     steps:
-      - uses: vedantmgoyal2009/vedantmgoyal2009/winget-pkgs-automation/releaser-action@v1.0.0
+      - uses: vedantmgoyal2009/winget-releaser@latest
         with:
           identifier: Balena.Etcher
           installer-regex: 'balenaEtcher-Setup.*.exe$'
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -6,7 +6,7 @@ jobs:
   publish:
     runs-on: windows-latest # action can only be run on windows
     steps:
-      - uses: vedantmgoyal2009/winget-releaser@latest
+      - uses: vedantmgoyal2009/winget-releaser@v1
         with:
           identifier: Balena.Etcher
           installer-regex: 'balenaEtcher-Setup.*.exe$'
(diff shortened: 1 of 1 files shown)


mcraa:
closing in favor of #3819

ab77:
@mcraa do we want to do this still?

ab77:
@mcraa @builder555  @zwhitchcox this looks like a great first step to automating some more of these: https://en.wikipedia.org/wiki/List_of_software_package_management_systems

I would suggest we also consider `brew` and potentially `macports` (especially given our community brew maintainer has pulled out).

The ideal place for these steps to live would be in `.github/actions/finalize/action.yml`, since this action finalises GH releases and moves the latest pointer.

mcraa:
@ab77 yes,
need to fix the commithash of the action (instead of master) for security reasons, and need to get a token, or  PR an update to be able to pass a token instead using always `secrets.Winget_token`

ab77:
> @ab77 yes, need to fix the commithash of the action (instead of master) for security reasons, and need to get a token, or PR an update to be able to pass a token instead using always `secrets.Winget_token`

Does the WInget token ever change @mcraa ?

mcraa:
> Does the Winget token ever change @mcraa ?

No it does not. it is just a gh token so the action can open a PR for the `microsoft/winget` repo.
The action in the PR looks for the secret with a specific name, IMO the token should be passed as a parameter / env var.

ab77:
> > Does the Winget token ever change @mcraa ?
> 
> No it does not. it is just a gh token so the action can open a PR for the `microsoft/winget` repo. The action in the PR looks for the secret with a specific name, IMO the token should be passed as a parameter / env var.

If the token is always the same, we need to store it in a github actions secrets. This way it will be fully protected and only accessible to PRs running in our org (not on forks). Please get the token setup with the ops team in bitwarden and share it with me so I can take of setting up GH secrets.

mcraa:
i don't know why i remembered wrong but here is the correction:
the `token` is passed as a parameter

ab77:
> i don't know why i remembered wrong but here is the correction: the `token` is passed as a parameter

@mcraa parameters aren't secure, so passing secrets via these is an anti-pattern IMO. Can we make it a secret please?

Also, we need to consider https://github.com/balena-io/etcher/issues/3853 in the context of package repositories which use PR flow (doesn't scale with auto release cadence).

ab77 (changes requested):
Can we please use FLOWZONE existing token here if it will suffice.

ab77 (approved):

PR #3819, opened 2022-10-13
Switch to Flowzone

ab77:
Change-type: patch


mcraa:
it builds the lzma-native package into `node_modules/etcher-sdk/node_modules/lzma-native...`
instead of `node_modules/lzma-native` where it is expected to be

mcraa:
@ab77 is it possible to rebase to `update-sdk-7.2.0` ?

lurch:
:wink: 

image

.resinci.yml
```diff
@@ -0,0 +1,2 @@
+---
+disabled: true
```
Page-, line 2: You can just remove this file altogether as balenaCI auto-detects flowzone and disables itself (unless you didn't add that logic for the etcher pipeline?)

package.json
```diff
@@ -115,6 +118,9 @@
     "webpack-cli": "4.2.0",
     "webpack-dev-server": "4.5.0"
   },
+  "engines": {
+    "node": ">=16 < 18"
```
Page-, line 122: The version of electron etcher currently uses (12.2.3) runs nodejs 14.16.0, so it probably makes sense for the install/compile/etc to at least match 14.x

ab77:
> @ab77 is it possible to rebase to `update-sdk-7.2.0` ?

Have you tested it? Does it work/build in the current system at least @mcraa?

This project could benefit from an overhaul of dependencies...

package.json
ab77: @Page- turns out there is a bug in npm < 8 where it changes git URLs to git:// vs. https://. Since there's now SSH keys injected, private repos fail.

mcraa:
`const files = readdirSync(path.join('node_modules', 'lzma-native'))` where does it look if not in `node_modules/lzma-native` ? 
I drop the node_modules folder and rebuild etcher about every other week and it works.
The `update-sdk-7.2.0` branch does not build on the current CI the, really old node was the (or at least one) problem there too.
Since the lzma-native is updated in that sdk version and it builds differently (and has prebuilt binaries) I thought two birds one stone.

ab77:
@zwhitchcox @mcraa @builder555 do you folks know how to stop node14 from resolving git URLs to ssh://... ?

e.g.

```
$ grep ssh package-lock.json 
      "resolved": "git+ssh://git@github.com/balena-io-modules/unbzip2-stream.git#4a54f56a25b58950f9e4277c56db2912d62242e7",
      "version": "git+ssh://git@github.com/balena-io-modules/unbzip2-stream.git#4a54f56a25b58950f9e4277c56db2912d62242e7",
```


cc @Page-

zwhitchcox:
@ab77 I was able to resolve the issue by simply deleting the package-lock.json file and rebuilding it, but I'm not sure what caused it in the first place.

zwhitchcox:
I can push my branch if you want with the correct lock file.

zwhitchcox:
Ok, same lzma binding...I will work on that

ab77:
> I can push my branch if you want with the correct lock file.

Looks like after running `npm i` locally, it fixed it.

.github/actions/publish/action.yml
```diff
@@ -0,0 +1,292 @@
...
+
+        [[ '${{ inputs.VERBOSE }}' =~ on|On|Yes|yes|true|True ]] && set -x
+
+        # generalise release asset names (latest) to avoid rebuilding/renaming on finalize
+        find dist -type f -maxdepth 1 \
+          -name "*${{ steps.package_release.outputs.version }}*" \
+          -exec sh -c 'mv {} $(echo {} | sed 's/${{ steps.package_release.outputs.version }}/latest/g')' \;
+
+    # https://www.electron.build/auto-update.html#staged-rollouts
+    - name: Configure staged rollout(s)
```
ab77, line 239: @Page- @mcraa @builder555 @dfunckt folks, sanity check please on this... This is reverse engineered `resinci-deploy editlatest ...` (from [here](https://github.com/product-os/scripts/blob/master/meta/meta-electron.sh), which never ran by the [looks](https://ci.product-os.io/teams/main/pipelines/github-events-balena-io/jobs/meta-electron) of it). Basically we now update all the electron update yaml(s) from repo.yml spec and upload to GH as releases (once!).

.github/actions/publish/action.yml
```diff
@@ -0,0 +1,292 @@
...
+        find dist -type f -maxdepth 1 \
+          -name "latest*.yml" \
+          -exec yq -i e .version=\"${{ steps.package_release.outputs.version }}\" {} \;
+
+        find dist -type f -maxdepth 1 \
+          -name "latest*.yml" \
+          -exec yq -i e .stagingPercentage=\"$percentage\" {} \;
+
+    # https://github.com/softprops/action-gh-release#-customizing
+    - name: Create draft GitHub (pre)release
```
ab77, line 257: @Page- @mcraa @builder555 @dfunckt the new approach is to create a draft release using the branch name and upload generalised (no version number in filenames) assets. This allows the PR to be built multiple times, rebuilt/rebased, etc. and not have the release polluted with multiple versions of assets potentially.

I am not sure if this is better than deleting the draft release and re-creating it with every PR run. I suppose either way will work, which way to you feel is better?

ab77:
@mcraa @builder555 we also need to test the Electron updater, can we test is 1.7.9 upgrades to 1.7.10 (rigging it somehow while the PR/release are still draft)? cc @konmouz

.github/actions/publish/action.yml
mcraa: well if it never ran, that explains why was `stagingPercentage` always `0` however `triggerNotification` was set

mcraa:
> @mcraa @builder555 we also need to test the Electron updater, can we test is 1.7.9 upgrades to 1.7.10 (rigging it somehow while the PR/release are still draft)? cc @konmouz

@ab77 
I think we can do it if the release is named `1.7.10-alpha.1` or similar. Then in a local custom build allowing `prerelease` updates.
Otherwise the update is already broken, see my comment above on `stagingPercentage` it can't break it more.

ab77:
> > @mcraa @builder555 we also need to test the Electron updater, can we test is 1.7.9 upgrades to 1.7.10 (rigging it somehow while the PR/release are still draft)? cc @konmouz
> 
> @ab77 I think we can do it if the release is named `1.7.10-alpha.1` or similar. Then in a local custom build allowing `prerelease` updates. Otherwise the update is already broken, see my comment above on `stagingPercentage` it can't break it more.

Since we build only once, the release is stamped with the future version (i.e. v1.7.10 for this PR). So @mcraa, you are saying we should just skip the electron updater testing and iterate/fix after this PR is merged?

builder555:
tested the latest build on intel and m1 macs - both work. @mcraa @ab77

mcraa:
> Since we build only once, the release is stamped with the future version (i.e. v1.7.10 for this PR). So @mcraa, you are saying we should just skip the electron updater testing and iterate/fix after this PR is merged?

In short yes.
The files (`latest*.yml`s) are looking good. The last issue with the prev CI was the `stagingPercentage`. That looks better as well.

mcraa:
I always forgot to tag @ab77 in the reply. So I have to ask another for the ping:

Does flowzone update https://github.com/balena-io/etcher/blob/gh-pages/scrutinizer.json in the finalize step?
That is where the etcher.io site queries the latest version from.

ab77:
> I always forgot to tag @ab77 in the reply. So I have to ask another for the ping:
> 
> Does flowzone update https://github.com/balena-io/etcher/blob/gh-pages/scrutinizer.json in the finalize step? That is where the etcher.io site queries the latest version from.

Flowzone doesn't know about this at all @mcraa

ab77:
@mcraa @builder555 do we have a way to build on `windows-latest` or are we limited to `windows-2019`?

mcraa:
> @mcraa @builder555 do we have a way to build on `windows-latest` or are we limited to `windows-2019`?

i think it is possible to install the same dependencies/tools what `'19` have making `latest` able to build, but I'd better iterate on etcher and make it build on "stock" `latest`, but that will take some time.

.github/actions/publish/action.yml
```diff
@@ -0,0 +1,249 @@
...
+        token: ${{ fromJSON(inputs.secrets).FLOWZONE_TOKEN }}
+        files: |
+          dist/*.AppImage
+          dist/*.blockmap
+          dist/*.deb
+          dist/*.dmg
+          dist/*.exe
+          dist/*.rpm
+          dist/*.zip
+          dist/latest*.yml
```
mcraa, line 232: They should be there but I don't see the `latest*.yml` files in the lats release.

mcraa: [🤔 ](https://github.com/balena-io/etcher/actions/runs/3373079085/jobs/5597497859#step:6:827)

mcraa (approved):

.github/workflows/flowzone.yml
```diff
@@ -0,0 +1,33 @@
...
+    branches:
+      - "main"
+      - "master"
+
+jobs:
+  flowzone:
+    name: Flowzone Linux
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@ab77/operational
+    # FIXME: (re)inherit once SENTRY_AUTH_TOKEN secret standardised
+    #secrets: inherit
```
Page-, line 15: Why can't you use `secrets: inherit` @ab77? It should work fine even with custom jobs, see https://github.com/balena-os/balena-supervisor/blob/v14.2.18/.github/workflows/flowzone.yml#L14 / https://github.com/balena-os/balena-supervisor/blob/v14.2.18/.github/actions/finalize/action.yml#L18 and lets you use sensible names in the custom jobs as well

.github/workflows/flowzone.yml
```diff
@@ -0,0 +1,33 @@
...
+on:
+  pull_request:
+    types: [opened, synchronize, closed]
+    branches:
+      - "main"
+      - "master"
+
+jobs:
+  flowzone:
+    name: Flowzone Linux
```
Page-, line 12: Using anything other than the expected `Flowzone` name invites weirdness, especially since the `runs_on` below specifies that this can also run on things other than linux

.github/workflows/flowzone.yml
```diff
@@ -0,0 +1,16 @@
...
+  pull_request:
+    types: [opened, synchronize, closed]
+    branches:
+      - "main"
+      - "master"
+
+jobs:
+  flowzone:
+    name: Flowzone
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@ab77/operational
```
ab77, line 13: don't merge until https://github.com/product-os/flowzone/pull/278 is merged and reset back to @master.

PR #3837, opened 2022-11-07
Update release notes on finalize

ab77:
Change-type: patch


ab77:
@balena-ci I self-certify!

PR #3838, opened 2022-11-07
Build targets individually

ab77:
Change-type: patch


ab77:
@balena-ci I self-certify!

PR #3839, opened 2022-11-07
Update FAQ.md

the-real-kenna:
Making the information from [this](https://jel.ly.fish/support-thread-1-0-0-front-cnv-db7lhzh) support ticket and [this](https://jel.ly.fish/pattern-user-gets-confused-sha512-representation-base64-cd74df4) Pattern more readily discoverable to end users for the time being.


ab77 (changes requested):
@the-real-kenna please make your commit message look like this to pass tests: 

```
Update FAQ.md

Change-type: patch
```

lurch:
There seems to be a bit of an overlap between https://github.com/balena-io/etcher/blob/master/FAQ.md and https://github.com/balena-io/etcher/blob/master/docs/USER-DOCUMENTATION.md ?

PR #3859, opened 2022-11-08
CI: generalise artefact handling

ab77:
* on PR syncs, delete draft releases on Linux runners only
* delete draft releases when unmerged PRs are closed

Change-type: patch


ab77:
@balena-ci I self-certify!

Issue #3853, opened 2022-11-08
Release cadence


ppfeufer:
You guys trying to break a new record? Looks like your CI is running amok.

![image](https://user-images.githubusercontent.com/2989985/200519368-b3929083-2644-462c-8706-3abeec6a8322.png)

ulab:
It makes it impossible to handle repositories like brew too.

ppfeufer:
Same for the Gentoo ebuild I'm maintaining ...

molaeiali:
Same here on AUR

JohnTHaller:
I'm worried there's some sort of mental health issue going on. There have been 20 "stable" releases in less than 24 hours. Some just minutes apart.

lurch:
I'm not a package maintainer, but this **is** clogging up my GitHub Notifications view :slightly_frowning_face:

zwhitchcox:
Hi, we are currently switching to a new CI system, this issue should be temporary.

cc @ab77 for more context

ppfeufer:
> Hi, we are currently switching to a new CI system, this issue should be temporary. Sorry for the inconvenience.
> 
> cc @ab77 for more context

But even when switching to a new CI system, there does not have to be a release for every single change that's being made ... That suspiciously looks like a misconfiguration somewhere ...

ab77:
Folks, for some context, Etcher's CI pipeline has been broken for years. On Monday, Etcher has been moved to our new GH actions pipeline (Flowzone) and wired up to our renovate config. Renovate will run ~ every hour and if there is a dependency update, it will build, test and publish a new Etcher release and make it latest.

This is by design, enjoy new Etcher releases without critical vulnerabilities, etc.

p-linnane:
Thank you for the clarification @ab77. Speaking as a Homebrew maintainer, we have decided to discontinue support for future updates of Etcher due to this new release cadence. It is too difficult to keep up with. I wish the other package maintainers luck in finding a solution that works for them.

molaeiali:
@ab77 thanks for the clarification, I am maintaining `etcher-bin` on AUR, and thinking of possible solutions for handling the situation, I can automate the process, but I think it will become frustrating for users to have an etcher update, each time they check for an update

I don't think it's really necessary to have hourly releases to be "secure"?? I don't think etcher is more of a target than web browsers and Linux kernel! And they don't release hourly!

I really hope for a reconsideration, but I understand if you don't wanna do anything about the issue as it's "by design"

[...]

PR #3875, opened 2022-11-16
Remove Homebrew instructions

p-linnane:
We have frozen etcher at `v1.8.14` and discontinued support for it in Homebrew.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,22 +170,6 @@ yay -S balena-etcher
 yay -R balena-etcher
 ```
 
-#### Brew (macOS)
-
-**Note**: Etcher has to be updated manually to point to new versions,
-so it might not refer to the latest version immediately after an Etcher
-release.
-
-```sh
-brew install balenaetcher
-```
-
-##### Uninstall
-
-```sh
-brew uninstall balenaetcher
-```
-
 #### Chocolatey (Windows)
 
 This package is maintained by [@majkinetor](https://github.com/majkinetor), and


ab77:
@p-linnane please check previous commit history how to format your commit message correctly.

ab77 (approved):

ab77:
@klutchell FYI: here's one of the first external contributions to a Flowzone enabled project, which requires to be signed by our keys; I guess we need to figure out a way to build/test/package these..

chrisys:
> balenaetcher is no longer updated in homebrew-cask due to the maintenance demand of multiple stable releases per day.

@p-linnane I saw this message in the brew info output for Etcher, but I'm just wondering if there have been some crossed wires here. I checked #3853 which I'm guessing is what this relates to but if I'm understanding it correctly this was simply a temporary occurrence to get dependencies up to date. As @ab77 mentioned on the issue Renovate has now been disabled again so there won't be continued releases every day based on dependency updates.

I personally value having an up to date package available in Homebrew as I know @klutchell does and likely many others as well so just wondering if there's any reason we can't resume updates now?

ab77:
@chrisys IMO we really need to implement/manage this ourselves, same for other package repositories.

There needs to be some careful thought given as to how this is implemented, in terms of cadence for one. Since HB, Macports, etc. are based on PR workflows, for every update, a PR needs to be open against these repositories and reviewed/merged manually. This is incompatible with automatic release cadence where versions may change multiple times per day (even with bulletproof tests).

Also, with Etcher is particular, passing unit tests and merging code does not necessarily mean a working release, since the tests don't actually test the final product (ie. no e2e tests). There is a deprecated Spectron framework test stub for only one platform (Windows?) in there, which tests for an app window appearing, but doesn't actually inspect its contents. So a blank window will pass tests. We can't ever guarantee that the releases we generate will work on all platforms. It currently requires manual testing, which we don't have a lot of (any?) bandwidth for.

There will be a feature soon in flowzone to skip finalising GH releases on every merge and defer to a manual intervention. Once that ships, we can re-enable renovate on Etcher and continue it pumping out new "untested" pre-release minor/patch versions. Then someone from the Etcher team can validate a release manually, mark it final and move the latest pointer. If the Etcher team gets bored manually testing releases, they should consider implementing e2e tests using Selenium/Webdriver framework in Electron to test a user journey through the software on all supported platforms (i.e. flash image to USB). We can provide Github actions self-hosted runners with access to real USB devices for this.

We should then be able to wire an action for that "mark latest" event to open PRs against HB, Macports, Chocolatey, etc. and try to convince their gatekeepers that we are not submitting garbage.

p-linnane:
@chrisys If a regular release cadence is back on the table, then I am open to restoring support for it in Homebrew. That said, it appears every commit is still generating a new release (4 in the last 24 hours), so I'm hesitant to readd support for now.

@ab77 If you'd like a little bit of inspiration there is a GitHub Action I see that creates cask PR's for new releases. It was adapter from a more popular action made for formulae: https://github.com/macauley/action-homebrew-bump-cask.

The bottom line is that there is no animosity on my end. I like Etcher and use it myself. I am happy to work with you on getting support back into Homebrew when the time is right.

PR #3877, opened 2022-11-17
Check deps monthly

jlosito:
This will check node related dependencies with dependabot and submit pull requests with version bumps on a monthly basis in order to keep packages up-to-date.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


ab77:
@jlosito we use [renovate](https://github.com/balena-io/renovate-config) for this, and it is presently disabled, because:

https://github.com/balena-io/etcher/issues/3853
https://github.com/balena-io/etcher/issues/3835

PR #3890, opened 2022-11-28
Adding EtcherPro device serial number to the Settings modal

wolvi-lataniere:
Adding EtcherPro device serial number to the Settings modal
Pulls Serial from Supervisor API when available, displays only Serial number when available, otherwise displays only UUID.

Change-type: patch
Signed-off-by: Aurelien VALADE <aurelien.valade@balena.io>

---
##### Contributor checklist

- [x] Build output been rebuilt and tested (on linux aarch64 only)
- [ ] Introduces security considerations
- [ ] Tests are included
- [ ] Documentation is added or changed
- [ ] Affects the development, build or deployment processes of the component

##### Reviewer Guidelines
- When submitting a review, please pick:
  - '*Approve*' if this change would be acceptable in the codebase (even if there are minor or cosmetic tweaks that could be improved).
  - '*Request Changes*' if this change would not be acceptable in our codebase (e.g. bugs, changes that will make development harder in future, security/performance issues, etc).
  - '*Comment*' if you don't feel you have enough information to decide either way (e.g. if you have major questions, or you don't understand the context of the change sufficiently to fully review yourself, but want to make a comment)
---

### DIFF
--- a/lib/gui/app/utils/etcher-pro-specific.ts
+++ b/lib/gui/app/utils/etcher-pro-specific.ts
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2022 balena.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Dictionary } from 'lodash';
+
+export class EtcherPro {
+	private supervisorAddr: string;
+	private supervisorKey: string;
+	private tags: Dictionary<string> | undefined;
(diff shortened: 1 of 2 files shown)


lib/gui/app/utils/etcher-pro-specific.ts
```diff
@@ -0,0 +1,73 @@
...
+	constructor(supervisorAddr: string, supervisorKey: string) {
+		this.supervisorAddr = supervisorAddr;
+		this.supervisorKey = supervisorKey;
+		this.uuid = (process.env.BALENA_DEVICE_UUID ?? 'NO-UUID').substring(0, 7);
+		this.tags = undefined;
+		this.get_tags().then((tags) => (this.tags = tags));
+	}
+
+	async get_tags(): Promise<Dictionary<string>> {
+		const result = await fetch(
```
aethernet, line 40: I'd wrap this in a try/catch otherwise all etcher not running on a pro (almost everyone) would throw an exception which we'll get in Sentry

aethernet (changes requested):

lib/gui/app/utils/etcher-pro-specific.ts
aethernet: Actually, not sure fetch would throw an exception for 404 but at least an error, which it shouldn't if we're not on pro.

aethernet (approved):

PR #3893, opened 2022-12-01
Patch: run linux build on ubuntu-18.04

aethernet:
Running on ubuntu-latest means you need a more recent version of glibc which breaks on older ubuntu.

Thanks to @theofficialgman for suggesting the fix.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -20,5 +20,5 @@ jobs:
       (github.event.pull_request.head.repo.full_name != github.repository && github.event_name == 'pull_request_target')
     secrets: inherit
     with:
-      tests_run_on: '["ubuntu-latest","macos-latest","windows-2019"]'
+      tests_run_on: '["ubuntu-18.04","macos-latest","windows-2019"]'
       restrict_custom_actions: false


theofficialgman:
@aethernet looks like more changes are necessary
there is a prolematic PPA added that ends up bumping the glibc requirement higher than bionic.

```
The following packages will be upgraded:
  libgcc1
1 upgraded, 0 newly installed, 0 to remove and 11 not upgraded.
Need to get 41.7 kB of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu bionic/main amd64 libgcc1 amd64 1:11.1.0-1ubuntu1~18.04.1 [41.7 kB]
```
the ubuntu-toolchain-r ppa includes an updated libgcc1 package that when installed causes gcc/g++ to produce builds that are not compatible with bionic and focal

idk where that ppa is getting added in the buildsystem but it needs to be removed and is not necessay to build etcher

theofficialgman:
@aethernet its not the existance of libgcc1, that is a normal package, its the launchpad ppa that is in the buildsystem that causes the issue

the ppa is still there after your latest commit 
https://github.com/balena-io/etcher/actions/runs/3594586798/jobs/6053071258#step:5:319

aethernet:
mmh i see :(

aethernet:
Yeah i don't get where it comes from either, maybe it's a dep of something else, or part of the flow, still looking

aethernet:
The weird thing beeing it happens exactly between the installation of libgbm1 and lbgconf-2.4, ...
```
    - libgbm1
    - libgcc1
    - libgconf-2-4
```

theofficialgman:
it looks to be part of the official github runner image... <https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/gcc.sh>

thats not good, you could just remove the ppa programatically before any apt installs run, that would solve it

I'll look into this again tonight and double check that this is an issue.

aethernet:
why would it install it at that exact moment ? that's very suspicious to me.

btw i got confirmation that the build made on ubuntu 18.04 runs fine on unbuntu 20.04. (i won't have a 18.04 avail. tonight but i'll test that tomorrow myself)

aethernet:
I confirm it works on 18.04
I'm rebasing and will retest, but I believe we're good here.

aethernet:
@balena-ci I self-certify!

PR #3896, opened 2022-12-01
Build and package ARM releases

ab77:
Dependencies:
* https://github.com/balena-io-modules/etcher-sdk/pull/267

Showstoppers

* https://github.com/balena-io/etcher/pull/3896/files#diff-1e6a5ce8ae6f024f8d6a5e3443a05f0921f3f5aa0c789b816349b4c148638e4aR10-R18

image

Showstoppers (windows-arm64):
* https://github.com/addaleax/lzma-native/issues/132
  https://github.com/RockLakeGrass/lzma-native-Windows-ARM64/blob/42be008b8a657eae4a9506a90b96fec3e86d5582/package.json
* https://github.com/nodejs/build/issues/2450#issuecomment-1367786829
* https://github.com/actions/runner-images/issues/768
* https://github.blog/changelog/2022-09-28-github-actions-self-hosted-runners-now-support-windows-arm-hardware/


ab77:
So it looks like the main common culprit to getting a working arm64 build for Etcher is `ronomon/direct-io`:

macOS-arm64 (electron-builder error):

```
fatal error: /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/electron-universal-92tV9I/Tmp.app/Contents/Resources/app/generated/modules/@ronomon/direct-io/binding.node and /Users/runner/work/etcher/etcher/dist/mac-universal--arm64/balenaEtcher.app/Contents/Resources/app/generated/modules/@ronomon/direct-io/binding.node have the same architectures (x86_64) and can't be in the same fat output file
```

Windows-arm64 (dev tools console error)

```
Uncaught Error: \\?\C:\Users\belodetek\AppData\Local\Programs\balena-etcher\resources\app\generated\modules\@ronomon\direct-io\binding.node is not a valid Win32 application.
\\?\C:\Users\belodetek\AppData\Local\Programs\balena-etcher\resources\app\generated\modules\@ronomon\direct-io\binding.node
    at process.func [as dlopen] (electron/js2c/asar_bundle.js:5)
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:1138)
    at Object.func [as .node] (electron/js2c/asar_bundle.js:5)
    at Module.load (internal/modules/cjs/loader.js:935)
    at Module._load (internal/modules/cjs/loader.js:776)
    at Function.f._load (electron/js2c/asar_bundle.js:5)
    at Function.o._load (electron/js2c/renderer_init.js:33)
    at Module.require (internal/modules/cjs/loader.js:959)
    at require (internal/modules/cjs/helpers.js:88)
    at Object.<anonymous> (gui.js:1)
```

Hopefully we can get that dependency updated/built for multiple architectures. Could be also as simple as webpack picking up the wrong binaries?

ab77:
Taking over from https://github.com/balena-io/etcher/pull/3868

aethernet:
@ab77 i had no issue building direct-io for arm64 on my m1 with node-gyp, but the version bundled in the npm package is x86_64

ab77:
> Does this build with an DMG now? I've been able to build the code but no DMG is made. I can publish a community build if a DMG is built.

unfortunately not, see https://github.com/balena-io/etcher/pull/3896#issuecomment-1334500206

mcraa:
We had the `ext2fs` error in previous PRs and should be [solved on master](https://github.com/balena-io/etcher/commit/b59b171e43a170c0ccf84d36ab4ba2db403e25c5) already, I don't know why is it coming back here.
The node-gyp error on windows is new for me, found quite some issues mentioning it for example 
https://github.com/nodejs/node-gyp/issues/1371

package.json
```diff
@@ -123,9 +123,185 @@
     "webpack-dev-server": "4.11.1"
   },
   "engines": {
-    "node": ">=14 < 16"
+    "node": "16"
+  },
+  "build": {
```
mcraa, line 128: what is the point having the whole `electron-builder.yml` in the package json?

ab77: one file instead of two.

ab77:
> We had the `ext2fs` error in previous PRs and should be [solved on master](https://github.com/balena-io/etcher/commit/b59b171e43a170c0ccf84d36ab4ba2db403e25c5) already, I don't know why is it coming back here.

workaround by setting `strict: false`. The real fix is to maybe figure out why a replacement isn't being made?

> The node-gyp error on windows is new for me, found quite some issues mentioning it for example [nodejs/node-gyp#1371](https://github.com/nodejs/node-gyp/issues/1371)

Maybe because we are trying to bundle [drivelist v9.2.x](https://github.com/balena-io-modules/drivelist/tree/3a000d24715b9914498b48aa53f86aab46c42563), which is ~ 2-4 years old and build/re-build under node 16?

mcraa:
> > We had the `ext2fs` error in previous PRs and should be [solved on master](https://github.com/balena-io/etcher/commit/b59b171e43a170c0ccf84d36ab4ba2db403e25c5) already, I don't know why is it coming back here.
> 
> workaround by setting `strict: false`. The real fix is to maybe figure out why a replacement isn't being made?
> 

to remove the reliance on node stuff like `binding` and the `node_modules` folder which are not available in the packaged electron environment

> > The node-gyp error on windows is new for me, found quite some issues mentioning it for example [nodejs/node-gyp#1371](https://github.com/nodejs/node-gyp/issues/1371)
> 
> Maybe because we are trying to bundle [drivelist v9.2.x](https://github.com/balena-io-modules/drivelist/tree/3a000d24715b9914498b48aa53f86aab46c42563), which is ~ 2-4 years old and build/re-build under node 16?

then it should fail on the other two platforms too

ab77:
> then it should fail on the other two platforms too

Sure, it should, but it doesn't. Maybe Windows is special. Maybe trying to build code years out of date with modern tools is the problem. Maybe both - I have no idea. I am ignoring this issue for now and hoping by the time I cycle back to it, all the dependencies are bumped up and it just builds.

ab77:
> to remove the reliance on node stuff like binding and the node_modules folder which are not available in the packaged electron environment

I am afraid I don't understand what that means. the problem I found is that `fetchWasm()` throws an error with `with string-replace-loader { strict: true }`. I've not dug into it, but from reading the docs, this would happen if no replacement is made, so to catch no-ops I am guessing.

package.json
```diff
@@ -74,14 +74,14 @@
...
-    "electron-builder": "^23.0.9",
+    "electron": "13.6.9",
+    "electron-builder": "23.6.0",
     "electron-mocha": "9.3.3",
     "electron-notarize": "1.2.2",
     "electron-rebuild": "3.2.9",
     "electron-updater": "5.3.0",
     "esbuild-loader": "2.20.0",
-    "etcher-sdk": "^7.4.6",
+    "etcher-sdk": "7.4.7-build-ab77-operational-baddef0b8af269203a8ed1681c1ef49a6ab6451b-1",
```
mcraa, line 84: see comment at 
https://github.com/balena-io-modules/etcher-sdk/pull/267#discussion_r1060799772

package.json
```diff
@@ -21,17 +21,18 @@
     "lint-css": "prettier --write lib/**/*.css",
     "lint-ts": "balena-lint --fix --typescript typings lib tests scripts/clean-shrinkwrap.ts webpack.config.ts",
     "lint": "npm run lint-ts && npm run lint-css",
-    "postinstall": "electron-rebuild -t prod,dev,optional",
+    "preinstall": "bash scripts/ci/preinstall.sh",
```
mcraa, line 24: Add `cross-env` or similar instead of `bash`.
On line 17 `flowzone-preinstall-linux` still has `electron-builder.yml` but that file is removed.

ab77: > Add `cross-env` or similar instead of `bash`.
https://github.com/kentcdodds/cross-env is archived and the package is not gettign any support. Do we really need to use it? My preference would be not to include depracated packages if at all possible, give we already have a ton of those. what does `cross-env` give us over `bash`, if bash is standardised across platforms?

> On line 17 `flowzone-preinstall-linux` still has `electron-builder.yml` but that file is removed.
Cheers, fixed.

mcraa: That is why I wrote "cross-env **_or similar_**"
however from the author:
> So I'm declaring cross-env as finished. No new features will be added (and old features will not likely be removed). I don't plan to do any more than fix security/critical bugs and keep the Node.js support up-to-date.

It still looks like a safe bet.
bash is not available out of the box on many platforms, with a package like this we don't need to write if windows then cmd if linux then bash but only if not zsh, etc and so when a new contributor tries to run `npm i` it will just run even if they don't have bash.

dfunckt:
Superseded by #4132

Issue #3942, opened 2022-12-12
Re-add setting to disable trim


net147:
- **Etcher version:** v1.10.2
- **Operating system and architecture:** Windows x64
- **Image flashed:** Raw SD image with ext4 partition that has unallocated blocks
- **What do you think should have happened:** The entire image is written as-is to the SD card
- **What happened:** Writing of unallocated blocks in the ext{2,3,4} partitions are skipped
- **Do you see any meaningful error information in the DevTools?** No

Currently the option can only be disabled in config.json by setting autoBlockmapping to false. There was previously a setting with the text "trim ext{2,3,4} partitions before writing (raw images only)" but it was removed. The data contained in the unallocated blocks of the image may be important in some use cases such as verifying the checksum of partitions or images.

aviks1919:
i have a 16GB SD card firmware image writing to 32GB SD card. After the firmware flashing, the SD card EXT4 partition has some files corruption. However, if i use Win32DiskImager, i did not see this issue. Do you think the issue i encountered is related to your mentioned "unallocated blocks in the ext4 partition are skipped"?

lurch:
Might be related to #3892 ? (which is also about Etcher writing corrupt images, which work fine with other image-flashing tools)

net147:
For people having issues with Etcher on Windows writing corrupt images, can you please try the following?
1. Close Etcher
2. Press Win+R to open the run dialog
3. Type the following and press enter: notepad %AppData%\balena-etcher\config.json
4. Change "autoBlockmapping": true, to "autoBlockmapping": false,
5. Save the file
6. Start Etcher
7. Try writing the image again

If you want to enable trim again, do the same as above but change autoBlockmapping back to true.

Thanks.

net147:
> i have a 16GB SD card firmware image writing to 32GB SD card. After the firmware flashing, the SD card EXT4 partition has some files corruption. However, if i use Win32DiskImager, i did not see this issue. Do you think the issue i encountered is related to your mentioned "unallocated blocks in the ext4 partition are skipped"?

It could be related. Try what I suggested in https://github.com/balena-io/etcher/issues/3942#issuecomment-1348601617.

aviks1919:
> For people having issues with Etcher on Windows writing corrupt images, can you please try the following?
> 
> 1. Close Etcher
> 2. Press Win+R to open the run dialog
> 3. Type the following and press enter: notepad %AppData%\balena-etcher\config.json
[...]

PR #3947, opened 2022-12-16
Patch: expose trim ext{2,3,4} setting

builder555:
Change-type: patch

### DIFF
--- a/lib/gui/app/components/settings/settings.tsx
+++ b/lib/gui/app/components/settings/settings.tsx
@@ -37,6 +37,10 @@ async function getSettingsList(): Promise<Setting[]> {
 			name: 'errorReporting',
 			label: i18next.t('settings.errorReporting'),
 		},
+		{
+			name: 'autoBlockmapping',
+			label: 'Trim ext{2,3,4} partitions before writing (raw images only)',
+		},
 	];
 	if (['appimage', 'nsis', 'dmg'].includes(packageType)) {
 		list.push({


net147:
Fixes https://github.com/balena-io/etcher/issues/3942

net147:
> > Fixes #3942
> 
> Can you test if the issue is fixed? I put the setting back but didn’t get a chance to flash a raw image yet. @net147 https://github.com/balena-io/etcher/releases/tag/untagged-568286a262c6719031c0

I can test if you can provide a Windows build. The link above doesn't work.

builder555:
> Both those links give a Github 404 error, which I guess means it's a link only available to Balena employees?

i think you're right. i don't have access to change permissions right now, but i uploaded it here: https://ufile.io/486m2h0t the link will expire in 7 days.

net147:
> > Both those links give a Github 404 error, which I guess means it's a link only available to Balena employees?
> 
> i think you're right. i don't have access to change permissions right now, but i uploaded it here: https://ufile.io/486m2h0t the link will expire in 7 days.

Works fine for me. With trim enabled the SD card doesn't match the image. With trim disabled the SD card matches the image.

builder555:
That’s good news, I’ll test on Etcher Pro and make a release

mcraa (approved):

lib/gui/app/components/settings/settings.tsx
```diff
@@ -37,6 +37,10 @@ async function getSettingsList(): Promise<Setting[]> {
 			name: 'errorReporting',
 			label: i18next.t('settings.errorReporting'),
 		},
+		{
+			name: 'autoBlockmapping',
+			label: 'Trim ext{2,3,4} partitions before writing (raw images only)',
```
lurch, line 42: Looking at the lines above and below this change, should the label-text here have been using `i18next.t()` ? :man_shrugging:

builder555: good call @lurch

PR #3948, opened 2022-12-16
Patch: replaced plain text with i18n in settings

builder555:
Change-type: patch

### DIFF
--- a/lib/gui/app/i18n/en.ts
+++ b/lib/gui/app/i18n/en.ts
@@ -138,6 +138,7 @@ const translation = {
 			autoUpdate: 'Auto-updates enabled',
 			settings: 'Settings',
 			systemInformation: 'System Information',
+			trimExtPartitions: 'Trim ext{2,3,4} partitions before writing (raw images only)',
 		},
 		menu: {
 			edit: 'Edit',
--- a/lib/gui/app/components/settings/settings.tsx
+++ b/lib/gui/app/components/settings/settings.tsx
@@ -39,7 +39,7 @@ async function getSettingsList(): Promise<Setting[]> {
 		},
 		{
 			name: 'autoBlockmapping',
-			label: 'Trim ext{2,3,4} partitions before writing (raw images only)',
+			label: i18next.t('settings.trimExtPartitions'),
 		},
 	];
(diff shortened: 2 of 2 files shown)


lib/gui/app/i18n/en.ts
```diff
@@ -138,6 +138,7 @@ const translation = {
 			autoUpdate: 'Auto-updates enabled',
 			settings: 'Settings',
 			systemInformation: 'System Information',
+			trimExtPartitions: 'Trim ext{2,3,4} partitions before writing (raw images only)',
```
lurch, line 141: I wonder if "ext{2,3,4}" is too much of a "developer shorthand", which won't make sense to normal people? :man_shrugging:  Perhaps "ext2/ext3/ext4" would be better?

mcraa: Probably the settings page will get crowded if everything is exposed there so this would move into an advanced settings accordion.

mcraa (approved):

lib/gui/app/i18n/en.ts
lurch: > if everything is exposed there

I wasn't suggesting having separate tickboxes for ext2, ext3 and ext4; merely changing the text-label to:
```
'Trim ext2/ext3/ext4 partitions before writing (raw images only)'
```
as I'm not sure if "normal people" will understand what ext{2,3,4} is supposed to mean? :man_shrugging:   (but this is obviously only a minor issue, so I'll shut up now :laughing: )

mcraa: Agree, minor stuff, but don't shut up 😄 
IIUC we have kind of aligned opinion.
The "advanced settings" section would be for not "normal people" so if you dare to go there you should understand.
Btw options like this are available already by modifying the `config.json`, that is hidden enough for "really not normal" people.

lurch: Cool cool.
And just to clarify further, I wasn't concerned about the "Trim partitions ..." language, but _specifically_ about the curly-braces part of the `ext{2,3,4}` nomenclature :wink: 

BTW, just out of curiosity... did you manage to figure out why this trim setting was causing filesystem corruption? AIUI it _shouldn't_ cause corruption when left enabled?

PR #3967, opened 2023-01-05
Unifying sentry reports in a single project

aethernet:
The DSN is hardcoded for a single `balenaetcher` project in Sentry.
Before this PR, one project was created for each PR by Flowzone using the name of the PR branch.

Change-type: patch


mcraa:
Would it make sense to merge it into the sentry PR?

aethernet:
totally

PR #3987, opened 2023-01-23
Add support for Node 18

dfunckt:
The Electron version we’re currently using is on Node 14 but this is a step forward to upgrading to a newer Electron and Node version.

Updates etcher-sdk and switches the redundant aws4-axios dependency to just axios.

Also changed bundler to stop trying to bundle wasm files — they must be included inline with JS code as data — and removed some now redundant code.

The crucial changes that enable support are:

1. The update to etcher-sdk@8 where some dependency fixes and updates took place
2. The downgrade and pinning of "electron-rebuild" to v3.2.3 until we’re able to update to Electron >= 14.2. The patch we need to avoid is https://github.com/electron/rebuild/pull/907. Also see: https://github.com/nodejs/node-gyp/issues/2673 and https://github.com/electron/rebuild/issues/913
3. A rule in webpack.config to ignore `aws-crt` which is a dependency of (ultimately) `aws4-axios` which is used by etcher-sdk and does a runtime check to its availability. We’re not currently using the “assume role” functionality (AFAIU) of aws4-axios and we don’t care that it’s not found, so force webpack to ignore the import. See https://github.com/aws/aws-sdk-js-v3/issues/3025

Change-type: minor

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -236,9 +236,9 @@
       "dev": true
     },
     "@aws-crypto/ie11-detection": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/ie11-detection/-/ie11-detection-2.0.2.tgz",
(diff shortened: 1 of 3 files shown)


dfunckt:
I self-certify!

PR #4000, opened 2023-02-01
Switch to `@electron/remote`

dfunckt:
Electron 12 deprecated `electron.remote` and the functionality was removed in Electron 14, but became available as a separate `@electron/remote` module. This commit makes the transition to the external module as an intermediary step to enable updating to a newer Electron version.

Change-type: patch


builder555:
Tested locally and on EtcherPro - works with CM4 without issues (also flashes regular sd cards) 👍  @dfunckt

PR #4002, opened 2023-02-01
Remove configuration remote update

aethernet:
Remove the check for an updated configuration from the server.

Change-type: patch

### DIFF
--- a/lib/gui/etcher.ts
+++ b/lib/gui/etcher.ts
@@ -26,7 +26,6 @@ import './app/i18n';
 
 import { packageType, version } from '../../package.json';
 import * as EXIT_CODES from '../shared/exit-codes';
-import { delay, getConfig } from '../shared/utils';
 import * as settings from './app/models/settings';
 import { buildWindowMenu } from './menu';
 import * as i18n from 'i18next';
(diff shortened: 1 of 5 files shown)


aethernet:
I self-certify!

lib/gui/etcher.ts
```diff
@@ -206,32 +205,6 @@ async function createMainWindow() {
...
-				const autoUpdaterConfig: AutoUpdaterConfig = onlineConfig?.autoUpdates
-					?.autoUpdaterConfig ?? {
-					autoDownload: false,
-				};
-				for (const [key, value] of Object.entries(autoUpdaterConfig)) {
-					autoUpdater[key as keyof AutoUpdaterConfig] = value;
-				}
-				const checkForUpdatesTimer =
-					onlineConfig?.autoUpdates?.checkForUpdatesTimer ?? 300000;
-				checkForUpdates(checkForUpdatesTimer);
```
lurch, line 229: Looks like this also removes the call to `checkForUpdates` - was that intentional?

aethernet: aouch! No not intentional. Good catch!

aethernet: fixed in https://github.com/balena-io/etcher/pull/4006

Issue #3792, opened 2022-08-02
New Pi Compute Modules Not Initializing


bhjelstrom:
- 1.7.9
- Ubuntu 20.04, Win10
- No Dev Tools output

I have tested 7 new CM4s with HW revision ```c03141``` and new bootloaders. The progress bar appears for the cm, but the cm is never mounted as a storage device and the progress bar remains at 0%.
![image](https://user-images.githubusercontent.com/13728935/182279034-6efc91de-558a-495c-aec3-5500bcbf8cd6.png)

The new CM4s function as expected only when I run the latest rpiboot manually. Assuming https://github.com/balena-io-modules/node-raspberrypi-usbboot needs an update.

lsusb:
```Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 
  bDeviceSubClass         0 
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x0a5c Broadcom Corp.
  idProduct          0x2711 
  bcdDevice            0.00
  iManufacturer           1 
  iProduct                2 
  iSerial                 3 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0020
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0xc0
      Self Powered
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass       255 Vendor Specific Class
      bInterfaceSubClass      0 
      bInterfaceProtocol      0 
      iInterface              0 
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0200  1x 512 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x02  EP 2 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0200  1x 512 bytes
        bInterval               0
```

mcraa:
[...]

PR #4003, opened 2023-02-01
Update to Electron 14

dfunckt:
Depends on https://github.com/balena-io/etcher/pull/4000


builder555:
Tested this version locally and on EtcherPro, it seems to have a regression on #3792

dfunckt:
It’s unlikely this PR is at fault @builder555, can you test [v1.15.1](https://github.com/balena-io/etcher/releases/tag/v1.15.1) instead? That’s far more likely to have broken something.

builder555:
just tested it - 1.15.1 works fine with a CM4 @dfunckt

dfunckt:
That’s actually great news. There’s nothing that immediately comes to mind about what might have caused the regression but the update wasn’t straightforward either, so I may have missed something. I’ll take another tomorrow and reach out in  Zulip.

In the meantime, @builder555 can you please give #4000 a go too? That would narrow down things further.

mcraa:
Fin + CM3 or CM4s does not mount on windows. :/

lib/gui/etcher.ts
```diff
@@ -209,6 +210,7 @@ async function createMainWindow() {
 	);
 
 	const page = mainWindow.webContents;
+	remoteMain.enable(page);
```
mcraa, line 214: on line 233, `allowRendererProcessReuse`, that property is removed: https://www.electronjs.org/blog/electron-14-0/#removed-appallowrendererprocessreuse

does this remoteMain add that back, or we can delete that line? (same for tests)

dfunckt: Yeah, we should remove the flag, I expect it has no effect but for some reason typings allow it to be set which is misleading.

dfunckt:
I self-certify!

PR #4005, opened 2023-02-02
Add reference to etcher-efp in publishing.md

aethernet:
Add reference to etcher-efp in publishing.md

Change-type: patch


Lizzieepton:
LGTM

PR #4025, opened 2023-03-09
patch: add apt-get update in flowzone preinstall

aethernet:
libudev package has changed and cannot be installed if we not update apt cache


aethernet:
I Self-Certify!

Issue #3501, opened 2021-04-30
New deb|rpm PPA


thundron:
We're transitioning from Bintray to [Cloudsmith](https://cloudsmith.com), as such the old PPA won't work anymore.
Instead, to install deb/rpm packages you will have to follow the new instructions of the README, which for simplicity I'll also paste here:

# debian package

add new PPA:
```sh
curl -1sLf \
      'https://dl.cloudsmith.io/public/balena/etcher/setup.deb.sh' \
      | sudo -E bash
```
other ubuntu/debian based distros may add overrides for `distro` and `codename` as described in the [cloudsmith guide](https://cloudsmith.io/~balena/repos/etcher/setup/#formats-deb)

install etcher:
```sh
sudo apt install -y balena-etcher-electron
```

# rpm package

add new PPA:
```sh
curl -1sLf \
  'https://dl.cloudsmith.io/public/balena/etcher/setup.rpm.sh' \
  | sudo -E bash
```
different distros using rpm may add overrides for `distro` and `codename` as described in the [cloudsmith guide](https://cloudsmith.io/~balena/repos/etcher/setup/#formats-rpm)

install etcher:
```sh
sudo yum install -y balena-etcher-electron
```

PR #4055, opened 2023-04-25
Update docs

aethernet:
Removes reference to cloudsmith as a valid installation method for deb and rpm

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,132 +32,41 @@ installers for all supported operating systems.
 
 ## Packages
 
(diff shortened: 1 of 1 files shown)


lurch:
Might be worth updating #3501 as well, which is one of the sticky issues at https://github.com/balena-io/etcher/issues  ?

aethernet:
good call @lurch 
just closed it

aethernet:
I Self-Certify!

PR #4021, opened 2023-03-05
Update link to moved SUPPORT.md

olyop:
Updates the link of the help button to the moved SUPPORT.md markdown file.

Screenshot from 2023-03-05 12-42-46

PR #4056, opened 2023-04-26
Patch: Fix support link

aethernet:
replace https://github.com/balena-io/etcher/pull/4021


aethernet:
I self-certify!

lurch:
This will fix "new" versions of Etcher, but (as I mentioned in #4021 ) I still think it'd be worth adding a redirect or something at https://github.com/balena-io/etcher/blob/master/SUPPORT.md for the benefit of "existing" versions of Etcher? :man_shrugging:

PR #4060, opened 2023-04-27
patch: update instructions for installing deb file

jcapona:
Update install & uninstall instructions for `.deb` files in `README`. 

Changes:
- On install instructions, use `apt` instead of `dpkg` since `apt` takes care of managing dependencies when installing a `.deb` file.
- On uninstall instructions, removed some of the steps since they're not necessary.

PR #4061, opened 2023-05-03
Update README.md

haipnh:
- Accomplish some commands for Debian and Ubuntu based Package Repository.


lurch:
Overlaps with #4060 ?

haipnh:
@lurch , I agree with the solution in #4060.

PR #4065, opened 2023-05-12
Update link of electron-supported-platforms in README.md

PeterDaveHello:
This doc has moved for a while ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ the [license].
 
 [etcher]: https://balena.io/etcher
 [electron]: https://electronjs.org/
-[electron-supported-platforms]: https://electronjs.org/docs/tutorial/support#supported-platforms
+[electron-supported-platforms]: https://github.com/electron/electron/blob/main/README.md#platform-support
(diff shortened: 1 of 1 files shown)


dfunckt:
Hey @PeterDaveHello -- thanks for the PR. Etcher is currently stuck at an older Electron version so links to their docs for the supported platforms are misleading -- for example, we currently do not provide Apple Silicon builds, despite what their docs says. I'd much rather you deleted the complete `Note:` about supported platforms and the link -- the list of actually supported platforms is provided just above that note.

dfunckt:
BTW, make sure to provide a `Change-type` with your commit message so that the PR becomes mergeable. For example, here is how to format a commit message that is acceptable by our CI:

```
Updated foo with bar

Change-type: patch
```

dfunckt:
Superseded by #4132

PeterDaveHello:
Oh, sorry that I didn't notice and revise on time, but it's good to see some updates here.

PR #4105, opened 2023-09-07
Create jay mcisaac

jaydawg133:
sudo yum localinstall balena-etcher-***.x86_64.rpm


jay mcisaac
```diff
@@ -0,0 +1 @@
+
```
jaydawg133, line 1: jay mcisaac

PR #4122, opened 2023-10-19
patch: bump deps

aethernet:
Bump outdated dependencies


dfunckt:
Is it really necessary to switch from single quotes to double quotes right now? This will make rebasing any other work impossible.

aethernet:
Ah shit, i probably broke the auto-linter while upgrading husky :(

PR #4145, opened 2023-12-08
Update dependencies

aethernet:
bump most deps on top of the new electron-forge builder

Still need to upgrade @balena-lint to 7 this requires a lot of fixes that should be cherry-picked from the poc.

Not upgreadable without some deeper work:
- immutable -> 5 (it's a pre-release, 4.3.4 is the actual latest)
- css-loader -> 6 (breaks loading the fonts)
- node-ipc -> 11 (no need to mess with the breaking changes, let's switch to another transport instead)
- react -> 18 (unknown breaking changes)
- react-dom -> 18 (same)
- redux -> 5 (unknown breaking changes)
- styled-components -> 6 (unknown breaking changes)
- pretty-bytes -> 6 (it's an *esm only* package now, which is not compatible with test current suite)


dfunckt:
Updating `immutable` needs some looking into so I reverted. It's otherwise good to merge.

dfunckt (approved):

dfunckt:
I added a commit on top that coalesces all `test-<platform>` scripts into one.

Interestingly, even before this commit, a bunch of errors are printed during Linux tests but they seem to run successfully to completion:

```
[5300:1222/133804.075080:ERROR:bus.cc(407)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[5323:1222/133804.332045:ERROR:viz_main_impl.cc(196)] Exiting GPU process due to errors during initialization
[5333:1222/133804.352286:ERROR:command_buffer_proxy_impl.cc(127)] ContextResult::kTransientFailure: Failed to send GpuControl.CreateCommandBuffer.
```

aethernet:
LGTM

PR #4132, opened 2023-11-01
Modernize build pipeline

dfunckt:
Switch to electron-forge which is now officially adopted by the Electron project. Solves a bunch of issues along the way.

Issues related to experimental v1.18.13:

Closes #4138 
Closes #4129
Closes #4119 

Build related issues:

Closes #4136 
Closes #4125 
Closes #4103 
Closes #4087 
Closes #3767 
Closes #3728 
Closes #3653 
Closes #3400 
Closes #3292 
Closes #3269 
Closes #3263 
Closes #2870 
Closes #2826 

(Admittedly, some of the issues are linked opportunistically -- please reopen them if the issue is not actually solved)

PR #4147, opened 2023-12-17
fix: required node version

Rotzbua:
min required node version was changed to v18 in #4102


dfunckt:
Hi, thanks for your PR -- can you please amend your commit message so that it includes the Change-type header and force-push? Like so:

```
fix: required node version

Change-type: patch
```

Same for your other PR as well.

dfunckt:
Superseded by #4132

PR #4167, opened 2024-01-26
Aethernet/fix screensaver error

aethernet:
should prevent error for not being able to enable/disable screensaver when not running on an etcher-pro


aethernet:
I agree on the principle, however It's more tricky than it looks. It's not an normal event listener and I couldn't figure how to deal with the fact that it comes from a wrapper on etcher-pro :/

dfunckt:
Ugh:

> If you do not need a response to the message, consider using [ipcRenderer.send](https://www.electronjs.org/docs/latest/api/ipc-renderer#ipcrenderersendchannel-args).

I think we can just switch to `send` and be done with it.

aethernet:
awaiting the promise indeed did the trick + switch to .send as we don't care about the response.

lib/gui/app/models/flash-state.ts
```diff
@@ -45,10 +45,10 @@ export function isFlashing(): boolean {
  * The flag is used to signify that we're going to
  * start a flash process.
  */
-export function setFlashingFlag() {
+export async function setFlashingFlag() {
 	// see https://github.com/balenablocks/balena-electron-env/blob/4fce9c461f294d4a768db8f247eea6f75d7b08b0/README.md#remote-methods
 	try {
-		electron.ipcRenderer.invoke('disable-screensaver');
+		await electron.ipcRenderer.send('disable-screensaver');
```
dfunckt, line 51: Docs don't mention the return type of `send` but I suspect it doesn't return a promise like `invoke`, so there's no need to `await` anything.

aethernet: indeed it doesn't care there's no listener

dfunckt (approved):

PR #4175, opened 2024-02-14
patch: Correct MacOS devtools hotkey

gshpychka:
MacOS does not have an Alt key, it has an Option key instead. Correcting the hotkey in the docs.


aethernet (changes requested):
The commit title must start with `patch: ` to make versionist happy, having it in the PR title is not enough

aethernet (dismissed):

gshpychka:
@aethernet CI seems broken for Windows? https://github.com/balena-io/etcher/actions/runs/8372914804/job/24030888363?pr=4175

PR #4176, opened 2024-02-14
Replace deprecated flowzone input tests_run_on

klutchell:
The `custom_runs_on` array supports multiple runner labels in nested arrays.

Change-type: patch

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -18,7 +18,7 @@ jobs:
       (github.event.pull_request.head.repo.full_name != github.repository && github.event_name == 'pull_request_target')
     secrets: inherit
     with:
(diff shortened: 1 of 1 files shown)


klutchell:
Nothing helpful for this repo, but it allows us to cleanup old Flowzone code. I had to jump through some hoops recently in Flowzone because I saw repos were still using this field.

dfunckt:
> Nothing helpful for this repo, but it allows us to cleanup old Flowzone code. I had to jump through some hoops recently in Flowzone because I saw repos were still using this field.

👍

> Unless of course we wanted to start using self-hosted runners for tests on this repo, in which case multiple runner labels would be required and not supported by the old input.

Gotcha, no, I think there’s no reason to switch away from GitHub’s for the time being.

dfunckt (approved):

PR #4202, opened 2024-04-19
patch: prevent rebuild of native modules by @electron/rebuild

aethernet:
Prevent the rebuilding of native modules by @electron/rebuild.
This was useless and waste times during the build as those modules are only used in the sidecar using another non-electron version of node (and packaged by pkg).
Now we only rebuild once for the correct target instead of twice.

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -52,7 +52,9 @@ const config: ForgeConfig = {
 		},
 		...osxSigningConfig,
 	},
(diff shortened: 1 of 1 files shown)


aethernet:
waits for https://github.com/balena-io/etcher/pull/4193 to be merged, then should be ready to go

dfunckt (approved):

PR #4207, opened 2024-04-24
patch: setup wdio and port most tests

aethernet:
This is a similar move that what we did from electron-build to electron-forge but for testing.
This tool is more modern, can test the real build e2e (and not only do some partial unit/internal testing).
And removes a lot of complexity (or at least move it out of our way)

Most of the old tests runs fine in the new context (the real unit one).
I actually found and fix a couple bugs that were in the way.

But the more integrated tests are failing, and I temporarily moved them out of the way (the `spec.todo.ts`) as they will become redundant once we have proper E2E.

Once https://github.com/balena-io-modules/etcher-sdk/pull/321 is in, and merged here, we can start adding E2E tests with virtual drives, so that's for a next PR.

Note that this 🐰 has been driven by the ~~bad~~ weird support for ESM and typescript in electron-mocha, which is blocking https://github.com/balena-io/etcher/pull/4205 (prettyBytes to 6.1.1) which fixes some important bugs.

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -7,6 +7,7 @@ import { MakerDMG } from '@electron-forge/maker-dmg';
 import { MakerAppImage } from '@reforged/maker-appimage';
 import { AutoUnpackNativesPlugin } from '@electron-forge/plugin-auto-unpack-natives';
 import { WebpackPlugin } from '@electron-forge/plugin-webpack';
+import { exec } from 'child_process';
 
 import { mainConfig, rendererConfig } from './webpack.config';
 import * as sidecar from './forge.sidecar';
@@ -152,6 +153,23 @@ const config: ForgeConfig = {
 			// packageJson.packageType = 'dmg' | 'AppImage' | 'rpm' | 'deb' | 'zip' | 'nsis' | 'portable'
 
 			return packageJson;
+		},
+		postPackage: async (_forgeConfig, options) => {
+			if (options.platform === 'linux') {
+				// symlink the etcher binary from balena-etcher to balenaEtcher to ensure compatibility with the wdio suite and the old name
+				await new Promise<void>((resolve, reject) => {
+					exec(
+						`ln -s "${options.outputPaths}/balena-etcher" "${options.outputPaths}/balenaEtcher"`,
+						(err) => {
+							if (err) {
+								reject(err);
+							} else {
+								resolve();
+							}
+						},
+					);
+				});
+			}
 		},
 	},
 };
--- a/package.json
+++ b/package.json
@@ -16,12 +16,11 @@
   "scripts": {
     "prettify": "prettier --write lib/**/*.css && balena-lint --fix --typescript typings lib tests forge.config.ts forge.sidecar.ts webpack.config.ts",
     "lint": "npm run prettify && catch-uncommitted",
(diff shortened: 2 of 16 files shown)


forge.config.ts
```diff
@@ -153,6 +154,23 @@ const config: ForgeConfig = {
...
+							if (err) {
+								reject(err);
+							} else {
+								resolve();
+							}
+						},
+					);
+				});
+			}
+		},
```
aethernet, line 173: I don't know about this one.

WDIO knows that we build with forge and find the build without problem for the two other platforms.
On linux as the name is different, it can't find it.

I thought of 4 potential solutions:
- set a conditional absolute path to the build in wdio.conf
- use a conditional env var in the action.yml
- create a symlink in the action.yml
- this one, create a symlink in forge post-package hook

This has my preference as it's the authoritative file for both the name of the package and the output path

WDYT?

package.json
```diff
@@ -54,10 +53,12 @@
     "redux": "4.2.1",
     "rendition": "35.1.2",
     "semver": "7.5.4",
+    "sinon": "^17.0.1",
     "styled-components": "5.3.6",
     "sys-class-rgb-led": "3.0.1",
     "uuid": "9.0.1",
-    "ws": "^8.16.0"
+    "ws": "^8.16.0",
+    "xvfb-maybe": "^0.2.1"
```
dfunckt, line 61: This should be a dev dependency.

package.json
```diff
@@ -16,12 +16,11 @@
   "scripts": {
     "prettify": "prettier --write lib/**/*.css && balena-lint --fix --typescript typings lib tests forge.config.ts forge.sidecar.ts webpack.config.ts",
     "lint": "npm run prettify && catch-uncommitted",
-    "test-gui": "xvfb-maybe electron-mocha --recursive --reporter spec --window-config tests/gui/window-config.json --require ts-node/register/transpile-only --require-main tests/gui/allow-renderer-process-reuse.ts --full-trace --no-sandbox --renderer tests/gui/**/*.ts",
-    "test-shared": "xvfb-maybe electron-mocha --recursive --reporter spec --require ts-node/register/transpile-only --require-main tests/gui/allow-renderer-process-reuse.ts --full-trace --no-sandbox tests/shared/**/*.ts",
-    "test": "npm run test-gui && npm run test-shared",
+    "test": "echo 'Only use custom tests; if you want to test locally, use `npm run wdio`' && exit 0",
```
dfunckt, line 19: I don’t understand why you needed to make the test script a noop? From what I see the wdio command could just as well run in the test script. Can you elaborate?

dfunckt (approved):

package.json
aethernet: There's no point at running tests in against node as this is not a valid target. 
For the wdio test to run, we need a valid build. So it would be `npm run make && npm run wdio`. Which would be just a duplicate of the custom tests.

package.json
```diff
@@ -54,6 +53,7 @@
     "redux": "4.2.1",
     "rendition": "35.1.2",
     "semver": "7.5.4",
+    "sinon": "^17.0.1",
```
dfunckt, line 56: This should be a dev dependency, right?

wdio.conf.ts
```diff
@@ -0,0 +1,309 @@
...
+	// worker process. In order to have a group of spec files run in the same worker
+	// process simply enclose them in an array within the specs array.
+	//
+	// The path of the spec files will be resolved relative from the directory of
+	// of the config file unless it's absolute.
+	//
+	specs: ['./tests/**/*.spec.ts'],
+	// Patterns to exclude.
+	exclude: [
+		// 'path/to/excluded/files'
```
dfunckt, line 37: Can you please use this directive to exclude the test files instead of moving them around?

dfunckt (approved):

PR #4212, opened 2024-04-26
patch: hold request for metadata while waiting for flasher

aethernet:
This is a poor man wait while loading to prevent a potential race condition without completely blocking the interface
This should be properly addressed when we do a proper GUI refactoring.

### DIFF
--- a/lib/gui/app/components/source-selector/source-selector.tsx
+++ b/lib/gui/app/components/source-selector/source-selector.tsx
@@ -423,6 +423,14 @@ export class SourceSelector extends React.Component<
 						// this will send an event down the ipcMain asking for metadata
 						// we'll get the response through an event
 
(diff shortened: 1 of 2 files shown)


lib/gui/app/components/source-selector/source-selector.tsx
```diff
@@ -423,6 +423,14 @@ export class SourceSelector extends React.Component<
 						// this will send an event down the ipcMain asking for metadata
 						// we'll get the response through an event
 
+						// FIXME: This is a poor man wait while loading to prevent a potential race condition without completely blocking the interface
+						// This should be addressed when refactoring the GUI
+						let retryRequestMetada = 10;
```
dfunckt, line 428:
```suggestion
						let retryRequestMetadata = 10;
```

Or even better, just “retriesLeft” or similar.

dfunckt (approved):

PR #4228, opened 2024-05-20
patch: fix sentry DSN

aethernet:
Sentry DNS was only set for renderer and not main process.
(it needs both to work)

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -61,6 +61,13 @@ const rules: Required<ModuleOptions>['rules'] = [
 	},
 ];
 
+const injectAnalyticsToken = new DefinePlugin({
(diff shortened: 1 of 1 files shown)


webpack.config.ts
```diff
@@ -109,4 +109,14 @@ export const mainConfig: Configuration = {
 	resolve: {
 		extensions: ['.js', '.ts', '.jsx', '.tsx', '.css', '.json'],
 	},
+	plugins: [
+		new DefinePlugin({
```
dfunckt, line 113: The same exact definitions occur for the renderer process, I think it’s worth defining the vars once in module-global scope and just passing them as plugins for each process

dfunckt (approved):

PR #4333, opened 2024-10-08
patch: etcher-util is corrupted in RPM package

aethernet:
Cherry-picking https://github.com/balena-io/etcher/pull/4304

---

rpmbuild strips executables by default when generating an rpm packge. This was causing the JavaScript code bundled in the etcher-util file to be removed, causing "Pkg: Error reading from file." whenever etcher-util was called.

This in turn caused balena-etcher to generate the error message `Error: (0, h.requestMetadata) is not a function` when attempting to write an SD card.

This fixes the issue for RPM builds by replacing the `strip` command with `true` so that rpmbuild no longer strips the executables and the embeded code stays intact.

See: https://github.com/balena-io/etcher/issues/4150

### DIFF
--- a/.github/actions/publish/action.yml
+++ b/.github/actions/publish/action.yml
@@ -53,6 +53,13 @@ runs:
       shell: bash
       run: sudo apt-get install -y --no-install-recommends fakeroot dpkg rpm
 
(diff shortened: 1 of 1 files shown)


dfunckt (commented):
👍

PR #4411, opened 2025-02-26
Remove stale secrets

ab77:
these are long expired/gone/deprecated/unused


aethernet (approved):

klutchell (approved):

lurch:
I wonder if https://github.com/balena-io/etcher/blob/master/.gitignore#L110-L116 ought to be updated too?

PR #4417, opened 2025-03-17
Add Swedish translation

Sopor:


lurch:
Looks like this ought to be part of #4416 , rather than being a separate PR?

PR #4435, opened 2025-04-11
patch: remove analytics code

aethernet:
Removes usage analytics code (keep sentry crash reports)

[Fibery](https://balena.fibery.io/Work/Project/remove-unused-analytics-from-etcher-1224)


otaviojacobi:
@aethernet looks good, I think you can also `npm uninstall analytics-client` as it is no longer used aswell

PR #4484, opened 2025-06-28
Update README.md install instructions

fusaty:
Installing on Fedora throws error: Unknown argument "localinstall" for command "dnf5".  Rewritten to "install"


fusaty (commented):
/deploy - it is a simple readme file change

PR #4559, opened 2025-12-09
build: update workflow file to use MacOS 15

ghost:
MacOS 13 is no longer supported on Runners (https://github.com/actions/runner-images/issues/13046)

Change-type: patch


tqpcharlie:
Potential duplicate of #4445 depending on what version maintainers want?

PR #4560, opened 2025-12-09
node: upgrade to Node 22 latest

### LINKED ISSUE #4518: "error starting flasher sidecar process" when trying to make a bootable usb for linux
I tried both on MacOS and Windows and i get the same error

ghost:
Fixes #4518 and the underlying nodejs issue, which was fixed in CLI by upgrading to node 22.20.0

Change-Type: major

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@types/chai": "4.3.14",
     "@types/debug": "^4.1.12",
     "@types/mime-types": "2.1.4",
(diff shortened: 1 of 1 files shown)


tqpcharlie:
maybe this project should also [upgrade to Electron 39](https://www.electronjs.org/docs/latest/tutorial/electron-timelines) (newest Node/Electron version) before 37 goes EoL in January?

Update: see #4568

PR #4576, opened 2026-01-06
docs: keep yum instructions, add dnf install for Fedora 39+

DataArchitectPro:
Keep the existing yum-based installation instructions and add dnf-based commands for modern Fedora and RHEL-derived systems using dnf5, where `yum localinstall` is no longer supported.


DataArchitectPro:
Kept the existing yum-based installation instructions for legacy systems and added dnf-based install commands for Fedora 39+/dnf5 environments where yum localinstall is no longer supported.

PR #4636, opened 2026-05-16
Fix drag-and-drop and macOS sidecar startup

guitar24t:
## Summary

Restores drag-and-drop image loading and gets the flasher sidecar
running again on recent macOS (tested on macOS 26). Two user-visible
bugs were addressed, plus several smaller issues found while debugging.

### Drag-and-drop

Previously, dragging an image onto the window opened a browser-style
download instead of loading the image, and dropping on certain spots
(notably the "Flash from file" button) silently did nothing.

- **`lib/gui/app/app.ts`** — window-level `dragenter`/`dragover`/`drop`
  listeners (capture phase) that `preventDefault()` and force
  `dataTransfer.dropEffect = 'copy'`. This stops Chromium from
  navigating/downloading the dropped file, and the explicit
  `dropEffect` makes the `drop` event fire even over interactive
  `<button>` elements (Chromium otherwise defaults the effect to
  `'none'` there and suppresses the event).
- **`lib/gui/etcher.ts`** — `setWindowOpenHandler` denies renderer
  new-window opens and a `will-download` handler cancels unexpected
  downloads, as a safety net for the old fall-through behaviour.
- **`source-selector.tsx`** — the dropped path is now resolved with
  `webUtils.getPathForFile()`; `File.path` was removed in Electron
  32+ and the app runs on Electron 37, so it was always empty.
- **`source-selector.tsx`** — the drop is accepted anywhere on the
  window via a capture-phase `window` listener, plus a stable
  full-window overlay shown only during a drag. The overlay keeps the
  drop from landing on a button whose underlying DOM node is swapped
  out on hover (rendition `ColouredButton` ↔ `Plain`), which was
  disrupting the drag and preventing the drop from registering.

### macOS sidecar startup

After drag-and-drop worked, the flasher sidecar (`etcher-util`) failed
with `requestMetadata is not a function`. The root cause: the
`pkg`-built sidecar embedded Node 20.11.1, which segfaults on
macOS 26. Newer Node 20 builds segfault too; Node 22 runs cleanly.

- **`forge.sidecar.ts`** — pin the `pkg` target to `node22.22.2`.
- **`package.json`** — `overrides` bump `@yao-pkg/pkg-fetch` to
  `^3.5.33` (ships the Node 22 build) and `node-abi` to `^4.31.0`
  (older versions do not recognise Electron 37).

### Additional fixes in `api.ts`

Found while debugging the sidecar:

- Parenthesise the `ETCHER_SERVER_PORT` fallback — operator
  precedence meant an explicit port env var was discarded and `3435`
  was always used.
- Remove the dead Windows env-reset block (it spawned a throwaway
  `cmd` subprocess that had no effect).
- Wrap `ws.on('message')` in try/catch and stop `throw`ing on an
  unknown message type, so a bad frame can't crash the connection.
- Close the WebSocket and stop the heartbeat on connection error.
- Show a clear "sidecar failed to start" error instead of calling an
  undefined `requestMetadata`.

## Test plan

- [x] `npm run make` builds a macOS arm64 DMG
- [x] Sidecar binary starts and listens (no segfault) on macOS 26
- [x] Drop an image on the window background — loads instead of
      downloading
- [x] Drop on the "Flash from file" button, highlighted and not —
      loads
- [ ] Verify Windows and Linux builds via CI
- [ ] Verify an actual flash end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)


puhitaku:
+1
Broken DnD prevents me from upgrading from an ancient x86 build so please be merged soon ...

